PR #2, opened 2018-07-25
feat: initial organization, some db code, and the full request path

bbangert:
Co-authored-by: Phil Jenvey <pjenvey@underboss.org>


src/db/schema.sql
```diff
@@ -0,0 +1,69 @@
+-- why is collection_id before id in BSO?
```
rfk: It's just a convention to match the primary key order (collection_id, id); the order is important there because it allows `SELECT * FROM bso WHERE collection_id=X` to execute efficiently using a prefix of the primary key index.

src/db/schema.sql
```diff
@@ -0,0 +1,69 @@
...
+CREATE TABLE bso (
+  collection_id   INTEGER NOT NULL,
+  id              VARCHAR(64) NOT NULL,
+
+  sortindex       INTEGER DEFAULT 0,
+
+  payload         TEXT NOT NULL DEFAULT '',
+  payload_size    INTEGER NOT NULL DEFAULT 0,
+
+  -- milliseconds since unix epoch. Sync 1.5 spec says it shoud
```
rfk: sp: "shoud"

src/db/schema.sql
```diff
@@ -0,0 +1,69 @@
...
+        ( 4, "history"),
+        ( 5, "keys"),
+        ( 6, "meta"),
+        ( 7, "bookmarks"),
+        ( 8, "prefs"),
+        ( 9, "tabs"),
+        (10, "passwords"),
+        (11, "addons"),
+        -- forces new collections to start at 100
+        (99, "-push-");
```
rfk: Is there a way we can do this without creating a "sentinel" collection with a special name, perhaps by inserting and then deleting the record?

src/db/schema.sql
```diff
@@ -0,0 +1,69 @@
...
+);
+INSERT INTO keyvalues (key, value) VALUES ("SCHEMA_VERSION", 0);
+
+
+INSERT INTO collections (id, name) VALUES
+        (12, "addresses"),
+        (13, "creditcards");
+
+-- begin using user_version to track schema changes
+-- skip user_version=1 as that *should have been* set by 'SCHEMA_0'
```
rfk: Starting at 2 may not be necessary for us now that we're starting over from scratch.

PR #5, opened 2018-07-26
feat: more db code and tests

pjenvey:
also addressed or at least added comments for all of rfk's comments from yesterday

### DIFF
--- a/src/db/models.rs
+++ b/src/db/models.rs
@@ -2,9 +2,9 @@ use std::collections::HashMap;
 use std::ops::Deref;
 
 use diesel::{
-    delete, dsl::sql, insert_into, replace_into, result::Error as DieselError,
-    result::Error::NotFound, sql_query, sql_types::Integer, sqlite::SqliteConnection, update,
-    Connection, ConnectionError, ExpressionMethods, QueryDsl, RunQueryDsl, Table,
+    delete, dsl::sql, insert_into, replace_into, result::Error as DieselError, sql_query,
+    sql_types::Integer, sqlite::SqliteConnection, update, Connection, ConnectionError,
+    ExpressionMethods, OptionalExtension, QueryDsl, RunQueryDsl, Table,
 };
 
 use super::schema::{bso, collections, keyvalues};
@@ -19,7 +19,7 @@ pub const MAX_TIMESTAMP: i64 = 4070822400000;
 
 pub const STORAGE_LAST_MODIFIED: &'static str = "Storage Last Modified";
 
-#[derive(Default, Copy, Clone)]
+#[derive(Default, Debug, Copy, Clone)]
 pub struct DBConfig {
     pub cache_size: i64,
 }
@@ -30,6 +30,7 @@ pub struct DBManager {
     config: DBConfig,
 }
 
+#[derive(Debug)]
 pub enum Sorting {
     None,
     Newest,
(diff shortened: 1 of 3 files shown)


src/db/models.rs
```diff
@@ -173,31 +179,59 @@ impl DBManager {
...
+
+    pub fn get_bso_modified(&self, collection_id: i64, bso_id: &str) -> Result<i64, DieselError> {
+        let last_modified = bso::table
+            .select(bso::last_modified)
+            .filter(bso::collection_id.eq(&collection_id))
+            .filter(bso::id.eq(&bso_id))
+            .filter(bso::expiry.ge(&ms_since_epoch()))
+            .first(&self.conn)
+            .optional()?;
+        // XXX: returning 0 when not found or prefer Option?
```
rfk: FWIW, my gut feeling is it'll be better off returning an `Option`, but my rust-related gut feelings are not exactly well developed yet...

src/db/models.rs
```diff
@@ -173,31 +179,59 @@ impl DBManager {
...
+    }
+
+    pub fn delete_bsos(&self, collection_id: i64, bso_id: &[&str]) -> Result<i64, DieselError> {
+        self.conn.transaction(|| {
+            delete(bso::table)
+                .filter(bso::collection_id.eq(&collection_id))
+                .filter(bso::id.eq_any(bso_id))
+                .execute(&self.conn)?;
+
+            let last_modified = ms_since_epoch();
```
rfk: There's an interesting architectural choice here, which maybe won't ever come up in practice but that I wanted to mention.

From the point of view of the Firefox Sync client, an individual request to the syncstorage API is supposed to be atomic and occur at a specific instant in time.  For example, if the client does a `POST /storage/collection` and sends three items, then it expects that:

* The last-modified time of each of those three items will be the same
* The new last-modified time of the collection will be the same as those items just added
* The new last-modified time of the storage as a whole will be the same as that of the collection
* The response will include an `X-Last-Modified` header that gives this new timestamp

So is the "instant" at which all those changes occur a property of the individual call into the backend storage API (as it is here)?  Or should it be a property of the request as a whole?  Should some higher-level code be passing in a notion of "this is the current timestamp" in order to help ensure this consistency?

If each web API request corresponds to exactly one backend storage API request, then it probably doesn't matter.  In the python code, I believe we fix a timestamp when we open a database transaction, and use it for all operations done inside that transaction.  I don't know what the best approach will be for this new code.

Just something to think about as this code grows :-)

src/db/models_test.rs
```diff
@@ -140,26 +158,215 @@ fn test_put_bso_updates() {
...
+
+    let cid = 1;
+    let size = 12;
+    for i in 0..size {
+        let mut bso = pbso(
+            cid,
+            &i.to_string(),
+            Some(&format!("payload-{}", i)),
+            Some(i),
+            Some(DEFAULT_BSO_TTL),
```
rfk: Should passing in `None` here cause it to default to `DEFAULT_BSO_TTL`?

src/db/models.rs
pjenvey: In general I agree (it's an Option) but I suspect 0 might work ok for how the view layer uses this. Left this as a note to revisit after we implement those views

src/db/models.rs
pjenvey: I did notice this difference in the Python version vs go (go timestamps in the db layer). We'll keep this in mind, the latter might be simpler but we should only do it if it works in the same consistent way as the Python version.

PR #8, opened 2018-08-03
feat: stub out some more endpoints

philbooth:
This afternoon I started stubbing out some more of the API endpoints, the `info/...` ones and `delete_bso`. Handlers and dispatchers are added, but obviously there's no attempt to wire anything into the db layer here.

While doing that, all of the boilerplate started to get a bit annoying so I tried to eliminate some of it using macros. Not sure how intuitive the macro format I ended up with is, so I'm interested to see whether you guys have suggestions to improve those.

I also had to update the `cargo fmt` invocation in travis because of the 1.28.0 release (as predicted by Phil J in #6). That change is included here too along with some related `rustfmt` fixes to `db/models.rs` and `db/models_test.rs`, just so that the build would succeed in Travis.

@bbangert @pjenvey r?

### DIFF
--- a/src/db/models.rs
+++ b/src/db/models.rs
@@ -324,7 +324,8 @@ impl DBManager {
     }
 
     pub fn last_modified(&self) -> Result<i64, DieselError> {
-        Ok(self.get_key(STORAGE_LAST_MODIFIED)?
+        Ok(self
+            .get_key(STORAGE_LAST_MODIFIED)?
             .map_or(0, |last_modified| last_modified.parse().unwrap()))
     }
 }
--- a/src/dispatcher.rs
+++ b/src/dispatcher.rs
@@ -17,7 +17,7 @@ pub struct CollectionInfo {
 }
 
 impl Message for CollectionInfo {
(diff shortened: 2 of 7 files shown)


src/db/models.rs
```diff
@@ -324,7 +324,8 @@ impl DBManager {
     }
 
     pub fn last_modified(&self) -> Result<i64, DieselError> {
-        Ok(self.get_key(STORAGE_LAST_MODIFIED)?
+        Ok(self
+            .get_key(STORAGE_LAST_MODIFIED)?
```
philbooth: This and all of the changes in `db/models_test.rs` were rustfmt-inflicted.

src/dispatcher.rs
```diff
@@ -72,14 +84,54 @@ impl DBExecutor {
     }
 }
 
-impl Handler<CollectionInfo> for DBExecutor {
+impl Handler<Collections> for DBExecutor {
```
philbooth: Changed the name of this to better match the endpoint name and be more consistent with the others, hope that's cool.

PR #9, opened 2018-08-08
feat: add handler stubs for collection endpoints

philbooth:
The substantive change here is just to add the remaining handler stubs for the collection endpoints. That should mean all the endpoints listed in the API docs now have a handler, unless I've missed any.

Note that `DELETE /storage/{collection}` has two entries in the docs but only one handler here. It behaves differently depending on whether there is an `ids` query param (or rather, it *will* behave differently when we actually hook it up to some behaviour). But obviously that's all just one resource in actix-web's view of things.

In addition to that there's some other small fixes:

* A rustfmt fix for some `extern crate` ordering.

* Some of the repetition between similar macro rules has been eliminated by moving it to a common base macro. This took me longer to get right than I anticipated it would, for a couple of reasons:

  * The ambiguity of trying to parse repeating groups in nested macros; is the repetition part of the child's match arm or the parent's macro body? That ruled out using a kind of higher-order macro approach.

  * Macro hygiene rules, which I hadn't come across before.

* Some of the repetition in the tests has been extracted to a common `test_endpoint` function.

* Mea culpa from #8, I forgot to include some of the `/info` routes in the server setup. ~~This was a side-effect of the really annoying fact that there's apparently no way to share the setup between test server and real server, or at least not one that I've figured out yet (suggestions welcome for that if you have any).~~

@bbangert @pjenvey r?

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -1,8 +1,5 @@
 //! Main application server
 
-#[cfg(test)]
-mod test;
-
 use std::collections::HashMap;
 use std::sync::{Arc, RwLock};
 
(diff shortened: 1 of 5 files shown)


philbooth:
As this PR hasn't had any comments yet, I took the liberty of amending the route-fix commit to eliminate the duplication between `src/server/mod.rs` and `src/server/test.rs`.

It's just a small tweak to pull the route initialisation code into a macro, and didn't seem worthy of a separate PR or whatever.

* **BEFORE:** b889fd3
* **AFTER:** a89a24b
* **DIFF:** b889fd3...a89a24b

src/server/mod.rs
```diff
@@ -16,6 +13,41 @@ use handlers;
 use handlers::ServerState;
 use settings::Settings;
 
+macro_rules! init_routes {
+    ($app:expr) => {
+        $app.resource("{uid}/info/collections", |r| {
+            r.method(http::Method::GET).with(handlers::collections);
+        }).resource("{uid}/info/collection_counts", |r| {
+                r.method(http::Method::GET)
```
philbooth: The indentation here looks all kinds of wrong, I know, but `rustfmt` insists this is how it should be. 😕

PR #14, opened 2018-08-14
fix: rip out the DBExecutor code

### LINKED ISSUE #10: Remove dbexecutor abstraction
The dbexecutor abstraction was needed for talking to blocking databases, we're going to be working against a database we can talk to with futures, so this abstraction should be removed.

philbooth:
Fixes #10.

Not sure if I did this right but I deleted the entire `dispatcher` module because from Ben's email it sounded like we'll be calling async db methods directly from the handlers, rather than sending messages to an actor. I'm happy to put it back though if I misunderstood.

I opted to retain the structs that encapsulate the args for db methods, even though we're not doing anything with them yet. Keeping them around meant there was a target for the data-marshalling code in the handlers, which presumably we're still going to need in due course. I've put those structs temporarily in a module called `src/data.rs`, even though ultimately I'd expect them to live somewhere in the `db` directory. I just figured there's still a lot of unknowns about what that directory is going to look like, so it's better to leave a clean slate in there for now.

@bbangert @pjenvey @rfk r?


src/handlers.rs
```diff
@@ -1,21 +1,17 @@
...
 
-use dispatcher;
+use data;
 
 /// This is the global HTTP state object that will be made available to all
 /// HTTP API calls.
-pub struct ServerState {
-    pub db_executor: Addr<dispatcher::DBExecutor>,
-}
+pub struct ServerState;
```
philbooth: I kept `ServerState` around even though it's empty now because presumably it's going to be populated with some kind of db property in due course.

src/handlers.rs
```diff
@@ -59,30 +55,23 @@ impl<S> FromRequest<S> for HawkHeader {
...
+                $($property: $value,)*
+            };
+            Box::new(result(Ok(HttpResponse::Ok().json(()))))
         }
     }
 }
 
 macro_rules! info_endpoints {
-    ($($handler:ident: $dispatcher:ident),+) => ($(
+    ($($handler:ident: $data:ident,)+) => ($(
```
philbooth: I tweaked the match arms for all these macros so that trailing items keep their comma. It makes our macro invocations better match the real Rust code (not to mention all the other reasons that trailing commas are a Good Thing).

philbooth:
(and no rush on the review front btw, I realise you guys have got more interesting things to focus on this week!)

PR #17, opened 2018-08-23
feat: define and mock a db trait

### LINKED ISSUE #16: Create db abstraction trait
We need to eventually support Google Spanner, but for local testing/hosting we need something we can run as well as our users. To handle this, we should abstract the db commands we need into a trait, and perhaps provide some default generic SQL queries that Spanner/SQL specializations can extend as needed.

The generic sql queries for Sync are here:
https://github.com/mozilla-services/server-syncstorage/blob/master/syncstorage/storage/sql/queries_generic.py

Schemas:
https://github.com/mozilla-services/server-syncstorage/blob/master/syncstorage/storage/sql/dbconnect.py#L105

philbooth:
Fixes #16.

Continuing from the conversation about my WIP branch in IRC yesterday, this leaves the existing db code mostly untouched but defines a new `Db` trait, plus a `MockDb` stubbed implementation thereof.

I say "mostly" untouched because the old `src/db/mod.rs` has been completely clobbered, so that old code is not actually being compiled any more. I realise that's not very different from deleting it all, but there were enough points of difference here in both the struct definitions and where I've moved them to that it really seemed cleaner to start anew. I'll call out some of those differences in the inline comments and hopefully they make sense to you guys. If not, I can revert some stuff, no worries.

I hooked the new trait into a mock implementation just so the tests would have something to assert against really. It's not a very useful mock apart from that and I expect us to delete it as soon as we have an SQLite implementation for local testing. But the upside of plugging in a mock implementation now was that it gave some flesh to the test cases that need to assert result properties.

@bbangert @pjenvey r?

### DIFF
--- a/src/db/mock.rs
+++ b/src/db/mock.rs
@@ -1,0 +1,40 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Mock db implementation with methods stubbed to return default values.
+
+use futures::future;
+
+use super::*;
+
+#[derive(Debug)]
+pub struct MockDb;
+
+macro_rules! mock_db_method {
+    ($name:ident, $type:ident) => {
+        fn $name(&self, _params: params::$type) -> DbFuture<results::$type> {
+            Box::new(future::result(Ok(results::$type::default())))
+        }
+    }
+}
+
+impl Db for MockDb {
+    fn new() -> Box<Db> {
+        Box::new(MockDb)
+    }
+
+    mock_db_method!(get_collections, GetCollections);
+    mock_db_method!(get_collection_counts, GetCollectionCounts);
+    mock_db_method!(get_collection_usage, GetCollectionUsage);
+    mock_db_method!(get_quota, GetQuota);
+    mock_db_method!(delete_all, DeleteAll);
+    mock_db_method!(delete_collection, DeleteCollection);
+    mock_db_method!(get_collection, GetCollection);
+    mock_db_method!(post_collection, PostCollection);
+    mock_db_method!(delete_bso, DeleteBso);
+    mock_db_method!(get_bso, GetBso);
+    mock_db_method!(put_bso, PutBso);
+}
+
+unsafe impl Send for MockDb {}
--- a/src/db/mod.rs
+++ b/src/db/mod.rs
@@ -1,5 +1,67 @@
-pub mod models;
-#[cfg(test)]
-mod models_test;
-pub mod schema;
-pub mod util;
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Generic db abstration.
+
+pub mod mock;
+pub mod params;
+pub mod results;
+
+use std::{
+    error::Error,
+    fmt::{self, Display, Formatter},
+};
+
+use futures::future::Future;
+
+type DbFuture<T> = Box<Future<Item = T, Error = DbError>>;
+
+pub trait Db: Send {
+    fn new() -> Box<Db>
+    where
+        Self: Sized;
+
+    fn get_collections(&self, params: params::GetCollections) -> DbFuture<results::GetCollections>;
+
+    fn get_collection_counts(
+        &self,
+        params: params::GetCollectionCounts,
+    ) -> DbFuture<results::GetCollectionCounts>;
+
+    fn get_collection_usage(
+        &self,
+        params: params::GetCollectionUsage,
+    ) -> DbFuture<results::GetCollectionUsage>;
+
+    fn get_quota(&self, params: params::GetQuota) -> DbFuture<results::GetQuota>;
+
+    fn delete_all(&self, params: params::DeleteAll) -> DbFuture<results::DeleteAll>;
+
+    fn delete_collection(
+        &self,
+        params: params::DeleteCollection,
+    ) -> DbFuture<results::DeleteCollection>;
+
+    fn get_collection(&self, params: params::GetCollection) -> DbFuture<results::GetCollection>;
+
+    fn post_collection(&self, params: params::PostCollection) -> DbFuture<results::PostCollection>;
+
+    fn delete_bso(&self, params: params::DeleteBso) -> DbFuture<results::DeleteBso>;
+
+    fn get_bso(&self, params: params::GetBso) -> DbFuture<results::GetBso>;
+
+    fn put_bso(&self, params: params::PutBso) -> DbFuture<results::PutBso>;
+}
+
+// HACK: temporary placeholder until we have proper error-handling
+#[derive(Debug)]
+pub struct DbError(String);
+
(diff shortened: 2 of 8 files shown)


src/db/params.rs
```diff
@@ -71,16 +70,16 @@ bso_data! {
     DeleteBso {},
     GetBso {},
     PutBso {
-        sortindex: Option<i64>,
+        sortindex: Option<i32>,
```
philbooth: The API docs define `sortindex` as max 9 digits, hence I scaled this down to `i32` everywhere.

Ref: https://mozilla-services.readthedocs.io/en/latest/storage/apis-1.5.html#basic-storage-object

src/db/params.rs
```diff
@@ -71,16 +70,16 @@ bso_data! {
...
         payload: Option<String>,
-        ttl: Option<i64>,
+        ttl: Option<u32>,
```
philbooth: The API docs define `ttl` as unsigned with up to 9 digits, so I switched this to `u32` everywhere.

Ref: https://mozilla-services.readthedocs.io/en/latest/storage/apis-1.5.html#basic-storage-object

src/db/results.rs
```diff
@@ -0,0 +1,47 @@
...
+//! Result types for database methods.
+
+use std::collections::HashMap;
+
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+pub type GetCollections = HashMap<String, u64>;
+pub type GetCollectionCounts = HashMap<String, u64>;
+pub type GetCollectionUsage = HashMap<String, u32>;
+pub type GetQuota = Vec<u32>;
```
philbooth: Specified units for usage and quotas in the API are kilobytes, so I figured `u32` was plenty big enough for these two?

Ref: https://mozilla-services.readthedocs.io/en/latest/storage/apis-1.5.html#general-info

src/db/params.rs
```diff
@@ -2,7 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, you can obtain one at https://mozilla.org/MPL/2.0/.
 
-//! Data definitions.
+//! Parameter types for database methods.
```
philbooth: I mentioned this in IRC yesterday, but having the input and output type definitions in separate modules just made it feel a bit easier to keep track of things in my head.

src/handlers.rs
```diff
@@ -187,7 +197,32 @@ pub struct BsoParams {
...
     pub payload: Option<String>,
-    pub ttl: Option<i64>,
+    pub ttl: Option<u32>,
 }
+
+pub fn get_configuration(_state: State<ServerState>) -> FutureResponse<HttpResponse> {
+    // TODO: populate from static config?
+    Box::new(future::result(Ok(
+        HttpResponse::Ok().json(Configuration::default())
+    )))
```
philbooth: I'm assuming this endpoint has no need to hit the db, but let me know if that's wrong. (/cc @rfk)

src/db/params.rs
philbooth: (another benefit is that it also makes the consuming code more uniform, which improves the refactorability of some of the boilerplate, e.g. everything can just use `params::$type` and `results::$type`)

src/db/results.rs
```diff
@@ -0,0 +1,47 @@
...
+#[derive(Debug, Deserialize, Serialize)]
+#[serde(untagged)]
+pub enum GetCollectionItem {
+    Id(String),
+    Bso(GetBso),
+}
+
+#[derive(Debug, Default, Deserialize, Serialize)]
+pub struct GetBso {
+    pub id: String,
```
philbooth: The old code was also returning `collection_id` here, but I don't think we ever need to expose those ids externally since all the endpoints take a collection name instead, right?

src/handlers.rs
```diff
@@ -54,22 +62,25 @@ impl<S> FromRequest<S> for HawkHeader {
...
-            let _data = data::$data {
-                $($property: $value,)*
-            };
-            Box::new(result(Ok(HttpResponse::Ok().json(()))))
+            Box::new(
+                state.db.$handler(params::$data {
+                    $($property: $value,)*
+                })
+                .map_err(From::from)
+                .map(|result| HttpResponse::Ok().json(result))
```
philbooth: Fwiw this macro implementation is not going to work long-term because some of the endpoints need to do extra marshalling, e.g. setting `payload_size` on BSOs and so on.

philbooth: (or we could keep this simple and do that stuff inside the db layer, but it would be common to both implementations so outside seems more appropriate)

src/db/mock.rs
```diff
@@ -0,0 +1,34 @@
...
+macro_rules! mock_db_method {
+    ($name:ident, $type:ident) => {
+        fn $name(&self, _params: params::$type) -> Box<Future<Item = results::$type, Error = DbError>> {
+            Box::new(future::result(Ok(results::$type::default())))
+        }
+    }
+}
+
+impl Db for MockDb {
+    mock_db_method!(get_collections, GetCollections);
```
bbangert: I'm a bit wary of macro's for trait definitions as they impair the readability of the trait with fairly minimal gain.

src/db/mock.rs
```diff
@@ -0,0 +1,34 @@
...
+use futures::future;
+
+use super::*;
+
+#[derive(Debug)]
+pub struct MockDb;
+
+macro_rules! mock_db_method {
+    ($name:ident, $type:ident) => {
+        fn $name(&self, _params: params::$type) -> Box<Future<Item = results::$type, Error = DbError>> {
```
bbangert: It's already become pretty common for Box Future's with common Error types to type alias it away, for example in autopush-rs we have:
https://github.com/mozilla-services/autopush-rs/blob/master/src/errors.rs#L66

actix-web used the trait system all-out to not only type alias away the Box Future shenanigans, but also put a variety of traits around Futures in general for a nicer more ergonomic API internally and externally:
https://github.com/actix/actix-web/blob/master/src/handler.rs#L157

Using such a setup should get us better readability, without the boiler-plate that made macro's slightly more appealing in this case.

src/db/mock.rs
philbooth: Understood. Should I get rid of the `db::params` and `db::results` abstractions too in that case, because the the method signatures aren't actually that useful otherwise, all the meaningful definitions are tied up in those two modules?

src/handlers.rs
philbooth: Oh yeah, that is a splendid suggestion, thanks!

philbooth:
Have pushed with changes for the inline comments and meeting feedback just now.

src/db/mod.rs
```diff
@@ -1,5 +1,56 @@
...
+
+use futures::future::Future;
+
+type DbFuture<T> = Box<Future<Item = T, Error = DbError>>;
+
+pub trait Db: Send {
+    fn new() -> Box<Db>
+    where
+        Self: Sized;
+    fn get_collections(&self, params: params::GetCollections) -> DbFuture<results::GetCollections>;
```
pjenvey: nit: some newlines around all these methods

src/handlers.rs
rfk: That's correct, it should be able to pull everything it needs from static app config.

src/db/results.rs
pjenvey: yup

PR #20, opened 2018-08-29
feat: implement hawk authentication

### LINKED ISSUE #19: Support hawk authentication
We'll need to figure out the details of hawk authentication and build it via the https://github.com/taskcluster/rust-hawk crate (we have some notes from rfk in handlers::HawkHeader).

(Despite the crate's warning of production use, a couple different projects in the https://github.com/mozilla/application-services repo have already used it successfully)

philbooth:
Fixes #19.

I didn't quite manage to get this finished today, mostly because I wasted a few hours writing parsing logic for the Hawk header, only to eventually RTFM and notice the hawk crate already implements that via the `FromStr` trait. Sigh.

Anyway, I wanted to post this as a WIP so that @rfk can hopefully cast his eye over the crypto logic while I'm away. As you'll see Ryan, I've left your excellent (and long) comment in place for now, interjected with my attempts to implement each section of it. There are a couple of places where I'm not sure I understood you correctly, so I'll call those out in the inline comments.

If I have time after I've packed my bag tomorrow, I might see if I can wire it into the request mechanism and write some tests for it. If I do get to that point I'll comment here and remove the WIP. But, being realistic, I probably won't get to that point.

See you all next week! 😄


src/auth.rs
```diff
@@ -0,0 +1,155 @@
...
+        if (payload.expires.round() as i64) < now {
+            return Err(AuthError);
+        }
+
+        //   * Derive `token_secret = HKDF-SHA256(master_secret, size=32, salt=payload["salt"], info="services.mozilla.com/tokenlib/v1/derive/" + id)`
+        let hkdf: Hkdf<Sha256> = Hkdf::extract(Some(payload.salt.as_bytes()), &hex::decode(&settings.master_secret)?);
+        let mut token_secret = [ 0u8; 32 ];
+        hkdf.expand(format!("services.mozilla.com/tokenlib/v1/derive/{}", id).as_bytes(), &mut token_secret)?;
+
+        //   * Use `token_secret` as the secret key for calculating the Hawk request MAC
```
philbooth: I wasn't sure about the significance of `calculating the Hawk request MAC` here. Does it mean something more than the direct comparison of `mac` and `token_secret` I'm doing below?

src/auth.rs
```diff
@@ -0,0 +1,170 @@
...
+            Some(payload.salt.as_bytes()),
+            &hex::decode(&settings.master_secret)?,
+        );
+        let mut token_secret = [0u8; 32];
+        hkdf.expand(
+            format!("services.mozilla.com/tokenlib/v1/derive/{}", id).as_bytes(),
+            &mut token_secret,
+        )?;
+
+        //   * Use `token_secret` as the secret key for calculating the Hawk request MAC
```
philbooth: I wasn't sure about the significance of calculating the Hawk request MAC here. Does it mean something more than the direct comparison of mac and token_secret I'm doing below?

src/auth.rs
```diff
@@ -0,0 +1,170 @@
...
+        let expected_signature = check.result().code();
+        if !constant_time_compare(signature, &expected_signature) {
+            return Err(AuthError);
+        }
+
+        //   * JSON decode `payload` to give an object like: {
+        //       'userid': 42,
+        //       'expires': 1329875384.073159
+        //       'salt': '1c033f'
+        //     }
```
philbooth: Is this a comprehensive definition or can other properties be present?

src/auth.rs
```diff
@@ -0,0 +1,170 @@
...
+        let payload: HawkPayload = serde_json::from_slice(payload)?;
+
+        //   * Check that the "expires" timestamp is not in the past.
+        let now = Utc::now().timestamp_millis();
+        if (payload.expires.round() as i64) < now {
```
philbooth: Whoops, just realised that this is all kinds of wrong, I need to multiply so that I'm not comparing apples to oranges.

src/auth.rs
philbooth: I notice the linked code has a `uid` property but your comment had `userid`. I'm presuming that was just a typo in the comment, am changing to `uid` here.

philbooth:
> If it helps, I generated some test data using a local syncserver instance.

Thanks for the test data @rfk, I started to write a test case with it today but I'm still having a bit of trouble getting it to work.

From reading the docs (both the protocol spec and the hawk-rs docs), I have the impression that I need some request data in order to validate the mac from the header: method, host, port and path. Obviously I can just read those from the actix request object during real execution, but what are they for your test data?

I tried a few things using host and port from the `node` property and some variations including `uid` as the path, but without success... what am I missing?

rfk:
> I need some request data in order to validate the mac from the header [...]
> what are they for your test data

Oh yeah, right, sorry.  I didn't write these down at the time, so I generated some fresh ones:

```
REQUEST:  GET http://localhost:5000/storage/1.5/1/storage/col2
AUTHZ:    Hawk nonce="omxLZWE=", mac="+1oGdzqpxYndK5ejQLdnZpXgGSt/IlxNh5MvcR6j7t4=", id="eyJub2RlIjogImh0dHA6Ly9sb2NhbGhvc3Q6NTAwMCIsICJ1aWQiOiAxLCAiZXhwaXJlcyI6IDE1MzYxOTkyNzQsICJmeGFfdWlkIjogIjMxOWI5OGY5OTYxZmYxZGJkZDA3MzEzY2Q2YmE5MjVhIiwgInNhbHQiOiAiYjAyNjBlIiwgImRldmljZV9pZCI6ICJjMDlkMjZmYWYyYjQ5YWI2NGEyODgyOTA3MjA2ZDBiNSJ96drmQ_KNFOe7U3g1D8ZX5-he2Bv2aRvKZzBPrCjHKO4=", ts="1536198980"
DATA:     {'node': 'http://localhost:5000', 'uid': 1, 'expires': 1536199274, 'fxa_uid': '319b98f9961ff1dbdd07313cd6ba925a', 'salt': 'b0260e', 'device_id': 'c09d26faf2b49ab64a2882907206d0b5'}
```

```
REQUEST:  POST http://localhost:5000/storage/1.5/1/storage/col2?batch=MTUzNjE5ODk3NjkyMQ==&commit=true
AUTHZ:    Hawk nonce="1d4mRs0=", mac="xRVjP7607eZUWCBxJKwTo1CsLcNf4TZwUUNrLPUqkdQ=", id="eyJub2RlIjogImh0dHA6Ly9sb2NhbGhvc3Q6NTAwMCIsICJ1aWQiOiAxLCAiZXhwaXJlcyI6IDE1MzYxOTkyNzQsICJmeGFfdWlkIjogIjMxOWI5OGY5OTYxZmYxZGJkZDA3MzEzY2Q2YmE5MjVhIiwgInNhbHQiOiAiYjAyNjBlIiwgImRldmljZV9pZCI6ICJjMDlkMjZmYWYyYjQ5YWI2NGEyODgyOTA3MjA2ZDBiNSJ96drmQ_KNFOe7U3g1D8ZX5-he2Bv2aRvKZzBPrCjHKO4=", ts="1536198978"
DATA:     {'node': 'http://localhost:5000', 'uid': 1, 'expires': 1536199274, 'fxa_uid': '319b98f9961ff1dbdd07313cd6ba925a', 'salt': 'b0260e', 'device_id': 'c09d26faf2b49ab64a2882907206d0b5'}
```

philbooth:
> So that leaves me thinking that the key must be wrong.

(well either that, or I've got your test data wrong somehow...)

src/auth/mod.rs
```diff
@@ -0,0 +1,202 @@
...
+        // Derive token secret from master secret + salt
+        let hkdf: Hkdf<Sha256> = Hkdf::extract(
+            Some(payload.salt.as_bytes()),
+            &settings.master_secret,
+        );
+        let mut token_secret = [0u8; 32];
+        hkdf.expand(
+            format!("services.mozilla.com/tokenlib/v1/derive/{}", id).as_bytes(),
+            &mut token_secret,
+        )?;
```
rfk: I missed it in my initial description, but re-reading the python `tokenlib` source code, I believe that `token_secret` here needs to be b64urlsafe encoded before using it to verify the mac below.  I'm having a bit of trouble building the project right now, but if I can get that working I'll put an encode step in here to check.

Cargo.lock
```diff
@@ -1301,6 +1403,18 @@ dependencies = [
  "quick-error 1.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
+[[package]]
+name = "ring"
+version = "0.8.1"
```
rfk: Both travis-ci and I are getting build failures like:

```
          /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/ring-0.12.1/crypto/curve25519/curve25519.c:4693: multiple definition of `GFp_x25519_public_from_private'
          /home/travis/build/mozilla-services/syncstorage-rs/target/debug/deps/libring-ae014aa1941bdf5c.rlib(curve25519.o):/home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/ring-0.8.1/crypto/curve25519/curve25519.c:4712: first defined here
```

Could be because this is introducing a second version of `ring`, and both versions can't be compiled together into a single binary?

src/auth/mod.rs
mostlygeek: I believe it is supposed to be base64encoded. 

https://github.com/mozilla-services/go-syncstorage/blob/master/token/token.go#L101

src/auth/mod.rs
```diff
@@ -0,0 +1,202 @@
...
+        // Verify signature
+        let mut expected_signature: Hmac<Sha256> = Hmac::new_varkey(&signing_secret)?;
+        expected_signature.input(&payload);
+        expected_signature.verify(signature)?;
+
+        // Parse payload JSON
+        let payload: HawkPayload = serde_json::from_slice(payload)?;
+        println!("{:?}", payload);
+
+        // Check payload is not expired
```
mostlygeek: FYI in [this commit](https://github.com/mozilla-services/go-syncstorage/commit/ba4f08c668d742754906ef950c91bb9c670bf687)  I made go-syncstorage accept an environment configurable timestamp skew.  In reality some client clocks are off by months and every sync request they sent was 403'd ...  I wound up setting this to 1 year just so they could sync.

Cargo.lock
philbooth: > Could be because this is introducing a second version of `ring`...

Looks like it, yep. Thanks for the tip, I'll make sure I sort it before de-wipping this.

src/auth/mod.rs
philbooth: 👍 Thanks for the tip, will do the same here!

src/auth/mod.rs
philbooth: 🎉 Woop woop! It's validating successfully now, thanks for the help with this!

philbooth:
Okay, I think this is close to being reviewable now, there's just one last thing that I'm unsure about.

As @rfk pointed out in https://github.com/mozilla-services/syncstorage-rs/pull/20#discussion_r215828757, the hawk crate depends on a different version of ring to actix. Although that doesn't seem to affect my build on MacOS, it causes linker errors about conflicting symbols in other environments.

Initially I just forked `hawk` to make it use the same version of ring as actix, and that allows the CI build to succeed. But thinking it through, is that setting us up for a maintenance nightmare as we risk the same thing happening again every time we run `cargo update`? @bbangert & @pjenvey, is there some clever Rust/cargo magic you guys know of for managing this kind of situation differently?

An alternative, which I've fleshed out in a separate branch, would be to ditch the `hawk` crate and replace it with some internal logic. That doesn't mean implementing all the crypto ourselves of course, it just means doing a thin layer of parsing and validation on top of external crypto libs that I was already using anyway. Concretely, here's the diff from this PR to what that looks like:

https://github.com/philbooth/syncstorage-rs/compare/hawk-authentication...philbooth:hawk-authentication-internal?expand=1

The exact same tests pass for both branches, so it seems like a low-risk proposal to me (but I'm a hawk/crypto novice so I realise that doesn't count for much).

bbangert:
The hawk Rust crate is maintained by a Mozillian on the Taskcluster team, perhaps we can just ask @djmitche to upgrade the ring dependency? Or does actix use an older version?

philbooth:
> ...perhaps we can just ask @djmitche to upgrade the ring dependency?

Well yep, of course. My question was more about whether there's a mechanism that works longer term than that. I don't know, some kind of name-mangling magic that lets different versions of the lib co-exist in the same binary? If not that's cool, it just seems annoying that we have to synchronise the sub-dependencies of two unrelated external dependencies every time we want to run `cargo update`.

> ...I'd also be happy to merge a PR...

Ref: https://github.com/taskcluster/rust-hawk/pull/10

bbangert:
> Well yep, of course. My question was more about whether there's a mechanism that works longer term than that. I don't know, some kind of name-mangling magic that lets different versions of the lib co-exist in the same binary? If not that's cool, it just seems annoying that we have to synchronise the sub-dependencies of two unrelated external dependencies every time we want to run cargo update.

It depends, sometimes the crate will work just fine depending on what constructs we need to pass it. If we need to pass it something impl a trait from one version of the dep, and the crate uses a different version we'll end up with a difficult to understand error. If we just call it with more basic types though, it should be fine that it uses one version of ring, while we use another elsewhere.... as long as they don't both try and bind to different versions of say, openssl.

An alternative approach would be to see if it makes sense to change the API of the rust-hawk crate more substantially, such that the user must pass in an object impl a crypto trait for the operations hawk needs. This way the user can choose what ring/crypto libs are used, and the hawk crate can avoid being bound to any particular crypto dep version. We could add this as a lower-level API with feature flags, so that we could turn off the ring dep in rust-hawk and pass in our own object.

thomcc:
Possibly worth noting that this would be very useful for the rust-hawk client lib too.

In different places in mozilla/application-services we're using https://github.com/eoger/rust-hawk/tree/use-openssl and https://github.com/eoger/rust-hawk/tree/use-ring-latest. We planned on doing something like that (exposing a trait) before upstreaming these, but never got around to it...

djmitche:
I'd love to see those upstreamed.  I didn't totally understand hawk at the
time I was writing rust-hawk (and anyway I think it's gotten several major
revisions since then).

src/auth/mod.rs
```diff
@@ -0,0 +1,171 @@
...
+            Some(payload.salt.as_bytes()),
+            &settings.master_secret,
+        )?;
+        let token_secret = base64::encode_config(&token_secret, base64::URL_SAFE);
+
+        let request = RequestBuilder::new(method, host, port, path).request();
+        if request.validate_header(
+            &header,
+            &Key::new(token_secret.as_bytes(), &ring::digest::SHA256),
+            Duration::weeks(52),
```
rfk: I think we should add a comment to explain what this `weeks(52)` is doing here, because is could easily raise eyebrows.

src/auth/mod.rs
```diff
@@ -0,0 +1,171 @@
...
+                .get("authorization")
+                .ok_or(AuthError)?
+                .to_str()?,
+            request.method().as_str(),
+            request.path(),
+            request.uri().host().ok_or(AuthError)?,
+            request.uri().port().ok_or(AuthError)?,
+            settings,
+            // Allow a generous amount of client skew for expiry
+            Utc::now().timestamp() as u64 + YEAR,
```
rfk: This `expiry` date applies to the `expires` timestamp that is decoded from inside the Hawk id JSON payload, right?  That timestamp is generated server-side by the tokenserver (and is covered by the HMAC signature check in `extract_and_validate`) so we do not need to allow for client-side clock skew in that check.  We should assume that the servers have well-synchronized clocks.

As written, this would allow tokenserver tokens to be valid for up to a year longer than intended, which we definitely don't want.

(The timestamp from the hawk header *is* client-generated and we can allow some generous clock skew on that one).

philbooth: Oh, my bad. I only added this after the suggestion in https://github.com/mozilla-services/syncstorage-rs/pull/20#discussion_r215839404.

philbooth:
(just in case anyone looks at this PR overnight and wonders what's going on with it, I'm still trying to make the server tests pass with authentication turned on)

philbooth:
The test failures appear to be related to the actix-web upgrade, so I'm waiting to see what comes out of actix/actix-web#508 before spending any more time on this.

philbooth:
> The test failures appear to be related to the actix-web upgrade...

Problem will be fixed in actix-web 0.7.7, so I should have something to review here tomorrow.

src/server/test.rs
```diff
@@ -31,13 +38,63 @@ fn test_endpoint(method: http::Method, path: &str, expected_body: &str) {
...
+        &[0u8; 32],
+    );
+    let token_secret = base64::encode_config(&token_secret, base64::URL_SAFE);
+    let request = RequestBuilder::new(method, "127.0.0.1", 8000, path).request();
+    let credentials = Credentials {
+        id,
+        key: Key::new(token_secret.as_bytes(), &ring::digest::SHA256),
+    };
+    let header = request.make_header(&credentials).unwrap();
+    format!("Hawk {}", header)
```
philbooth: I opted for a manual implementation in the tests here, rather than leaning on code in `HawkPayload`, because I didn't want the possibility of a bug in production code leaking into the tests and preventing them from picking up on it.

src/settings.rs
```diff
@@ -9,6 +11,21 @@ pub struct Settings {
     pub database_pool_max_size: Option<u32>,
     #[cfg(test)]
     pub database_use_test_transactions: bool,
+    pub master_token_secret: Vec<u8>,
+}
+
+impl Default for Settings {
```
philbooth: `Default` is required for the implementation of `FromRequest` on `HawkPayload`.

philbooth:
Also, side-note, the hawk crate includes some logging that shows up when authenticating requests (including in the tests). I've opened taskcluster/rust-hawk#12 to investigate the possibility of turning it off.

src/auth/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+        }
+    }
+
+    fn extract_and_validate(
+        id: &str,
+        master_secret: &[u8],
+        expiry: u64,
+    ) -> AuthResult<HawkPayload> {
+        let decoded_id = base64::decode_config(id, base64::URL_SAFE)?;
+        let payload_length = decoded_id.len() - 32;
```
pjenvey: could decoded_id be < 32 here?

philbooth: I was about to say "no", but then I realised I really mean "not if it's a valid header". But of course we still need to fail gracefully for invalid input so I'll add a length check before this step.

src/settings.rs
pjenvey: The FromRequest Config's a little odd, it always gets the from_request() cfg via Config::default() (so HawkPayloads from_request will always get this bad master_token_secret returned here). I guess it's more of a hardcoded (at build time) configuration

What we should do instead is put Settings (or some subset of it) in the actix_web state and from_request can get it from req.state()

src/auth/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+
+        let payload_length = decoded_id.len() - 32;
+        let payload = &decoded_id[0..payload_length];
+        let signature = &decoded_id[payload_length..];
+
+        let signing_secret = hkdf_expand_32(
+            b"services.mozilla.com/tokenlib/v1/signing",
+            None,
+            master_secret,
+        )?;
```
rfk: In a future refactor, you could consider pre-calculating this and caching it somewhere globally.  The overhead will be pretty minimal in practice though so NBD.

philbooth:
Yeah, so re: the `Settings` stuff, I noticed that too and it's part of what I mentioned in https://github.com/mozilla-services/syncstorage-rs/pull/20#issuecomment-420628861. Putting it in the request state is a fine suggestion and I'm happy to do that, but I also had another idea which I'll expand on now to see what you think.

Over [in fxa-email-service we have a kind of cascade of configs](https://github.com/mozilla/fxa-email-service/blob/master/src/settings/mod.rs#L376-L387) that let us commit different defaults for local dev, tests, CI etc to the repo, without forcing the consuming code to complicate itself with the correct initialisation of `Settings`:

```rust
config.merge(File::with_name("config/default"))?;

let current_env = match env::var("FXA_EMAIL_ENV") {
        Ok(env) => env,
        _ => String::from("dev"),
};
config.merge(File::with_name(&format!("config/{}", current_env)).required(false))?;
 config.set_default("env", "dev")?;

config.merge(File::with_name("config/local").required(false))?;
let env = Environment::with_prefix("fxa_email");
config.merge(env.separator("_"))?;
```

If we did this it would mean that we just need to update those default configs when new settings get added, rather than what I had to do in this PR which was go and make changes in the db tests as well as in the auth tests and in that implementation of the `Default` trait. So the tests can stay a bit cleaner, but also it means that implementation of `Default` could be made useful.

Instead of hard-coded default values, it could invoke something like the above code and let the `Settings` struct concern itself with loading the correct settings for each environment. Then we wouldn't need to include `Settings` in the request state and it would keep things a bit more separated (I think).

Thoughts?

philbooth:
> Thoughts?

But of course, the thing I didn't consider there is that the linked code returns a `Result` and the `Default` trait doesn't, so it would only work if we allowed for the possibility of panics. I guess that means it's a non-starter for the `Default` trait, so I'll do the request thing.

But fwiw, I'm still interested in whether the above proposal is of interest for separating the test code from it's dependency on `Settings` (and providing a sensible default value for `SYNC_DATABASE_URL` for local dev).

philbooth:
I've made a change in 4f01d96c6d32d42636f312775255bfb6469865ae to put `master_token_secret` in `ServerState`. Note that I opted to wrap it in an `Arc`, to avoid calling `clone()` on the entire array each time `ServerState` is instantiated.

Fwiw I also tried giving `ServerState` an explicit lifetime and borrowing `master_token_secret` as a slice instead, but borrowck complained about a conflicting lifetime requirement on the `App` instance (which requires `'static`). Lifetimes are hard. 🤔

(although if you guys can suggest a way to make it work, I'm keen to learn!)

r?

src/handlers.rs
```diff
@@ -9,13 +9,8 @@ use futures::future::{self, Future};
...
 use auth::HawkPayload;
-use db::{params, util::ms_since_epoch, Db, DbError};
-
-/// This is the global HTTP state object that will be made available to all
-/// HTTP API calls.
-pub struct ServerState {
-    pub db: Box<Db>,
-}
+use db::{params, util::ms_since_epoch, DbError};
+use server::ServerState;
```
philbooth: Now that `ServerState` is referenced in a few different places, it seemed to belong more in the `server` module than in `handlers`.

src/auth/mod.rs
philbooth: We're going to need to implement a custom `Deserialize` for this property on `Settings` anyway, so could do it in there.

philbooth: Ref: #26

pjenvey:
The Settings per environ setup sounds good Phil (you can nuke the Settings Default when adding it)

Arc looks good here too. It's not worth setting up lifetimes for data like this setup once (or in the case of actix_web State, at least infrequently) -- the clones or the like just don't matter

Issue #18, opened 2018-08-23
Support a mysql backend


pjenvey:
Create a Db trait that works w/ mysql. This will be used for users who want their own local syncstorage instances and either potentially to speak to Aurora OR as a local test suite database.

We'll use diesel to handle connection management and mapping sql results to rust structs.

Let's prefer raw sql query strings vs the diesel query builder for now. Our next db target might be Auorora but a database that could potentially reuse many of the mysql query strings but wouldn't be supported underneath diesel (it's still up in the air so let's prefer this route while we're figuring it out)

pjenvey:
We're now dropping the raw strings requirement for the most part. Spanner in particular only supports DQL (SELECT statements) and some standard (mostly) DDL (create/alter table). DML INSERT/UPDATE/DELETEs are not supported.

Thus we'd only be able to share SELECT statements at best, and that's not necessarily guaranteed with spanner's many other differences.

pjenvey:
Closed out in #51

PR #22, opened 2018-09-10
feat: beginnings of a mysql backend

pjenvey:
w/ some initial calls and a test suite migrated from the sqlite version

- prefers raw DQL (note: not DML) queries vs diesel's query builder for
potential reuse for other backends (spanner)

- TODO: further fleshing out of the types, likely wanting i64 or wrappers
everywhere (as all spanner has is INT64) -- nor should the db layer be
responsible for conversions from unsigned

Issue #18

### DIFF
--- a/migrations/2018-08-28-010336_init/up.sql
+++ b/migrations/2018-08-28-010336_init/up.sql
@@ -1,0 +1,71 @@
+CREATE DATABASE IF NOT EXISTS `syncstorage` /*!40100 DEFAULT CHARACTER SET latin1 */;
+
+USE `syncstorage`;
+
+-- DROP TABLE IF EXISTS `bso`;
+CREATE TABLE `bso` (
+    `user_id` INT                           NOT NULL,
+    `collection_id` INT                     NOT NULL,
+    `id` VARCHAR(64)                        NOT NULL,
+
+    `sortindex` INT DEFAULT NULL,
+
+    `payload` MEDIUMTEXT                    NOT NULL,
+    `payload_size` INT DEFAULT '0'          NOT NULL,
+
+    -- last modified time in milliseconds since epoch
+    `modified` BIGINT                       NOT NULL,
+    -- expiration in seconds milliseconds since epoch
+    `expiry` BIGINT DEFAULT '3153600000000' NOT NULL,
+
+    PRIMARY KEY (`user_id`, `collection_id`, `id`),
+    KEY `bso_expiry_idx` (`expiry`),
+    KEY `bso_usr_col_mod_idx` (`user_id`, `collection_id`, `modified`)
+) ENGINE=InnoDB DEFAULT CHARSET=latin1;
+
+
+-- DROP TABLE IF EXISTS `collections`;
+CREATE TABLE `collections` (
+    `id` INT PRIMARY KEY      NOT NULL AUTO_INCREMENT,
+    `name` VARCHAR(32) UNIQUE NOT NULL
+) ENGINE=InnoDB AUTO_INCREMENT=100 DEFAULT CHARSET=latin1;
+INSERT INTO collections (id, name) VALUES
+    ( 1, "clients"),
+    ( 2, "crypto"),
+    ( 3, "forms"),
+    ( 4, "history"),
+    ( 5, "keys"),
+    ( 6, "meta"),
+    ( 7, "bookmarks"),
+    ( 8, "prefs"),
+    ( 9, "tabs"),
+    (10, "passwords"),
+    (11, "addons"),
+    (12, "addresses"),
+    (13, "creditcards");
+
+
+-- DROP TABLE IF EXISTS `user_collections`;
+CREATE TABLE `user_collections` (
+    `user_id` INT       NOT NULL,
+    `collection_id` INT NOT NULL,
+    -- last modified time in milliseconds since epoch
+    `modified` BIGINT   NOT NULL,
+    PRIMARY KEY (`user_id`, `collection_id`)
+) ENGINE=InnoDB DEFAULT CHARSET=latin1;
+
+
+-- XXX: based on the go version (bsos is a concatenated blob of BSO jsons separated by newlines)
+-- DROP TABLE IF EXISTS `batches`;
+CREATE TABLE `batches` (
+    `user_id` INT       NOT NULL,
+    `collection_id` INT NOT NULL,
+    `id` varchar(64)    NOT NULL,
+
+    `modified` BIGINT   NOT NULL,
+
+    -- XXX: possibly largetext?
+    `bsos` MEDIUMTEXT   NOT NULL,
+
+    PRIMARY KEY (`user_id`, `collection_id`, `id`)
+) ENGINE=InnoDB DEFAULT CHARSET=latin1;
--- a/src/db/error.rs
+++ b/src/db/error.rs
@@ -1,0 +1,74 @@
+use std::fmt;
+
+use diesel;
+use diesel_migrations;
+use failure::{Backtrace, Context, Fail};
+
+#[derive(Debug)]
+pub struct DbError {
+    inner: Context<DbErrorKind>,
+}
+
+#[derive(Debug, Fail)]
+pub enum DbErrorKind {
+    #[fail(display = "A database error occurred")]
+    Query(#[cause] diesel::result::Error),
+
+    #[fail(display = "A database pool error occurred")]
+    Pool(diesel::r2d2::PoolError),
+
+    #[fail(display = "Error migrating the database")]
+    Migration(diesel_migrations::RunMigrationsError),
+}
+
+impl DbError {
+    pub fn kind(&self) -> &DbErrorKind {
+        self.inner.get_context()
+    }
+}
+
+impl Fail for DbError {
+    fn cause(&self) -> Option<&Fail> {
+        self.inner.cause()
+    }
+
+    fn backtrace(&self) -> Option<&Backtrace> {
+        self.inner.backtrace()
+    }
+}
+
+impl fmt::Display for DbError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        fmt::Display::fmt(&self.inner, f)
+    }
+}
+
+impl From<DbErrorKind> for DbError {
+    fn from(kind: DbErrorKind) -> DbError {
+        Context::new(kind).into()
+    }
+}
+
+impl From<Context<DbErrorKind>> for DbError {
+    fn from(inner: Context<DbErrorKind>) -> DbError {
+        DbError { inner }
+    }
+}
+
+impl From<diesel::result::Error> for DbError {
+    fn from(inner: diesel::result::Error) -> DbError {
+        DbErrorKind::Query(inner).into()
+    }
+}
+
+impl From<diesel::r2d2::PoolError> for DbError {
+    fn from(inner: diesel::r2d2::PoolError) -> DbError {
+        DbErrorKind::Pool(inner).into()
+    }
+}
+
+impl From<diesel_migrations::RunMigrationsError> for DbError {
+    fn from(inner: diesel_migrations::RunMigrationsError) -> DbError {
+        DbErrorKind::Migration(inner).into()
+    }
+}
--- a/src/db/mysql/models.rs
+++ b/src/db/mysql/models.rs
@@ -1,0 +1,352 @@
+use std::{self, ops::Deref};
+
+use diesel::{
+    insert_into,
+    mysql::MysqlConnection,
+    r2d2::{ConnectionManager, Pool, PooledConnection},
+    sql_types::{BigInt, Integer, Text},
+    update, Connection, ExpressionMethods, OptionalExtension,
+};
+use diesel::{sql_query, QueryDsl, RunQueryDsl};
+#[cfg(test)]
+use diesel_logger::LoggingConnection;
+use futures::future;
+
+use super::schema::{bso, collections, user_collections};
+#[cfg(test)]
+use super::test::TestTransactionCustomizer;
+use db::{
+    error::DbError, get_std_collection_id, get_std_collection_name, params, results,
+    util::ms_since_epoch, Db, DbFuture, Sorting,
+};
+use settings::Settings;
+
+embed_migrations!();
+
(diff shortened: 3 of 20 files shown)


src/db/error.rs
```diff
@@ -0,0 +1,74 @@
...
+    Migration(diesel_migrations::RunMigrationsError),
+}
+
+impl DbError {
+    pub fn kind(&self) -> &DbErrorKind {
+        self.inner.get_context()
+    }
+}
+
+impl Fail for DbError {
```
bbangert: Nice!

src/db/mysql/test.rs
```diff
@@ -0,0 +1,419 @@
...
+use env_logger;
+use settings::Settings;
+
+use diesel::{
+    mysql::MysqlConnection,
+    r2d2::{CustomizeConnection, Error as PoolError},
+    Connection,
+};
+
+// 2099 ... somebody else's problem by then (I hope)
```
bbangert: What does this mean?

migrations/2018-08-28-010336_init/up.sql
```diff
@@ -0,0 +1,70 @@
+CREATE DATABASE IF NOT EXISTS `syncstorage` /*!40100 DEFAULT CHARACTER SET latin1 */;
```
rfk: The BSO payload can in theory be arbitrary unicode, since it's JSON.  Should we set an explicit encoding like `utf8mb4` to guard against any weirdness in unicode handling?

migrations/2018-08-28-010336_init/up.sql
```diff
@@ -0,0 +1,70 @@
...
+-- DROP TABLE IF EXISTS `bso`;
+CREATE TABLE `bso` (
+    `user_id` INT                           NOT NULL,
+    `collection_id` INT                     NOT NULL,
+    `id` VARCHAR(64)                        NOT NULL,
+
+    `sortindex` INT DEFAULT NULL,
+
+    `payload` MEDIUMTEXT                    NOT NULL,
+    `payload_size` INT DEFAULT '0'          NOT NULL,
```
rfk: This `payload_size` is a separate column to (in theory) make it quicker to calculate total size of items stored in a collection.  IIRC we don't ever do that in practice, so it may be worth considering whether that optimization still makes sense for us here.

migrations/2018-08-28-010336_init/up.sql
```diff
@@ -0,0 +1,70 @@
...
+    KEY `bso_expiry_idx` (`expiry`),
+    KEY `bso_usr_col_mod_idx` (`user_id`, `collection_id`, `modified`)
+) ENGINE=InnoDB DEFAULT CHARSET=latin1;
+
+
+-- DROP TABLE IF EXISTS `collections`;
+CREATE TABLE `collections` (
+    `id` INT PRIMARY KEY          NOT NULL AUTO_INCREMENT,
+    `name` VARCHAR(32) UNIQUE     NOT NULL,
+    `modified` BIGINT DEFAULT '0' NOT NULL
```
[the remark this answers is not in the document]
rfk: I don't think we need a `modified` timestamp here, because any modification timestamp is going to be per-user rather than global.

migrations/2018-08-28-010336_init/up.sql
```diff
@@ -0,0 +1,70 @@
...
+    (11, "addons"),
+    (12, "addresses"),
+    (13, "creditcards");
+
+
+-- DROP TABLE IF EXISTS `user_collections`;
+CREATE TABLE `user_collections` (
+    `user_id` INT       NOT NULL,
+    `collection_id` INT NOT NULL,
+    `modified` BIGINT   NOT NULL,
```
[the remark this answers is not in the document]
rfk: This should be the last-modified timestamp of the collection as a whole, and should change whenever an item is added, modified or deleted in the collection. It will be greater-than-or-equal-to the MAX() of the modified column for that collection in the bso table, with the greater-than case being because an item was deleted.

FWIW, the python server has a bug here in its handling of deleted collections:

  https://github.com/mozilla-services/server-syncstorage/pull/62

Basically, it tries to calculate the last-modified time of the storage as a whole by doing `SELECT MAX(modified) FROM user_collections WHERE uid = X`.  That's incorrect in the case of a deleted collection, which should cause the last-modified time of the storage as a whole to increase, but won't affect the last-modified time of any remaining collections.

It's clearly an edge-case, because we haven't bothered to actually fix it in the python version.  But for greenfields code it's probably worth doing it right the first time.  I suggest adding a `deleted_at` timestamp field to the `user_collections` table, so that we can explicit track the deletion of collections.  But that can be a follow-up issue if necessary.

migrations/2018-08-28-010336_init/up.sql
```diff
@@ -0,0 +1,70 @@
...
+    PRIMARY KEY (`user_id`, `collection_id`)
+) ENGINE=InnoDB DEFAULT CHARSET=latin1;
+
+
+-- DROP TABLE IF EXISTS `batches`;
+CREATE TABLE `batches` (
+    `user_id` INT       NOT NULL,
+    `collection_id` INT NOT NULL,
+    `id` varchar(64)    NOT NULL,
```
rfk: The "id" here is the batch id, right?  It may be worth naming it "batch_id" or similar to avoid confusion.

src/db/mysql/models.rs
```diff
@@ -0,0 +1,352 @@
...
+pub fn run_embedded_migrations(settings: &Settings) -> Result<()> {
+    let conn = MysqlConnection::establish(&settings.database_url).unwrap();
+    Ok(embedded_migrations::run(&conn)?)
+}
+
+pub struct MysqlDbPool {
+    pool: Pool<ConnectionManager<MysqlConnection>>,
+}
+
+// XXX: to become a db::DbPool trait
```
rfk: How would this trait be used in practice?  It's not obvious to me whether the calling code really wants to know about pooling, or whether the connection pool could be a hidden implementation detail of a particular DB backend.

src/db/mysql/models.rs
```diff
@@ -0,0 +1,352 @@
...
+                .id
+        };
+        Ok(id)
+    }
+
+    pub fn get_collections_sync(
+        &self,
+        params: &params::GetCollections,
+    ) -> Result<results::GetCollections> {
+        sql_query("SELECT id, modified FROM collections WHERE user_id = ?")
```
rfk: ISTM this should be selecting from `user_collections` rather than `collections`.  You could get the names by joining onto the `collections` table, although that may be less efficient when it's only the standard collections for which you already have the names in memory.

src/db/mysql/models.rs
```diff
@@ -0,0 +1,352 @@
...
+            })
+            .collect()
+    }
+
+    pub fn create_collection_sync(&self, name: &str) -> Result<i32> {
+        let collection_id = self.conn.transaction(|| {
+            sql_query("INSERT INTO collections (name, modified) VALUES (?, ?)")
+                .bind::<Text, _>(name)
+                .bind::<BigInt, _>(ms_since_epoch())
+                .execute(&self.conn)?;
```
rfk: This might fail due to two concurrent attempts to insert the same collection; you could automatically recover in that case by detecting the uniqueness constraint violation and reading the inserted row, similar to https://github.com/mozilla-services/server-syncstorage/blob/master/syncstorage/storage/sql/__init__.py#L921

src/db/mysql/models.rs
```diff
@@ -0,0 +1,352 @@
...
+
+pub struct MysqlDb {
+    #[cfg(not(test))]
+    pub(super) conn: PooledConnection<ConnectionManager<MysqlConnection>>,
+    #[cfg(test)]
+    pub(super) conn: LoggingConnection<PooledConnection<ConnectionManager<MysqlConnection>>>,
+}
+
+impl MysqlDb {
+    pub fn get_collection_id_sync(
```
rfk: It's not obvious to me why some of these are suffixed with `_sync` and some are not; what's the significance?

src/db/mysql/models.rs
```diff
@@ -0,0 +1,352 @@
...
+    pub fn put_bso_sync(&self, bso: &params::PutBso) -> Result<results::PutBso> {
+        /*
+        if bso.payload.is_none() && bso.sortindex.is_none() && bso.ttl.is_none() {
+            // XXX: go returns an error here (ErrNothingToDo), and is treated
+            // as other errors
+            return Ok(());
+        }
+        */
+
+        // XXX: consider mysql ON DUPLICATE KEY UPDATE?
```
rfk: Please do :-)

src/db/mysql/models.rs
```diff
@@ -0,0 +1,352 @@
...
+            if exists {
+                update(bso::table)
+                    .filter(bso::user_id.eq(bso.user_id as i32)) // XXX:
+                    .filter(bso::collection_id.eq(&bso.collection_id))
+                    .filter(bso::id.eq(&bso.id))
+                    .set(put_bso_as_changeset(&bso))
+                    .execute(&self.conn)?;
+            } else {
+                let payload = bso.payload.as_ref().map(Deref::deref).unwrap_or_default();
+                let sortindex = bso.sortindex.unwrap_or_default();
```
rfk: It's not obvious to me whether this defaults sortindex to `NULL` or to `0`; I believe `NULL` is the correct behaviour.

src/db/mysql/models.rs
```diff
@@ -0,0 +1,352 @@
...
+        query = match sort {
+            Sorting::Index => query.order(bso::sortindex.desc()),
+            Sorting::Newest => query.order(bso::modified.desc()),
+            Sorting::Oldest => query.order(bso::modified.asc()),
+            _ => query,
+        };
+
+        // fetch an extra row to detect if there are more rows that
+        // match the query conditions
+        query = query.limit(if limit >= 0 { limit + 1 } else { limit });
```
rfk: The docs aren't clear on what happens when `limit < 0`, does it default to "no limit"?

src/db/mysql/test.rs
```diff
@@ -0,0 +1,419 @@
...
+    let payload = "Updated payload";
+    let bso2 = pbso(uid, cid, bid, Some(payload), None, None);
+    db.put_bso_sync(&bso2).unwrap();
+
+    let bso = db.get_bso_sync(&gbso(uid, cid, bid)).unwrap().unwrap();
+    assert_eq!(bso.modified, bso2.modified);
+    assert_eq!(bso.payload, payload);
+    assert_eq!(bso.sortindex, Some(sortindex));
+    // XXX: go version assumes ttl was updated here?
+    //assert_eq!(bso.expiry, modified + ttl);
```
rfk: FWIW I wouldn't expect the ttl to be updated unless you explicitly sent a new TTL in the update.

src/db/mysql/models.rs
bbangert: Agreed, given how funky the Spanner backend is likely to be, pooling and such should prolly be hidden entirely behind the Db trait.

src/db/mysql/models.rs
```diff
@@ -0,0 +1,352 @@
...
+        if ids.len() > 100 {
+            // spec says only 100 ids at a time
+            ids = &ids[0..100];
+        }
+
+        // XXX: convert to raw SQL for use by other backends
+        let mut query = bso::table
+            //.select(bso::table::all_columns())
+            .select((bso::id, bso::modified, bso::payload, bso::sortindex, bso::expiry))
+            .filter(bso::user_id.eq(user_id as i32)) // XXX:
```
bbangert: I assume the // XXX are to maybe do a convert operation with ? so we can ensure we catch casting errors?

migrations/2018-08-28-010336_init/up.sql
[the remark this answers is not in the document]
pjenvey: https://github.com/mozilla-services/syncstorage-rs/issues/23

src/db/mysql/test.rs
```diff
@@ -0,0 +1,419 @@
...
+use settings::Settings;
+
+use diesel::{
+    mysql::MysqlConnection,
+    r2d2::{CustomizeConnection, Error as PoolError},
+    Connection,
+};
+
+// 2099 ... somebody else's problem by then (I hope)
+pub const MAX_TIMESTAMP: i64 = 4070822400000;
```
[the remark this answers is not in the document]
pjenvey: types are still in flight but u64 makes sense here (changed)

migrations/2018-08-28-010336_init/up.sql
pjenvey: payload_size does appear to be utilized in the info/quota/collectionion_usage calls

src/db/mysql/models.rs
pjenvey: Think of it as a basic factory for creating Dbs. I think regardless we require some kind of separation:

Db: directs calls, Send, !Sync
Factory: creates Dbs, Send + Sync

src/db/mysql/models.rs
pjenvey: the higher level db interface (trait) supplies async calls. So the MysqlDb impl of this trait will call sync methods via the [tokio blocking](https://docs.rs/tokio-threadpool/0.1/tokio_threadpool/fn.blocking.html) wrapping call.

src/db/mysql/models.rs
pjenvey: good catch, this is leftover from the port from go which instead defaulted to 0.

src/db/mysql/models.rs
pjenvey: added a TODO

pjenvey:
Will address further things later (likely switching get_bsos limit to u64, table encoding, batches table/architecture is still up in the air).

I'm already intending to have the transactions work like you suggest @rfk -- a lot like the python version: one transaction started (TODO: a transaction() call added to the Db trait) per handler request, having all db calls taking place within it.

modified timestamp likely following the same pattern

migrations/2018-08-28-010336_init/up.sql
rfk: It might be interesting to compare `SELECT SUM(payload_size)` vs `SELECT SUM(LENGTH(payload))` for this purpose to see whether having it as a separate column really provides much value in practice.

To be clear, I don't have any particular objection to keeping it, just wondering whether the extra complexity pays for itself or not.

PR #27, opened 2018-09-19
feat: implement Deserialize for the master_secret setting

### LINKED ISSUE #26: Implement a custom deserializer for Settings::master_token_secret
#20 adds `master_token_secret` to `Settings`, giving it type `Vec<u8>`. Afaik, there is no way to deserialize that without writing a custom deserializer, so we'll need to add one before we can set it from environment variable or config file.

philbooth:
Fixes #26.

The `master_secret` setting is a `Vec<u8>`, for which there is no default deserialization. This change pulls it out to a struct and implements `Deserialize` so that it can be set with a string via environment variable or config file.

That change also represented an opportunity to cache the value of `signing_secret`, which was being freshly calculated on every request even though the value is a constant derived from `master_secret`. So the struct is actually called `Secrets` and has two properties.

This does leave a slightly weird naming mismatch that I wasn't sure what to do with. `master_secret` makes sense from the perspective of setting e.g. a `SYNC_MASTER_SECRET` environment variable, but it looks kind of funny when the code accesses that property internally as a `Secrets` struct with its own `master_secret` and `signing_secret` properties. I tried to minimise the weirdness by passing around the `Secrets` instance and elsewhere referring to that variable/argument as `secrets`. I hope that doesn't seem wilfully confusing.

@bbangert @pjenvey r?

### DIFF
--- a/src/auth/mod.rs
+++ b/src/auth/mod.rs
@@ -18,7 +18,7 @@ use actix_web::{error::ResponseError, http::header::ToStrError, FromRequest, Htt
 use base64::{self, DecodeError};
 use chrono::offset::Utc;
 use hawk::{Error as HawkError, Header as HawkHeader, Key, RequestBuilder};
-use hkdf::{Hkdf, InvalidLength};
+use hkdf::Hkdf;
 use hmac::{
     crypto_mac::{InvalidKeyLength, MacError},
     Hmac, Mac,
@@ -29,7 +29,7 @@ use sha2::Sha256;
 use time::Duration;
 
 use server::ServerState;
-use settings::Settings;
+use settings::{Secrets, Settings};
 
 #[derive(Debug, Deserialize, PartialEq, Serialize)]
 pub struct HawkPayload {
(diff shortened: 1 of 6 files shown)


src/auth/mod.rs
```diff
@@ -127,17 +117,18 @@ impl FromRequest<ServerState> for HawkPayload {
...
 }
 
-fn hkdf_expand_32(info: &[u8], salt: Option<&[u8]>, key: &[u8]) -> AuthResult<[u8; 32]> {
+pub fn hkdf_expand_32(info: &[u8], salt: Option<&[u8]>, key: &[u8]) -> [u8; 32] {
     let mut result = [0u8; 32];
     let hkdf: Hkdf<Sha256> = Hkdf::extract(salt, key);
-    hkdf.expand(info, &mut result)?;
-    Ok(result)
+    // This unwrap will never panic because 32 bytes is a valid size for Hkdf<Sha256>
+    hkdf.expand(info, &mut result).unwrap();
```
philbooth: Potentially contentious change alert: This is 100% safe and will not fail as long as `result` is a `[u8;32]`. But also it's an `unwrap` in production code and that's not cool, I know.

I was hoping to avoid having to propagate the error type out through the call stack, because this function is called more widely (anywhere that calls `Setting::new` would need to implement error handling, even though we know the error won't happen in practice).

philbooth:
The CI failure is rustfmt errors from the recent stable release. I didn't want to pollute this diff with a bunch of unrelated changes so have opened #28 to fix those separately. I'll rebase this after that lands.

PR #35, opened 2018-09-20
feat: load default and env-specific settings from the file system

philbooth:
I wonder whether you guys will think this is over-engineered nonsense, so fwiw I'm totes cool with just closing it if you're thinking YAGNI or whatever.

Anyway, this change pulls out the default settings to `config/default.json` and also tries to load environment-specific settings from `config/$SYNC_ENV.json`. My thinking is that this will be useful when we want to point at a standard GCP region/zone for dev and CI environments.

There's also an attempt to load from `config/local.json`, which is added to `.gitignore` so that there is the option to locally set `master_secret` or a password for the MySQL connection string without accidentally committing them to the repo.

Lastly there is a quick check added to make sure `master_secret` is set to something in prod.

I've retained the `filename` argument (and command-line arg) because I wasn't sure whether they're being used. If they're not and there's no plan to use them we could also just delete.

@bbangert @pjenvey r?

### DIFF
--- a/config/default.toml
+++ b/config/default.toml
@@ -1,0 +1,5 @@
+debug = false
+port = 8000
+database_url = "mysql://root@127.0.0.1/syncstorage"
+database_use_test_transactions = false
+master_secret = ""
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -1,11 +1,11 @@
 //! Application settings objects and initialization
 
+use std::env::var;
+
 use config::{Config, ConfigError, Environment, File};
 use serde::de::{Deserialize, Deserializer};
 
 use web::auth::hkdf_expand_32;
-
-static DEFAULT_PORT: u16 = 8000;
 
 #[derive(Debug, Deserialize)]
 pub struct Settings {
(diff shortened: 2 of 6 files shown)


philbooth:
Also, wasn't sure about the preferred config file format. If YAML or something is more your cup of tea I'm happy to change it.

config/default.toml
```diff
@@ -0,0 +1,5 @@
+debug = false
+port = 8000
+database_url = "mysql://root@127.0.0.1/syncstorage"
+database_use_test_transactions = false
+master_secret = ""
```
bbangert: Do we want to provide a default one or include instructions as a comment about how to do so?

src/settings.rs
```diff
@@ -20,38 +20,45 @@ pub struct Settings {
...
+    ///
+    ///   1. Environment variables: `$SYNC_<UPPERCASE_KEY_NAME>`
+    ///   2. File: `filename` argument
+    ///   3. File: `config/local.toml`
+    ///   4. File: `config/<$SYNC_ENV>.toml`
+    ///   5. File: `config/default.toml`
+    pub fn with_env_and_config_file(filename: Option<String>) -> Result<Self, ConfigError> {
+        let mut config = Config::new();
+
+        config.merge(File::with_name("config/default"))?;
```
bbangert: We don't need to include the .toml?

config/default.toml
philbooth: I'll add a comment. The concern with providing a default is that it actually gets used in production by accident, using the empty string makes it easy to check that hasn't happened in the code.

src/settings.rs
bbangert: Also, this means that the binary will no longer run at all unless you make a default config file in the right spot, right?

So we'll be going from a fairly pure '12-factor app' that requires no files in the docker and can have defaults tweaked with env vars, to a binary that requires files in the right spots. Is this what we want?

philbooth: No, the config crate tries TOML, JSON, YAML, HJSON and INI extensions automatically. I'll add a comment to that effect too.

philbooth: > So we'll be going from a fairly pure '12-factor app' that requires no files in the docker and can have defaults tweaked with env vars, to a binary that requires files in the right spots.

Everything can still be set with env vars fwiw, but...

> Is this what we want?

It sounds like the answer to that is no. I'm happy to close this.

PR #36, opened 2018-09-21
feat: add collection id/name caching at the pool level

pjenvey:
and a couple more db calls that utilize it

Issue #18

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -65,7 +65,7 @@ impl Server {
             let state = ServerState {
                 // TODO: replace MockDb with a real implementation
                 db: Box::new(MockDb::new()),
-                secrets: secrets.clone(),
+                secrets: Arc::clone(&secrets),
             };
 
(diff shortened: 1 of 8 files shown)


src/server/mod.rs
```diff
@@ -65,7 +65,7 @@ impl Server {
             let state = ServerState {
                 // TODO: replace MockDb with a real implementation
                 db: Box::new(MockDb::new()),
-                secrets: secrets.clone(),
+                secrets: Arc::clone(&secrets),
```
philbooth: Purely for my own education, I'm curious about this change.

Was `secrets.clone()` not doing what I thought it was before, or is this about code style, or is there some other underlying difference between the two?

pjenvey: This became the idiomatic way of calling smart pointer methods last year https://github.com/rust-lang/rust/pull/42137 -- intention was to clarify you're calling the typically cheap smart pointer clone vs a potentially heavier inner clone operation

PR #37, opened 2018-09-22
feat: port python sync tween for weave timestamp

### LINKED ISSUE #31: Implement sync middlewares
Sync has a batch of middle-wares that populate common response headers and other basic spec handling. These should be relatively easy to port to actix middleware.

bbangert:
Only the timestamp middleware seems relevant, as the other two in
Python are for handling Python web framework implementation details.
The Rust version will likely require different approaches.

Closes #31

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,4 +1,4 @@
-//! Main application
+//! Sync Storage Server for Sync 1.6
 extern crate actix;
 extern crate actix_web;
 extern crate base64;
--- a/src/web/middleware.rs
+++ b/src/web/middleware.rs
@@ -1,0 +1,145 @@
+//! # Web Middleware
+//!
+//! Matches the [Sync Storage middleware](https://github.com/mozilla-services/server-syncstorage/blob/master/syncstorage/tweens.py) (tweens).
+use actix_web::middleware::{Middleware, Response, Started};
+use actix_web::{error::ErrorInternalServerError, http::header, HttpRequest, HttpResponse, Result};
+use chrono::Utc;
+
+/// Default Timestamp used for WeaveTimestamp middleware.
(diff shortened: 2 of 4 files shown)


src/web/middleware.rs
```diff
@@ -0,0 +1,142 @@
...
+            Ok(Response::Done(resp)) => resp,
+            _ => panic!(),
+        };
+        let weave_hdr = resp
+            .headers()
+            .get("X-Weave-Timestamp")
+            .unwrap()
+            .to_str()
+            .unwrap();
+        assert_ne!(weave_hdr, hts);
```
philbooth: Is there any value in asserting `<` here rather than a plain `!=`?

src/web/middleware.rs
```diff
@@ -0,0 +1,145 @@
...
+            None => return Ok(Response::Done(resp)),
+        };
+
+        let weave_ts = if let Some(val) = resp.headers().get("X-Last-Modified") {
+            let resp_ts = val
+                .to_str()
+                .map_err(|_| ErrorInternalServerError("Unable to extract header value"))?
+                .parse::<f64>()
+                .map_err(|_| ErrorInternalServerError("Invalid last-modified set"))?;
+            if resp_ts > ts.0 {
```
pjenvey: nit: could use cmp::max

bbangert: Actually, cannot, floats only implement PartialOrd, I tried that first. :grinning:

src/main.rs
```diff
@@ -1,4 +1,4 @@
-//! Main application
+//! Sync Storage Server for Sync 1.6
```
rfk: I see what you did there...

Issue #29, opened 2018-09-19
Implement custom datatype for common extractions


bbangert:
A lot of boiler-plate exists for the handlers as they frequently require at leas 3 different extractors. This can be done in a single object for a cleaner API with less boilerplate with its own extractor.

PR #43, opened 2018-10-01
feat: refactor extractors to separate file and begin refinement

bbangert:
As Spanner needs a different identifier, this introduces a HawkIdentifer
object that will provide a NoSql compatible random string identifier for
users as well as a legacy compatible identifier for testing.

Due to lifetime issues and the actix frameworks types, moving all the
validation/extraction logic to extractors didn't work as intended. This
contains a commented out version to begin another PR from to attempt
it.

Issue #29


src/web/extractors.rs
```diff
@@ -0,0 +1,107 @@
...
+
+#[derive(Deserialize, Serialize)]
+pub struct PostCollectionBody {
+    pub id: String,
+    pub sortindex: Option<i32>,
+    pub payload: Option<String>,
+    pub ttl: Option<u32>,
+}
+
+impl From<PostCollectionBody> for params::PostCollectionBso {
```
pjenvey: will this extractor include more validation later?

The PostCollectionBody struct is a duplicate of PostCollectionBso (minus the ability to Deserialize) -- wondering if we can get rid of the duplication

bbangert: No, the validation should really occur in the extractor, this From looks like its just a transform. I'm not entirely sure why its needed at all.

src/web/auth.rs
```diff
@@ -24,6 +24,42 @@ use time::Duration;
...
+/// Represents a user-identifier that is extract from the authentication token
+///
+/// This token should be adapted as needed for the storage system to store data
+/// for the user.
+#[derive(Debug, Default, Deserialize, PartialEq, Serialize)]
+pub struct HawkIdentifier {
+    /// For MySQL database backends as the primary key
+    pub legacy_id: u64,
+    /// For NoSQL database backends that require randomly distributed primary keys
+    pub fxa_id: String,
```
philbooth: I'm a bit confused by the naming here, maybe we can chat about it in this morning's meeting?

But anyway, fwiw there is also an `fxa_uid` property available in the `HawkPayload`, which I didn't bother deserialising because I didn't have an immediate use for. Does this risk confusion with that, or could we even just use that instead?

philbooth: > I'm a bit confused...

Oh wait, ignore me, it's early here and I just woke up. This is all realted to mozilla-services/tokenserver#121 isn't it? I just remembered it's a thing.

PR #45, opened 2018-10-02
feat: add post_bsos and get_collection_counts/sizes

pjenvey:
prefer test fns returning Results (now that rust supports it) vs unwraps

switch the CollectionCache to an Arc for convenience

Issue #18

### DIFF
--- a/src/db/mod.rs
+++ b/src/db/mod.rs
@@ -8,32 +8,32 @@ pub mod params;
 pub mod results;
 pub mod util;
 
-use std::collections::HashMap;
-
 use futures::future::Future;
 
 pub use self::error::DbError;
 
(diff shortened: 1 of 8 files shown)


philbooth:
> prefer test fns returning Results (now that rust supports it) vs unwraps

Noted. Will do the same next time I'm touching some tests.

pjenvey:
> Noted. Will do the same next time I'm touching some tests.

I'm not sure it's always preferable (it's still new, added in rust 1.28 so not used much yet). It does require adding the -> Result<..> signature and Ok(..) at the end of each test.

Seemed like the right tradeoff here though w/ the numerous amount of unwrap()s.

PR #46, opened 2018-10-04
feat: add a BSO query string extractor/validator and foundation

### LINKED ISSUE #29: Implement custom datatype for common extractions

bbangert:
Add's a variety of validation machinery to handle validation errors and
their coercion to appropriate JSON actix-web errors that match the
format of the Cornice error JSON from the Python version.

Closes #29


bbangert:
Now that the scope of the remaining validators/extractors is clearer, I'll open separate issues to port the remaining request validation/extraction routines over.

src/web/extractors.rs
```diff
@@ -2,19 +2,61 @@
...
+use serde_json::from_value;
+use validator::{Validate, ValidationError, ValidationErrors};
 
-use db::{params, Db};
 use server::ServerState;
 use settings::Settings;
-use web::auth::{HawkIdentifier, HawkPayload};
+use web::auth::HawkIdentifier;
+
+const BATCH_MAX_IDS: usize = 100;
```
philbooth: Long term, I guess we want to read the value of this from `Settings`, in case it's been set to something different to the default value?

bbangert: Most likely, yea.

src/web/extractors.rs
```diff
@@ -42,44 +84,305 @@ impl FromRequest<ServerState> for MetaRequest {
...
+
+/// Convenience function to create a `ValidationError` with additional context
+fn request_error(message: &'static str, location: RequestErrorLocation) -> ValidationError {
+    let mut err = ValidationError::new(message);
+    err.add_param("location".into(), &location);
+    err
+}
+
+/// Verifies that the supplied value is >= 0
+fn validate_qs_positive_value<T: PartialOrd + Zero>(ts: T) -> Result<(), ValidationError> {
```
pjenvey: looks good though this module is already becoming pretty large (and with more to come). any pieces make sense to split off?

bbangert: Yup, definitely going to be splitting it up.

PR #56, opened 2018-10-09
refactor: create a db Session w/ a canonical timestamp

### LINKED ISSUE #55: Create a canonical session timestamp
The initial mysql support more matches go-syncstorage's handling of what's essentially a per session 'modified' timestamp.

We should more closely match the python version by creating the timestamp in one place per API call (per SqlStorageSession). Likely living in a similar DbSession struct within the Db impl.

pjenvey:
and quiet a warning under #[cfg(test)]

Closes #55


pjenvey:
merged via #58

PR #58, opened 2018-10-10
feat: push collection id lookup down into the db layer

### LINKED ISSUE #57: Switch db interface back to collection: String (vs collection_id: i32)
Passing collection_ids was somewhat inherited from the go version and additionally I missed the python version's handling of collection_ids via its global collection_id cache. Now that this cache is in place it makes sense to switch the higher level db interface to take collection names as Strings everywhere.

This also saves the handlers an extra future callback they'd need to make by pushing it down into the db layer

pjenvey:
and in a separate commit:

add get_or_create_collection_id for put/post_bso

and ensure collection gets/creates cache their results

Closes #57


bbangert:
So to verify, this makes translation of the collection name to an id something for a lower layer? Doesn't the collection come in as an id?

PR #59, opened 2018-10-11
feat: add lock_for_read/write

### LINKED ISSUE #52: Implement collection-level locking
Build the equivalent of the lock_for_{read/write} context managers and trigger their use for each handler transaction. read's triggered for GET/HEAD requests, otherwise write.

pjenvey:
and basic tests to just ensure it works

Closes #52

### DIFF
--- a/src/db/mysql/models.rs
+++ b/src/db/mysql/models.rs
@@ -18,6 +18,7 @@ use diesel_logger::LoggingConnection;
 use futures::future;
 
 use super::{
+    diesel_ext::LockInShareModeDsl,
     pool::CollectionCache,
     schema::{bso, collections, user_collections},
 };
@@ -49,12 +50,19 @@ pub fn run_embedded_migrations(settings: &Settings) -> Result<()> {
     Ok(embedded_migrations::run(&conn)?)
 }
 
+enum CollectionLock {
+    Read,
+    Write,
+}
+
 /// Per session Db metadata
 pub(super) struct MysqlDbSession {
     /// The "current time" on the server used for this session's operations
     pub(super) timestamp: i64,
     /// Cache of collection modified timestamps per (user_id, collection_id)
     coll_modified_cache: HashMap<(u32, i32), i64>,
+    /// Currently locked collections
+    coll_locks: HashMap<(u32, i32), CollectionLock>,
 }
 
 impl Default for MysqlDbSession {
@@ -62,6 +70,7 @@ impl Default for MysqlDbSession {
         Self {
             timestamp: ms_since_epoch(),
             coll_modified_cache: Default::default(),
+            coll_locks: Default::default(),
         }
     }
 }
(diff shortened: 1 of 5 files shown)


src/db/mysql/test.rs
```diff
@@ -837,3 +838,32 @@ fn delete_storage() -> Result<()> {
...
     Ok(())
 }
+
+#[test]
+fn lock_for_read() -> Result<()> {
+    let mut db = db()?;
+
+    let uid = 1;
+    let coll = "clients";
+    db.conn.transaction_manager().begin_transaction(&db.conn)?;
```
pjenvey: I'll further flesh out the handling of transactions/locking later when we start integrating w/ handlers

src/db/mysql/models.rs
```diff
@@ -49,19 +50,27 @@ pub fn run_embedded_migrations(settings: &Settings) -> Result<()> {
...
+}
+
 /// Per session Db metadata
 pub(super) struct MysqlDbSession {
     /// The "current time" on the server used for this session's operations
     pub(super) timestamp: i64,
     /// Cache of collection modified timestamps per (user_id, collection_id)
     coll_modified_cache: HashMap<(u32, i32), i64>,
+    /// Currently locked collections
+    coll_locks: HashMap<(u32, i32), CollectionLock>,
```
pjenvey: I'm not totally convinced we need this lock metadata as long as we're careful enough about handling the transactions/locking -- which we should be if it's to be managed by actix_web middleware/extractors (like how the python's handled by decorators).

but I've ported it over anyway

src/db/mysql/models.rs
```diff
@@ -90,16 +99,80 @@ impl MysqlDb {
...
+            .filter(user_collections::collection_id.eq(collection_id))
+            .lock_in_share_mode()
+            .first(&self.conn)
+            .optional()?;
+        if let Some(modified) = modified {
+            self.session
+                .coll_modified_cache
+                .insert((user_id, collection_id), modified);
+        }
+        // XXX: who's responsible for unlocking
```
rfk: The Python version implements these as context-managers.  What's the rust equivalent, some sort of guard object in the spirit of `MutexGuard`?

In practice, the python version only ever takes the lock at the start of a request, and releases it at the end.  The actual unlocking is implicit in committing the transaction.  So perhaps there's no need for an explicit unlock, the lock is owned by the session/transaction/whatever and it exists through until that gets committed.

src/db/mysql/models.rs
```diff
@@ -90,16 +99,80 @@ impl MysqlDb {
...
+            .insert((user_id, collection_id), CollectionLock::Read);
+        Ok(())
+    }
+
+    pub fn lock_for_write(&mut self, user_id: HawkIdentifier, collection: &str) -> Result<()> {
+        let user_id = user_id.legacy_id as u32;
+        let collection_id = self.get_or_create_collection_id(collection)?;
+        if let Some(CollectionLock::Read) = self.session.coll_locks.get(&(user_id, collection_id)) {
+            Err(DbError::internal("Can't escalate read-lock to write-lock"))?
+        }
```
rfk: Should this be a `match` so that you can handle the case of already having a `CollectionLock::Write` as well?

src/db/mysql/models.rs
```diff
@@ -90,16 +99,80 @@ impl MysqlDb {
...
+                DbErrorKind::CollectionNotFound => Ok(0),
+                _ => Err(e),
+            })?;
+        // If we already have a read or write lock then it's safe to
+        // use it as-is.
+        if let Some(_) = self.session.coll_locks.get(&(user_id, collection_id)) {
+            return Ok(());
+        }
+
+        // XXX: begin transaction here like python?
```
rfk: If this is called outside of a transaction, it won't do anything useful, so I think we do need to resolve transaction-handling as part of this.  Would it make sense for the `MysqlDbSession` to "own" a transaction that is started when creating the session, and committed (or rolled back) when the session is dropped?  I'm not sure how that plays with ownership throughout the rest of the code here...

src/db/mysql/models.rs
pjenvey: Rust equivalents are either a Guard w/ a Drop (but lacking error handling during the dropping) or wrapping everything (via a closure) in another separate call. Diesel encourages running transactions via a closure.

However I don't know if actix_web provides a hook to call into a handler function allowing us to run the transaction in that manner (I doubt it does) -- hence deferring these details until I begin integrating the db w/ handlers

I suspect these lock methods will end up beginning the transaction (like Python) w/ an accompanying actix_web middleware responsible for committing them (or rolling back + rendering an error response)

src/db/mysql/models.rs
bbangert: This probably makes sense in the prior model where only a single node can access a db at once, and is fine for the MySQL case at least.

src/db/mysql/models.rs
bbangert: A MutexGuard or passing a closure to a function that executes it.

src/db/mysql/models.rs
```diff
@@ -49,19 +50,27 @@ pub fn run_embedded_migrations(settings: &Settings) -> Result<()> {
...
 
+enum CollectionLock {
+    Read,
+    Write,
+}
+
 /// Per session Db metadata
 pub(super) struct MysqlDbSession {
     /// The "current time" on the server used for this session's operations
     pub(super) timestamp: i64,
```
bbangert: Is this in seconds, milliseconds, etc? The timestamp we get from clients is f64, with two digits of precision.

pjenvey: modified is stored in the db as milliseconds https://github.com/mozilla-services/syncstorage-rs/blob/master/migrations/2018-08-28-010336_init/up.sql#L15

the storage layer doesn't have any direct interaction with client supplied timestamps. the modified field on BSOs is always set by the server. the handler layer gets timestamps back from the database and handles the comparisons against the client supplied headers.

Though the one thing the python db code does is make its result timestamps floats (via bigint2ts). I'll worry about that when we start integrating the db into the handlers (I think it might make more sense for the handler to make the conversion/rendering into a float as both mysql/spanner are likely storing int64s)

rfk: Strong 👍 to treating the "float with two-decimal places" encoding as a kind of legacy encoding issue and dealing with it as close to the edge as possible.

bbangert: Should we immediately convert to milliseconds and use that everywhere internally then? Will existing clients handle 3 digits for milliseconds in the Last-(Un)Modified?

rfk: > Will existing clients handle 3 digits for milliseconds in the Last-(Un)Modified?

Probably, but I wouldn't count on it :-/

I was actually remembering incorrectly here, the python server uses a quantized `Decimal` instance (not floats) as its canonical timestamp format internally:

  https://github.com/mozilla-services/server-syncstorage/blob/master/syncstorage/util.py#L13

Integer milliseconds with the lowest digit always zero will probably be an easier format to deal with in rust code.

Issue #48, opened 2018-10-05
Port default validators


bbangert:
The default validators from Python:
```
DEFAULT_VALIDATORS = (
    extract_target_resource,
    extract_precondition_headers,
    extract_query_params,
)
```

Issue #29's last PR brought over extract_query_params and the machinery needed. Closing this requires implementing the other 2 default validators.

bbangert:
I'm considering this closed by the last one, as the collection/item extraction done in one of the prior default validators is not ideal for Rust. It's better to require it when needed for the endpoints, so I'll integrate that into the validator/extractor for handlers that need the collection/item.

PR #61, opened 2018-10-13
feat: add path uid validation to HawkIdentifier extractor

bbangert:
Per extract_target_resource extraction from the Python version, this
add's uid validation to the HawkIdentifier extraction.

Issue #48

### DIFF
--- a/src/web/extractors.rs
+++ b/src/web/extractors.rs
@@ -17,7 +17,7 @@ use validator::{Validate, ValidationError, ValidationErrors};
 
 use server::ServerState;
 use settings::Settings;
-use web::auth::HawkIdentifier;
+use web::auth::HawkPayload;
 
 const BATCH_MAX_IDS: usize = 100;
 
@@ -34,7 +34,7 @@ pub type BsoRequest = (Path<BsoParams>, HawkIdentifier, State<ServerState>);
 #[derive(Deserialize)]
 pub struct UidParam {
     #[allow(dead_code)] // Not really dead, but Rust can't see the deserialized use.
-    uid: String,
+    uid: u64,
 }
 
 #[derive(Deserialize)]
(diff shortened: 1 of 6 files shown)


src/web/extractors.rs
```diff
@@ -425,6 +523,38 @@ mod tests {
...
+        hmac.input(payload.as_bytes());
+        let payload_hash = hmac.result().code();
+        let mut id = payload.as_bytes().to_vec();
+        id.extend(payload_hash.to_vec());
+        let id = base64::encode_config(&id, base64::URL_SAFE);
+        let token_secret = hkdf_expand_32(
+            format!("services.mozilla.com/tokenlib/v1/derive/{}", id).as_bytes(),
+            Some(salt.as_bytes()),
+            &SECRETS.master_secret,
+        );
```
rfk: It feels like you should be able to import some higher-level function to do this, rather than duplicating all the details of how to make a token into the test here.

src/web/extractors.rs
```diff
@@ -500,4 +630,65 @@ mod tests {
             .unwrap();
         assert_eq!(result, PreConditionHeader::IfUnmodifiedSince(32.14));
     }
+
+    #[test]
+    fn valid_header_with_valid_path() {
+        let payload = HawkPayload {
+            expires: Utc::now().timestamp() as f64 + 200000.0,
+            node: "friendly-node".to_string(),
+            salt: "saltysalt".to_string(),
```
rfk: 😆

src/web/extractors.rs
```diff
@@ -500,4 +630,65 @@ mod tests {
...
+            "GET",
+            "/storage/1.5/5/storage/col2",
+            "localhost",
+            5000,
+        );
+        let req = TestRequest::with_state(state)
+            .header("authorization", header)
+            .method(Method::GET)
+            .uri("http://localhost:5000/storage/1.5/5/storage/col2")
+            .param("uid", "5")
```
rfk: Do you have to repeat the `5` here because `TestRequest` won't parse the uid out of the supplised URI?

src/web/extractors.rs
bbangert: Indeed, I didn't find any function that would do it though. I'll probably move this out to be that higher level function elsewhere.

src/web/extractors.rs
bbangert: Yup, TestRequest assumes the params were parsed by the Server and put in the the params object on the request.

PR #63, opened 2018-10-15
chore: deploy docs to gh-pages branch

### LINKED ISSUE #42: Build the docs in CI and push them to the gh-pages branch?
Over in `fxa-email-service` we have a Travis job that builds the docs and then pushes them to the `gh-pages` branch. This means we have automagically refreshing docs available here:

https://mozilla.github.io/fxa-email-service/fxa_email_service/

Are you guys interested in something similar for this repo?

philbooth:
Fixes #42.

The script is only active on `master` so you won't actually see it working if you look at the build log for this PR. But it will work when it's merged, I promise!

@bbangert @pjenvey r?


.travis.yml
```diff
@@ -21,6 +21,8 @@ script:
   - cargo fmt -- --check
   - cargo build
   - cargo test
+  - cargo doc
```
bbangert: This needs the ```--document-private-items``` flag.

philbooth:
> The --no-deps option might help the build time. But deps are nice to have documented for the final published docs (maybe we could only specify it for non master PRs?)

Good point, I'll make it `--no-deps` for pull requests.

PR #67, opened 2018-10-17
wip: use failure crate in src/web/auth.rs

philbooth:
Opening this to get some early feedback because, I'm not sure if the hierarchy I'm moving towards here is the right way to do it or madness.

Essentially I have a top level `ApiError`, which can base its kind on `HawkError` or the existing `DbError`. But that leaves me with two layers of `From` to traverse through when returning from the hawk methods, which may be wrong. There's also a couple of places where the type inference doesn't work and I have to explicitly annotate the interim type. I'll call this out in an inline comment in case it's not clear...

Feedback sought!

### DIFF
--- a/src/web/auth.rs
+++ b/src/web/auth.rs
@@ -2,25 +2,18 @@
 //! Matches the [Python logic](https://github.com/mozilla-services/tokenlib).
 //! We may want to extract this to its own repo/crate in due course.
 
-use std::{
-    error::Error,
-    fmt::{self, Display, Formatter},
-};
-
-use actix_web::{error::ResponseError, http::header::ToStrError, FromRequest, HttpRequest};
-use base64::{self, DecodeError};
+use actix_web::{FromRequest, HttpRequest};
+use base64;
 use chrono::offset::Utc;
-use hawk::{Error as HawkError, Header as HawkHeader, Key, RequestBuilder};
+use hawk::{Header as HawkHeader, Key, RequestBuilder};
 use hkdf::Hkdf;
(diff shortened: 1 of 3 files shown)


src/web/error.rs
```diff
@@ -0,0 +1,143 @@
...
+unsafe impl Sync for ApiErrorKind {}
+unsafe impl Send for HawkError {}
+unsafe impl Sync for HawkError {}
+unsafe impl Send for HawkErrorKind {}
+unsafe impl Sync for HawkErrorKind {}
+
+impl From<HawkErrorKind> for ApiError {
+    fn from(kind: HawkErrorKind) -> Self {
+        let hawk_error: HawkError = Context::new(kind).into();
+        hawk_error.into()
```
philbooth: This is my crazy second layer of `From` madness.

src/web/auth.rs
```diff
@@ -62,11 +55,14 @@ impl HawkPayload {
...
         if &header[0..5] != "Hawk " {
-            return Err(AuthError);
+            return Err(HawkErrorKind::MissingPrefix.into());
         }
 
         let header: HawkHeader = header[5..].parse()?;
-        let id = header.id.as_ref().ok_or(AuthError)?;
+        let id = header.id.as_ref().ok_or_else(|| {
+            let error: ApiError = HawkErrorKind::MissingId.into();
+            error
```
philbooth: Here's an example of where the double-layer of `From` meant I had to explicitly annotate the interim type.

PR #68, opened 2018-10-17
feat: add collection/bso extractors for get/delete requests

### LINKED ISSUE #49: Port get/delete collection/bso validators
Port the Python post validators:
```
POST_VALIDATORS = DEFAULT_VALIDATORS + (
    extract_batch_state,
    parse_multiple_bsos,
    check_for_known_bad_payloads
)
```

bbangert:
Closes #49

### DIFF
--- a/src/web/extractors.rs
+++ b/src/web/extractors.rs
@@ -25,11 +25,8 @@ lazy_static! {
     static ref KNOWN_BAD_PAYLOAD_REGEX: Regex =
         Regex::new(r#"IV":\s*"AAAAAAAAAAAAAAAAAAAAAA=="#).unwrap();
     static ref VALID_ID_REGEX: Regex = Regex::new(r"^[ -~]{1,64}$").unwrap();
+    static ref VALID_COLLECTION_ID_REGEX: Regex = Regex::new(r"^[a-zA-Z0-9._-]{1,32}$").unwrap();
 }
-
-// XXX: Convert these to full extractors.
-pub type GetCollectionRequest = (Path<CollectionParams>, HawkIdentifier, State<ServerState>);
-pub type BsoRequest = (Path<BsoParams>, HawkIdentifier, State<ServerState>);
 
 #[derive(Deserialize)]
 pub struct UidParam {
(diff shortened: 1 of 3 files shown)


src/web/extractors.rs
```diff
@@ -57,7 +54,60 @@ pub struct BsoBody {
...
-/// Request arguments needed for Information Requests
+/// Bso id parameter extractor
+#[derive(Deserialize, Validate)]
+pub struct BsoParam {
+    #[validate(regex = "VALID_ID_REGEX")]
+    pub bso: String,
+}
+
+impl FromRequest<ServerState> for BsoParam {
+    type Config = Settings;
```
pjenvey: maybe we should use type Config = () for most of these since it's ignored?

src/web/extractors.rs
```diff
@@ -584,6 +702,116 @@ mod tests {
...
+        );
+        let req = TestRequest::with_state(state)
+            .header("authorization", header)
+            .method(Method::GET)
+            .uri("http://localhost:5000/storage/1.5/1/storage/tabs")
+            .param("uid", "1")
+            .param("collection", INVALID_COLLECTION_NAME)
+            .finish();
+        let result = CollectionRequest::extract(&req);
+        assert!(result.is_err());
```
pjenvey: Can save a line here and elsewhere via

`let response: HttpResponse = result.unwrap_err().into();`

src/web/extractors.rs
```diff
@@ -584,6 +702,116 @@ mod tests {
...
+            "/storage/1.5/1/storage/tabs",
+            "localhost",
+            5000,
+        );
+        let req = TestRequest::with_state(state)
+            .header("authorization", header)
+            .method(Method::GET)
+            .uri("http://localhost:5000/storage/1.5/1/storage/tabs")
+            .param("uid", "1")
+            .param("collection", INVALID_COLLECTION_NAME)
```
rfk: The name of the test is `test_invalid_bso_request`, but this suggests you're testing an invalid *collection*; should this be adding some sort of invalid bso id param as well?

Issue #38, opened 2018-09-24
Use failure crate everywhere


philbooth:
`api/handlers.rs` and `api/auth.rs` do their error-marshalling manually. The db code uses `failure` and I know it has magic in it for the marshalling stuff, I just never got round to reading about it yet.

philbooth:
Repurposing this as a general "use failure everywhere" issue.

philbooth:
Taking my face off this for now, as I'm not going to get to it tomorrow and am on PTO next week. If it's still available when I get back, I'll happily put my face on it again...

PR #69, opened 2018-10-18
feat: map db and hawk errors to a common top-level error type

philbooth:
Opened as a first step to fixing #38, only affecting the db and hawk stuff for now. I started to map the validation errors too but there's a lot of code there and it'll probably take me a few attempts to get right, so this seemed worth reviewing in the meantime.

As per #67, there is a top-level `ApiError` that `DbError` and `HawkError` get mapped to. Having one top-level error means we only have to `impl ResponseError` on one thing.

I moved `ApiError` up to `src/error.rs` because it contains a couple of macros that are used in both `src/db/error.rs` and `src/web/error.rs`, so I figured it made sense. And yeah, I got busy with the macros again. 😱

I'm happy to remove them if they're not welcome, I just thought they tidied up some boring and repetitive type-related hoops that we have to jump through.

@bbangert @pjenvey r?

### DIFF
--- a/src/web/error.rs
+++ b/src/web/error.rs
@@ -1,0 +1,82 @@
+//! Error types for `web` modules.
+
+use std::fmt;
+
+use actix_web::http::header::ToStrError;
+use base64::DecodeError;
+use failure::{Backtrace, Context, Fail, SyncFailure};
+use hawk::Error as ParseError;
+use hmac::crypto_mac::{InvalidKeyLength, MacError};
+use serde_json::Error as JsonError;
+
+use error::ApiError;
+
+/// An error occurred during HAWK authentication.
+#[derive(Debug)]
+pub struct HawkError {
+    inner: Context<HawkErrorKind>,
+}
+
+/// Causes of HAWK errors.
+#[derive(Debug, Fail)]
+pub enum HawkErrorKind {
+    #[fail(display = "{}", _0)]
+    Base64(#[cause] DecodeError),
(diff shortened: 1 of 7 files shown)


src/web/error.rs
```diff
@@ -0,0 +1,79 @@
...
+from_error!(ParseError, ApiError, HawkErrorKind::Parse);
+
+impl From<HawkErrorKind> for ApiError {
+    fn from(kind: HawkErrorKind) -> Self {
+        let hawk_error: HawkError = Context::new(kind).into();
+        hawk_error.into()
+    }
+}
+
+unsafe impl Sync for HawkErrorKind {}
```
pjenvey: Mentioned this on IRC: here's an alternative to the unsafe marker:

https://gist.github.com/pjenvey/c3b71ce0d684f07a6bf4b28f5e66b81b

src/db/mod.rs
```diff
@@ -36,7 +36,7 @@ lazy_static! {
     };
 }
 
-type DbFuture<T> = Box<Future<Item = T, Error = DbError>>;
+type DbFuture<T> = Box<Future<Item = T, Error = ApiError>>;
```
pjenvey: Can we keep this as Error = DbError for now? I think I'll probably need to keep it this way when I begin plugging in the MysqlDb into Future calls (and maybe ApiError does make more sense here, I'll have to play around with it)

PR #70, opened 2018-10-19
refactor: convert validation errors to use failure

philbooth:
~~Note this branch is based off the branch for #69, so if you want to review both commits in this PR that's cool, or if it's easier to review them separately I can wait until that one is merged and then rebase this against `master`.~~ (rebased now)

Anyway, this is another step on the road to fixing #38. (I guess just the middleware is left after this?)

It updates the validation errors in `src/web/extractors.rs` to map to `ApiError`. In addition to mapping the error types, it also meant bringing across the status code and serialization stuff, hence `ApiError` has grown a `status` property and there are some implementations of `Serialize` introduced to make sure we return all the error details.

Additionally, because a couple of the db errors seemed like 400s rather than 500s, `DbError` has also grown its own `status` property to model that duality. I'm not sure, is that a layering violation, or is it okay for the db layer to know that some of its errors translate to 400s at the API boundary? I'm happy to do something different for this if you have suggestions.

The other contentious part of these changes is that the validation errors aren't strictly uniform. Some of them go through `WithLocation` and some of them are just custom error strings like the `DbError` and `HawkError` cases. This is purely a result of how I progressed through the module, I tackled the `WithLocation` stuff last and had already done the others. If it's important for them all to use the same structure, I can go back in and change them, no worries.

@bbangert @pjenvey r?

### DIFF
--- a/src/web/error.rs
+++ b/src/web/error.rs
@@ -2,13 +2,16 @@
 
 use std::fmt;
 
-use actix_web::http::header::ToStrError;
+use actix_web::{self, http::header::ToStrError};
 use base64::DecodeError;
 use failure::{Backtrace, Context, Fail, SyncFailure};
 use hawk::Error as ParseError;
 use hmac::crypto_mac::{InvalidKeyLength, MacError};
+use serde::ser::{Serialize, SerializeMap, Serializer};
 use serde_json::Error as JsonError;
+use validator;
 
+use super::extractors::RequestErrorLocation;
 use error::ApiError;
 
 /// An error occurred during HAWK authentication.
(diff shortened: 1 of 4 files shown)


philbooth:
(re: the uniformity thing, not sure I made sense, but the difference is that some have a `ValidationErrors` instance and others don't, the ones that don't I just left as plain strings)

src/web/extractors.rs
```diff
@@ -63,19 +60,15 @@ pub struct BsoParam {
...
     fn from_request(req: &HttpRequest<ServerState>, _: &Self::Config) -> Self::Result {
         let bso = Path::<BsoParam>::extract(req)
-            .map_err(|_| {
-                RequestValidationErrors::with_error(
-                    "bso",
-                    "Invalid bso id",
-                    RequestErrorLocation::Path,
-                )
+            .map_err(|e| {
+                ValidationErrorKind::InvalidPathComponent("bso".to_owned(), e.to_string())
```
philbooth: Example of the non-uniform thing, I didn't have a `ValidationErrors` here...

src/web/extractors.rs
```diff
@@ -63,19 +60,15 @@ pub struct BsoParam {
...
             })?.into_inner();
         bso.validate()
-            .map_err(|e| RequestValidationErrors::with_location(e, RequestErrorLocation::Path))?;
+            .map_err(|e| ValidationErrorKind::WithLocation(e, RequestErrorLocation::Path))?;
```
philbooth: ...whereas here I did.

philbooth:
I've pushed another commit to this branch to make it compatible the validation error structure used by cornice.

A validation error now looks like this:

```json
{
  "status": 400,
  "reason": "Bad Request",
  "errors": [
    {
      "description": "Invalid sort option",
      "location": "querystring",
      "name": "sort",
      "value": "whatever"
    }
  ]
}
```

I also tried to preserve some level of uniformity in the serialised db error structure, even though they obviously don't have all of the same rich error information. So a db error now looks like this:

```json
{
  "status": 500,
  "reason": "Internal Server Error",
  "errors": [
    "A database error occurred: foo"
  ]
}
```

It's not perfect, the errors array contains strings rather than objects, but then the Python code doesn't make these uniform either IIUC so I'm assuming that's fine?

src/web/error.rs
```diff
@@ -80,3 +112,71 @@ impl From<ParseError> for ApiError {
...
+                seq.serialize_element(&SerializedValidationError {
+                    description,
+                    location,
+                    name: name.as_ref().map(|name| &**name),
+                    value: None,
+                })?;
+            }
+
+            ValidationErrorKind::FromValidationErrors(ref errors, ref location) => {
+                errors.clone().field_errors().iter().fold(
```
philbooth: Oh, almost forgot! This `clone()` is because of a borrowck error I was getting.

`errors.field_errors()` takes `self`, but I want to call it on a reference to `errors` so was getting the dreaded "cannot move out of borrowed content". Doing the extra `clone()` feels like it should be unnecessary though, is there something else I should be doing?

pjenvey: ref https://github.com/Keats/validator/issues/65

pjenvey: I think this block would be improved by just a couple for loops. Since we're not really accumulating anything via fold (if I'm reading it right, we just use the accumulator to break out of the fold after an error)

```
                for (field, field_errors) in errors.clone().field_errors() {
                    for field_error in field_errors {
                        seq.serialize_element(&SerializedValidationError {
                            description: &field_error.code,
                            location,
                            name: Some(field),
                            value: field_error.params.get("value"),
                        })?;
                    }
                }
```

(I think that's equivalent?)

Issue #65, opened 2018-10-16
Integrate db layer and handlers w/ a Middleware/Extractor


pjenvey:
The Handlers should be provided a Db instance created from the pool (probably from an extractor). An associated middleware will ensure an appropriate transaction + lock is established and committed (or rolled back) at the end of the request

PR #71, opened 2018-10-24
feat: initial integration of db w/ handlers

pjenvey:
adds a transaction/collection lock middleware + a Db extractor

tokio_threadpool::blocking won't work under actix, so:

- manage our own ThreadPool, satisfying our async interface via cloning the
MysqlDb before sending it over to the ThreadPool
- unfortunately requires wrapping the inner state in Arc (and RefCell for the
mutable parts) and an accompanying unsafe Send and

Issue #65

### DIFF
--- a/src/db/mysql/models.rs
+++ b/src/db/mysql/models.rs
@@ -1,8 +1,9 @@
 #![allow(proc_macro_derive_resolution_fallback)]
 
-use std::{self, collections::HashMap, ops::Deref, sync::Arc};
+use std::{self, cell::RefCell, collections::HashMap, ops::Deref, sync::Arc};
 
 use diesel::{
+    connection::TransactionManager,
     delete,
     dsl::max,
     expression::sql_literal::sql,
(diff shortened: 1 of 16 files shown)


src/db/mysql/models.rs
```diff
@@ -573,16 +649,56 @@ impl MysqlDb {
...
 }
 
 impl Db for MysqlDb {
+    fn commit(&self) -> DbFuture<()> {
+        let db = self.clone();
+        Box::new(
+            self.thread_pool
+                .spawn_handle(lazy(move || future::result(db.commit_sync()))),
+        )
+    }
```
rfk: For my own edification, why can't the `sync_db_method!` macro be used with `commit` and `rollback` here?

src/web/auth.rs
```diff
@@ -124,15 +124,15 @@ impl FromRequest<ServerState> for HawkPayload {
     ///
     /// Not hugely useful, all of the configurable settings
     /// can be found on the [request state](../../server/struct.ServerState.html) instead.
-    type Config = Settings;
+    type Config = ();
```
philbooth: :tada:

src/db/mysql/models.rs
philbooth: Looks like it's the lack of a `params` argument. Macros get messy pretty quickly when you try to make things optional I've found, because you have to use `*` so you're also introducing repeating groups to the macro body. The simpler the better.

PR #72, opened 2018-10-26
feat: plugin and cleanup various trait methods into MysqlDb:

pjenvey:
- kills get_collection_id (only used within the db layer now)
- simplify uid_data! (just alias HawkIdentifier)
- phrasing

Issue #65

### DIFF
--- a/src/db/mock.rs
+++ b/src/db/mock.rs
@@ -30,20 +30,14 @@ impl MockDb {
 
 macro_rules! mock_db_method {
     ($name:ident, $type:ident) => {
-        fn $name(&self, _params: &params::$type) -> DbFuture<results::$type> {
-            Box::new(future::ok(results::$type::default()))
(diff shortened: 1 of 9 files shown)


src/db/mock.rs
```diff
@@ -30,20 +30,14 @@ impl MockDb {
 
 macro_rules! mock_db_method {
     ($name:ident, $type:ident) => {
-        fn $name(&self, _params: &params::$type) -> DbFuture<results::$type> {
-            Box::new(future::ok(results::$type::default()))
+        mock_db_method!($name, $type, results::$type);
+    };
+    ($name:ident, $type:ident, $result:ty) => {
```
philbooth: Ah, of course, thus disproving my comment yesterday about having to use `*`. I hadn't thought about multiple match arms, nice one.

philbooth:
Should we just disable the docs script in CI until it's fixed?

PR #73, opened 2018-10-26
refactor: map to failure errors in X-Weave-Timestamp middleware

### LINKED ISSUE #38: Use failure crate everywhere

philbooth:
Fixes #38 (I think?)

I was not feeling inspired when naming the new error type so ended up having a minor collision with actix's `ResponseError` trait. If one of you has a better name suggestion that avoids the need to alias that, let me know.

@bbangert @pjenvey r?

### DIFF
--- a/src/web/middleware.rs
+++ b/src/web/middleware.rs
@@ -4,7 +4,6 @@
 use std::rc::Rc;
 
 use actix_web::{
-    error::ErrorInternalServerError,
     http::{header, Method},
     middleware::{Middleware, Response, Started},
     FromRequest, HttpRequest, HttpResponse, Result,
@@ -13,6 +12,7 @@ use chrono::Utc;
 use futures::{future, Future};
 
 use db::{params, Db};
+use error::{ApiError, ApiErrorKind};
 use server::ServerState;
 use web::extractors::{CollectionParam, HawkIdentifier};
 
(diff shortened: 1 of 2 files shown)


pjenvey:
I consider these 3 more of a blanket "internal server error" as they're the result parsing data we were responsible for putting in the outgoing headers (so they should never happen).

I like having a blanket InternalServer error that takes a string for these. In the rare case it happens we can uniquely identify it (in a log or even in the outgoing error response) -- being a string its uniqueness isn't enforced by the type system -- but good enough for something that should never happen anyway *fingers crossed*

(a lot like what was already there, but we could have our own version of actix's internal server error or at the very least a custom json response for it if that's possible)

philbooth:
> I like having a blanket InternalServer error that takes a string for these.

Understood, I can do that...

philbooth:
> Understood, I can do that...

🧙

src/web/middleware.rs
```diff
@@ -43,9 +43,20 @@ impl<S> Middleware<S> for WeaveTimestamp {
...
             let resp_ts = val
                 .to_str()
-                .map_err(|_| ErrorInternalServerError("Unable to extract header value"))?
-                .parse::<f64>()
-                .map_err(|_| ErrorInternalServerError("Invalid last-modified set"))?;
+                .map_err(|e| {
+                    let error: ApiError = ApiErrorKind::Internal(format!(
+                        "Invalid X-Last-Modified response header: {}",
+                        e
+                    )).into();
```
bbangert: I kind of prefer this over turbofishing it.

pjenvey:
Sorry for the lag I didn't see the last update

PR #80, opened 2018-11-02
fix: prefix urls w/ 1.5 and fix auth's host/port handling

pjenvey:


pjenvey:
FYI @philbooth I forgot to mention here, we ran the python test suite against syncstorage-rs yesterday and these 2 were the first bugs to shake out (with the fixes making now pass 1 instead of 0 tests out of 60 :sunglasses: )

Issue #74, opened 2018-10-29
Cleanup db types


pjenvey:
Cleanup many of the various XXX comments around type signatures around the mysql db

- [x] fixing modified/'sizes' to become u64s
- [ ] fix the user_id column to be 64 bits (it's a BigInt in token server yet a regular Int in server-syncstorage for some reason)
- [x] may want to revisit handling of u64s passed into the database: they must be casted to i64s so we may want to bounds check them beforehand **(fixed for user_id)**

pjenvey:
AFAICT item #1 (fixing modified/'sizes' to become u64s) already happened?

Extracted #2 into #470, so closing.

PR #82, opened 2018-11-06
feat: refactor sync timestamp

bbangert:
Co-authored by: Phil Jenvey <pjenvey@underboss.org>

Issue #74

### DIFF
--- a/src/db/util.rs
+++ b/src/db/util.rs
@@ -1,6 +1,105 @@
+#![allow(proc_macro_derive_resolution_fallback)]
+
 use chrono::offset::Utc;
+use diesel::{
+    backend::Backend,
+    deserialize::{self, FromSql},
+    sql_types::BigInt,
+};
+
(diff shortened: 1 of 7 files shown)


src/db/util.rs
```diff
@@ -1,6 +1,105 @@
...
+                }
+            }).map(|v: f64| (v * 1_000f64) as u64)
+            .map(SyncTimestamp::from_milliseconds)
+            .map_err(|_| "Invalid value")
+    }
+
+    /// Create a `SyncTimestamp` from an i64
+    ///
+    /// Only called from the db module
+    pub(super) fn from_i64(val: i64) -> Result<Self, DbError> {
```
philbooth: Ah, it's a shame the `TryFrom` trait is nightly-only, it would have been perfect for this right?

bbangert: Yup, I was wishing that as we added it.

PR #83, opened 2018-11-09
feat: add precondition check middleware and handler headers

### LINKED ISSUE #78: Create precondition header middleware
To avoid doing queries we shouldn't, the precondition header check should be done in middleware in front of the db middlware, to verify whether the query has any chance of succeeding.

bbangert:
Most handler functions now have the appropriate header added. The
precondition middleware returns the request early if the modified
header and database differ.

Closes #78

### DIFF
--- a/src/db/mysql/models.rs
+++ b/src/db/mysql/models.rs
@@ -29,7 +29,7 @@ use db::{
     util::SyncTimestamp,
     Db, DbFuture, Sorting,
 };
-use web::extractors::HawkIdentifier;
+use web::extractors::{BsoQueryParams, HawkIdentifier};
 
 no_arg_sql_function!(last_insert_id, Integer);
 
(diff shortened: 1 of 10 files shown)


src/web/middleware.rs
```diff
@@ -131,6 +131,50 @@ impl Middleware<ServerState> for DbTransaction {
...
+        let collection = CollectionParam::from_request(&req, &())
+            .ok()
+            .map(|v| v.collection);
+        let bso = BsoParam::from_request(&req, &()).ok().map(|v| v.bso);
+        let fut = db
+            .extract_resource(user_id, collection, bso)
+            .and_then(move |resource_ts| {
+                let (status, modified) = match precondition {
+                    PreConditionHeader::IfModifiedSince(header_ts) if resource_ts <= header_ts => {
+                        (StatusCode::NOT_MODIFIED, header_ts)
```
pjenvey: Unless I'm mistaken python's rendering resource_ts as the X-Last-Modified in these cases, not header_ts?

src/web/middleware.rs
```diff
@@ -131,6 +131,50 @@ impl Middleware<ServerState> for DbTransaction {
...
+                    }
+                    PreConditionHeader::IfUnmodifiedSince(header_ts) if resource_ts > header_ts => {
+                        (StatusCode::PRECONDITION_FAILED, header_ts)
```
pjenvey: same here

src/db/mysql/models.rs
```diff
@@ -433,7 +432,7 @@ impl MysqlDb {
         Ok(results::GetBsos {
             bsos,
             more,
-            offset: next_offset,
+            offset: next_offset as i64,
```
pjenvey: nit I forgot to revert this cast, it's no longer needed

src/web/middleware.rs
bbangert: This is purely the status code for the request, not the header.

pjenvey: I meant the second item of these tuple set as X-Last-Modified below

src/db/mysql/models.rs
bbangert: Fixed.

src/web/middleware.rs
bbangert: Ah! Right.

PR #85, opened 2018-11-13
fix: implement batches on top of mysql

### LINKED ISSUE #51: Implement batch db calls
Solidify the batches table schema (base it off the python version's schema or can we use the go version's simpler design?) and implement its calls.

pjenvey:
like go-syncstorage, we just store the bsos' json in a text column,
concatenating batches together w/ newlines

also:

- fix load_collection_names handling of cache hits
- get_bsos limit handling after b066fba9
- kill the old schema files I forgot to remove

Closes #51, #18


src/db/mysql/batch.rs
```diff
@@ -0,0 +1,272 @@
...
+
+pub fn create(db: &MysqlDb, params: params::CreateBatch) -> Result<results::CreateBatch> {
+    let user_id = params.user_id.legacy_id as i32;
+    let collection_id = db.get_collection_id(&params.collection)?;
+    let timestamp = db.timestamp().as_i64();
+    // Sync timestamps are in seconds and quantized to two decimal places, so
+    // when we convert one to a bigint in milliseconds, the final digit is
+    // always zero. But we want to use the lower digits of the batchid for
+    // sharding writes via (batchid % num_tables), and leaving it as zero would
+    // skew the sharding distribution.
```
rfk: Do you in fact do this sharding across multiple "batches" tables?

src/db/mysql/batch.rs
```diff
@@ -0,0 +1,272 @@
...
+pub fn validate(db: &MysqlDb, params: params::ValidateBatch) -> Result<bool> {
+    let user_id = params.user_id.legacy_id as i32;
+    let collection_id = db.get_collection_id(&params.collection)?;
+    let exists = batches::table
+        .select(sql::<Integer>("1"))
+        .filter(batches::user_id.eq(&user_id))
+        .filter(batches::collection_id.eq(&collection_id))
+        .filter(batches::id.eq(&params.id))
+        .get_result::<i32>(&db.conn)
+        .optional()?;
```
rfk: This feels like it should also check whether the batch has expired.

src/db/mysql/batch.rs
```diff
@@ -0,0 +1,272 @@
...
+    Ok(exists.is_some())
+}
+
+pub fn append(db: &MysqlDb, params: params::AppendToBatch) -> Result<()> {
+    // XXX: this is possibly an upsert
```
rfk: Not sure I agree about an upsert here, you should always know whether you have a new batch or an existing batch.

src/db/mysql/batch.rs
```diff
@@ -0,0 +1,272 @@
...
+    let user_id = params.user_id.legacy_id as i32;
+    let collection_id = db.get_collection_id(&params.collection)?;
+    let bsos = bsos_to_batch_string(&params.bsos)?;
+    // XXX: also update expiry?
+    update(batches::table)
+        .filter(batches::user_id.eq(&user_id))
+        .filter(batches::collection_id.eq(&collection_id))
+        .filter(batches::id.eq(&params.id))
+        .set(batches::bsos.eq(batches::bsos.concat(&bsos)))
+        .execute(&db.conn)?;
```
rfk: If the batch has already expired, should this update it or return an error?  Also if the specified batch id does not exist, this should likely return an error rather updating zero rows.

src/db/mysql/batch.rs
```diff
@@ -0,0 +1,272 @@
...
+    Ok(())
+}
+
+pub fn commit(db: &MysqlDb, params: params::CommitBatch) -> Result<results::CommitBatch> {
+    let bsos = batch_string_to_bsos(&params.batch.bsos)?;
+    db.post_bsos_sync(params::PostBsos {
+        user_id: params.user_id,
+        collection: params.collection,
+        bsos,
+    })
```
rfk: Ideally we should atomically delete the batch after commiting it, to ensure that it can't be committed twice.

src/db/mysql/batch.rs
```diff
@@ -0,0 +1,272 @@
...
+
+        let bsos = vec![
+            postbso("b0", Some("payload 0"), Some(10), None),
+            postbso("b1", Some("payload 1"), Some(1000000000), None),
+        ];
+        append(&db, ab(uid, coll, id, bsos))?;
+
+        let batch = get(&db, gb(uid, coll, id))?.unwrap();
+        assert_ne!(batch.bsos, "".to_owned());
+        // XXX: expiry updated?
```
rfk: I don't have strong opinions about this, but my first instance is that expiry should not be updated.  Otherwise there's no finite bound on the ultimate lifetime of a batch, a misbehaving client could slow-drip add items to a batch and keep it alive indefinitely.

src/db/mysql/batch.rs
```diff
@@ -0,0 +1,272 @@
...
+        .filter(batches::user_id.eq(&user_id))
+        .filter(batches::collection_id.eq(&collection_id))
+        .filter(batches::id.eq(&params.id))
+        .execute(&db.conn)?;
+    Ok(())
+}
+
+pub fn commit(db: &MysqlDb, params: params::CommitBatch) -> Result<results::CommitBatch> {
+    let bsos = batch_string_to_bsos(&params.batch.bsos)?;
+    db.post_bsos_sync(params::PostBsos {
```
rfk: Defering the work here to an existing `post_bsos_sync` method should give you a lot of edge-cases for free, such as proper handling of partial updates and ttl bumping.  I like it!  Handling those was a source of some complexity in the python code.

src/db/mysql/batch.rs
pjenvey: Nope, so I removed this comment and the associated code

src/db/mysql/batch.rs
pjenvey: Sounds good, I've added expiry checks here and other places

src/db/mysql/models.rs
```diff
@@ -665,13 +670,31 @@ impl MysqlDb {
...
+    }
+
     pub fn timestamp(&self) -> SyncTimestamp {
         self.session.borrow().timestamp
     }
 
     #[cfg(test)]
-    pub(super) fn set_timestamp(&self, timestamp: i64) {
-        self.session.borrow_mut().timestamp = SyncTimestamp::from_i64(timestamp).unwrap();
+    pub(super) fn with_delta<T, E, F>(&self, delta: i64, f: F) -> std::result::Result<T, E>
```
rfk: This is a neat little abstraction!

PR #91, opened 2018-11-21
Two tiny tweaks for Sentry

### LINKED ISSUE #90: Don't send 4xx errors to Sentry
This just bit us [in fxa-email-service](https://github.com/mozilla/fxa-email-service/pull/245) and the same mistake I made there, I made here:

https://github.com/mozilla-services/syncstorage-rs/blob/master/src/error.rs#L136

We should only send `500` errors to Sentry.

philbooth:
Fixes #90.

These are both things I had to fix recently in fxa-email-service, I presume we want to do the same here.

* In b2935f5, I stop capturing every error in Sentry because it's too noisy. Currently the only 500 errors we return are from the db error module, so I moved the capture into there.

* In 9d99c01, I include debug symbols in the release builds so that Sentry can include line numbers in stack traces.

@bbangert @pjenvey r?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,10 @@ authors = [
   "Phil Jenvey <pjenvey@underboss.org>",
 ]
 
+[profile.release]
+# Enables line numbers in Sentry
+debug = 1
+
 [dependencies]
 actix = "0.7"
 actix-web = "0.7.7"
--- a/src/db/error.rs
+++ b/src/db/error.rs
@@ -65,7 +65,13 @@ impl From<Context<DbErrorKind>> for DbError {
             _ => StatusCode::INTERNAL_SERVER_ERROR,
         };
 
-        Self { inner, status }
+        let error = Self { inner, status };
+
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -7,6 +7,10 @@ authors = [
   "Phil Jenvey <pjenvey@underboss.org>",
 ]
 
+[profile.release]
+# Enables line numbers in Sentry
+debug = true
```
pjenvey: Took me a while to track this down, we came across it for [autopush-rs](https://github.com/mozilla-services/autopush-rs/issues/77). There's also a not well documented debug value of either 1 (or 2, but true is the equivalent of 2). This value's passed along to rustc -C debuginfo:

```
    -C            debuginfo=val -- debug info emission level, 0 = no debug info, 1 = line tables only, 2 = full debug info with variable and type information
```

`debug = 1` sounds ideal for release builds, just producing the line tables

There's been some reports of slowdowns with full release = 2 which may or may not be fixed:

https://github.com/rust-lang/rust/issues/49227

debug = 1 should at the very least reduce the binary size

PR #95, opened 2018-11-27
fix: return a 415 for unsupported content-type headers

### LINKED ISSUE #76: Supported other statuses in ValidationErrorKind::FromDetails
Not all validation errors should be a 400, in this case we want a 415 Unsupported Media Type:

https://github.com/mozilla-services/syncstorage-rs/pull/75/files#diff-9de2327cbb67a90931e98ff70715da9dR205

philbooth:
Fixes #76, returning a 415 when the `Content-Type` header is wrong.

@pjenvey @bbangert r?

### DIFF
--- a/src/web/error.rs
+++ b/src/web/error.rs
@@ -2,7 +2,7 @@
 
 use std::fmt;
 
-use actix_web::http::header::ToStrError;
+use actix_web::http::{header::ToStrError, StatusCode};
 use actix_web::Error as ActixError;
 use base64::DecodeError;
 use failure::{Backtrace, Context, Fail, SyncFailure};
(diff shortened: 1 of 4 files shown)


src/web/error.rs
```diff
@@ -97,7 +98,21 @@ impl From<Context<HawkErrorKind>> for HawkError {
 
 impl From<Context<ValidationErrorKind>> for ValidationError {
     fn from(inner: Context<ValidationErrorKind>) -> Self {
-        Self { inner }
+        let status = match inner.get_context() {
+            ValidationErrorKind::FromDetails(ref _description, ref location, ref name) => {
+                if *location == RequestErrorLocation::Header
+                    && name.is_some()
+                    && name.as_ref().unwrap() == "Content-Type"
```
bbangert: Can't you compare without the unwrap?
```rust
if *location == RequestErrorLocation::Header && name == Some("Content-Type") {
```

philbooth: Ah, didn't know that, thanks.

PR #104, opened 2018-11-29
feat: fix extractors, middleware for e2e tests

### LINKED ISSUE #97: Improper response status codes
Several requests return a 400 instead of 415, or a 400 instead of 404, or a 200 instead of a 400.

bbangert:
Extractors now use proper limits from the state object. BSO handling
during deserialization/validation matches Python more closely.

Closes #97

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -7,7 +7,26 @@ use failure::{Backtrace, Context, Fail};
 use serde::ser::{Serialize, SerializeMap, SerializeSeq, Serializer};
 
 use db::error::{DbError, DbErrorKind};
-use web::error::{HawkError, ValidationError};
+use web::error::{HawkError, ValidationError, ValidationErrorKind};
+use web::extractors::RequestErrorLocation;
+
+/// Legacy Sync 1.1 error codes, which Sync 1.5 also returns by replacing the descriptive JSON
+/// information and replacing it with one of these error codes.
+#[derive(Serialize)]
+enum WeaveError {
+    /// Unknown error
+    UnknownError = 0,
+    /// Illegal method/protocol
+    IllegalMethod = 1,
+    /// Json parse failure
+    MalformedJson = 6,
+    /// Invalid Weave Basic Object
+    InvalidWbo = 8,
+    /// User over quota
+    OverQuota = 14,
+    /// Size limit exceeded
+    SizeLimitExceeded = 17,
+}
 
 /// Common `Result` type.
 pub type ApiResult<T> = Result<T, ApiError>;
@@ -75,6 +94,24 @@ impl ApiError {
         }
         false
     }
+
+    fn weave_error_code(&self) -> WeaveError {
+        match self.kind() {
+            ApiErrorKind::Validation(ver) => match ver.kind() {
+                ValidationErrorKind::FromDetails(ref _description, ref location, name) => {
+                    let name = name.clone().unwrap_or("".to_owned());
+                    if *location == RequestErrorLocation::Body
+                        && ["bso", "bsos"].contains(&name.as_str())
+                    {
+                        return WeaveError::InvalidWbo;
+                    }
+                    WeaveError::UnknownError
+                }
+                _ => WeaveError::UnknownError,
+            },
+            _ => WeaveError::UnknownError,
+        }
+    }
 }
 
 impl From<ApiError> for HttpResponse {
@@ -98,10 +135,16 @@ impl From<Context<ApiErrorKind>> for ApiError {
 
 impl ResponseError for ApiError {
     fn error_response(&self) -> HttpResponse {
+        // To return a descriptive error response, this would work. We do not
+        // unfortunately do that so that we can retain Sync 1.1 backwards compatibility
+        // as the Python one does.
+        // HttpResponse::build(self.status).json(self)
+        //
+        // So instead we translate our error to a backwards compatible one
         HttpResponse::build(self.status)
             .if_true(self.is_conflict(), |resp| {
                 resp.header("Retry-After", RETRY_AFTER.to_string());
-            }).json(self)
+            }).json(self.weave_error_code() as i32)
     }
 }
 
--- a/src/web/extractors.rs
+++ b/src/web/extractors.rs
@@ -2,6 +2,7 @@
 //!
 //! Handles ensuring the header's, body, and query parameters are correct, extraction to
 //! relevant types, and failing correctly with the appropriate errors if issues arise.
+use std::collections::HashMap;
 use std::str::FromStr;
 
 use actix_web::http::header::{HeaderValue, CONTENT_TYPE};
@@ -13,7 +14,7 @@ use actix_web::{
 use futures::{future, Future};
 use regex::Regex;
 use serde::de::{Deserialize, Deserializer, Error as SerdeError};
-use serde_json;
+use serde_json::Value;
 use validator::{Validate, ValidationError};
 
 use db::{util::SyncTimestamp, Db, Sorting};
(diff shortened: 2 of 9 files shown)


src/error.rs
```diff
@@ -75,6 +94,31 @@ impl ApiError {
...
+                ValidationErrorKind::FromDetails(ref description, ref location, name) => {
+                    let name = name.clone().unwrap_or("".to_owned());
+                    if *location == RequestErrorLocation::Body
+                        && ["bso", "bsos"].contains(&name.as_str())
+                    {
+                        if description.starts_with("Invalid JSON") {
+                            // Python's tests feel differently about when its invalid JSON
+                            // In Python's case, its a JSON deserialization failure, not merely
+                            // invalid JSON per the desired output.
+                            return WeaveError::InvalidWbo;
```
pjenvey:
```suggestion
                            return WeaveError::MalformedJson;
```

src/web/extractors.rs
```diff
@@ -1232,12 +1309,16 @@ mod tests {
             assert_eq!(response.status(), 400);
             let body = extract_body_as_str(&response);
 
+            assert_eq!(body, "0");
```
pjenvey: this method's emitting unused var warnings now, can you s/error_/_error_/g it

src/web/extractors.rs
```diff
@@ -46,18 +42,49 @@ pub struct UidParam {
...
     pub sortindex: Option<i32>,
     pub payload: Option<String>,
+    #[validate(custom = "validate_body_bso_ttl")]
     pub ttl: Option<u32>,
 }
 
+impl BatchBsoBody {
+    /// Function to convert valid raw JSON BSO body to a BatchBsoBody
+    fn from_raw_bso(val: &Value) -> Option<BatchBsoBody> {
+        let map = match val.as_object() {
```
pjenvey: nit can replace this whole block w/ `let map = val.as_object()?;`

src/web/extractors.rs
```diff
@@ -46,18 +42,49 @@ pub struct UidParam {
...
+            Some(m) => m,
+            None => return None,
+        };
+        // Verify all the keys are valid
+        let valid_keys = ["id", "sortindex", "payload", "ttl"];
+        for key_name in map.keys() {
```
pjenvey: nit this could be a one liner any()

src/web/extractors.rs
```diff
@@ -120,55 +150,67 @@ impl FromRequest<ServerState> for BsoBodies {
...
+
+            for bso in bsos {
+                // Error out if its not a JSON mapping type
+                if !bso.is_object() {
+                    return future::err(make_error());
+                }
+                match BatchBsoBody::from_raw_bso(&bso) {
+                    Some(b) => {
+                        // Is this record too large? Deny if it is.
+                        let payload_size = b.payload.clone().map(|v| v.len()).unwrap_or_default();
```
pjenvey: avoid cloning these 'larger' payload chunks
```suggestion
                        let payload_size = b.payload.as_ref().map(|v| v.len()).unwrap_or_default();
```

src/web/extractors.rs
```diff
@@ -220,7 +264,16 @@ impl FromRequest<ServerState> for BsoBody {
                     Some("bso".to_owned()),
                 ).into();
                 err.into()
-            }).and_then(|bso: Json<BsoBody>| {
+            }).and_then(move |bso: Json<BsoBody>| {
+                // Check the max payload size manually with our desired limit
+                if bso.payload.clone().map(|s| s.len()).unwrap_or_default() > max_payload_size {
```
pjenvey:
```suggestion
                if bso.payload.as_ref().map(|s| s.len()).unwrap_or_default() > max_payload_size {
```

src/web/extractors.rs
```diff
@@ -358,6 +411,23 @@ pub struct CollectionPostRequest {
     pub bsos: BsoBodies,
 }
 
+impl CollectionPostRequest {
+    /// Return bsos that failed to parse as appropriate failed hashmap
+    pub fn failed_bsos(&self) -> HashMap<String, String> {
+        // Leave room for any new errors when attempting to save them
+        let mut result = HashMap::with_capacity(self.bsos.invalid.len() + self.bsos.valid.len());
+        for raw in self.bsos.invalid.iter() {
```
pjenvey:
```suggestion
        for raw in &self.bsos.invalid {
```

src/web/extractors.rs
```diff
@@ -358,6 +411,23 @@ pub struct CollectionPostRequest {
...
+            if let Some(id) = raw
+                .get("id")
+                .and_then(|v| v.as_str().map(|s| s.to_string()))
```
pjenvey: moving the final to_string() call into the insert line below is a little clearer I think

src/error.rs
bbangert: While this would make sense, it would not match the Python tests, which consider this to be an Invalid Wbo due to the reason I doc'd here.

src/web/extractors.rs
bbangert: I guess I'm feeling dense as I'm not seeing what would be needed to make that happen.

src/web/extractors.rs
```diff
@@ -358,6 +408,20 @@ pub struct CollectionPostRequest {
...
 }
 
+impl CollectionPostRequest {
+    /// Return bsos that failed to parse as appropriate failed hashmap
+    pub fn failed_bsos(&self) -> HashMap<String, String> {
+        // Leave room for any new errors when attempting to save them
+        let mut result = HashMap::with_capacity(self.bsos.invalid.len() + self.bsos.valid.len());
+        for raw in &self.bsos.invalid {
+            if let Some(id) = raw.get("id").and_then(|v| v.as_str()) {
+                result.insert(id.to_string(), "invalid json".to_string());
```
pjenvey: we're cheating on the failed error messages here, python shows a more detailed message about what failed to validate

After a quick look I don't see any of our sync client code utilizing these messages, so maybe it's not too important, but I could see the messages being potentially useful for debugging (if they were echoed to a debug console log message e.g.). Very least we should log an issue about it?

src/error.rs
pjenvey: I think I misunderstood the comment then. do we even need the block then? it returns InvalidWbo in both cases

src/web/extractors.rs
```diff
@@ -120,55 +147,67 @@ impl FromRequest<ServerState> for BsoBodies {
...
+                }
+                match BatchBsoBody::from_raw_bso(&bso) {
+                    Some(b) => {
+                        // Is this record too large? Deny if it is.
+                        let payload_size = b.payload.as_ref().map(|v| v.len()).unwrap_or_default();
+                        if payload_size <= max_payload_size
+                            && total_payload_size + payload_size <= max_post_bytes
+                        {
+                            valid.push(b);
+                            total_payload_size += payload_size;
```
pjenvey: might not matter much but let's match the python and increase this tally *before* checking for overflow (this block). I can imagine an odd case where a certain payload overflows the total -- but the total tally wouldn't increase, then maybe the next payload in the sequence is small enough to be fit

src/web/extractors.rs
rfk: > After a quick look I don't see any of our sync client code utilizing these messages

I believe any useful error messages get converted into simple numeric codes before they reach the client anyway, so *shrug*

pjenvey: > I believe any useful error messages get converted into simple numeric codes before they reach the client anyway, so _shrug_

everywhere *except* here in POST's failed mapping, the API docs even have an example of it: https://mozilla-services.readthedocs.io/en/latest/storage/apis-1.5.html#individual-collection-interaction

src/web/extractors.rs
```diff
@@ -120,55 +142,95 @@ impl FromRequest<ServerState> for BsoBodies {
...
+                if !bso.is_object() {
+                    return future::err(make_error());
+                }
+                // Save all id's we get, check for missing id, or duplicate.
+                let bso_id = if let Some(id) = bso.get("id").and_then(|s| s.as_str()) {
+                    let id = id.to_string();
+                    if bso_ids.contains(&id) {
+                        return future::err(
+                            ValidationErrorKind::FromDetails(
+                                "Input BSO has no ID".to_owned(),
```
pjenvey: copy pasta

Issue #86, opened 2018-11-15
Port batch validator


pjenvey:
Originally part of #49: port the `extract_batch_state` validator from Python to an actix-web extractor

pjenvey:
closed in #108

PR #108, opened 2018-12-12
fix: add the batch extractor + handler

### LINKED ISSUE #105: Port batch handler
Dependent on its extractor (#86)

pjenvey:
Closes #105, #86

and a cargo fmt in a separate commit

### DIFF
--- a/src/web/extractors.rs
+++ b/src/web/extractors.rs
@@ -2,8 +2,7 @@
 //!
 //! Handles ensuring the header's, body, and query parameters are correct, extraction to
 //! relevant types, and failing correctly with the appropriate errors if issues arise.
-use std::collections::HashMap;
-use std::str::FromStr;
+use std::{self, collections::HashMap, str::FromStr};
 
 use actix_web::http::header::{HeaderValue, ACCEPT, CONTENT_TYPE};
 use actix_web::{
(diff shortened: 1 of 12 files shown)


src/web/extractors.rs
```diff
@@ -729,14 +753,113 @@ impl FromRequest<ServerState> for BsoQueryParams {
...
         })?;
         Ok(params)
     }
 }
 
+#[derive(Debug, Default, Clone, Deserialize, Validate)]
+#[serde(default)]
+pub struct BatchParams {
+    pub batch: Option<String>,
+    #[validate(custom = "validate_commit")]
```
bbangert: Is this query or header params? I was including a bit in the validate function to indicate where they came from so the error would properly reflect the location. It'd be nice if the validate was capable of passing args as a partial so we could avoid funcs for each location.

src/web/extractors.rs
```diff
@@ -869,6 +996,17 @@ fn validate_body_bso_ttl(ttl: u32) -> Result<(), ValidationError> {
     Ok(())
 }
 
+/// Verifies the batch commit field is valid
+fn validate_commit(commit: &String) -> Result<(), ValidationError> {
```
bbangert: validate_qs_commit maybe

src/web/extractors.rs
```diff
@@ -1314,6 +1453,66 @@ mod tests {
         assert_eq!(result.bsos.invalid.len(), 2);
     }
 
+    #[test]
+    fn test_valid_collection_batch_post_request() {
```
bbangert: A test for invalid?

PR #129, opened 2019-04-26
feat: Add cargo-audit

### LINKED ISSUE #127: Run cargo-audit in CI
This project does not appear to be using cargo-audit (https://github.com/RustSec/cargo-audit) - that’s strongly recommended in order to be warned about dependencies with security vulnerabilities.

jrconlin:
* Includes a core .circleci config file
  TODO: Needs dockerhub config settings if we want to publish this
      to dockerhub
  * Local unit tests are currently commented out of the config, but
    can be enabled later.
* Some quick clippy fixes. Clippy does NOT like post macro formatted
  code.

Closes #127


jrconlin:
The clippy bits are there because I originally was going to include `cargo fmt` and `cargo clippy` in the test suite. There are still some clippy warnings being generated, but I didn't want to address them because I wasn't sure if there was some reliant behaviors that might break because of lossy behaviors.

bug-mirror:
cc @pjenvey

PR #130, opened 2019-05-03
feat add dockerflow endpoints

### LINKED ISSUE #123: Support dockerflow endpoints
Implement the dockerflow `__{version,lbheartbeat,heartbeat}__` endpoints.

channelserver's already implemented them under actix.

jrconlin:
Closes #123

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -3,7 +3,7 @@
 use std::sync::Arc;
 
 use actix::{System, SystemRunner};
-use actix_web::{http, middleware::cors::Cors, server::HttpServer, App};
+use actix_web::{http, middleware::cors::Cors, server::HttpServer, App, HttpResponse};
 //use num_cpus;
 
 use db::{mysql::MysqlDbPool, DbError, DbPool};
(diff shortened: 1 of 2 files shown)


src/server/mod.rs
```diff
@@ -96,7 +127,14 @@ impl Server {
                 port,
             };
 
-            build_app(state)
+            let dfstate = ServerState {
```
jrconlin: I really hope I'm being dumb here and there's a better way to send the ServerState to the second App instance. I tried a few things, but they got ugly really fast. Setting up the pool and Arc clones seems wasteful, since the app never uses them.

pjenvey: ServerState could support a clone with some effort. It would have to manually call Db's box_clone.

The dockerflow APIs don't need ServerState currently, so you could not pass it at all (apps can have [different state via .boxed()](https://actix.rs/docs/application/#combining-applications-with-different-state)) but at some point `__heartbeat__` should be doing a database check anyway..

pjenvey: actix-web 1.0 might change this anyway? It doesn't have an App::with_state any more (unfortunately the user guide hasn't been updated yet https://github.com/actix/actix-website/tree/master/content/docs)

jrconlin: Ah, yeah, things might change a lot with the actix update. This PR was mostly just trying to get ahead of that update so things weren't joined and the PR was even uglier. 

I agree that `__heartbeat__` should do a bit more than just return Ok, but again, that could be added later.

PR #140, opened 2019-05-30
feat: Add docker config

### LINKED ISSUE #122: Build docker images
- Setup a Dockerfile according to Dockerflow guidelines to build a syncstorage-rs image
- Setup circleci to build it

See autopush-rs's or channelserver's Dockerfile and the [Dockerflow repo's](https://github.com/mozilla-services/Dockerflow) examples for good bases to start from.

jrconlin:
WIP

Closes #122

NOTES:
* Probably want to run `docker-compose up` on this to get both the dummy database and app running
* ~App is complaining that "System is not running". I'm guessing that this is a different error, but will investigate more.~
* Added note about making damn sure that `actix-web` and `actix` versions are compatible when updating. This means no dependbot or auto-updates for a while... 😭


src/server/mod.rs
```diff
@@ -137,7 +137,7 @@ impl Server {
 
             vec![build_app(state), build_dockerflow(dfstate)]
         })
-        .bind(format!("127.0.0.1:{}", settings.port))
+        .bind(format!("0.0.0.0:{}", settings.port))
```
pjenvey: is the new settings.host actually used anywhere (looks like it should be here)?

jrconlin: oh for... forgot to commit that change.

src/server/mod.rs
```diff
@@ -137,7 +137,7 @@ impl Server {
 
             vec![build_app(state), build_dockerflow(dfstate)]
         })
-        .bind(format!("127.0.0.1:{}", settings.port))
+        .bind(format!("{}:{}", settings.host, settings.port))
```
pjenvey: FYI `.bind((settings.host.as_ref(), settings.port))` also works, there's a ToSocketAddr's impl for that kind of 2 item tuple

jrconlin: Yeah, wasn't sure if "127.0.0.1" was something explicitly set, otherwise would have done something like that.

Issue #113, opened 2018-12-13
Allow text/plain json bodies


pjenvey:
According to test_storage::test_set_item_input_formats this is a bw-compat case we need to support:

```python
        # Unless we use text/plain, which is a special bw-compat case.
        self.app.put(self.root + '/storage/col2/TEST', body, headers={
            "Content-Type": "text/plain"
        })
```

philbooth:
@pjenvey I don't see `test_set_item_input_formats` in the server-syncstorage repo. Am I missing something?

philbooth:
> Am I missing something?

Only my own foolishness it seems. Forgot that I deleted a load of tests in my local copy, sorry, ignore the above.

philbooth:
This is because of the implementation of [`actix_web::dev::JsonBody`](https://actix.rs/api/actix-web/stable/actix_web/dev/struct.JsonBody.html), which fails if content type is not `application/json`.

`JsonBody::from_request` is called internally when we do the following in `BsoBody::from_request`:

```rust
let fut = <Json<BsoBody>>::from_request(req, &config)
```

The docs mention [`actix_web::dev::JsonConfig`](https://actix.rs/api/actix-web/stable/actix_web/dev/struct.JsonConfig.html) as a way to configure the built-in JSON extraction, but you can't disable the header check that way. So it looks like we'll need to deserialize the payload ourselves instead, I'll aim to wrap that up in something that looks the same as the actix's `Json`.

philbooth:
Fwiw, I've opened a PR to see if we can fix this on the actix side: actix/actix-web#637

jrconlin:
Looks like this was merged in https://github.com/actix/actix-web/pull/637.

Is this still a relevant bug? Should we open a new one to remove our hack and apply a "proper" fix using the actix lib?

pjenvey:
> Is this still a relevant bug? Should we open a new one to remove our hack and apply a "proper" fix using the actix lib?

Yep, this is still an issue (we don't have a hack for it, test_set_item_input_formats still fails in the same spot).

The fix required is to begin utilizing the JsonConfig::content_type method that pb added to actix-web. The example in that PR's docs/tests appear to be exactly what we need (allowing `mime.type_() == mime::TEXT && mime.subtype() == mime::PLAIN`

PR #141, opened 2019-06-05
bug: Allow text/plain, application/json, or no content type to be accepted

jrconlin:
Closes 115

### DIFF
--- a/src/web/extractors.rs
+++ b/src/web/extractors.rs
@@ -286,11 +286,9 @@ impl FromRequest<ServerState> for BsoBody {
         }
         let mut config = JsonConfig::default();
         let max_request_size = req.state().limits.max_request_bytes as usize;
-        config.limit(max_request_size).content_type(|ct| {
-            ct.to_string() == "".to_owned()
-                || ct == mime::TEXT_PLAIN
-                || ct == mime::APPLICATION_JSON
-        });
+        config
+            .limit(max_request_size)
+            .content_type(|ct| ct == mime::TEXT_PLAIN);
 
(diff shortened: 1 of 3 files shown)


pjenvey:
> Closes 115

it's #113

src/web/extractors.rs
```diff
@@ -285,7 +286,11 @@ impl FromRequest<ServerState> for BsoBody {
         }
         let mut config = JsonConfig::default();
         let max_request_size = req.state().limits.max_request_bytes as usize;
-        config.limit(max_request_size);
+        config.limit(max_request_size).content_type(|ct| {
+            ct.to_string() == "".to_owned()
+                || ct == mime::TEXT_PLAIN
```
pjenvey: nit: I think we only need this line (at least according to test_storage.py). The mime crate is strongly typed and doesn't appear to parse an empty string content_type. Also JsonConfig always allows application/json

pjenvey:
Unfortunately @philbooth's content_type feature somehow was lost in [actix-web 1.0](https://docs.rs/actix-web/1.0.0/actix_web/web/struct.JsonConfig.html), so we'll also need to redo that PR :(

src/web/extractors.rs
jrconlin: Honestly, I was being super paranoid by that point. My test worked when I didn't include `mime::APPLICATION_JSON`, but I had no idea if that was a fluke or not.

jrconlin:
Also filed https://github.com/actix/actix-web/issues/894

PR #143, opened 2019-06-07
bug: exclude empty-ish application/newline content.

### LINKED ISSUE #110: Reject application/newlines body of "\n"
We currently accept this expected failure from test_storage::test_batch_empty_commit:

```python
 testEmptyCommit("application/newlines", "\n", status=400)
```

jrconlin:
Closes #110

### DIFF
--- a/src/web/extractors.rs
+++ b/src/web/extractors.rs
@@ -158,6 +158,12 @@ impl FromRequest<ServerState> for BsoBodies {
             // Get all the raw JSON values
             let bsos: Vec<Value> = if newlines {
                 let mut bsos = Vec::new();
+                if body.len() < 3 {
+                    match body.clone().as_str() {
+                        "\n" | "{}" | "[]" => return future::err(make_error()),
+                        _ => (),
+                    };
+                }
                 for item in body.lines() {
                     // Skip any blanks
                     if item == "" {
--- a/src/web/extractors.rs
+++ b/src/web/extractors.rs
@@ -18,7 +18,7 @@ use serde::{
     de::{Deserializer, Error as SerdeError},
     Deserialize, Serialize,
 };
-use serde_json::{json, Value};
+use serde_json::Value;
 use validator::{Validate, ValidationError};
 
 use crate::db::{util::SyncTimestamp, Db, DbError, DbErrorKind, Sorting};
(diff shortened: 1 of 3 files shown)


src/web/extractors.rs
```diff
@@ -158,13 +158,20 @@ impl FromRequest<ServerState> for BsoBodies {
             // Get all the raw JSON values
             let bsos: Vec<Value> = if newlines {
                 let mut bsos = Vec::new();
+                if body.len() < 2 && body.as_str() == "\n" {
+                    return future::err(make_error());
+                }
                 for item in body.lines() {
                     // Skip any blanks
                     if item == "" {
```
pjenvey: Looking at the Python code, I don't see a similar line for skipping blank lines, looking here and the test's commit in particular:

https://github.com/mozilla-services/server-syncstorage/blob/master/syncstorage/views/validators.py#L253

https://github.com/mozilla-services/server-syncstorage/commit/6f50ebe3f

I think we need to remove it: the is_object() check below ensures we get a mapping type and following that the check for an id field ensures nothing empty can be passed along.

I think the empty string case produces no lines and we pass thru returning a 200

jrconlin: I'd agree, only for that specific test case, we never hit `bso.is_object()`. No BSOs are generated, so the for loop is skipped.

The problem is that "\n" is functionally equivalent to "" as far as the rust code is concerned, only in the test case we are given, they're not. The above patch pretty much just handles the case where we're passed a new line.

jrconlin: I'd be fine dropping the test case from the python side, but if we want 100% compatibility, we probably should handle this case too.

pjenvey: > I'd agree, only for that specific test case, we never hit `bso.is_object()`. No BSOs are generated, so the for loop is skipped.

ah you're right but in rust what I'm seeing is:

- "".lines() -> empty iterator, passes thru
- "\n".lines() -> iterator of one item of "": errors on serde_json::from_str
- "{}".lines() -> errors on id check
- "[]" ->  errors on is_object check

so we can match the python almost exactly (no blank line check) -- the one difference in the python from above is python `"".split("\n")` produces a one item iterator of "", hence python needing the `if request.body` check, which rust doesn't

Issue #146, opened 2019-06-11
DB Lock Failures not being properly reported


jrconlin:
The `middleware::DbTransaction` is not reporting when a database lock failure occurs. This cascades to cause a 500 error when there's no request extension for web/extractors `FromRequest<ServerState> for Box<dyn Db>::from_request()` fails with a generic "Unexpected Db[sic] error".

The 500 response is technically correct, but the reason can be easily lost.

pjenvey:
This is another middleware related issue,

- The `DbTransaction` middleware may trigger ConflictErrors on lock_for_write, by design
- `DbTransaction::start` erroring out appears to prevent `PreconditionCheck::start` from running, but `PreconditionCheck::response` still runs.. (not what I would have expected)
- Regardless, PreconditionCheck start or response (actually the Box<dyn Db> extractor) always assumes it has a Db available, erroring out with an ErrorInternalServerError("Unexpected DB error") if not

So ultimately the ErrorInternalServerError propagates out as a 500. We'll fix this when updating the middleware to actix 1.0

tublitzed:
@jrconlin is this one still in progress?

jrconlin:
@tublitzed Nope.

Closed with https://github.com/mozilla-services/syncstorage-rs/pull/185

PR #149, opened 2019-06-13
feat: Add debugging to track database locking failures.

jrconlin:
Database locking failures can cause intermittent integration test
failures. The errors appear to get lost on the way out and eventually
get resolved to "Unexpected Db error". This logs the locking failure at
the time of occurrence, at least.

Issue #146

### DIFF
--- a/src/web/middleware.rs
+++ b/src/web/middleware.rs
@@ -129,7 +129,6 @@ impl Middleware<ServerState> for DbTransaction {
                     req.extensions_mut().insert((db, in_transaction));
                     future::ok(None)
                 })
-                .map_err(Into::into)
             })
             .map_err(Into::into);
         Ok(Started::Future(Box::new(fut)))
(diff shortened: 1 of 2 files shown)


src/web/middleware.rs
```diff
@@ -128,6 +129,7 @@ impl Middleware<ServerState> for DbTransaction {
                     req.extensions_mut().insert((db, in_transaction));
                     future::ok(None)
                 })
+                .map_err(Into::into)
```
pjenvey: this doesn't look needed (and we could use less of them oh god they're everywhere)

jrconlin: Honestly, I'm still not sure what's eating the error. That was more a hail Mary. Now that the clippy stuff is in place, I'll refactor and strip a lot of these out. 

My hunch is whatever is processing the future is eating the error being generated in the `future:err()` and we might be missing a handler at a higher level. I'm still trying to get my head around the futures stuff (my apologies for being dense), that's why this is still a WIP.

PR #159, opened 2019-06-27
Feat/test

jrconlin:
Tweaks for e2e tests
* return correct BSO error code
* return 404 instead of 400 
* fix collection id regex
* fix tests


src/error.rs
```diff
@@ -114,7 +114,9 @@ impl ApiError {
                     }
                     WeaveError::UnknownError
                 }
-                _ => WeaveError::UnknownError,
+                ValidationErrorKind::FromValidationErrors(ref _err, ref _location) => {
```
pjenvey: This seems potentially too broad (all FromValidationErrors are InvalidWbo)? In the python it's only emitted error["location"] = "body" and name in ("bso", "bsos") covered above

src/web/error.rs
```diff
@@ -117,6 +117,11 @@ impl From<Context<ValidationErrorKind>> for ValidationError {
                     _ => StatusCode::BAD_REQUEST,
                 }
             }
+            ValidationErrorKind::FromDetails(ref _description, ref location, Some(ref _name))
+                if *location == RequestErrorLocation::Path =>
```
pjenvey: Same here (too broad) but I think this one can be further restricted to && name == "bso" || name == "collection"

Otherwise it hits another ::Path case (triggered by the HawkIdentifier)

src/web/extractors.rs
```diff
@@ -37,7 +37,7 @@ lazy_static! {
     static ref KNOWN_BAD_PAYLOAD_REGEX: Regex =
         Regex::new(r#"IV":\s*"AAAAAAAAAAAAAAAAAAAAAA=="#).unwrap();
     static ref VALID_ID_REGEX: Regex = Regex::new(r"^[ -~]{1,64}$").unwrap();
-    static ref VALID_COLLECTION_ID_REGEX: Regex = Regex::new(r"^[a-zA-Z0-9._-]{1,32}$").unwrap();
+    static ref VALID_COLLECTION_ID_REGEX: Regex = Regex::new(r"^[a-zA-Z0-9\._\-]{1,32}$").unwrap();
```
pjenvey: nit: not actually necessary to escape these 2 here (in a character set and when - is the last one)

src/web/error.rs
jrconlin: Well, there's no name, but I agree that I can safely limit to location == body.

pjenvey: The `Some(ref _name)` I meant

Issue #160, opened 2019-06-27
Adapt mysql batch tests to the generic Db trait


pjenvey:
Followup to #156:

I won't tackle the batch tests in #156 because they live in a separate file. Though they need to be moved/adapted over as well.

pjenvey:
We should also cleanup the duplicated helpers when moving these and consider [small improvements](https://github.com/mozilla-services/syncstorage-rs/pull/161#discussion_r299232868) to the helpers as well

jrconlin:
@pjenvey Was this addressed by #199?

PR #161, opened 2019-06-27
test: convert the db_tests to async/await 

pjenvey:
and switch them to work against the generic Db trait

I attempted to keep the commits split to help review and also maintain the test history (don't squash this!)


db-tests/src/db_tests.rs
```diff
@@ -0,0 +1,1079 @@
...
+fn gbso(user_id: u32, coll: &str, bid: &str) -> params::GetBso {
+    params::GetBso {
+        user_id: hid(user_id),
+        collection: coll.to_owned(),
+        id: bid.to_owned(),
+    }
+}
+
+#[allow(clippy::too_many_arguments)]
+fn gbsos(
```
jrconlin: nit: and `get_bsos(...)`

db-tests/src/db_tests.rs
```diff
@@ -0,0 +1,1079 @@
...
+        coll,
+        bid,
+        Some("initial value"),
+        Some(sortindex),
+        Some(ttl),
+    );
+    db.put_bso(bso1).compat().await?;
+
+    let payload = "Updated payload";
+    let bso2 = pbso(uid, coll, bid, Some(payload), None, None);
```
jrconlin: nit; purely for readability, but I wonder if enclosing these sub tests in blocks might be a good idea. It would help with possible cross contamination.

db-tests/src/db_tests.rs
```diff
@@ -0,0 +1,1079 @@
...
+        limits: ServerLimits::default(),
+        master_secret: Secrets::default(),
+    };
+
+    // XXX: let pool = syncstorage::db:pool_from_settings(&settings)?;
+    let pool = MysqlDbPool::new(&settings)?;
+    pool.get().compat().await
+}
+
+fn pbso(
```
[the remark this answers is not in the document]
pjenvey: These helpers are only used on the previous line of their associated calls (making them self explanatory?). We could collapse their calls into the same line so it's even more obvious and that would reduce the local vars in the test (maybe helping avoid the potential cross pollination)

They're currently duplicated for the batch tests so I'll consider these changes when cleaning those up later (#160)

PR #162, opened 2019-06-27
bug: Allow expired auth for `/info/collection` endpoints

### LINKED ISSUE #152: /info/collections should grant access to expired tokens
The python has a special case for expired tokens, granting them access to /info/collections (acl=expired_token_acl):

```python
@info_timestamps.get(accept="application/json", renderer="sync-json",
                     acl=expired_token_acl)
@default_decorators
def get_info_timestamps(request):
   ...
```

This causes test_accessing_info_collections_with_an_expired_token to fail (though it's not normally ran: it's always skipped with the test suite setting self.distant to True)

jrconlin:
Closes #152


src/web/auth.rs
```diff
@@ -156,22 +156,28 @@ impl FromRequest<ServerState> for HawkPayload {
             80
         };
 
+        let path = request
+            .uri()
+            .path_and_query()
+            .ok_or(HawkErrorKind::MissingPath)?;
+        let expiry = if path.path().ends_with("/info/collections") {
```
pjenvey: I was going to suggest a stricter check by looking into the request.resource() -- the ResourceDef's pattern inside of that -- but the resource method appears to be removed in actix-web 1.

I wonder if there's another way to determine what route triggered the request in 1.0

jrconlin: Good question! Looking at the path like this does feel a bit hacky, but it kinda makes sense. We're fetching the path of the request URI, which returns a string that we're doing a comparison on. I don't know if there's a more reliable way of fetching the HTTP request element, but I suspect it's probably private.

pjenvey: No current 1.0 way of getting that resource def (but maybe in the future): https://github.com/actix/actix-web/issues/911

PR #164, opened 2019-06-30
Inline init_routes back inside the body of build_app

### LINKED ISSUE #117: Looks like we can ditch the init_routes! macro now
I noticed that the `init_routes!` macro isn't used by the tests any more so it's probably just unnecessary indirection at this point. Maybe we should just inline it back inside the body of `build_app`.

lyuyuan92:
fixes #117


pjenvey:
Thanks, this looks good!

Could you configure commit signing for this repo and update the commit (I remember you already did for application-services)?

lyuyuan92:
Done!

PR #171, opened 2019-07-20
Use INSERT ON DUPLICATE KEY UPDATE in put_bso

### LINKED ISSUE #150: mysql's put_bso should use mysql's ON DUPLICATE KEY UPDATE
It currently falls back to a manual select + insert or an update. In rare cases this behaves slightly different than the python version using on duplicate key update.

The rare case was writing the bso `{"id": "1", "modified": 1500000150.22}` (the modified should be ignored anyway) twice via post_json (which calls put_json internally in the rust version). The second write fails with "'A database error occurred: There are no changes to save. This query cannot be built" which might partly be diesel's fault. Whereas the python version succeeds

lyuyuan:
fixes #150

### DIFF
--- a/src/db/mysql/models.rs
+++ b/src/db/mysql/models.rs
@@ -5,12 +5,11 @@ use diesel::{
     delete,
     dsl::max,
     expression::sql_literal::sql,
-    insert_into,
     mysql::MysqlConnection,
     r2d2::{ConnectionManager, PooledConnection},
     sql_query,
     sql_types::{BigInt, Integer, Nullable, Text},
-    update, Connection, ExpressionMethods, GroupByDsl, OptionalExtension, QueryDsl, RunQueryDsl,
+    Connection, ExpressionMethods, GroupByDsl, OptionalExtension, QueryDsl, RunQueryDsl,
 };
 #[cfg(any(test, feature = "db_test"))]
 use diesel_logger::LoggingConnection;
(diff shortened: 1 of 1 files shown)


src/db/mysql/models.rs
```diff
@@ -327,43 +326,31 @@ impl MysqlDb {
...
         self.conn.transaction(|| {
+            let payload = bso.payload.as_ref().map(Deref::deref).unwrap_or_default();
+            let sortindex = bso.sortindex;
+            let ttl = bso.ttl.map_or(DEFAULT_BSO_TTL, |ttl| ttl);
             let q = r#"
-                SELECT 1 as count FROM bso
-                WHERE user_id = ? AND collection_id = ? AND id = ?
+                INSERT INTO bso (user_id, collection_id, id, sortindex, payload, modified, expiry)
+                VALUES (?, ?, ?, ?, ?, ?, ?)
+                ON DUPLICATE KEY UPDATE
```
pjenvey: A couple issues here:

- ON DUPLICATE KEY UPDATE targets the existing row (by primary key) for an UPDATE if it already exists. So we want to indicate an UPDATE of all the bso fields in the `PutBso`/`UpdateBso` structs *except* the primary key (user_id/collection_id/id, which are already implied): so they shouldn't be specified here. This would look more like `ON DUPLICATE KEY UPDATE sortindex = VALUES(sortindex) payload = ..etc`

You can think of the UPDATE clause being generated as `UPDATE bso set sortindex = ? payload = ..etc WHERE user_id = ? AND collection_id = ..etc`

- Additionally all 4 of the fields in question for update are *optional* -- they're only included in the update clause if they're not a `None` value in PutBso. So we need to generate the update part of the clause dynamically. With the `VALUES()` feature of ON DUPLICATE KEY UPDATE we shouldn't need to dynamically bind the update values (which probably requires a boxed query)

src/db/mysql/models.rs
```diff
@@ -867,12 +854,6 @@ struct UserCollectionsResult {
...
 }
 
-#[derive(Debug, QueryableByName)]
-struct Count {
-    #[sql_type = "BigInt"]
-    count: i64,
-}
-
 /// Formats a BSO for UPDATEs
 #[derive(AsChangeset)]
```
pjenvey: You can kill UpdateBSO along with this change as well

src/db/mysql/models.rs
lyuyuan: Thanks for the explanation!

src/db/mysql/models.rs
lyuyuan: Done

lyuyuan:
Thanks for reviewing @pjenvey ! I have updated my branch.

src/db/mysql/models.rs
```diff
@@ -327,43 +326,53 @@ impl MysqlDb {
...
         self.conn.transaction(|| {
+            let payload = bso.payload.as_ref().map(Deref::deref).unwrap_or_default();
+            let sortindex = bso.sortindex;
+            let ttl = bso.ttl.map_or(DEFAULT_BSO_TTL, |ttl| ttl);
             let q = r#"
-                SELECT 1 as count FROM bso
-                WHERE user_id = ? AND collection_id = ? AND id = ?
+                INSERT INTO bso (user_id, collection_id, id, sortindex, payload, modified, expiry)
+                VALUES (?, ?, ?, ?, ?, ?, ?)
+                ON DUPLICATE KEY UPDATE sortindex = VALUES(sortindex)
```
pjenvey: `sortindex` like the others is also optional.

We also need to handle the case of something issuing an empty update (all update fields being None). Though I mentioned earlier that we're not required to issue updates to the primary key fields (user_id/collection_id/id) let's include them anyway so we're *always* issuing at least something to the UPDATE clause.

It's kind of cheating but it'll keep the logic here simpler. (I just checked what the Python version does for empty updates and it does this, so this cheat works for me).

src/db/mysql/models.rs
```diff
@@ -327,43 +326,53 @@ impl MysqlDb {
...
+                if bso.ttl.is_some() {
+                    ", expiry = VALUES(expiry)"
+                } else {
+                    ""
+                },
+            );
+            let q = format!(
+                "{}{}",
+                q,
+                if bso.payload.is_some() || bso.ttl.is_some() {
```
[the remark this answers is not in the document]
lyuyuan: Sorry for this, change to `bso.sortindex`

lyuyuan:
Sorry for the delay! I have made the changes requested.

pjenvey:
Thanks lyuyuan!

lyuyuan:
Thanks for your help and reviews!

PR #172, opened 2019-07-23
Implement a version of syncstorage-rs with spanner as the backend, using google-apis-rs in a threadpool.

fzzzy:


src/db/spanner/models.rs
```diff
@@ -0,0 +1,1302 @@
...
+        sql.params = Some(params);
+
+        let results = spanner
+            .hub
+            .projects()
+            .instances_databases_sessions_execute_sql(sql, session)
+            .doit();
+        if let Ok(results) = results {
+            if let Some(rows) = results.1.rows {
+                let modified = SyncTimestamp::from_i64(rows[0][0].parse().unwrap())?;
```
pjenvey: So Spanner TIMESTAMP types come back as i64 (probably in the raw json)?

The Python Spanner client brings them back as Python datetimes -- probably because it's using grpc underneath w/ more type info

src/db/spanner/models.rs
```diff
@@ -0,0 +1,1302 @@
...
+        //     .borrow_mut()
+        //     .coll_locks
+        //     .insert((user_id, collection_id), CollectionLock::Write);
+        Ok(())
+    }
+
+    pub(super) fn begin(&self) -> Result<()> {
+        let spanner = &self.conn;
+        let session = spanner.session.name.as_ref().unwrap();
+        let options = Some(TransactionOptions::default());
```
pjenvey: TransactionOptions should differ depending on the call from lock_for_write/read (read/write vs plain read). Maybe begin should take a TransactionOptions argument

pjenvey:
This'll need a cargo fmt run before merging.

Here I've noted some TODOs (not exhaustive) for after this merges. We should probably log them as separate issues:

SpannerPool:
- refactor our 2 pool impls. to remove code duplication
- cache the service account Access? (or whatever it is so that we're not doing more auth work than we should for every `connect()` call
- implement is_valid (and has_broken which can just call it)

SpannerDb:
- kill unwraps and improve error handling
- ensure transaction ids are passed around (if it's necessary)

src/db/spanner/models.rs
fzzzy: We'll have to double check a lot of assumptions like this made by the code once the tests are running against spanner and exposing errors to fix.

Right now the tests run against spanner, but all fail silently. Need to figure out why no errors are being printed anywhere.

src/db/mod.rs
```diff
@@ -44,8 +47,15 @@ lazy_static! {
 
 type DbFuture<T> = Box<Future<Item = T, Error = ApiError>>;
 
-pub trait DbPool: Sync + Debug {
+pub trait DbPool: Send + Debug {
```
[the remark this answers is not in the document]
fzzzy: I tried making it just Sync again and it complained that it was not threadsafe. I made it Sync + Send.

Must be because of the clone? Not sure why

src/db/spanner/spanner.rs
```diff
@@ -0,0 +1,74 @@
+extern crate google_spanner1 as spanner1;
```
pjenvey: A couple more nits: 
- we should be able to kill all the extern declarations here
- let's rename this module now (maybe r2d2 or manager.rs?)
- maybe some import cleanups (import ManageConnection and the hyper bits directly instead of using their full namespaces inline and kill the unneeded std::result from std::result::Result)

src/db/spanner/models.rs
```diff
@@ -0,0 +1,1292 @@
...
+        //    .bind::<Text, _>(name)
+        //    .get_result::<IdResult>(&self.conn)
+        //    .optional()?
+        //    .ok_or(DbErrorKind::CollectionNotFound)?
+        //    .id;
+        //self.coll_cache.put(id, name.to_owned())?;
+    }
+
+    pub(super) fn create_collection(&self, name: &str) -> Result<i32> {
+        // XXX: handle concurrent attempts at inserts
```
jrconlin: Is this still valid? Do we run the risk of collisions? I see that we have locking down below, would those functions help here?

src/db/spanner/models.rs
```diff
@@ -0,0 +1,1292 @@
...
+                    Some(row) => Ok(row[0][0].parse().unwrap()),
+                    None => Ok(0),
+                }
+            }
+        };
+        if let Ok(val) = rv {
+            self.coll_cache.put(val, name.to_owned())?;
+        };
+        rv
+        //let id = sql_query("SELECT id FROM collections WHERE name = ?")
```
jrconlin: Is this needed? Could we move it to a declared comment and add a bit of context to prevent it from being dropped as "dead code" later?

src/db/spanner/models.rs
```diff
@@ -0,0 +1,1292 @@
...
+        let mut sql = ExecuteSqlRequest::default();
+        sql.sql = Some("SELECT MAX(collectionid) from collections".to_string());
+        let result = spanner
+            .hub
+            .projects()
+            .instances_databases_sessions_execute_sql(sql, session)
+            .doit();
+        let id = match result {
+            Err(e) => {
+                println!("err {}", e);
```
jrconlin: Might want to make these `dbg!` instead to prevent the logs from filling up.

src/db/spanner/models.rs
```diff
@@ -0,0 +1,1292 @@
...
+                    .map(|ttl| timestamp + (i64::from(ttl) * 1000))
+                    .unwrap()
+                    .to_string(),
+            );
+
+            sql.params = Some(sqlparams);
+            sql
+        };
+
+        let results = spanner
```
jrconlin: Is locking a concern here like it was above with inserts?

src/db/spanner/models.rs
pjenvey: This is copy pasted from the mysql code but let's leave it for now, there's similar but different concerns about it under spanner.

The python spanner runs this create in a separate transaction (I'm not recalling off hand why). I think at the very least we need to handle potential ALREADY_EXISTS errors

src/db/spanner/models.rs
pjenvey: All the writes are guarded by lock_for_write (a Spanner read/write transaction) so we should be "locked here". When a transaction can't succeed we will get an ABORT (on commit, but possibly on other API calls I believe).

We'll have a separate issue to cleanup the error handling. Part of that will be ensuring ABORTs are converted to DbError::ConflictErrors (503 to the client to retry)

src/db/spanner/models.rs
fzzzy: I just copied that from the previous code, but if it is a concern we should open a bug.

src/db/spanner/models.rs
fzzzy: I had intended to delete these blocks actually, thanks for the reminder. I'll open a followup bug for it.

src/db/spanner/models.rs
fzzzy: All the error handling needs to be combed over once the tests are running against this code. I'll switch to debug when that happens.

PR #185, opened 2019-08-01
Feat/126

jrconlin:


src/web/middleware.rs
```diff
@@ -1,151 +1,273 @@
...
+                            .body("Err: invalid collection".to_owned())
+                            .into_body(),
+                    ),
+                ));
+            }
+        };
+        let ci = &sreq.connection_info().clone();
+        let headers = &sreq.headers();
+        let auth = match headers.get("authorization") {
+            Some(a) => a.to_str().unwrap(),
```
pjenvey: There's a few `unwrap()`s in this file that need cleanup, I'll log an issue to fix these after this merges

src/web/handlers.rs
```diff
@@ -1,65 +1,59 @@
...
+                .collect();
+            HttpResponse::build(StatusCode::OK)
+                .header(X_WEAVE_RECORDS, usage.len().to_string())
+                .json(usage)
+        })
 }
 
-pub fn get_quota(meta: MetaRequest) -> FutureResponse<HttpResponse> {
+pub fn get_quota(meta: MetaRequest) -> impl Future<Item = HttpResponse, Error = Error> {
     Box::new(
```
fzzzy: I see a lot of these handlers remove the Box::new, but this one doesn't. Is there a reason for that?

src/web/handlers.rs
```diff
@@ -68,16 +62,19 @@ pub fn get_quota(meta: MetaRequest) -> FutureResponse<HttpResponse> {
     )
 }
 
-pub fn delete_all(meta: MetaRequest) -> FutureResponse<HttpResponse> {
+pub fn delete_all(meta: MetaRequest) -> impl Future<Item = HttpResponse, Error = Error> {
+    #![allow(clippy::unit_arg)]
     Box::new(
```
fzzzy: Same here

src/server/mod.rs
```diff
@@ -29,84 +38,120 @@ pub struct ServerState {
...
+                web::resource(&cfg_path("/storage/{collection}"))
+                    .data(
+                        // Declare the payload limit for "normal" collections.
+                        web::PayloadConfig::new($limits.max_request_bytes as usize),
+                    )
+                    .data(
+                        // Declare the payload limits for "JSON" payloads
+                        web::JsonConfig::default()
+                            .limit($limits.max_request_bytes as usize)
+                            .content_type(|ct| ct == mime::TEXT_PLAIN),
```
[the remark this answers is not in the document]
pjenvey: Yup we need it for legacy (we even had to add this feature to actix-web specifically for it)

src/web/handlers.rs
pjenvey: good catch, we probably just forgot these

src/server/mod.rs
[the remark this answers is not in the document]
jrconlin: Added a comment to indicate why.

src/server/mod.rs
```diff
@@ -29,84 +38,120 @@ pub struct ServerState {
...
+                    )
+                    .route(web::delete().to_async(handlers::delete_bso))
+                    .route(web::get().to_async(handlers::get_bso))
+                    .route(web::put().to_async(handlers::put_bso)),
+            )
+            // Dockerflow
+            .service(
+                web::resource("/__heartbeat__").route(web::get().to(|_: HttpRequest| {
+                    // if addidtional information is desired, point to an appropriate
```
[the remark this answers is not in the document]
jrconlin: thadnks!

src/web/extractors.rs
```diff
@@ -284,14 +289,11 @@ impl FromRequest<ServerState> for BsoBody {
...
-        let max_request_size = req.state().limits.max_request_bytes as usize;
-        config
-            .limit(max_request_size)
-            .content_type(|ct| ct == mime::TEXT_PLAIN);
-
-        let max_payload_size = req.state().limits.max_record_payload_bytes as usize;
-        let fut = <Json<BsoBody>>::from_request(req, &config)
+        let state = req.app_data::<ServerState>().unwrap();
+
+        let max_payload_size = state.limits.max_record_payload_bytes as usize;
```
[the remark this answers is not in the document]
jrconlin: Probably, still, since `from_request` is a magic method, we can't always assume that the only path to this method is from `build_app`. It may be safer to check again.

src/web/handlers.rs
jrconlin: These don't respond with `HttpResponse::build()`. Might be some argument that we should normalize all of these to do that.

jrconlin: Huh, looks like we can drop the boxes, so I will.

fzzzy: Yeah that was confusing me that the return type was impl Future and it was returning a Box. I guess Box implements Future by dereferencing itself first?

pjenvey: Not quite: the future module implements Future for all Box<F> where F: Future

fzzzy: Right, that makes sense. Thanks.

PR #192, opened 2019-08-09
Feat: remove unwraps from production path

### LINKED ISSUE #186: Remove unwraps in new actix 1.0 middlwares
We have =~ 5 `unwrap()`s in our new middlware impls which should be more carefully handled (propagating the errors out).

Though the middlwares all involve async bits now, I previously found (in the WeaveTimestamp middlware) that large sections of middle code aren't async and can be split out into separate functions w/ sane error handling returning regular Results. Then Middlewares can call into them and translate their Results into async Results

jrconlin:
Closes #186

### DIFF
--- a/src/web/extractors.rs
+++ b/src/web/extractors.rs
@@ -12,7 +12,7 @@ use actix_web::{
         Uri,
     },
     web::{Json, Query},
-    Error, FromRequest, HttpRequest,
+    Error, FromRequest, HttpMessage, HttpRequest,
 };
 use futures::{future, Future};
 use lazy_static::lazy_static;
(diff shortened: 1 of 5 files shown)


src/web/extractors.rs
```diff
@@ -289,7 +302,20 @@ impl FromRequest for BsoBody {
                 ));
             }
         }
-        let state = req.app_data::<ServerState>().unwrap();
+        let state = match req.app_data::<ServerState>() {
+            Some(s) => s,
+            None => {
+                dbg!("⚠️ Could not load the app state");
+                return Box::new(future::err(
+                    ValidationErrorKind::FromDetails(
```
pjenvey: nit: these should probably be ApiErrorKind::Internal but we don't have a `From<ApiErrorKind> for actix_web::Error` (however ValidationErrorKind does).

Our error hierarchy and these from conversions are a bit involved so I'm fine ok w/ validation errors for now

src/web/extractors.rs
```diff
@@ -838,7 +894,23 @@ impl FromRequest for HawkIdentifier {
 
     /// Use HawkPayload extraction and format as HawkIdentifier.
     fn from_request(req: &HttpRequest, _: &mut Payload) -> Self::Future {
-        Self::extrude(req)
+        let state = match req.app_data::<ServerState>() {
```
pjenvey: Just occurred to me now: I suppose we could have a macro for only the future::err case. Whereas non-future returns can opt for the ok_or_else (like I have in get_hawk_id).

(to just throw another alternative to the match statements)

jrconlin: I tried that at first, but the problem was wrapping the error types. We'd wind up with several macros (or I suppose adding a block to return the error, but then it felt like "why bother with a macro" so I just reverted back to the matches.

PR #204, opened 2019-08-30
Add rust dependency diff

g-k:
WIP

refs: https://github.com/mozilla-services/find-package-rugaru/issues/69

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -235,11 +235,15 @@ jobs:
             tar -xz -C /tmp -f /tmp/docker-$VER.tgz
             mv /tmp/docker/* /usr/bin
       - run:
+          name: pull find-package-rugaru image
+          command: |
+            docker pull gguthemoz/fpr:latest
+      - run:
           name: Check Rust Dependencies
           command: |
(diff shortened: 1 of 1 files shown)


jrconlin:
@g-k What's the status on this one? Can we close this PR?

g-k:
@jrconlin still working on it! I was on PTO earlier this week and the last couple weeks.

.circleci/config.yml
```diff
@@ -238,6 +263,11 @@ workflows:
           filters:
             tags:
               only: /.*/
+        # ping @g-k / gguthe on slack with feedback
+      - rust-changelog:
```
g-k: Could put this in a separate workflow too if that's preferable.

jrconlin: actually, is there a more generic address that could be used? I assume you want to take vacations or possibly do other things in the future, maybe take on a minion you can assign stuff like this too, etc.

.circleci/config.yml
```diff
@@ -220,6 +220,31 @@ jobs:
...
+          command: |
+            set -x
+            VER="18.06.3-ce"
+            curl -L -o /tmp/docker-$VER.tgz https://download.docker.com/linux/static/stable/x86_64/docker-$VER.tgz
+            tar -xz -C /tmp -f /tmp/docker-$VER.tgz
+            mv /tmp/docker/* /usr/bin
+      - run:
+          name: pull find-package-rugaru image
+          command: |
+            docker pull gguthemoz/fpr:latest
```
jrconlin: honestly, this makes me a tiny bit concerned. I frequently chase down folks and packages that are no longer maintained by one person. Is it possible to put this under a larger org so that someone else can take it over if need be?

g-k: Absolutely, that was just to get something out the door. Let me get added to the mozilla docker org.

Issue #173, opened 2019-07-23
Fix batch handling for larger batches


pjenvey:
Copied from [server-syncstorage#98](https://github.com/mozilla-services/server-syncstorage/issues/98):

> The Spanner impl uses a single bsos field for batches, this is limited to 2.5 MB. Using an array or bytes would allow 10MB, but that's still relatively low. We should switch to an impl. like the SQL storage and use a table of bso's.

We'll probably just implement it here in rust at this point (but not in Python)

pjenvey:
We should probably consider #21 as we revamp this

PR #208, opened 2019-09-06
Fix batch handling on spanner by using multiple records instead of one.

fzzzy:
There are a few type errors that need to be fixed, but I think the general approach is correct.

This assumes that the batches table has a new "timestamp" field, which I added to the test database. Since id is used as part of the primary key, it can't contain duplicates. The code generates a unique timestamp for every bso in the batch to use as an id, but then selects based on the timestamp field instead of the id field.

### DIFF
--- a/src/db/spanner/support.rs
+++ b/src/db/spanner/support.rs
@@ -215,26 +215,21 @@ impl SyncResultSet {
         }
     }
 
+    pub fn all_or_none(&mut self) -> Option<Vec<ListValue>> {
+        if self.result.rows.is_empty() {
+            None
+        } else {
+            Some(self.result.rows.clone().into_vec())
+        }
+    }
+
     #[cfg(feature = "google_grpc")]
     pub fn affected_rows(self: &SyncResultSet) -> Result<i64> {
         let stats = self
             .stats()
             .ok_or_else(|| DbError::internal("Expected result_set stats"))?;
-        Ok(stats.get_row_count_exact())
-    }
-
-    #[cfg(not(feature = "google_grpc"))]
-    pub fn affected_rows(self: &SyncResultSet) -> Result<i64> {
-        let stats = self
-            .stats()
-            .ok_or_else(|| DbError::internal("Expected result_set stats"))?;
-        let row_count_exact = stats
-            .row_count_exact
-            .as_ref()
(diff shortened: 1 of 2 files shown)


fzzzy:
I am on PTO next week. If someone wants to pick this up and finish it, it should only require a bit of type conversion. Otherwise, I will finish it when I get back.

fzzzy:
Here are some things that I think should happen before this lands:

  - The `timestamp` field should be renamed to `batch_id`
  - `sortindex` and `payload` columns should be added, and the `bsos` column should be dropped. `id`, `sortindex`, `payload` and `expiry` should be stored directly in the table, instead of in a json string in the `bsos` column.

fzzzy:
The changes to the spanner batches table need to be recorded in the python file @pjenvey mentioned to me at the workweek. Where is that file?

tublitzed:
Closes [Issue 173](https://github.com/mozilla-services/syncstorage-rs/issues/173)

src/db/spanner/support.rs
```diff
@@ -215,26 +215,21 @@ impl SyncResultSet {
         }
     }
 
-    #[cfg(feature = "google_grpc")]
-    pub fn affected_rows(self: &SyncResultSet) -> Result<i64> {
-        let stats = self
-            .stats()
-            .ok_or_else(|| DbError::internal("Expected result_set stats"))?;
-        Ok(stats.get_row_count_exact())
+    pub fn all_or_none(&mut self) -> Option<Vec<ListValue>> {
```
pjenvey: We can revisit this later but I'm not sure why this method is necessary w/ SyncResult supporting Iteration. That let's you do something like `let result: Vec<_> = sql(..)?.execute(..)?.collect();`, giving you a `Vec<Vec<Value>>` vs `Vec<ListValue>`

pjenvey:
> The changes to the spanner batches table need to be recorded in the python file @pjenvey mentioned to me at the workweek. Where is that file?
> 
> (Edit: I think the file is https://github.com/mozilla-services/server-syncstorage/blob/master/syncstorage/storage/sql/queries_spanner.py)

JR's including a version of the schema in his PR, we might as well just wait until that lands and apply the changes there

src/db/spanner/support.rs
fzzzy: Ok, awesome. I did not know that.

PR #210, opened 2019-09-06
bug: Add 1000 years for test cases to prevent premature Hawk expiration

### LINKED ISSUE #209: Hawk test cases are expiring prematurely
It's been a year, so all the hard coded Hawk stuff is now failing.

jrconlin:
Closes #209


jrconlin:
ah, cool. This was mostly just a friday fix because the tests were failing. I kinda stole the 1000 year fix from the Hawk test suite.

PR #214, opened 2019-09-12
feat: make mysql use existing column names

### LINKED ISSUE #194: Audit the Schema
- STRING for payloads? We may need to continue using STRING for compatibility with the Python version, but if we were able to use BYTES instead we could save storage and bandwidth
- Secondary indices: we may requirement multiple indicies depending on the order specified on get_bsos

jrconlin:
* for continuity, ensure that we use the original column names for mysql
* added non-destructive migration for existing tables.

Closes #194

### DIFF
--- a/src/db/mysql/schema.rs
+++ b/src/db/mysql/schema.rs
@@ -1,6 +1,8 @@
 table! {
     batches (user_id, collection_id, id) {
+        #[sql_name="userid"]
         user_id -> Integer,
+        #[sql_name="collection"]
         collection_id -> Integer,
         id -> Bigint,
         bsos -> Longtext,
@@ -10,11 +12,14 @@ table! {
 
 table! {
     bso (user_id, collection_id, id) {
+        #[sql_name="userid"]
         user_id -> Integer,
+        #[sql_name="collection"]
         collection_id -> Integer,
         id -> Varchar,
         sortindex -> Nullable<Integer>,
         payload -> Mediumtext,
+        #[sql_name="modified"]
         modified -> Bigint,
         expiry -> Bigint,
     }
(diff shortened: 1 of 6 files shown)


src/db/mysql/schema.rs
```diff
@@ -10,11 +12,14 @@ table! {
...
 table! {
     bso (user_id, collection_id, id) {
+        #[sql_name="userid"]
         user_id -> Integer,
+        #[sql_name="collection"]
         collection_id -> Integer,
         id -> Varchar,
         sortindex -> Nullable<Integer>,
         payload -> Mediumtext,
+        #[sql_name="modified"]
```
pjenvey: kill?

src/db/mysql/schema.rs
```diff
@@ -10,11 +12,14 @@ table! {
...
         modified -> Bigint,
         expiry -> Bigint,
```
pjenvey: `expiry` is `ttl` in the old schema :(

also collection's `id` column is `collectionid`

and of course, mysql's batch schema/implementation is completely different, which is really a whole other issue entirely (should we match its schema + old impl.?)

src/db/mysql/schema.rs
jrconlin: Probably, mostly there because I was debating that we should be consistent about using `last_modified` as the variable name.

src/db/mysql/schema.rs
jrconlin: hmm. I was using `server-syncstorage` repo for the schema. Is there a better source I can use to verify against?

pjenvey: Yup, I'm looking at its sqlalchemy definitions:

https://github.com/mozilla-services/server-syncstorage/blob/5e9045b/syncstorage/storage/sql/dbconnect.py#L132

jrconlin: Ah, cool. I just missed that one. Triple checking now. Thanks!

PR #220, opened 2019-09-17
feat: Set tombstone on deleted data for user.

### LINKED ISSUE #23: Add a deleted_at to user_collections table
See rfk's comments here:

https://github.com/mozilla-services/syncstorage-rs/pull/22#discussion_r216541111

I believe the go version gets away without this field because it stores the top level modified timestamp separately in its per user KeyValues store.
https://github.com/mozilla-services/server-syncstorage/pull/62

jrconlin:
Closes #23

### DIFF
--- a/src/db/mysql/models.rs
+++ b/src/db/mysql/models.rs
@@ -266,7 +266,7 @@ impl MysqlDb {
     }
 
     // Delete all data associated with the user. Erect a tombstone because legacy
-    // systems expect a `200 []` instead of a 404 if no data is present.
+    // systems expect a `200 []`.
     pub fn delete_storage_sync(&self, user_id: HawkIdentifier) -> Result<()> {
         let user_id = user_id.legacy_id as i32;
         self.begin()?;
--- a/src/db/mysql/models.rs
+++ b/src/db/mysql/models.rs
@@ -265,8 +265,6 @@ impl MysqlDb {
         Ok(())
     }
 
-    // Delete all data associated with the user. Erect a tombstone because legacy
-    // systems expect a `200 []`.
     pub fn delete_storage_sync(&self, user_id: HawkIdentifier) -> Result<()> {
         let user_id = user_id.legacy_id as i32;
         self.begin()?;
@@ -281,7 +279,9 @@ impl MysqlDb {
         Ok(())
     }
 
-    // Delete just a collection for a given user.
+    // Deleting the collection should result in:
+    //  - collection does not appear in /info/collections
+    //  - X-Last-Modified timestamp at the storage level changing
     pub fn delete_collection_sync(
         &self,
         params: params::DeleteCollection,
--- a/src/db/mysql/models.rs
+++ b/src/db/mysql/models.rs
@@ -37,6 +37,7 @@ type Conn = PooledConnection<ConnectionManager<MysqlConnection>>;
 /// The ttl to use for rows that are never supposed to expire (in seconds)
 pub const DEFAULT_BSO_TTL: u32 = 2_100_000_000;
 
+pub const TOMBSTONE: i32 = 0;
 /// SQL Variable remapping
 /// These names are the legacy values mapped to the new names.
 pub const COLLECTION_ID: &str = "collection";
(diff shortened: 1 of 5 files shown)


src/db/spanner/models.rs
```diff
@@ -558,11 +561,17 @@ impl SpannerDb {
...
-        // XXX: should delete from bso table too
         let user_id = user_id.legacy_id as u32;
         self.sql("DELETE FROM user_collections WHERE userid=@userid")?
             .params(params! {"userid" => user_id.to_string()})
             .execute(&self.conn)?;
+        self.sql("DELETE FROM bso WHERE userid=@userid")?
+            .params(params! {"userid" => user_id.to_string()})
+            .execute(&self.conn)?;
+        // Erect a Tombstone
+        self.sql("INSERT INTO user_collections (userid, collection, last_modified) values (@userid, \"0\", @modified)")?
```
pjenvey: We need to track collection deletion (delete_collection_sync) whereas this (delete_storage_sync) is tracking user deletion.

jrconlin: Ah, I misunderstood the requirement. If we're noting that a given user record is modified by the removal of collections, should we keep a tombstone when all collections are deleted as well, or if everything is deleted does it not really matter?

pjenvey: It doesn't matter when everything's deleted. storage's last_modified in that case resets to 0

src/db/spanner/models.rs
```diff
@@ -558,11 +561,17 @@ impl SpannerDb {
     }
 
     pub fn delete_storage_sync(&self, user_id: params::DeleteStorage) -> Result<()> {
-        // XXX: should delete from bso table too
```
pjenvey: Sorry, please ignore this bogus comment I added! I forgot to remove it

(bso is interleaved into user_collections w/ ON DELETE CASCADE, so the query removing the row from user_collections automatically does bso deletion for us)

src/db/mysql/models.rs
```diff
@@ -249,17 +250,38 @@ impl MysqlDb {
...
+        )
+        .bind::<Integer, _>(user_id)
+        .bind::<Integer, _>(TOMBSTONE)
+        .bind::<BigInt, _>(self.timestamp().as_i64())
+        .execute(&self.conn)?;
+        Ok(())
+    }
+
+    // Delete all data associated with the user. Erect a tombstone because legacy
+    // systems expect a `200 []` instead of a 404 if no data is present.
```
pjenvey: the last part of this comment's outdated now

src/db/spanner/models.rs
```diff
@@ -647,12 +650,21 @@ impl SpannerDb {
         }
     }
 
+    fn erect_tombstone(&self, user_id: u32) -> Result<()> {
+        self.sql("INSERT INTO user_collections (userid, collection, last_modified) values (@userid, \"0\", @modified)")?
```
pjenvey: Should be
```suggestion
        self.sql("INSERT INTO user_collections (userid, collection, last_modified) values (@userid, 0, @modified)")?
```
but might as well bind the TOMBSTONE const instead since it's defined (surprised clippy didn't complain about it being unused)

src/db/spanner/models.rs
```diff
@@ -647,12 +650,21 @@ impl SpannerDb {
...
+            .params(params!{"userid" => user_id.to_string(), "modified" => self.timestamp().as_rfc3339()?})
```
pjenvey: need an explicit type for timestamps:
```suggestion
            .params(params!{"userid" => user_id.to_string(), "modified" => self.timestamp().as_rfc3339()?})
            .param_types(param_types! {
                    "modified" => SpannerType::Timestamp,
                })            
```

src/db/spanner/models.rs
pjenvey: We need an UPSERT here like the mysql version. Spanner only has one in its mutation API.

Feel free to just issue a DELETE of the tombstone beforehand for now (way easier than  writing select + if exists insert else update)

I have some uncommitted code for using Mutations that we can change this to later

src/db/spanner/models.rs
```diff
@@ -667,12 +670,37 @@ impl SpannerDb {
         }
     }
 
+    fn erect_tombstone(&self, user_id: u32) -> Result<()> {
+        // Delete the old tombstone (if it exists)
+        self.sql("DELETE from user_collection where userid=@userid and collection=@collection")?
+            .params(params! {
+                "userid" => user_id.to_string(),
```
pjenvey: forgot the collection param here again

src/db/spanner/models.rs
```diff
@@ -667,12 +670,37 @@ impl SpannerDb {
         }
     }
 
+    fn erect_tombstone(&self, user_id: u32) -> Result<()> {
+        // Delete the old tombstone (if it exists)
+        self.sql("DELETE from user_collection where userid=@userid and collection=@collection")?
```
pjenvey:
```suggestion
        self.sql("DELETE from user_collections where userid=@userid and collection=@collection")?
```

Issue #226, opened 2019-09-20
Do not require Hawk Auth header for Dockerflow functions


jrconlin:
Hawk headers should not be required for dockerflow commands.

PR #227, opened 2019-09-20
feat add fn to purge old bsos

### LINKED ISSUE #119: Add a purgettl command line tool
We haven't ported the purgettl.py script from python to rust. It's ran continuously in deployments.

There's a slight possibility we could use the python version ([when and if it's implemented](https://github.com/mozilla-services/server-syncstorage/issues/100)) but:

- we'd need to ensure the python spanner schema matches ours exactly
- it requires a python style config file when ran which our rust code doesn't otherwise use

It may make sense (at the very least for the sake of convenience) to reimplement it in rust.

jrconlin:
* fix up DockerFlow to not require HawkHeader

Closes #119, #226


jrconlin:
closed in favor of #229 and ops cron job

Issue #231, opened 2019-09-25
`collection` table IDs below 100 need to be reserved for system use.


jrconlin:
See [bug 1364005](https://bugzilla.mozilla.org/show_bug.cgi?id=1364005) The collections table reserves ids < 100 for system use. We need to ensure that any new collections occur after id 100. <insert JR grumbling about name collisions and system flags or at least prefixes because, really... >

fzzzy:
This was fixed by https://github.com/mozilla-services/syncstorage-rs/commit/46c650a1794881e58cf1af5abdefadbea9b72aec

PR #233, opened 2019-09-25
fix: get the session timestamp from Spanner

### LINKED ISSUE #223: SpannerDb::timestamp should return CURRENT_TIMESTAMP()
SpannerDb inherited MysqlDb's behavior where it creates its self.timestamp() value from the current server's time.

For the sake of sanity we should be using `CURRENT_TIMESTAMP()` from Spanner itself instead (as the python does)

pjenvey:
and fix reservation of the std collection id space w/:

- a small adjustment to spanner's autoincrement
- a dummy value on mysql (nothing else is preserved between mysqld restarts)

and enable docker hub deploying

Closes #223, #231

### DIFF
--- a/db-tests/src/support.rs
+++ b/db-tests/src/support.rs
@@ -27,7 +27,11 @@ pub async fn db() -> Result<Box<dyn Db>> {
     };
 
     let pool = syncstorage::db::pool_from_settings(&settings)?;
-    pool.get().compat().await
+    let db = pool.get().compat().await?;
+    // Spanner won't have a timestamp until lock_for_xxx are called: fill one
+    // in for it
+    db.set_timestamp(SyncTimestamp::default());
+    Ok(db)
 }
 
 macro_rules! with_delta {
--- a/migrations/2018-08-28-010336_init/up.sql
+++ b/migrations/2018-08-28-010336_init/up.sql
@@ -26,7 +26,7 @@ CREATE TABLE `bso` (
 CREATE TABLE `collections` (
     `id` INT PRIMARY KEY      NOT NULL AUTO_INCREMENT,
     `name` VARCHAR(32) UNIQUE NOT NULL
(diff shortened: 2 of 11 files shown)


db-tests/src/support.rs
```diff
@@ -27,7 +27,11 @@ pub async fn db() -> Result<Box<dyn Db>> {
     };
 
     let pool = syncstorage::db::pool_from_settings(&settings)?;
-    pool.get().compat().await
+    let db = pool.get().compat().await?;
+    // Spanner won't have a timestamp until lock_for_xxx are called: fill one
+    // it for it
```
jrconlin: it's it

migrations/2018-08-28-010336_init/up.sql
```diff
@@ -26,7 +26,7 @@ CREATE TABLE `bso` (
 CREATE TABLE `collections` (
     `id` INT PRIMARY KEY      NOT NULL AUTO_INCREMENT,
     `name` VARCHAR(32) UNIQUE NOT NULL
-) ENGINE=InnoDB AUTO_INCREMENT=100 DEFAULT CHARSET=latin1;
```
jrconlin: Isn't it bad form to update an old migration?

migrations/2019-09-25-174347_min_collection_id/up.sql
```diff
@@ -0,0 +1,3 @@
+-- Reserve space for additions to the standard collections
+INSERT INTO collections (id, name)
```
jrconlin: stupid thought? We already filter for "xxx*" maybe use that for this?

migrations/2018-08-28-010336_init/up.sql
pjenvey: Definitely but with AUTO_INCREMENT being ephemeral it doesn't really matter -- which is an argument for leaving it -- but I'm nuking it so no one might mistake it for being needed/useful

PR #234, opened 2019-09-26
docs: include details on golang requirement in setup instructions

tublitzed:
Adding details about need to install go to support the newly added grpc lib.


jrconlin:
Pulling master should solve the failing `ci: checks`

Issue #170, opened 2019-07-12
Adopt the new user_id scheme for the Spanner backend


pjenvey:
Mysql uses the "legacy" user_id (an autoincremented id from the token server). Spanner requires a non serial user_id primary key: ([discussion](https://github.com/mozilla-services/tokenserver/issues/121)).

Our HawkPayload hasn't stubbed these values out yet (fxa_uid/fxa_kid and we'll likely utilize hash_fxa_uid for metrics). They need to be added and the tests should be adjusted to include such values (if they don't already).

Finally Spanner should switch from legacy_id over to the new values.

pjenvey:
Additionally the db_tests should somehow setup non-legacy HawkIdentifiers when testing the spanner backend

jrconlin:
Is this value ever exposed? (Wondering if we could use an augmented time value or something since we may not have a central location to draw from.)

pjenvey:
Assuming you mean the legacy_id: various client implementations assume an integer user_id. We need to continue providing one for them and they'd continue using it for constructing the {uid} field in the sync storage urls.

But underneath the Spanner impl. would use the new uuid-style id from the Hawk header.

(we currently have a centralized location to draw the legacy ids from -- the tokenserver -- but we want the tokenserver to eventually go away)

rfk:
> But underneath the Spanner impl. would use the new uuid-style id from the Hawk header for its user_id.

So, there's one thing to be careful of which I'm not sure is clear from the discussion in https://github.com/mozilla-services/tokenserver/issues/124, so I'm going to re-iterate it here just in case.

The tokenserver currently changes the numeric uid for a user whenever their FxA encryption key changes. Because the current storage nodes index their storage by this numeric userid, the effect is that the user's storage gets "wiped" automagically when their encryption key changes - they start writing into a different bucket on the storage node that doesn't have any of the old data in it encrypted under the old key.

There is also a background job on the tokenserver which periodically asks the storage nodes to delete data stored under numberic userids that have become obsolete.

If we switch to indexing data storage by the (stable and immutable) fxa_uid instead, we'll need to make sure that we have some other way to:

1. Separate data encrypted with different encryption keys into distinct storage buckets
2. Purge old data encrypted with old encryption keys as a background cleanup job

I don't recall if we settled on a precise scheme for how to achieve those two things in a durable-storage world.  The tokenserver sends an `fxa_kid` field that uniquely identifies the user's current encryption key and should be helpful here.

[...]

PR #237, opened 2019-09-30
Fix/170

jrconlin:
*  Added "kid" column to database:
* May want to use alternate "sync_kid" db for local testing
 DSN: spanner://projects/sync-spanner-dev-225401/instances/spanner-test/databases/sync_kid
* Switched over indexing to use "fxa_id" (kid) and pass HawkParameter when possible.

Issue https://github.com/mozilla-services/syncstorage-rs/issues/170

### DIFF
--- a/src/db/spanner/models.rs
+++ b/src/db/spanner/models.rs
@@ -60,10 +60,10 @@ struct SpannerDbSession {
     /// CURRENT_TIMESTAMP() from Spanner, used for timestamping this session's
     /// operations
     timestamp: Option<SyncTimestamp>,
-    /// Cache of collection modified timestamps per (fxa_kid, legacy_id, collection_id)
-    coll_modified_cache: HashMap<(String, i32, i32), SyncTimestamp>,
+    /// Cache of collection modified timestamps per (fxa_kid, fxa_uid, collection_id)
+    coll_modified_cache: HashMap<(String, String, i32), SyncTimestamp>,
     /// Currently locked collections
-    coll_locks: HashMap<(String, i32, i32), CollectionLock>,
+    coll_locks: HashMap<(String, String, i32), CollectionLock>,
     #[cfg(feature = "google_grpc")]
     transaction: Option<googleapis_raw::spanner::v1::transaction::TransactionSelector>,
     #[cfg(not(feature = "google_grpc"))]
@@ -194,7 +194,7 @@ impl SpannerDb {
         self.begin(false)?;
 
         let fxa_kid = params.user_id.fxa_kid;
-        let legacy = params.user_id.legacy_id as i32;
+        let fxa_uid = params.user_id.fxa_uid;
 
         let collection_id =
             self.get_collection_id(&params.collection)
@@ -211,14 +211,14 @@ impl SpannerDb {
             .session
             .borrow()
             .coll_locks
-            .get(&(fxa_kid.clone(), legacy, collection_id))
+            .get(&(fxa_kid.clone(), fxa_uid.clone(), collection_id))
             .is_some()
         {
             return Ok(());
         }
 
         self.session.borrow_mut().coll_locks.insert(
-            (fxa_kid.clone(), legacy, collection_id),
+            (fxa_kid.clone(), fxa_uid.clone(), collection_id),
             CollectionLock::Read,
         );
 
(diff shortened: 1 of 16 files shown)


jrconlin:
I'm fine rebasing this into one PR. There were a few messier merges in play, but not sure if we want to lose history.

src/server/test.rs
```diff
@@ -267,7 +267,7 @@ fn delete_collection() {
         http::Method::DELETE,
         "/1.5/42/storage/bookmarks",
         None,
-        Some("0.00"),
+        None, // deleting a collection may return a tombstone rather than "0.00"
```
pjenvey: This check might as well switch to `test_endpoint_with_response` w/ the exact same `result > start` closure in the other 2 checks in this test

src/web/auth.rs
```diff
@@ -116,6 +116,7 @@ impl HawkPayload {
         // Comment the following to disable auth
         verify_hmac(payload, &secrets.signing_secret, signature)?;
 
+        dbg!(std::str::from_utf8(payload).unwrap());
```
pjenvey: Couple things:

- we don't want to remove user_id from our primary key, but it should be switched from the `legacy_id` u64 for mysql to the `fxa_uid` String from the hawk header (you can see it in this dbg! statement)
- HawkIdentifier already has a field for it: `fxa_id`, let's rename it to `fxa_uid`
- And the rest of this patch pretty much gets the second part: the addition of `fxa_kid` to the primary key. The "kid" column name sounds good but its value should come into the db methods as `params.user_id.fxa_kid` to be clear and distinguished from `params.user_id.fxa_uid`
- HawkParameter (which passes through the hawk json) also needs these new fields added

So our primary key will generally be`(user_id, kid, collection)`.

jrconlin: ok, added a DDL file containing the new spanner schema to this PR.

src/server/test.rs
jrconlin: Not quite, but I did chase down an inconsistency with how mysql and spanner were dealing with this.

src/web/extractors.rs
```diff
@@ -906,7 +908,8 @@ impl HawkIdentifier {
 
         let user_id = HawkIdentifier {
             legacy_id: payload.user_id,
-            fxa_id: "".to_string(),
+            fxa_uid: "".to_string(),
```
pjenvey: we wanna pass thru fxa_{u,k}id from HawkPayload into HawkIdentifier here

src/db/spanner/models.rs
```diff
@@ -193,7 +193,9 @@ impl SpannerDb {
         // Begin a transaction
         self.begin(false)?;
 
-        let user_id = params.user_id.legacy_id as u32;
+        let fxa_kid = params.user_id.fxa_kid;
+        let legacy = params.user_id.legacy_id as i32;
```
pjenvey: we should be using fxa_uid everywhere in here now instead of mysql's legacy_id

src/db/spanner/models.rs
```diff
@@ -243,11 +246,13 @@ impl SpannerDb {
             .sql(
                 "SELECT CURRENT_TIMESTAMP(), last_modified
                    FROM user_collections
-                  WHERE userid = @userid
+                  WHERE fxa_kid = @fxa_kid
```
pjenvey: This is an annoying ask but can we keep the order of columns intact everywhere (user_id, kid)? w/ kids being a piece of metadata on users

src/db/spanner/models.rs
```diff
@@ -781,38 +830,50 @@ impl SpannerDb {
...
+                )?
+                .params(params! {
+                    "fxa_kid" => fxa_kid.to_string(),
+                    "legacy" => legacy.to_string(),
+                    "collectionid" => collection_id.to_string(),
+                })
+                .execute(&self.conn)?
+                .one_or_none()?
+                .is_some();
+        if result {
```
pjenvey: The go version never had a check like this, but I'm +1 for adding it, it makes DELETE more idempotent (depending on your definition of it)

But you'll need to use `affected_row_count()` on SyncResult for accurate DML statement result counts. I think you get back 0 rows for such statements (so this is always false).

You can also kill the first `DELETE FROM bso` statement for simplicity here, since we have ON DELETE CASCADE for the bso table, `DELETE FROM user_collections` automatically does that delete for us

src/db/spanner/models.rs
jrconlin: Ah! ok, I was confused about that part. thanks!

src/db/spanner/models.rs
jrconlin: turns out, kinda needed to make sure that spanner and mysql exhibit the same behaviors.

src/db/spanner/models.rs
jrconlin: ok, i've tried to restore the order of columns (userid first, then fxa_kid, etc. to match the indexes) also tried to make sure that the parameter lists match up as well, since we have two values that are only different by one letter.

PR #245, opened 2019-10-03
Fix #224 Only populate the CollectionCache when in a read transaction

### LINKED ISSUE #224: CollectionCache could populate w/ invalid values
The collection cache is populated when creating new collections or when loading collections not present in the cache. The former case is always within r/w transactions, the latter sometimes during a r/w transactions.

R/W transactions can potentially abort and rollback. In that case any newly cached collection_ids are bogus.

We should either not cache at all or more carefully cache only during only read transactions. Collection writes/reads could also potentially run in separate transactions, guaranteeing they're committed before the cache is populated.

fzzzy:


fzzzy:
> also there's an additional put in create_collection which should be removed (it's always during a write transaction and could potentially be rolled back)

Whoops, thank you. I thought that it was never in a write transaction but I misread it.

PR #246, opened 2019-10-04
fix: correct mistaken GROUP BY fxa_kid (vs userid)

pjenvey:
minor rearranging

### DIFF
--- a/src/db/spanner/models.rs
+++ b/src/db/spanner/models.rs
@@ -4,7 +4,6 @@ use futures::lazy;
 use diesel::r2d2::PooledConnection;
 
 use std::cell::RefCell;
-use std::cmp;
 use std::collections::HashMap;
 use std::convert::TryInto;
 use std::fmt;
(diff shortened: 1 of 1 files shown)


src/db/spanner/models.rs
```diff
@@ -777,18 +766,17 @@ impl SpannerDb {
...
         self.sql(
             "DELETE FROM user_collections
               WHERE userid = @fxa_uid
                 AND fxa_kid = @fxa_kid",
         )?
-        .params(params.clone())
+        .params(params! {
+            "fxa_uid" => user_id.fxa_uid,
+            "fxa_kid" => user_id.fxa_kid,
+        })
```
jrconlin: This was more me trying to be consistent about how params were being defined for functions. The clone was definitely unneeded. Thanks!

pjenvey: No worries, I think this is actually leftover from removing a second unneeded DELETE query anyway

PR #247, opened 2019-10-04
refactor: cleanup spanner's schema

pjenvey:
needs #246 merged first

it's deployed on the `sync_schema` database


jrconlin:
I do kinda wish that `fxa_uid` and `fxa_kid` weren't just one buried letter off, but that's not really something we can change.

pjenvey:
> I do kinda wish that `fxa_uid` and `fxa_kid` weren't just one buried letter off, but that's not really something we can change.

Agreed, it's `uid` in fxa's db table, `kid` in its scoped keys, and thus everywhere else

PR #256, opened 2019-10-08
feat: add tag support for metrics.

### LINKED ISSUE #222: Find a way to differentiate between old sync/new sync for tracking purposes.
We'll need to find a way to make sure we can bucket our tracking data for old vs. new Sync as we start to roll this out.

Not an absolute blocker for shipping, but it would be great to get this in very soon afterwards.

jrconlin:
Include a "version" tag with all metrics that uses the Cargo version

Closes #222


src/server/metrics.rs
```diff
@@ -39,10 +40,23 @@ impl Metrics {
...
-            match self.client.unwrap().incr(label) {
+            let client = self.client.unwrap();
+            let mut incr = client.incr_with_tags(label);
+            let tags = tags.unwrap_or_default();
+            let keys = tags.keys();
+            for tag in keys {
+                incr = incr.with_tag(tag, &tags.get(tag).unwrap())
+            }
+            // Include any "hard coded" tags.
+            incr = incr.with_tag("version", env!("CARGO_PKG_VERSION"));
```
tublitzed: if we bump the minor/patch version, we probably aren't going to want to change the metrics dashboards. Would you mind also grabbing the `CARGO_PKG_VERSION_MAJOR` here too so it'll make for less required tracking/dashboard changes when version numbers change?

pjenvey: I'm not sure passing along the version is necessary. I think we can distinguish between the old python cluster and new syncstorage-rs metrics easily: https://github.com/mozilla-services/syncstorage-rs/issues/222#issuecomment-540153554

Also our metrics tend to be configured to include tags for further distinguishing too, e.g. "host" and  "app" tags by the ingesting process (telegraf in syncstorage's case). We might wanna double check this w/ @bobm

jrconlin:
So, based on [pjenvey's comment](https://github.com/mozilla-services/syncstorage-rs/issues/222#issuecomment-540153554), I've added a `timer_with_tags()` metric function (it records the time when it drops out of scope). One problem is getting the metric object (or the app state) into the database so I can mirror the python decorators.

PR #268, opened 2019-10-11
feat: rewrite post/put_bsos w/ spanner mutations

### LINKED ISSUE #267: Optimize post_bsos
post_bsos currently calls put_bso for each of its bsos. This incurs a query to check it exists + an individual insert for every single bso.

It should be more efficient to utilize Spanner's mutations for a bulk write. It will still requires a check for presence of existing bsos, but the check for all of them can occur in one query. Additionally this is what the the Python prototype used.

(Spanner's insert_or_update mutation won't work as it assumes the columns touched in the insert vs the update clauses are the same which is not the case for us).

The caveat with the Mutations are they won't be seen inline during the db_tests' transactions until they are committed. For the time being we can leave the old post_bsos code for the db_tests, and hopefully come up with a better solution for the tests later. The e2e tests will test the new Mutations path.

pjenvey:
leaving the old impl. behind a cfg toggle for the db_tests (for now)

Closes #267

### DIFF
--- a/src/db/spanner/models.rs
+++ b/src/db/spanner/models.rs
@@ -17,24 +17,28 @@ use super::support::SpannerType;
 use crate::db::{
     error::{DbError, DbErrorKind},
     params, results,
-    util::{to_rfc3339, SyncTimestamp},
+    util::SyncTimestamp,
     Db, DbFuture, Sorting, FIRST_CUSTOM_COLLECTION_ID,
 };
 
 use crate::web::extractors::{BsoQueryParams, HawkIdentifier};
 
+#[cfg(not(any(test, feature = "db_test")))]
+use super::support::{bso_to_insert_row, bso_to_update_row};
 use super::{
     batch,
     support::{as_list_value, as_value, bso_from_row, ExecuteSqlRequestBuilder},
 };
 
-use googleapis_raw::spanner::v1::spanner::{
-    BeginTransactionRequest, CommitRequest, ExecuteSqlRequest, RollbackRequest,
-};
 use googleapis_raw::spanner::v1::transaction;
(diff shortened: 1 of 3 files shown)


src/db/spanner/models.rs
```diff
@@ -63,6 +67,9 @@ struct SpannerDbSession {
     /// Currently locked collections
     coll_locks: HashMap<(HawkIdentifier, i32), CollectionLock>,
     transaction: Option<TransactionSelector>,
+    /// Behind Vec so commit can take() it (maybe commit() should consume self
+    /// instead?)
```
jrconlin: I kind of agree that `commit()` consuming self is more in spirit with how I understand ownership should work. You shouldn't be able to do anything else with the session once the commit has happened.

src/db/spanner/support.rs
```diff
@@ -204,3 +211,76 @@ pub fn bso_from_row(row: Vec<Value>) -> Result<results::GetBso> {
...
+
+#[cfg(not(any(test, feature = "db_test")))]
+pub fn bso_to_insert_row(
+    user_id: &HawkIdentifier,
+    collection_id: i32,
+    bso: params::PostCollectionBso,
+    now: SyncTimestamp,
+) -> Result<ListValue> {
+    use super::models::DEFAULT_BSO_TTL;
+    use protobuf::well_known_types::NullValue;
```
jrconlin: nit: I wonder if we should put spaces around these function `use`s just to help them stand out a bit more.

pjenvey: I just switched them back up to the top, since there was the 2 left (wrapped in cfg! directives)

PR #288, opened 2019-10-17
bug: accept weighted content-type headers

### LINKED ISSUE #287: Improve Accept header parsing (revisited)
We missed another couple features of the Accept header format, causing potential 406 Not Accepted responses that shouldn't happen:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept

- They can potentially include multiple, comma separated values
- We don't correctly handle an `Accept: */*`, which should return the default of Json

The loadtester is currently hitting the latter issue, I believe its aiohttp underneath defaults to an */* Accept header value.

jrconlin:
Closes #287

### DIFF
--- a/src/web/extractors.rs
+++ b/src/web/extractors.rs
@@ -116,8 +116,8 @@ fn get_weighted_header(
             let weights: Vec<&str> = opt_weight[1].split('=').collect();
             // if the weight is malformed, ignore this choice.
             if weights.len() < 2 {
-                continue
-            } 
+                continue;
+            }
             f32::from_str(weights[1]).unwrap_or_else(|_| 0.0)
         } else {
             1.0
@@ -1630,7 +1630,6 @@ mod tests {
             "text/plain",
         );
         assert_eq!(selected, "text/plain".to_owned());
-
     }
 
     #[test]
--- a/src/web/extractors.rs
+++ b/src/web/extractors.rs
@@ -118,7 +118,7 @@ fn get_weighted_header(
             if weights.len() < 2 {
                 continue;
             }
-            f32::from_str(weights[1]).unwrap_or_else(|_| 0.0)
+            f32::from_str(weights[1]).unwrap_or(0.0)
         } else {
             1.0
         };
--- a/src/web/extractors.rs
+++ b/src/web/extractors.rs
@@ -42,6 +42,9 @@ const BSO_MAX_TTL: u32 = 31_536_000;
 const BSO_MAX_SORTINDEX_VALUE: i32 = 999_999_999;
 const BSO_MIN_SORTINDEX_VALUE: i32 = -999_999_999;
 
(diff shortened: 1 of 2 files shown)


tublitzed:
Hey, @jrconlin would you mind filling out the PR template here so it's clear what this is for? Including the link is really helpful as then Github will link them as well so we can easily go from issue to PR and back and forth.

jrconlin:
Hrm. Not really sure why github doesn't include the actual change as part of the top comment.

https://github.com/mozilla-services/syncstorage-rs/pull/288/commits/48a5d4a6f36eee2f591b8f77e9d8d017bb3a580a contains all the criteria. 

> commit 48a5d4a6f36eee2f591b8f77e9d8d017bb3a580a
> Author: jrconlin <me+crypt@jrconlin.com>
> Date:   Thu Oct 17 14:44:25 2019 -0700
> 
>    bug: accept weighted content-type headers
> 
>    Closes #287

That was generated from git CLI, so maybe github is injecting it's own? Not really sure.

src/web/extractors.rs
```diff
@@ -88,11 +91,39 @@ impl BatchBsoBody {
...
+    key: HeaderName,
+    accepted: &[&str],
+    default: &'static str,
+) -> String {
+    let def_hv = HeaderValue::from_static(default);
+    let hv_raw = std::str::from_utf8(headers.get(key).unwrap_or(&def_hv).as_bytes())
         .unwrap_or_else(|_| "invalid");
-    let ct_parts: Vec<&str> = ct_raw.split(';').collect();
-    ct_parts[0].trim_end().to_owned()
+    let hv_parts: Vec<&str> = hv_raw.split(',').collect();
```
pjenvey: nit: no need to collect since it's just iterated over

src/web/extractors.rs
```diff
@@ -88,11 +91,39 @@ impl BatchBsoBody {
...
+    let mut pick: String = default.to_owned();
+    for choice in hv_parts {
+        let opt_weight: Vec<&str> = choice.split(';').collect();
+        let opt = opt_weight[0].trim_end().to_lowercase();
+        if accepted.contains(&&opt.as_ref()) {
+            return opt;
+        }
+        let weight = if opt_weight.len() > 1 {
+            let weights: Vec<&str> = opt_weight[1].split('=').collect();
+            i64::from_str(weights[1]).unwrap_or_else(|_| 0)
```
pjenvey: weights.len() could be 1 here

nit: if this still exists, could use unwrap_or(0) instead? (or am I wrong, I thought clippy caught this?)

jrconlin: clippy usually wants the `unwrap_or_else()` rather than `unwrap_or()` so I defaulted to that.

pjenvey: only when the or arg is some kind of fn call, I guess it just doesn't notice the inverse

Issue #284, opened 2019-10-17
Write sql to get stg populated with data


tublitzed:

fzzzy:
I believe we use the load tester to do this, with an environment variable to tell it not to delete.

pjenvey:
> I believe we use the load tester to do this, with an environment variable to tell it not to delete.

We're going to try implementing a more direct method of populating junk data, which should hopefully load faster than the load test would

jrconlin:
Tied to bug https://bugzilla.mozilla.org/show_bug.cgi?id=1589484

jrconlin:
![image](https://user-images.githubusercontent.com/35755/67507058-0f36c480-f643-11e9-89e6-0e85199ee7e0.png)

A few notes: 
* Ballast stuffing took about 4 hours (the CPUs for the EC2 and Spanner were NOT happy).
* I will probably release another tools patch that includes a few tweaks to the tool set.
* The original script stuffed a lot of records under a small set of keys. Next tool update will generate more fake users to keep the keys a bit cooler. We can still easily spot and remove ballast later if need be.
* Ops will keep the stage box in place for little longer, mostly to deal with potential ballast reloading.

PR #292, opened 2019-10-18
feat: add spanner tools

jrconlin:
Add a set of python based tools for managing data in the spanner
database.

Issue: #119, #284

- [ ] **Test cases** included in the change (if appropriate)

### DIFF
--- a/tools/spanner/purge_ttl.py
+++ b/tools/spanner/purge_ttl.py
@@ -5,16 +5,35 @@
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 import os
+from urllib import parse
+
 from google.cloud import spanner
 
-# Change these to match your install. 
-instance_id = os.environ.get("INSTANCE_ID", 'spanner-test')
-database_id = os.environ.get("DATABASE_ID", 'sync_stage')
+# Change these to match your install.
 
 client = spanner.Client()
 
 
+def from_env():
+    try:
+        url = os.environ.get("SYNC_DATABASE_URL")
+        if not url:
+            raise Exception("no url")
+        purl = parse.urlparse(url)
+        if purl.scheme == "spanner":
+            path = purl.path.split("/")
+            instance_id = path[-3]
+            database_id = path[-1]
+    except Exception as e:
+        # Change these to reflect your Spanner instance install
+        print("Exception {}".format(e))
+        instance_id = os.environ.get("INSTANCE_ID", "spanner-test")
+        database_id = os.environ.get("DATABASE_ID", "sync_stage")
+    return (instance_id, database_id)
+
+
 def spanner_read_data(request=None):
+    (instance_id, database_id) = from_env()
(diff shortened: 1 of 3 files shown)


tools/spanner/purge_ttl.py
```diff
@@ -0,0 +1,24 @@
+# Purge Expired TTLs
+#
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+from google.cloud import spanner
+
+# Change these to match your install. 
+instance_id = 'spanner-test'
```
pjenvey: Let's log an issue for a couple enhancements to this I can think of:

- it should delete from the batch table (I'm fine deferring this for now as I might be altering its schema shortly)
- these should probably be read from env variables. I'm tempted to say it should match syncstorage-rs's SYNC_DATABASE_URL format for consistency sake (though that will require url parsing code, the python client doesn't use that format)

tools/spanner/purge_ttl.py
```diff
@@ -0,0 +1,24 @@
...
+database_id = 'sync_kid'
+
+client = spanner.Client()
+
+
+def spanner_read_data(request):
+    instance = client.instance(instance_id)
+    database = instance.database(database_id)
+
+    query = 'DELETE FROM bso WHERE ttl < TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 30 DAY)'
```
pjenvey:
```suggestion
    query = 'DELETE FROM bso WHERE expiry < TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 30 DAY)'
```

jrconlin: Oh, right! we changed the name. thanks!

jrconlin:
ok, tweaked up the tools a bit on the last commit.

Made the purge tool a bit more "real". Should still work if someone wanted to use it as a cloud function too.

tools/spanner/purge_ttl.py
jrconlin: Meh, I'm ok with altering this PR to at least read the sync data from two env vars. 
I'm not quite sure about parsing the sync_database_url, a bit since I don't know if the cloud func can pull in the urllib. I may mess around a bit to see.

pjenvey: Good point. These probably need a requirement file anyway, unless google-cloud-spanner is pre-installed on GCE nodes. write_batch.py may not necessarily run from there anyway.

I forgot to add this one to my orig. repo: https://github.com/pjenvey/spanner-tests/blob/master/requirements.txt

jrconlin: looks like `urllib` is available for cloud functions. Added DSN parser. 

thanks! Good idea.

PR #303, opened 2019-10-24
feat: workaround timeouts w/ larger db conn sizes

### LINKED ISSUE #302: Lengthen spanner's db pool connection_timeout
In our current pool impl, one grpc Channel is created per db connection.

Channel creation combined with Spanner session creation take some time. When configured with a larger db pool size (around 200 or more), establishing these connections may timeout with the default connection timeout of 30 seconds. This causes the pool initialization (and the whole app) to fail.

This problem should go away when we multiplex many connections over one grpc channel (establishing the grpc Channel takes a larger chunk of the time). Until that's finished, let's workaround this with a bump to the pool's connection timeout.

pjenvey:
by (temporarily) increasing the connection timeout value

Closes #302

### DIFF
--- a/src/db/spanner/pool.rs
+++ b/src/db/spanner/pool.rs
@@ -2,6 +2,7 @@ use std::{
     collections::HashMap,
     fmt,
     sync::{Arc, RwLock},
+    time::Duration,
 };
 
 use diesel::r2d2;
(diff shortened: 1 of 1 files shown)


pjenvey:
Probably this setting should be configurable, but I'm hoping it's temporary then it can go away

src/db/spanner/pool.rs
```diff
@@ -54,7 +55,9 @@ impl SpannerDbPool {
 
     pub fn new_without_migrations(settings: &Settings, metrics: &Metrics) -> Result<Self> {
         let manager = SpannerConnectionManager::new(settings)?;
-        let builder = r2d2::Pool::builder().max_size(settings.database_pool_max_size.unwrap_or(10));
+        let builder = r2d2::Pool::builder()
+            .max_size(settings.database_pool_max_size.unwrap_or(10))
+            .connection_timeout(Duration::from_secs(120));
```
jrconlin: nit: This should probably be a setting. Future feature.

PR #314, opened 2019-11-01
refactor: schema renames (again)

### LINKED ISSUE #313: Schema cleanups
Some last minute cleanups to the schema while we have the chance (mostly to our primary key columns, which we can't change later):

- Spanner best practice is prefixing any primary key column with the table name. As when interleaving tables share primary key columns, their names must match exactly (so a name like `id` becomes ambiguous in a child table)
- I guess `bso` -> `bsos` since other tables are plural
- batches could probably use an expiry index?

Unfortunately this requires a new data preload of stage.

pjenvey:
- always prefix primary key names with their table name (id -> bso_id,
  collection_id). a best practice in spanner, as when tables
  interleave, their shared id names must always match
- bso -> bsos (every other table name's plural)
- give batches an expiry index

Closes #313

- [ ] **Test cases** included in the change (if appropriate)

### DIFF
--- a/spanner-2019-10-01.ddl
+++ b/spanner-2019-10-01.ddl
@@ -17,34 +17,37 @@ CREATE TABLE user_collections (
   modified TIMESTAMP   NOT NULL,
 ) PRIMARY KEY(fxa_uid, fxa_kid, collection_id);
 
-CREATE TABLE bso (
+CREATE TABLE bsos (
   fxa_uid STRING(MAX)  NOT NULL,
   fxa_kid STRING(MAX)  NOT NULL,
   collection_id INT64  NOT NULL,
-  id STRING(64)        NOT NULL,
+  bso_id STRING(64)    NOT NULL,
+
   sortindex INT64,
+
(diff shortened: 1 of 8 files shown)


spanner-2019-10-01.ddl
```diff
@@ -68,20 +71,23 @@ CREATE TABLE batches (
...
 
   sortindex INT64,
   payload STRING(MAX),
   ttl INT64,
-)    PRIMARY KEY(fxa_uid, fxa_kid, collection_id, batch_id, id),
+)    PRIMARY KEY(fxa_uid, fxa_kid, collection_id, batch_id, batch_bso_id),
   INTERLEAVE IN PARENT batches ON DELETE CASCADE;
 
--- batch_bso's bso fields are nullable as the batch upload may or may
+-- batch_bsos bso fields are nullable as the batch upload may or may
```
jrconlin: nit: heh, I'm guessing that this should be `batch_bsos'` to indicate ownership.

PR #324, opened 2019-11-08
feat: check spanner commit size, error out if "too large"

### LINKED ISSUE #320: Audit max_total_bytes per Spanner's mutation per commit limit
Followup to #298:

I haven't double checked our max_total_bytes value, but Spanner has a 100mb commit size along w/ its 20k mutation limit.

Let's double check our value is under this limit for spanner

jrconlin:
* Also made "no_auth" a feature flag function
* included dependencies for golang, protobuf and libgrpcdev into setup
docs.
* may revisit some tools to make testing and debugging stuff like this
easier.

Closes #320

- [ ] **Test cases** included in the change (if appropriate) (unit tests code around this. live value is huge. may revisit)


src/web/auth.rs
```diff
@@ -87,27 +87,32 @@ impl HawkPayload {
         let request = RequestBuilder::new(method, host, port, path).request();
 
         // Toggle the following comments to disable auth (useful for local integration testing)
```
pjenvey: now dead comment

jrconlin:
hmm, good point. I'm not quite sure the relationship between MAX_TOTAL_BYTES and MAX_SPANNER_LOAD_SIZE (e.g. how would integers be compressed down? should this be variable based on if you're using MySQL vs Spanner, etc.) 

Gimme a bit.

src/db/spanner/models.rs
```diff
@@ -1153,19 +1158,34 @@ impl SpannerDb {
...
                     .or_insert_with(|| vec![])
                     .push(values);
             } else {
                 let values = bso_to_insert_row(&user_id, collection_id, bso, timestamp)?;
+                load_size += values.compute_size() as usize;
                 inserts.push(values);
             }
         }
+        if load_size > MAX_SPANNER_LOAD_SIZE {
+            self.metrics.clone().incr("sync.error.tooMuchData");
```
pjenvey: Almost forgot this (and it's probably a separate issue). Using the "sync" prefix here, but I've seen "syncstorage" elsewhere. Probably want to standardize on "syncstorage" like the python version uses

jrconlin: Agreed. Thanks!

jrconlin: (Also going to normalize the `timer.*` var calls as `metrics.*` to make them easier to find in the sources.)

PR #332, opened 2019-11-14
fix: bump the db worker thread pool size

### LINKED ISSUE #330: Improve handling of larger db POOL_MAX_SIZE values under load
With a DATABASE_POOL_MAX_SIZE value of 300, one particular node on staging emits an excessive amount of time outs when checking out a db conn from the pool:

https://sentry.prod.mozaws.net/operations/syncstorage-stage/issues/6625203/?query=is:unresolved
https://sentry.prod.mozaws.net/operations/syncstorage-stage/issues/6625204/?query=is:unresolved

It looks like we have a bad interaction between the limited size of the db worker thread pool (only 50 threads) combined with the recent bump of the connection timeout value to 2 minutes.

I think part of what we're seeing is: with the Pool waiting for 2 minutes for a conn that never becomes available within a thread, the thread worker pool becomes starved, preventing enough db work from progressing, exacerbating the lack of connections.

pjenvey:
for now, matching the db connection pool size

- and bump r2d2's separate, internal thread pool for connection
creation/reaping; a better solution for pool initialization timeouts
(#302) vs increasing the pool's connection_timeout.

the larger timeout could exacerbate worker thread pool starvation under
load

- don't reinit the grpc Env per conn

Closes #330, #302


db-tests/rust-toolchain
```diff
@@ -1 +1 @@
-beta-2019-10-31
+stable
```
jrconlin: heh, good point.

src/db/spanner/pool.rs
```diff
@@ -55,9 +55,15 @@ impl SpannerDbPool {
 
     pub fn new_without_migrations(settings: &Settings, metrics: &Metrics) -> Result<Self> {
         let manager = SpannerConnectionManager::new(settings)?;
+        let max_size = settings.database_pool_max_size.unwrap_or(10);
+        // r2d2 creates max_size count of db connections on creation via its
+        // own thread_pool. increase its default size to quicken their
+        // creation, accommodating large max_size values (otherwise it may
+        // timeout)
+        let r2d2_thread_pool = ScheduledThreadPool::new((max_size as f32 * 0.05) as usize);
```
jrconlin: I'm confused here. In the minimum case, we're taking 5% of 10, which rounds out to be 0. If we're trying to increase the available thread pools, why are we taking such a small percentage of the requested number?

pjenvey: thanks, was going to add a max here and forgot

PR #337, opened 2019-11-15
feat: Include user agent info in metric tags

### LINKED ISSUE #329: Include user agent info in Sentry events
For diagnosing Sentry issues, we should include information about the user agent that triggered the event.

autopush-rs has a good example of this, using the woothee-rust crate to parse the user agent information out into canonical os/browser family/versions

jrconlin:
* also update some dependencies

Closes #329


src/server/metrics.rs
```diff
@@ -23,6 +24,7 @@ pub struct MetricTimer {
 #[derive(Debug, Clone)]
 pub struct Metrics {
     client: Option<StatsdClient>,
+    tags: Option<Tags>,
```
pjenvey: FYI HashMap::new doesn't allocate, so the Option dance doesn't seem worth it

jrconlin: Ah, actually, thank you. I need to make sure that I consistently use type `Tags` here.

Issue #338, opened 2019-11-15
Review __heartbeat__ endpoint (and expand coverage if possible)


tublitzed:
Based on [Ops requirements](https://github.com/mozilla-services/services-engineering/issues/15) from @Micheletto, we should take another look at the heartbeat endpoint and make sure there's not any more useful information we can pass along.

Ie, does [this level of detail](https://sync-1-us-west1-g.sync.services.mozilla.com/__heartbeat__) include enough for us to feel confident that the service is running without issue? Do we need more checks around spanner availability/etc?

Let's add @Micheletto to the review here to make sure we're covering stuff that's going to be helpful from a monitoring perspective.

PR #345, opened 2019-11-19
feat: improve heartbeat to query db

jrconlin:
Closes [#338](https://github.com/mozilla-services/syncstorage-rs/issues/338)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,8 +31,9 @@ env_logger = "0.7.1"
 failure = "0.1.6"
 # do not upgrade futures!
 futures = "0.1.28"
+futures03 = {package = "futures", version="0.3", default-features=false, features=["compat"]}
 googleapis-raw = { version = "0", path = "../mozilla-rust-sdk/googleapis-raw" }
-grpcio = { version = "0.5.0-alpha.1" }
+grpcio = { version = "0.5.0-alpha.5" }
 lazy_static = "1.4.0"
 hawk = "3.0.0"
 hkdf = "0.8.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,6 @@ env_logger = "0.7.1"
 failure = "0.1.6"
 # do not upgrade futures!
 futures = "0.1.28"
-futures03 = {package = "futures", version="0.3", default-features=false, features=["compat"]}
 googleapis-raw = { version = "0", path = "../mozilla-rust-sdk/googleapis-raw" }
 grpcio = { version = "0.5.0-alpha.5" }
 lazy_static = "1.4.0"
--- a/src/db/mod.rs
+++ b/src/db/mod.rs
@@ -147,6 +147,8 @@ pub trait Db: Send + Debug {
 
     fn box_clone(&self) -> Box<dyn Db>;
 
+    fn check(&self, _: params::Check) -> DbFuture<results::Check>;
+
     /// Retrieve the timestamp for an item/collection
     ///
     /// Modeled on the Python `get_resource_timestamp` function.
--- a/src/db/mod.rs
+++ b/src/db/mod.rs
@@ -147,7 +147,7 @@ pub trait Db: Send + Debug {
 
     fn box_clone(&self) -> Box<dyn Db>;
 
-    fn check(&self, _: params::Check) -> DbFuture<results::Check>;
+    fn check(&self) -> DbFuture<results::Check>;
 
     /// Retrieve the timestamp for an item/collection
     ///
--- a/src/db/params.rs
+++ b/src/db/params.rs
@@ -99,6 +99,8 @@ bso_data! {
     GetBsoTimestamp {},
 }
 
+pub struct Check {}
+
 #[derive(Debug, Default, Queryable)]
 pub struct Batch {
     pub id: String,
--- a/src/db/params.rs
+++ b/src/db/params.rs
@@ -99,8 +99,6 @@ bso_data! {
     GetBsoTimestamp {},
 }
 
-pub struct Check {}
-
 #[derive(Debug, Default, Queryable)]
 pub struct Batch {
     pub id: String,
--- a/src/db/spanner/models.rs
+++ b/src/db/spanner/models.rs
@@ -1395,7 +1395,7 @@ impl SpannerDb {
         Ok(result)
     }
 
-    fn check_sync(&self, _: params::Check) -> Result<results::Check> {
+    fn check_sync(&self) -> Result<results::Check> {
         // TODO: is there a better check than just fetching UTC?
         self.sql("SELECT CURRENT_TIMESTAMP()")?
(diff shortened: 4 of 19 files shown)


jrconlin:
The new successful response to `__heartbeat__` will be:

```json
{"version": "*version info*",
 "state_data": "good",
 "database": "good",
 "status": "good"
}```
(item order is NOT guaranteed)
`status`: indicates overall system health. 
* `good`: all is well
* `fail`: system should not be used.

`state_data`: indicates if the application internal state is accessible. Bad state prevents the database from being accessible and is it's own error.

`database` can return
* `good`: all is well
* `warn`: the database is present, but returned a value that was unexpected
* `fail`: the database is not present or no connection is available to database

In addition, optional `database_msg` returns the associated error message with a `warn` or `fail` database state.

File: src/db/mod.rs [hunk 1]
```diff
@@ -147,6 +147,8 @@ pub trait Db: Send + Debug {
 
     fn box_clone(&self) -> Box<dyn Db>;
 
+    fn check(&self) -> Result<bool, DbError>;
```
pjenvey: shouldn't we go w/ async here like other db methods?

src/db/spanner/models.rs
```diff
@@ -1552,4 +1552,11 @@ impl Db for SpannerDb {
     fn clear_coll_cache(&self) {
         self.coll_cache.clear();
     }
+
+    fn check(&self) -> Result<bool> {
+        // TODO: is there a better check than just fetching UTC?
+        self.sql("SELECT UNIX_SECONDS(CURRENT_TIMESTAMP())")?
+            .execute(&self.conn)?;
```
pjenvey: I suppose you could also call `.one()?` here just to be strict.

nit: `UNIX_SECONDS` conversion isn't really needed

src/web/handlers.rs
```diff
@@ -376,3 +376,52 @@ pub fn put_bso(bso_req: BsoPutRequest) -> impl Future<Item = HttpResponse, Error
...
+    use crate::server::ServerState;
+
+    let mut checklist = HashMap::new();
+
+    checklist.insert(
+        "version",
+        Value::String(env!("CARGO_PKG_VERSION").to_owned()),
+    );
+    checklist.insert("status", Value::String("good".to_owned()));
+    checklist.insert("state_data", Value::String("good".to_owned()));
```
pjenvey: I don't think there's much value in reporting on ServerState. In practice it should always be accessible, only not if we've really badly misconfigured something and that surely would be caught much earlier than a healthcheck call.

Switching top level "status" to "fail" in that case is fine/accurate, though

src/web/handlers.rs
```diff
@@ -376,3 +376,52 @@ pub fn put_bso(bso_req: BsoPutRequest) -> impl Future<Item = HttpResponse, Error
...
+ *
+ * * version - Version information
+ * * state_data - Is the app state data present ("good" | "fail")
+ * * database - Database connection and status ("good" | "warn" | "fail")
+ * * status - Overall system status ("good" | "fail")
+ */
+pub fn heartbeat(hb: HeartbeatRequest) -> impl Future<Item = HttpResponse, Error = Error> {
+    use crate::server::ServerState;
+
+    let mut checklist = HashMap::new();
```
pjenvey: just an FYI: [serde's json!](https://docs.serde.rs/serde_json/macro.json.html) usually makes this kind of thing easier. You get back a `Value` which *I think* can be passed to the final `json()` call (as it implements Serialize)

src/db/mod.rs
jrconlin: probably, although not really looking forward to juggling either states.

src/web/handlers.rs
jrconlin: I'm kind of on the fence. I agree that "state_data" isn't super high value, until the server is returning fails and the db info looks correct and we're wondering what the heck is going on.

src/db/spanner/models.rs
jrconlin: Yeah, the original idea was to do the same as mysql and track the number of seconds that things were up. Kinda useless now, though.

Cargo.toml
```diff
@@ -31,8 +31,9 @@ env_logger = "0.7.1"
 failure = "0.1.6"
 # do not upgrade futures!
 futures = "0.1.28"
+futures03 = {package = "futures", version="0.3", default-features=false, features=["compat"]}
```
jrconlin: Crap. let me get rid of this...

src/web/extractors.rs
```diff
@@ -863,6 +863,30 @@ impl FromRequest for ConfigRequest {
...
+
+impl FromRequest for HeartbeatRequest {
+    type Config = ();
+    type Error = Error;
+    type Future = Box<dyn Future<Item = Self, Error = Self::Error>>;
+
+    fn from_request(req: &HttpRequest, _: &mut Payload) -> Self::Future {
+        let headers = req.headers().clone();
+
+        let state = req.app_data::<ServerState>().unwrap();
```
pjenvey: not terribly important but let's avoid this unwrap. You can copy one of the other ServerState gets in this file using a match statement so it can return a `future::err`,

src/db/params.rs
```diff
@@ -99,6 +99,8 @@ bso_data! {
     GetBsoTimestamp {},
 }
 
+pub struct Check {}
```
pjenvey: nit: this could be killed (have `check` take no args). Is there plans for it in the future?

jrconlin: Not really. For consistency (and a bit of sloth) I was using the  async to sync macros which require a matching param and result pair.

pjenvey: Ah right, I was thinking we have a few trait methods that take no args (e.g. commit). I forgot they can't/don't use the macro

jrconlin: No big, I removed the function from the macros. If we ever need to add input params, we can swap it back.

src/web/handlers.rs
```diff
@@ -402,3 +402,36 @@ pub fn put_bso(bso_req: BsoPutRequest) -> impl Future<Item = HttpResponse, Error
...
+                );
+            };
+            let status = if result { "Ok" } else { "Err" };
+            checklist.insert("status".to_owned(), Value::from(status));
+            HttpResponse::Ok().json(checklist)
+        }
+        Err(e) => {
+            checklist.insert("status".to_owned(), Value::from("Err"));
+            checklist.insert("database".to_owned(), Value::from("Unknown"));
+            checklist.insert("status_msg".to_owned(), Value::from(e.to_string()));
```
pjenvey: In megaphone I originally included the error like this in the json, then opted to remove it from there just in case the db error required sanitizing. A case like this is a good example, exposing the GCP project/instance info: https://sentry.prod.mozaws.net/operations/syncstorage-stage/issues/6625678/?query=session

Instead it's just logged

Issue #344, opened 2019-11-19
Combine docs/setup.md with README and add sample config file.


tublitzed:
Gonna combine into top level README to make instructions a bit more discoverable and add a sample git-ignored config file to help with local dev.

PR #349, opened 2019-11-21
docs: combine setup instructions into one master doc

tublitzed:
## Description

I noticed we had lots of useful info in docs/setup.md, so I opted to combine into the main README. You'll see I left some of the more detailed config options in there, but renamed the file, and am now linking to it from the top level README so it's more discoverable.

I also am exploring ways to make running this project locally a bit easier. While still a WIP, this PR does the following: adds instructions and example files to assist with config options and env vars that are required each time.

I really wanted to get _all_ the ENV vars into the config/local.toml file, but I found that the config Crate doesn't actually treat them the same as env vars. So, we need both for now. Eventually, I would love to get us over to something like [dotenv](https://github.com/dotenv-rs/dotenv) but holding on that for now since that project isn't quite stable yet, and there doesn't (yet) appear to be another Rust equivalent.

I'm also tweaking the PR template to reduce boilerplate. This PR serves as an example of what I'm thinking/going for. Feedback welcome :)

## Testing

Try testing locally by:

1. `cp config/local.example.toml config/local.toml`
2. Open your new `local.toml` file and replace DB with your valid MySQL creds.
3. Now, `make run-local`. Ensure you can run and are seeing log output.

## Issue(s)

Closes [#344](https://github.com/mozilla-services/syncstorage-rs/issues/344).


CONTRIBUTING.md
```diff
@@ -33,6 +33,7 @@ When submitting a PR:
 - Ensure your changes do not reduce code coverage of the test suite.
 - Please do not include merge commits in pull requests; include only commits
   with the new relevant code.
+- PR naming conventions - begins with type (fix, feature, doc, chore, etc) and a short description with no period.
```
tublitzed: I opted to just move this into the CONTRIBUTING.md file so we don't have to see it each time in PRs.

docs/config.md
```diff
@@ -0,0 +1,43 @@
+# Configuration
```
tublitzed: This is mainly just moved content from the old [setup.md file](https://github.com/mozilla-services/syncstorage-rs/blob/master/docs/setup.md)

PR #356, opened 2019-11-30
README.md: Add missing requirements

Mic92:
## Description

Just extend documentation; Nothing to test beyond.


jrconlin:
Thanks for adding these! We're planning on doing a clean up sweep once things are stable enough, but we definitely appreciate when folks add bits like this we miss. It's way too easy for us to miss things that our machines already have configured.

Issue #342, opened 2019-11-18
Firefox Preview doesn't sync with other FF browsers using a Spanner account


tublitzed:
Via Stefan from Bugzilla [1596378](https://bugzilla.mozilla.org/show_bug.cgi?id=1596378)

Devices:

    Nokia 6 (Android 7.1.1);
    Samsung Galaxy Note S8 (Android 9);

Build:

    Firefox Preview 2.3.0 (#13031728);

Prerequisites:
Login on multiple devices with the same Firefox account.

Steps to reproduce:

    Make sure that you are connected to the account in test;
    On both devices, spend some time to create some new browsing data;
    Sync first device than the second device;
    Pay attention to what data was synced between those two devices.

Expected results:
All of the new data created while browsing on those two devices should now be synced.

Actual results:
Fenix browser doesn't sync with the other types of browsers(FF for desktop, FF for iOS, FF for Android).

Notes:

    This issue is reproducible only with the Spanner accounts;
    The Fenix browser starts the sync and appears as synced on other devices on the manage account section.
    The Last sync data on the Firefox Preview browser it says an old data(ex. Last synced: Jan 1, 1970);
    This issue is not reproducible with FF for Android, FF for Desktop or FF for iOS.

jrconlin:
I think I've identified the error in the logs at least:
```
I/sync15::client: Request "/1.5/###/info/configuration" has incoming x-last-modified ServerTimestamp(1573697264890)
I/sync15::client: Request "/1.5/###/info/collections" has incoming x-last-modified ServerTimestamp(1573697264890)
I/sync15::client: Request "/1.5/###/storage/meta/global" has incoming x-last-modified ServerTimestamp(1573684707540)
D/sync15::client: Got meta global with modified = 1573684707.54; last-modified = 1573684707.54
I/sync15::state: Have info/collections and meta/global. Computing new engine states
D/sync15::state: compute_engine_states: input EngineStateInput { local_declined: {"greasemonkey", "passwords"}, remote: Some(RemoteEngineState { info_collections: {"forms", "meta", "bookmarks", "prefs", "passwords", "crypto", "tabs", "history", "addons", "clients"}, declined: {"greasemonkey", "passwords"} }), user_changes: {} }
W/sync15::state: Remote state contains engines which are in both info/collections and meta/global's declined: {"passwords"}
[...]

PR #359, opened 2019-12-03
 feat: Handle empty bso body on /meta/global

jrconlin:
## Description

This adds some additional debug around handling BSOs to try and identify
the source of the server returning 400 on PUT `/1.5/xxx/storage/meta/global`.
There may be a related client bug where it's sending no BSO on a
meta/global PUT, thus the very specific fix.

* also new clippy, new crap to clean up.

## Testing

Since we don't have a lot of tests that check full path, the test I've been using is to use `/tools/hawk/make_hawk_token.py` to generate a valid `Authorization` header

then call 
```
curl -v -X PUT "http://localhost:8000/1.5/1/storage/meta/global" \
    -H "content-type: application/json" \
    -H "Authorization: $AUTH"
```

## Issue(s)

Issue #342

### DIFF
--- a/tools/hawk/README.md
+++ b/tools/hawk/README.md
@@ -1,6 +1,6 @@
 # Make a Hawk compatible Auth header
 
-The best way to install this is probably to set up a python virtual
+1) The best way to install this is probably to set up a python virtual
 env.
 
 `python3 -m venv venv && venv/bin/pip install -r requirements.txt`
@@ -8,5 +8,17 @@ env.
 this will create a python virtual environment in the `/venv` directory.
 
 *Note* You may need to install `python3-venv` for the above to work.
+
+Once the virtual env is installed, run `. venv/bin/activate`. This
+will ensure that calls to python and python tools happen within this
+virutal environment.
+
+2) install the requirements using:
+
+`venv/bin/pip install -r requirements.txt`
+
+3) To create a Token Header:
+
+`venv/bin/python make_hawk_token.py`
 
 Use `-h` for help.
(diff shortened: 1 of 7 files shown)


tublitzed:
When I run `python3 -m venv venv && venv/bin/pip install -r requirements.txt` in `tools/hawk` and then `python3 make_hawk_token.py`, I see this: ```  File "make_hawk_token.py", line 21, in <module>
    import hawkauthlib
ModuleNotFoundError: No module named 'hawkauthlib'``` - I'm guessing I'm missing some setup instruction for running that script?

jrconlin:
@tublitzed Hrm. Interesting. 

Are you running python from inside of the virtualenv? (I'll add some additional instructions in the README to make sure that folks do that.)

tublitzed:
@jrconlin - that was it. `source venv/bin/activate` fixed it

tublitzed:
I just tried using both CURL and Postman, getting a 401:

Screen Shot 2019-12-04 at 1 52 43 PM

tools/hawk/README.md
```diff
@@ -9,4 +9,16 @@ this will create a python virtual environment in the `/venv` directory.
 
 *Note* You may need to install `python3-venv` for the above to work.
 
-Use `-h` for help.
\ No newline at end of file
+Once the virtual env is installed, run `. venv/bin/activate`. This
```
tublitzed: thanks for adding this!

jrconlin:
I've updated the test instructions to duplicate the local testing I did. I will follow up with another Tools PR to include scripts.

PR #361, opened 2019-12-04
bug: Don't run mysql tests when testing spanner.

### LINKED ISSUE #360: MySQL tests called when running spanner
the `db/mysql/test.rs` file is called by `cargo test` regardless of the DSN we're using, so the mysql specific test fails.

jrconlin:
## Description

Added logic to prevent mysql tests from running when tests use a Spanner DSN

## Testing

run `cargo test` with  `SYNC_DATABASE_URL` pointing to a Spanner DSN 

## Issue(s)

Closes #360

### DIFF
--- a/src/db/mysql/test.rs
+++ b/src/db/mysql/test.rs
@@ -9,6 +9,7 @@ use diesel::{
     Connection, ExpressionMethods, QueryDsl, RunQueryDsl,
 };
 use env_logger;
+use url::Url;
 
 use crate::db::mysql::{
     models::{MysqlDb, Result},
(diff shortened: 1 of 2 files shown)


fzzzy:
Hmm, except circle dies because of a lint thing.

tublitzed:
@jrconlin the testing instructions say to use MySQL, but this change is supposed to be for the spanner version of the tests, y?

jrconlin:
@tublitzed Ah, correct. I've modified the comment.

tublitzed:
Sweet, thanks JR.

Issue #286, opened 2019-10-17
Support Spanner migrations


pjenvey:
There should be an automated migration system for applying changes to the Spanner's schema.

Additionally we should be populating the initial collections table with the standard collections when creating the table.

I believe Spanner schema alterations require a different API call vs a regular sql/DML query/statement (which we'd need to use for populating the collections table).

I wonder if diesel's migrations could shoehorned for this, but is likely a misfit, especially with the above requirements.

PR #362, opened 2019-12-05
fix: oops, properly interleave this index

### LINKED ISSUE #318: Spanner's 20k mutation limit can be triggered by deletes
From the mutation limit error message on Commit:

_"Delete and delete range operations count as one mutation regardless of the number of columns affected."_

however,

**_"The total mutation count includes any changes to indexes that the transaction generates. Please reduce the number of writes, or use fewer indexes."_**

From https://cloud.google.com/spanner/docs/modify-mutation-api#deleting_rows_in_a_table

_"For example, you usually can't delete all of the rows in a large table in a single transaction if the table has a secondary index defined on it. The secondary index causes the transaction to exceed the transaction size limits."_

The bsos table currently has 2 secondary indices. I can trigger the mutation limit w/ a delete =~ 20k total rows. Partitioned DML doesn't seem to be an option here as it operates in its own separate, limited transactions.

This affects our 2 bulk delete operations:

- delete_storage (deletes all user's data)
- delete_collection

pjenvey:
Closes #318

### DIFF
--- a/spanner-2019-10-01.ddl
+++ b/spanner-2019-10-01.ddl
@@ -37,7 +37,7 @@ CREATE TABLE bsos (
 INTERLEAVE IN user_collections;
 
     CREATE INDEX BsoExpiry
-        ON bsos(expiry),
+        ON bsos(fxa_uid, fxa_kid, collection_id, expiry),
 INTERLEAVE IN user_collections;
 
 CREATE TABLE collections (


jrconlin:
Will this require a new DB instance?

spanner-2019-10-01.ddl
```diff
@@ -37,7 +37,7 @@ CREATE TABLE bsos (
 INTERLEAVE IN user_collections;
 
     CREATE INDEX BsoExpiry
-        ON bsos(expiry),
+        ON bsos(fxa_uid, fxa_kid, collection_id, expiry),
```
tublitzed: Would it make sense to make a new DDL file with a new date instead of changing the old one? (Just so we have a historical record of changes?)

pjenvey: I've been reluctant to do something like this (or applying changes as DDL modifications) until #286 is implemented. It might look similar to our [mysql backend migrations structure](https://github.com/mozilla-services/syncstorage-rs/tree/master/migrations) but that's not 100% clear it's implemented (and we always have git history? :sob:)

#286 should become higher priority once we're closer to a steadier prod and other even higher priority bits are finished.

pjenvey:
> Will this require a new DB instance

Nope, FYI: https://bugzilla.mozilla.org/show_bug.cgi?id=1601519

PR #363, opened 2019-12-05
Add a --method parameter to tools/hawk/make_hawk_token.py

fzzzy:
## Description

Currently, it is possible to change the uri being signed by make_hawk_token using the --uri command line argument, but not possible to change the http method. This patch adds a --method flag.

## Testing

How should reviewers test?

`cargo run` in one terminal

`venv/bin/python make_hawk_token.py --method PUT --uri /1.5/1/storage/meta/global --secret=$SYNC_MASTER_SECRET --as_header > auth.txt`

```curl -v -X PUT "http://localhost:8000/1.5/1/storage/meta/global"     -H "content-type: application/json"     -H "Authorization: `cat auth.txt`"```

### DIFF
--- a/tools/hawk/make_hawk_token.py
+++ b/tools/hawk/make_hawk_token.py
@@ -25,6 +25,7 @@ from webob.request import Request
 LEGACY_UID = 1
 COL = "col2"
 URI = "/1.5/{uid}/storage/{col}/".format(uid=LEGACY_UID, col=COL)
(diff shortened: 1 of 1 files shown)


jrconlin:
heh, still working out why my secrets aren't matching, but yeah, that's what I had.

tublitzed:
@fzzzy works for me too! Thanks for adding this!

PR #364, opened 2019-12-05
Bug/342

jrconlin:
## Description

Do not fail when BSOs include a "collection" parameter.
* Removed "no BSO" work around

## Testing
Included unit test `put_meta_storage` to test this bug.

## Issue(s)

Closes: #342

### DIFF
--- a/src/server/test.rs
+++ b/src/server/test.rs
@@ -372,6 +372,17 @@ fn put_bso() {
 }
 
 #[test]
+fn put_meta_storage() {
+    let start = SyncTimestamp::default();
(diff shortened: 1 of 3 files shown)


tublitzed:
@jrconlin are there details for how we test this to verify that it resolves #342?

jrconlin:
@tublitzed sorry for the late reply.

From what I was able to see during the preliminary debugging and some subsequent, the problem was that the client was sending an unexpected field in the JSON body that was causing the SyncServer to return an error that was causing the sync process to terminate. This fix should prevent that error. The best way to test is to verify that the client is able to perform a sync.

PR #368, opened 2019-12-06
perf: Port get_bsos' pagination optimization

### LINKED ISSUE #53: Port get_bsos' pagination optimization
A TODO item described here: https://github.com/mozilla-services/syncstorage-rs/blob/bea8032/src/db/mysql/models.rs#L242

fzzzy:
## Description

The python code had an optimization which the rust code did not. Ported this optimization to rust.

The `cargo test`s and db-tests pass now, but I'm still not completely sure how to test the new functionality.

## Testing

?

## Issue(s)

Fix #53

### DIFF
--- a/db-tests/src/db_tests.rs
+++ b/db-tests/src/db_tests.rs
@@ -128,12 +128,12 @@ async fn get_bsos_limit_offset() -> Result<()> {
             0,
             Sorting::Index,
             0,
-            0,
+            &"0".to_owned(),
         ))
         .compat()
         .await?;
     assert!(bsos.items.is_empty());
-    assert_eq!(bsos.offset, Some(0));
+    assert_eq!(bsos.offset, Some("0".to_string()));
 
     let bsos = db
         .get_bsos(gbsos(
@@ -144,7 +144,7 @@ async fn get_bsos_limit_offset() -> Result<()> {
             0,
             Sorting::Index,
             -1,
-            0,
+            &"0".to_owned(),
         ))
         .compat()
         .await?;
@@ -153,7 +153,7 @@ async fn get_bsos_limit_offset() -> Result<()> {
 
     let newer = 0;
     let limit = 5;
-    let offset = 0;
+    let offset = "0".to_owned();
     // XXX: validation?
     /*
     let bsos = db.get_bsos_sync(gbsos(uid, coll, &[], MAX_TIMESTAMP, 0, Sorting::Index, -1, 0))?;
@@ -169,12 +169,17 @@ async fn get_bsos_limit_offset() -> Result<()> {
             newer,
             Sorting::Newest,
             limit,
-            offset,
+            &offset,
         ))
         .compat()
         .await?;
     assert_eq!(bsos.items.len(), 5 as usize);
-    assert_eq!(bsos.offset, Some(5));
+    if let Some(ref offset) = bsos.offset {
+        if offset.chars().position(|c| c == ':').is_none() {
+            assert_eq!(offset, &"5".to_string());
+        }
+    }
+
     assert_eq!(bsos.items[0].id, "11");
     assert_eq!(bsos.items[4].id, "7");
 
@@ -185,14 +190,18 @@ async fn get_bsos_limit_offset() -> Result<()> {
             &[],
             MAX_TIMESTAMP,
             newer,
-            Sorting::Index,
+            Sorting::Newest,
             limit,
-            bsos.offset.unwrap(),
+            &bsos.offset.unwrap(),
         ))
         .compat()
         .await?;
     assert_eq!(bsos2.items.len(), 5 as usize);
-    assert_eq!(bsos2.offset, Some(10));
+    if let Some(ref offset) = bsos2.offset {
+        if offset.chars().position(|c| c == ':').is_none() {
+            assert_eq!(offset, &"10".to_owned());
+        }
+    }
     assert_eq!(bsos2.items[0].id, "6");
     assert_eq!(bsos2.items[4].id, "2");
 
@@ -203,9 +212,9 @@ async fn get_bsos_limit_offset() -> Result<()> {
             &[],
             MAX_TIMESTAMP,
             newer,
-            Sorting::Index,
+            Sorting::Newest,
             limit,
-            bsos2.offset.unwrap(),
+            &bsos2.offset.unwrap(),
         ))
         .compat()
         .await?;
@@ -245,7 +254,7 @@ async fn get_bsos_newer() -> Result<()> {
             timestamp as u64 - 30,
             Sorting::Newest,
             10,
-            0,
+            &"0".to_owned(),
         ))
         .compat()
         .await?;
@@ -263,7 +272,7 @@ async fn get_bsos_newer() -> Result<()> {
             timestamp as u64 - 20,
             Sorting::Newest,
             10,
-            0,
+            &"0".to_owned(),
         ))
         .compat()
         .await?;
@@ -280,7 +289,7 @@ async fn get_bsos_newer() -> Result<()> {
             timestamp as u64 - 10,
             Sorting::Newest,
             10,
-            0,
+            &"0".to_owned(),
         ))
         .compat()
         .await?;
@@ -296,7 +305,7 @@ async fn get_bsos_newer() -> Result<()> {
             timestamp as u64,
             Sorting::Newest,
             10,
-            0,
+            &"0".to_owned(),
         ))
         .compat()
         .await?;
@@ -334,7 +343,7 @@ async fn get_bsos_sort() -> Result<()> {
             0,
             Sorting::Newest,
             10,
-            0,
+            &"0".to_owned(),
         ))
         .compat()
         .await?;
@@ -352,7 +361,7 @@ async fn get_bsos_sort() -> Result<()> {
             0,
             Sorting::Oldest,
             10,
-            0,
+            &"0".to_owned(),
         ))
         .compat()
         .await?;
@@ -370,7 +379,7 @@ async fn get_bsos_sort() -> Result<()> {
             0,
             Sorting::Index,
             10,
-            0,
+            &"0".to_owned(),
         ))
         .compat()
         .await?;
@@ -843,7 +852,7 @@ async fn get_bsos() -> Result<()> {
             0,
             Sorting::Newest,
             10,
-            0,
+            &"0".to_owned(),
         ))
         .compat()
         .await?;
@@ -858,7 +867,7 @@ async fn get_bsos() -> Result<()> {
             0,
             Sorting::Newest,
             10,
-            0,
+            &"0".to_owned(),
         ))
         .compat()
         .await?;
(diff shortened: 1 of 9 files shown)


fzzzy:
Looks like `test_weird_args` fails and I think I know why. There were some sanity checks that didn't get ported over. Need to learn how to make it return the right error code.

db-tests/src/db_tests.rs
```diff
@@ -174,7 +174,7 @@ async fn get_bsos_limit_offset() -> Result<()> {
         .compat()
         .await?;
     assert_eq!(bsos.items.len(), 5 as usize);
-    assert_eq!(bsos.offset, Some(5));
+    assert_eq!(bsos.offset, Some("5".to_string()));
```
pjenvey: For testing, I would think some of these cases would already hit the new offset code? If not, we could add some and test for encoded offset values when running under spanner.

(The Db could have some elaborate `supports_encoded_offset() -> bool` method, but I'd be fine with a quick check like the one recently added here: https://github.com/mozilla-services/syncstorage-rs/blob/70bfcbb0d6c649d285b106edbf6bb3b06615fc19/src/db/mysql/test.rs#L57 )

src/db/spanner/models.rs
```diff
@@ -976,6 +977,48 @@ impl SpannerDb {
...
+                // Use a simple numeric offset for sortindex ordering.
+                return Some(
+                    Offset {
+                        offset: offset + modifieds.len() as i64,
+                        timestamp: None,
+                    }
+                    .to_string(),
+                );
+            }
+            Sorting::None => newer_eq.unwrap_or(0),
```
pjenvey: If I'm reading the python right, only Oldest uses `newer_eq`? So `None | Newest` would be `older_eq`?

src/db/spanner/models.rs
```diff
@@ -985,7 +1028,18 @@ impl SpannerDb {
...
+            .params
+            .offset
+            .clone()
+            .unwrap_or(Offset {
+                offset: 0,
+                timestamp: None,
+            })
+            .offset;
+        let sort = params.params.sort.clone();
+        let newer_eq = params.params.newer_eq;
```
pjenvey: This is missing integration of `newer/older_eq` into the `bsos_query_sync's` query:

https://github.com/mozilla-services/server-syncstorage/blob/576490a/syncstorage/storage/sql/queries_spanner.py#L96

db-tests/src/db_tests.rs
pjenvey: Additionally we would need some spanner specific tests for testing decoding of offsets

src/web/extractors.rs
```diff
@@ -1023,6 +1023,22 @@ impl FromRequest for Box<dyn Db> {
     }
 }
 
+#[derive(Debug, Default, Clone, Deserialize, Validate)]
+#[serde(default)]
+pub struct Offset {
+    pub timestamp: Option<SyncTimestamp>,
+    pub offset: i64,
```
pjenvey: Can this be u64?

src/web/extractors.rs
```diff
@@ -1046,7 +1062,11 @@ pub struct BsoQueryParams {
     pub limit: Option<u32>,
 
     /// position at which to restart search (string)
-    pub offset: Option<u64>,
+    #[serde(deserialize_with = "deserialize_offset")]
+    pub offset: Option<Offset>,
+
+    pub newer_eq: Option<i64>,
```
pjenvey: I don't think newer/older_eq belong here, they aren't ever populated. Looks their values are determined when looking at the sort type: https://github.com/mozilla-services/server-syncstorage/blob/af652cc/syncstorage/storage/sql/__init__.py#L480

pjenvey: Also related to the link above: there should be some kind of check on this timestamp value vs the `older/newer` params (see the `raise InvalidOffsetError` cases)

src/db/spanner/models.rs
fzzzy: Yep, guess I read it backwards. Thanks.

src/db/spanner/models.rs
fzzzy: 🤦‍♂️

src/web/extractors.rs
fzzzy: I did try that, but there's a lot of existing code that uses -1 as a sentinel, and fixing all the typeerrors seems like a whole lot of work. If you really think it is important, I will do it.

src/web/extractors.rs
fzzzy: It's pretty weird how the python code uses the params dictionary as a bag of state and puts newer_eq in it when it will never be in a url. But yeah I think these can be removed.

Yes, I think the missing raise InvalidOffsetError equivalent is what is causing the one integration test to fail.

pjenvey: I don't think I fully understand this optimization yet, but another piece from the Python that strikes me as maybe odd is it seems like it can create these `newer/older_eq` params when there may also be an `newer/older` as well?

Which would produce 2 comparisons on modified like `modified > value1 AND modified >= value2`.

src/web/extractors.rs
```diff
@@ -1269,6 +1291,43 @@ impl FromRequest for BsoQueryParams {
...
+                .clone()
+                .unwrap_or(Offset {
+                    timestamp: None,
+                    offset: 0,
+                })
+                .timestamp
+            {
+                None => 0,
+                Some(timestamp) => timestamp.as_i64(),
+            };
```
jrconlin: Not really sure you need the match here. You could have the `unwrap_or` set the `timestamp` to 0 and then convert `.timestamp.as_i64()`

src/web/extractors.rs
```diff
@@ -445,7 +445,12 @@ impl BsoParam {
         // path: "/1.5/{uid}/storage/{collection}/{bso}"
         let elements: Vec<&str> = uri.path().split('/').collect();
         let elem = elements.get(3);
-        if elem.is_none() || elem != Some(&"storage") || elements.len() != 6 {
+        if elem.is_none() {
```
pjenvey: Was this for debugging?

src/web/extractors.rs
```diff
@@ -445,7 +445,12 @@ impl BsoParam {
...
+            return Ok(Self { bso: "".to_owned() });
+        }
+        if elem != Some(&"storage")
```
pjenvey: I think this one is to quiet the warning log messages on the other hand?

src/web/extractors.rs
```diff
@@ -1269,6 +1291,43 @@ impl FromRequest for BsoQueryParams {
...
+                .unwrap_or(Offset {
+                    timestamp: None,
+                    offset: 0,
+                })
+                .timestamp
+            {
+                None => 0,
+                Some(timestamp) => timestamp.as_i64(),
+            };
+            if bound != 0 {
```
pjenvey: Seems like we only need these checks if there's a timestamp, so the above could simplify down to
```suggestion
            if let Some(timestamp) = params.offset.as_ref().and_then(|offset| offset.timestamp) {
                let bound = timestamp.as_i64();
```

src/db/mysql/models.rs
```diff
@@ -495,11 +495,15 @@ impl MysqlDb {
         // match the query conditions
         query = query.limit(if limit >= 0 { limit + 1 } else { limit });
 
-        let offset = offset.unwrap_or(0) as i64;
-        if offset != 0 {
+        let numeric_offset = match offset {
```
pjenvey: nit: alternatively
```suggestion
        let numeric_offset = offset.map_or(0, |offset| offset.offset);
```

src/web/extractors.rs
pjenvey: Might be worth doing as a separate issue from this one

src/db/spanner/models.rs
```diff
@@ -941,55 +944,110 @@ impl SpannerDb {
...
-            sqlparams.insert("newer".to_string(), as_value(newer.as_rfc3339()?));
-            sqltypes.insert("newer".to_string(), as_type(TypeCode::TIMESTAMP));
-        }
-
         if !ids.is_empty() {
             query = format!("{} AND bso_id IN UNNEST(@ids)", query);
             sqlparams.insert("ids".to_owned(), as_list_value(ids.into_iter()));
         }
 
+        if let Some(offset) = offset.clone() {
```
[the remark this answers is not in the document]
fzzzy: Wow, great, I had no idea about this

src/web/extractors.rs
fzzzy: This is because some of the tests expect to be able to do a DELETE on a weirdly spelled url to reset state. I'd really rather understand how handle this whole chunk better, but this works for now.

src/web/extractors.rs
fzzzy: Same here, some weird test behavior. I'd rather understand why the tests are expecting to be able to do what they do, or fix the tests so they don't do it if it doesn't make sense any more.

pjenvey: which tests?

fzzzy: The python tests. They do a DELETE that the rust code doesn't seem to expect.

src/web/extractors.rs
fzzzy: Filed https://github.com/mozilla-services/syncstorage-rs/issues/414

db-tests/src/db_tests.rs
fzzzy: Ok, in the process of fixing the last test I did add code which tests the results of the spanner offset optimization.

fzzzy:
Ok, I think this might be ready to merge now.

src/web/extractors.rs
pjenvey: Can you double check this? I can't reproduce it over here

src/db/spanner/models.rs
```diff
@@ -941,55 +944,108 @@ impl SpannerDb {
...
+                }
+                Sorting::Oldest => {
+                    sqlparams.insert("newer_eq".to_string(), as_value(timestamp.as_rfc3339()?));
+                    sqltypes.insert("newer_eq".to_string(), as_type(TypeCode::TIMESTAMP));
+                    format!("{} AND modified >= @newer_eq", query)
+                }
+                _ => query,
+            };
+        } else {
+            if let Some(older) = older {
```
pjenvey: One last thing: I'm wondering if we should just continue emitting `newer`/`older` when `newer_eq`/`older_eq` are, to match the python code.

Maybe it doesn't make sense for both clauses to be enabled, but if so maybe the validation layer should be enforcing it (`newer`/`older` parameters from the client would be disallowed when an `offset w/ timestamp` were specified). Maybe @rfk can chime in on this?

fzzzy: Sure, I think it makes sense to match the python code as much as possible even if the clauses are redundant.

src/web/extractors.rs
fzzzy: Ok, I was able to go back to the way the code was originally here.

src/web/extractors.rs
fzzzy: Double checked, the way the code was before passes the tests. Reverted that bit.

src/db/spanner/models.rs
rfk: I think it's reasonable for the client to continue sending `newer` even when it's using an `offset` token, because the offset token should be opaque to the client and may not encode all the info needed to reproduce the original query. As a client, I would expect to have to just append the given `offset` to the existing URL without modifying other query parameters.

Does that answer the concern? I'm not sure I fully understand what you're getting at here :-)

pjenvey:
don't forget to squash this one :)

src/db/spanner/models.rs
pjenvey: Thanks @rfk, that makes sense and we just ran into a test that confirms this.

PR #372, opened 2019-12-10
bug: add tag info to sentry error messages

jrconlin:
## Description

Not a super fan of this. It might be better as middleware, but that's
not always defined when we make sentry error calls. Passes a `Tags` structure around that contains the tags that will be added to sentry and metrics calls.

## Testing

There is an included unit test to check default tag generation. 

In addition, there is now a new test endpoint called `/__error__` that will generate a sentry error containing the ua tags. (see https://sentry.prod.mozaws.net/operations/syncstorage-dev/issues/6888305/ for sample output)

When running tests, note that the following tags should be added to sentry and metric calls:
* ua,os,ver
* ua.os.family
* ua.browser.ver
* ua.name
* ua.browser.family
* uri.path
* uri.method

## Issue(s)

Issue: [#329](https://github.com/mozilla-services/syncstorage-rs/issues/329)

### DIFF
--- a/src/web/error.rs
+++ b/src/web/error.rs
@@ -16,6 +16,7 @@ use serde_json::{Error as JsonError, Value};
 use validator;
 
 use super::extractors::RequestErrorLocation;
+use super::tags::Tags;
 use crate::error::ApiError;
 
 /// An error occurred during HAWK authentication.
@@ -70,8 +71,8 @@ pub enum HawkErrorKind {
 /// An error occurred in an Actix extractor.
 #[derive(Debug)]
 pub struct ValidationError {
-    inner: Context<ValidationErrorKind>,
     pub status: StatusCode,
+    inner: Context<ValidationErrorKind>,
 }
 
 impl ValidationError {
@@ -84,10 +85,14 @@ impl ValidationError {
 #[derive(Debug, Fail)]
 pub enum ValidationErrorKind {
     #[fail(display = "{}", _0)]
-    FromDetails(String, RequestErrorLocation, Option<String>),
+    FromDetails(String, RequestErrorLocation, Option<String>, Option<Tags>),
 
     #[fail(display = "{}", _0)]
-    FromValidationErrors(#[cause] validator::ValidationErrors, RequestErrorLocation),
+    FromValidationErrors(
+        #[cause] validator::ValidationErrors,
+        RequestErrorLocation,
+        Option<Tags>,
+    ),
 }
 
 failure_boilerplate!(HawkError, HawkErrorKind);
@@ -109,7 +114,7 @@ impl From<Context<ValidationErrorKind>> for ValidationError {
     fn from(inner: Context<ValidationErrorKind>) -> Self {
         debug!("Validation Error: {:?}", inner.get_context());
         let status = match inner.get_context() {
-            ValidationErrorKind::FromDetails(ref _description, ref location, Some(ref name))
+            ValidationErrorKind::FromDetails(ref _description, ref location, Some(ref name), _)
                 if *location == RequestErrorLocation::Header =>
             {
                 match name.to_ascii_lowercase().as_str() {
@@ -118,7 +123,7 @@ impl From<Context<ValidationErrorKind>> for ValidationError {
                     _ => StatusCode::BAD_REQUEST,
                 }
             }
-            ValidationErrorKind::FromDetails(ref _description, ref location, Some(ref name))
+            ValidationErrorKind::FromDetails(ref _description, ref location, Some(ref name), _)
                 if *location == RequestErrorLocation::Path
                     && ["bso", "collection"].contains(&name.as_ref()) =>
             {
@@ -175,16 +180,17 @@ impl Serialize for ValidationErrorKind {
         let mut seq = serializer.serialize_seq(None)?;
 
         match *self {
-            ValidationErrorKind::FromDetails(ref description, ref location, ref name) => {
+            ValidationErrorKind::FromDetails(ref description, ref location, ref name, ref tags) => {
                 seq.serialize_element(&SerializedValidationError {
                     description,
                     location,
                     name: name.as_ref().map(|name| &**name),
                     value: None,
+                    tags: tags.as_ref(),
                 })?;
             }
 
-            ValidationErrorKind::FromValidationErrors(ref errors, ref location) => {
+            ValidationErrorKind::FromValidationErrors(ref errors, ref location, ref tags) => {
                 for (field, field_errors) in errors.clone().field_errors().iter() {
                     for field_error in field_errors.iter() {
                         seq.serialize_element(&SerializedValidationError {
@@ -192,6 +198,7 @@ impl Serialize for ValidationErrorKind {
                             location,
                             name: Some(field),
                             value: field_error.params.get("value"),
+                            tags: tags.clone().as_ref(),
                         })?;
                     }
                 }
@@ -208,4 +215,5 @@ struct SerializedValidationError<'e> {
     pub location: &'e RequestErrorLocation,
     pub name: Option<&'e str>,
     pub value: Option<&'e Value>,
+    pub tags: Option<&'e Tags>,
 }
--- a/src/web/extractors.rs
+++ b/src/web/extractors.rs
@@ -461,7 +461,6 @@ impl BsoParam {
         }
         if let Some(v) = elements.get(5) {
             let sv = String::from_str(v).map_err(|_| {
-                // TODO: Why is the param for this block causing warn! to say it's unreachable?
                 warn!("⚠️ Invalid BsoParam Error: {:?}", v;
                 "ua.os.family" => tags.get("ua.os.family"),
                 "ua.browser.family" => tags.get("ua.browser.family"),
--- a/src/web/extractors.rs
+++ b/src/web/extractors.rs
@@ -2097,7 +2097,7 @@ mod tests {
                 Some(t) => t.clone(),
                 None => Tags::from_request_head(req.head()),
             };
-            let result = PreConditionHeaderOpt::extrude(&req.headers(), Some(tags.clone()));
+            let result = PreConditionHeaderOpt::extrude(&req.headers(), Some(tags));
             assert!(result.is_err());
             let response: HttpResponse = result.err().unwrap().into();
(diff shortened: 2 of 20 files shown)


jrconlin:
based on https://github.com/mozilla-services/syncstorage-rs/pull/373 need to make sure tag/"extra" fields are actually sent to sentry.

tublitzed:
@jrconlin - to confirm for the testing bit of this; have we verified manually in sentry that these tags appear as expected?

jrconlin:
@tublitzed They've not and thus the comment.

jrconlin:
Right, there needs to be an explicit `sentry::` capture call similar to https://github.com/mozilla-services/syncstorage-rs/pull/373/files#diff-b7289185b77ef7668febb777d5d33c5fR249. My initial misunderstanding was that the sentry tag data was being pulled from the logging messages directly, that's why I am currently trying to fix the issue with my local instance communicating with sentry directly.

tublitzed:
@jrconlin I'm not seeing any change between here and master related to errors appearing in Sentry. 

Here's what I'm doing:

1. Adding a panic to line 56 of main.rs.
2. Starting server like this (`SENTRY_DSN={DSN__HERE} make run_local`)

The errors generated from this branch appear identical to what's coming from master. I'm confused around what you've changed here specifically related Sentry tags; I'm assuming that I'm just missing something obvious here....would you mind clarifying specifically the new fields that you've added to Sentry, and how we can test this to make sure they're appearing as expected?

For reference, here is what I see in [this branch](https://sentry.prod.mozaws.net/operations/syncstorage-dev/issues/6452434/?query=is:unresolved) vs what's coming from [master into sentry](https://sentry.prod.mozaws.net/operations/syncstorage-dev/issues/6467367/?query=is:unresolved) - AFAIK there is no new information there?

Ie, all the tags look identical.

jrconlin:
@tublitzed Unfortunately, `panics` (and other macro based errors such as `error!`, `warn!`, etc.) require special effort in order to transmit tag info. This is because those tend to strip off tag info or happen outside of the contexts where the `Tags` structure are defined. There are some examples in the code base, and there might be additional ways that we could automate some of that, but for now, this patch adds the tagging info to the `ApiError` messages we generate normally.  (See https://github.com/mozilla-services/syncstorage-rs/pull/372/files#diff-9de2327cbb67a90931e98ff70715da9dR570 for an example of explicit tag set, however be aware that sentry currently does not transmit those additional tags values.)

While not perfect, it gets us further than where we were without the tag info.

tublitzed:
@jrconlin - makes sense to me. We can come back later to try and address panics etc if/when needed. Thanks for the example of the specific error; can you let those of us reviewing this PR know how to trigger that error so we can verify that it's passing along the correct info into Sentry?

jrconlin:
Yes, I updated the instructions above to indicate the new `/__error__` endpoint (sorry, the original got changed by markdown to /__error__)

tublitzed:
@JR can you highlight what's different in your [example event](https://sentry.prod.mozaws.net/operations/syncstorage-dev/issues/6888305/) than what we get from master? I'm not seeing any new tags in Sentry.

I tried it locally too, here's what I get:

this is from this branch, using the `__error__` endpoint:

Screen Shot 2019-12-17 at 3 53 31 PM

this is from this branch, in master using a panic:

Screen Shot 2019-12-17 at 3 55 17 PM

...AFAIK there isn't a difference in the tags sent. Maybe I'm looking in the wrong place in the Sentry UI?

jrconlin:
??? now I'm not seeing the added tags either.

src/web/handlers.rs
```diff
@@ -431,3 +431,31 @@ pub fn heartbeat(hb: HeartbeatRequest) -> impl Future<Item = HttpResponse, Error
...
+        e.g.
+        ```
+        error!("Something Bad {:?}", err;
+                "ua.os.family" => wtags.get("ua.os.family"),
+                "ua.browser.family" => wtags.get("ua.browser.family"),
+                "ua.name" => wtags.get("ua.name"),
+                "ua.os.ver" => wtags.get("ua.os.ver"),
+                "ua.browser.ver" => wtags.get("ua.browser.ver"))
+        ```
+        TODO: find some way to transform Tags into error::KV
```
pjenvey: KV's a trait that Tags can implement. [slog-derive](https://github.com/slog-rs/derive) makes it easy.

Probably better to tackle that in a separate issue.

src/web/extractors.rs
```diff
@@ -124,6 +131,91 @@ fn get_accepted(req: &HttpRequest, accepted: &[&str], default: &'static str) ->
     "invalid".to_string()
 }
 
+#[derive(Clone, Debug)]
```
pjenvey: Let's move this out of this file, extractors.rs is already too large IMO.

src/web/extractors.rs
```diff
@@ -124,6 +131,91 @@ fn get_accepted(req: &HttpRequest, accepted: &[&str], default: &'static str) ->
...
+        let mut seq = serializer.serialize_map(Some(self.tags.len()))?;
+        for tag in self.tags.clone() {
+            seq.serialize_entry(&tag.0, &tag.1)?;
+        }
+        seq.end()
+    }
+}
+
+impl Tags {
+    pub fn from_request_head(req_head: &RequestHead) -> Option<Tags> {
```
pjenvey: No need for an Option since this never returns None

src/web/middleware.rs
```diff
@@ -274,6 +288,93 @@ where
...
+        sreq.extensions_mut().insert(tags.clone());
+
+        Box::new(self.service.call(sreq).and_then(move |resp| {
+            // handed an actix_error::error::Error;
+            // Fetch out the tags (in case any have been added.)
+            // Tags associated with the sreq.extensions() may not
+            // always be preserved. Use the orignal set of
+            // tags and overwrite with any additional tags that may
+            // have been set in the response.
+            let mut tag_vals = tags.unwrap_or_default();
```
pjenvey: better to defer grabbing these tags till later in the Some(e) if 500 arm, only when they are needed for capturing

src/web/middleware.rs
```diff
@@ -274,6 +288,93 @@ where
...
+            if let Some(t) = resp.response().extensions().get::<Tags>() {
+                for (k, v) in t.tags.clone() {
+                    tag_vals.tags.insert(k, v);
+                }
+            };
+            match resp.response().error() {
```
pjenvey: We currently only capture errors if they're 500 responses, we should continue restricting what we capture to only things we care about (is there anything else other than 500s? probably not).

Also this PR should remove the old capturing in db/error.rs so it's only done once.

src/web/error.rs
```diff
@@ -84,10 +84,14 @@ impl ValidationError {
 #[derive(Debug, Fail)]
 pub enum ValidationErrorKind {
     #[fail(display = "{}", _0)]
-    FromDetails(String, RequestErrorLocation, Option<String>),
+    FromDetails(String, RequestErrorLocation, Option<String>, Option<Tags>),
```
pjenvey: Do we still need to pass Tags through everywhere now that the middleware's handling things? Removing it all would be a nice cleanup

jrconlin: If we want to add or alter tags in the future, yes.

jrconlin: Thinking about this more: I agree that we can probably do a bit of cleanup here and remove some of the pass-thru tag stuff. We can also pull from the `request.extensions().get::<Tags>()` and pass that along if need be.

src/web/middleware.rs
```diff
@@ -274,6 +288,93 @@ where
...
+            let mut tag_vals = tags.unwrap_or_default();
+            if let Some(t) = resp.response().extensions().get::<Tags>() {
+                    for (k, v) in t.tags.clone() {
+                        tag_vals.tags.insert(k, v);
+                    }
+            };
+            match resp.response().error() {
+                None => {}
+                Some(e) => {
+                    let mut event = Event::new();
```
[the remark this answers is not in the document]
jrconlin: Hmm. the problem with that is that I get a compiler error that it cannot be sent between threads safely.

jrconlin:
@pjenvey points out that it's possible to downcast to ApiError to avoid the problem:

```
    diff --git a/src/web/middleware.rs b/src/web/middleware.rs
    index 645ab41769..3eda353d59 100644
    --- a/src/web/middleware.rs
    +++ b/src/web/middleware.rs
    @@ -368,15 +368,12 @@ where
                                 tags.tags.insert(k, v);
                             }
                         };
    -                    let mut event = Event::new();
    -                    // The following cannot safely be shared between threads.
    -                    // note this does mean that the backtrace could be lost.
    -                    // let mut event = integrations::failure::event_from_fail(e);
    -                    if let Some(r_err) = e.as_error::<ApiError>() {
    -                        event.message = Some(format!("{:?}", r_err.kind()));
    +                    let apie: Option<&ApiError> = e.as_error();
    +                    if let Some(apie) = apie {
    +                        let mut event = sentry::integrations::failure::event_from_fail(apie);
    +                        event.tags = tags.as_btree();
    +                        sentry::capture_event(event);
                         }
    -                    event.tags = tags.as_btree();
    -                    sentry::capture_event(event);
                     }
                 }
                 resp
```
Also might be worth digging into `sentry_actix` to see how it handles passing the `cause.backtrace`

src/web/handlers.rs
jrconlin: see #387

src/web/extractors.rs
```diff
@@ -425,46 +440,71 @@ pub struct BsoParam {
...
                 RequestErrorLocation::Path,
                 Some("bso".to_owned()),
+                Some(tags.clone()),
             ))?;
         }
         if let Some(v) = elements.get(5) {
-            let sv = String::from_str(v).map_err(|e| {
-                warn!("⚠️ BsoParam Error element:{:?} error:{:?}", v, e);
+            let sv = String::from_str(v).map_err(|_| {
+                // TODO: Why is the param for this block causing warn! to say it's unreachable?
```
pjenvey: When does warn! say it's unreachable (compile time or what?)

jrconlin: so, after a bit of back and forth, this appears to have been an anomaly. I've not reproduced it. I removed the comment.

Issue #333, opened 2019-11-14
Seeing spanner 20k mutation limit errors for inserts in production


tublitzed:
I have a large collection of bookmarks(3078 bookmarks and 546 folders) that all of a sudden is failing to sync in production.

Sync id: `130387415`. Via bob, I got the following from the production logs: 

```A database error occurred: RpcFailure(RpcStatus { status: RpcStatusCode(3), details: Some("The transaction contains too many mutations. Insert and update operations count with the multiplicity of the number of columns they affect. For example, inserting values into one key column and four non-key columns count as five mutations total for the insert. Delete and delete range operations count as one mutation regardless of the number of columns affected. The total mutation count includes any changes to indexes that the transaction generates. Please reduce the number of writes, or use fewer indexes. (Maximum number: 20000)") }), status: 500 }```

Here's [a link](https://github.com/mozilla-services/services-engineering/blob/master/scripts/bookmarks/samples/large_anon_collection.json) to a sample anonymized collection of this same size.

**Comment by tublitzed** [comment]
What's interesting here is that I hit the Spanner 20k limit with only ~3500 items...

**Comment by pjenvey** [comment]
The 20k here translates into 2000 items (1 mutation per each column. bsos table has 8 columns + 2 extra mutations for our secondary indexes).

It looks like the sync client will print the server's limits in its log, can you search your log for "max_total_records"? Let's confirm your client's seeing a value of "2000".

E.g. (from my log, probably non-spanner current sync prod)
```
1573781473132	Sync.Engine.Tabs	TRACE	new PostQueue config (after defaults): : {"max_request_bytes":2101248,"max_record_payload_bytes":2097152,"max_post_bytes":2101248,"max_post_records":100,"max_total_bytes":209715200,"max_total_records":10000}
```

jrconlin:
Ah, I thought that it was one mutation per row too, not per column. That
really changes how I view things. We pretty much want to reduce the number
of columns to the bare minimum, but I get the feeling we'd get hit on data
transfer then.

tublitzed:
@pjenvey confirmed that my `max_total_records` is getting set correctly to 2k: 

```{"max_post_bytes":2097152,"max_post_records":100,"max_record_payload_bytes":2097152,"max_request_bytes":2101248,"max_total_bytes":209715200,"max_total_records":2000}```

tublitzed:
Here's my error log now in case that's also helpful.

[...]

PR #373, opened 2019-12-10
feat: add temporary sentry tags for the mutation limit issue

pjenvey:
a pile of of hacks which we'll hopefully revert shortly

Issue #333

Also see https://sentry.prod.mozaws.net/operations/syncstorage-prod/issues/6621760/events/36079823/ for what this is specifically after

### DIFF
--- a/src/db/spanner/batch.rs
+++ b/src/db/spanner/batch.rs
@@ -152,6 +152,107 @@ pub fn delete(db: &SpannerDb, params: params::DeleteBatch) -> Result<()> {
     Ok(())
 }
 
+pub fn _audit_commit_err<T>(
+    db: &SpannerDb,
+    params: &params::CommitBatch,
+    result: Result<T>,
+) -> Result<T> {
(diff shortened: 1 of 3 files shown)


src/db/spanner/batch.rs
```diff
@@ -152,6 +152,107 @@ pub fn delete(db: &SpannerDb, params: params::DeleteBatch) -> Result<()> {
...
+                .insert("mutations_len".to_owned(), mutations.len().to_string());
+            event.tags.insert(
+                "mutations_total_values_len".to_owned(),
+                mutations_total_values_len.to_string(),
+            );
+            event.tags.insert(
+                "mutations_insert_values_len".to_owned(),
+                mutations_insert_values_len.to_string(),
+            );
+            dbg!("ZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZ", &event);
```
jrconlin: drop

Issue #385, opened 2019-12-20
Rewrite purge_ttl in Rust


tublitzed:
We're adding python to the Docker image right now specifically to run the [purge_ttl](https://github.com/mozilla-services/syncstorage-rs/blob/master/tools/spanner/purge_ttl.py) script. It'd be nice to be able to shed the weight of python support; let's rewrite this script in Rust, and then make sure we've coordinated with ops around adjusting their logic to use it.

pjenvey:
It's a pretty small script, so I think a good start to this bug would be:

- Create a command line app in Rust that reads `SYNC_DATABASE_URL` from the environment
- Which then creates a Spanner client from, that then emits the two partitioned dml `DELETE` statements. This part would look similar to the code in the [spanner manager module](https://github.com/mozilla-services/syncstorage-rs/blob/2935846/src/db/spanner/manager.rs)

Once we have a the basic script running we can then add the extra features supported by the Python version:

- Emit timer metric to measure the duration of the calls (using the `cadence` metrics lib that syncstorage-rs already uses)
- Proper logging messages (`println!` would probably do to start with)

Issue #384, opened 2019-12-20
Add python to docker image


tublitzed:
So we can run the purge_ttl script

tublitzed:
I filed a followup issue [here](https://github.com/mozilla-services/syncstorage-rs/issues/385) to rewrite this script in Rust to avoid the need to do this, but for now this should be fine.

tublitzed:
Bumping the estimate a bit as I mess around with conditionally running the script based on env vars.

PR #388, opened 2019-12-27
chore: add python to docker image & bump rust version

tublitzed:
## Description

As per a recent Ops request, adding python to our docker image so we can automate the purge_ttl script. We've got [a followup](https://github.com/mozilla-services/syncstorage-rs/issues/385) to rewrite this in Rust so we can remove the python dep, but I don't think it's worth it right now.

While I was in here, I added a few details to the README on running this locally via Docker. I also bumped our Rust version to 1.40, and made a few small changes to the dockerfile to try to reduce it's size a bit.

## Testing

1. Follow the instructions I'm adding in this PR to the README for running via docker locally.
2. Once it's running, do a `docker ps` to grab your container id.
3. `docker exec -it {CONTAINER_ID_HERE} bash` - enter container, type `python3 tools/spanner/purge_ttl.py` - You should see an error related to missing `GOOGLE_APPLICATION_CREDENTIALS` - that's expected as they're not present, but verifies that you've hit this script via python3.

## Issue(s)

Closes [#384](https://github.com/mozilla-services/syncstorage-rs/issues/384).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,4 @@ service-account.json
 
 config/local.toml
 venv
+mozilla-rust-sdk
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,10 @@
-FROM rust:1.39.0-buster as builder
+FROM rust:1.40.0-buster as builder
 WORKDIR /app
 ADD . /app
 ENV PATH=$PATH:/root/.cargo/bin
 RUN apt-get -q update && \
-    apt-get -q install -y default-libmysqlclient-dev cmake golang-go && \
+    apt-get -q install -y --no-install-recommends default-libmysqlclient-dev cmake golang-go && \
+    rm -rf /var/lib/apt/lists/* && \
     cd /app && \
     mkdir -m 755 bin
 
@@ -18,11 +19,16 @@ RUN \
     groupadd --gid 10001 app && \
     useradd --uid 10001 --gid 10001 --home /app --create-home app && \
     apt-get -q update && \
-    apt-get -q install -y default-libmysqlclient-dev libssl-dev ca-certificates libcurl4 && \
-    rm -rf /var/lib/apt/lists
+    apt-get -q install -y --no-install-recommends default-libmysqlclient-dev libssl-dev ca-certificates libcurl4 python3-venv python3-pip && \
+    python3 -m pip install setuptools wheel && \
+    python3 -m pip install google-cloud-spanner && \
+    rm -rf /var/lib/apt/lists/*
 
 COPY --from=builder /app/bin /app/bin
 COPY --from=builder /app/version.json /app
 COPY --from=builder /app/spanner_config.ini /app
+COPY --from=builder /app/tools/spanner /app/tools/spanner
 
-CMD ["/app/bin/syncstorage", "--config=spanner_config.ini"]
+USER app:app
+
+ENTRYPOINT ["/app/bin/syncstorage", "--config=spanner_config.ini"]
--- a/docker-compose.e2e.yaml
+++ b/docker-compose.e2e.yaml
@@ -6,7 +6,7 @@ services:
         # itself a few times or should include a wait-for-it.sh script
         # inside its docker that would do this for us. Same (probably
         # the latter solution) for server-syncstorage below
-        command: >
+        entrypoint: >
           /bin/sh -c "
             sleep 15;
             /app/bin/syncstorage;
--- a/src/db/spanner/models.rs
+++ b/src/db/spanner/models.rs
@@ -1267,6 +1267,7 @@ impl SpannerDb {
                     "".to_string()
                 }
             );
+
             q = format!(
                 "{}{}",
                 q,
@@ -1279,6 +1280,7 @@ impl SpannerDb {
                     "".to_string()
                 }
             );
+
             q = format!(
                 "{}{}",
                 q,
@@ -1290,6 +1292,7 @@ impl SpannerDb {
                     "".to_string()
                 }
             );
+
             q = format!(
                 "{}{}",
                 q,
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,12 @@
+.git
+CHANGELOG.md
+CODE_OF_CONDUCT.md
+config/*.example.toml
+CONTRIBUTING.md
+Dockerfile
+docs/
+Makefile
+PULL_REQUEST_TEMPLATE.md
+README.md
+target
+tools/examples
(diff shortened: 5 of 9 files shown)


Dockerfile
```diff
@@ -1,9 +1,10 @@
-FROM rust:1.39.0-buster as builder
+FROM rust:1.40.0-buster as builder
```
tublitzed: Happy to back out this change since it's a bit unrelated, but figured I'd just grab it while I was in here.

Dockerfile
```diff
@@ -18,11 +19,12 @@ RUN \
     groupadd --gid 10001 app && \
     useradd --uid 10001 --gid 10001 --home /app --create-home app && \
     apt-get -q update && \
-    apt-get -q install -y default-libmysqlclient-dev libssl-dev ca-certificates libcurl4 && \
-    rm -rf /var/lib/apt/lists
+    apt-get -q install -y --no-install-recommends default-libmysqlclient-dev libssl-dev ca-certificates libcurl4 python3.7 python3-pip && \
```
tublitzed: @erkolson - do you need pip as well? I added it here, but can remove if you don't think it's necessary. I assumed you might since we still need to install the google SDK, but if you have another approach there, let me know and I'll remove `python3-pip` so we're not adding stuff we don't need.

src/db/spanner/models.rs
```diff
@@ -1266,8 +1266,8 @@ impl SpannerDb {
                 } else {
                     "".to_string()
                 }
-            )
-            .to_string();
```
tublitzed: these are all related to the same clippy error: [redundant clone](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone) - guessing it's the rust version bump that triggered these, since none of this is new code.

Dockerfile
erkolson: We came up with a slightly different plan at the Monday meeting, don't think we need python anymore.  Converting the tests to rust will take some time, the current tests will need python 2.7 and we don't want to put that in the rust app image.  The TTL purge script can be rewritten in rust quickly and I believe that is in progress.  To facilitate the automated tests, we will pull the old python app image to execute tests, the containers will be deleted immediately after tests pass.  This will keep the rust app image cleaner and smaller.

tublitzed: That's what I get for being on vacation for the Monday meeting :) I'll close this one out then. @pjenvey - who's handling the rewrite for the purge_ttl script?

erkolson: Sorry for the wasted time/effort!

tublitzed: no worries :)

jrconlin:
I'm kinda curious about this. I thought that Ops was going to use something like a Cloud Function to do purges. A good deal of this script is just reading stuff from the the environment variables and parsing out the actual database info.

The SQL is basically:
```DELETE FROM batches WHERE expiry < CURRENT_TIMESTAMP();```
and
```DELETE FROM bsos WHERE expiry <  CURRENT_TIMESTAMP();```

It's preferred if these are partitioned DML calls, but if I recall, that's an option when setting up cloud functions.

Is that no longer the case? Do we have to incorporate this into our own server?

**Comment by tublitzed** [comment]
Re-opening this based on our discussion in standup earlier today.

@erkolson - to clarify what you'd need around the bash script, you just want a one liner that calls this via python somewhere in the container? ie, ala `python3 tools/spanner/purge_ttl.py` ? Did you want to handle setting/dealing with the GOOGLE_APPLICATION_CREDENTIALS separately?

File: Dockerfile [hunk 4]
```diff
@@ -19,12 +19,14 @@ RUN \
     groupadd --gid 10001 app && \
     useradd --uid 10001 --gid 10001 --home /app --create-home app && \
     apt-get -q update && \
-    apt-get -q install -y --no-install-recommends default-libmysqlclient-dev libssl-dev ca-certificates libcurl4 python3.7 python3-pip && \
+    apt-get -q install -y --no-install-recommends default-libmysqlclient-dev libssl-dev ca-certificates libcurl4 python3-venv python3-pip && \
+    python3 -m pip install setuptools wheel && \
+    python3 -m pip install google-cloud-spanner && \
```
tublitzed: I'd have preferred all these on a single `pip install` line, but looks like the spanner install fails if I do that as the other required deps don't seem to finish in time.

tublitzed: (also - would a `requirements.txt` file be preferred here?) Seems like maybe it'd be cleaner.

erkolson: Huh, I get the same problem when I try to install via requirements.txt

erkolson:
> Re-opening this based on our discussion in standup earlier today.
> 
> @erkolson - to clarify what you'd need around the bash script, you just want a one liner that calls this via python somewhere in the container? ie, ala `python3 tools/spanner/purge_ttl.py` ? Did you want to handle setting/dealing with the GOOGLE_APPLICATION_CREDENTIALS separately?

*   Ideally, the Dockerfile would invoke the main app via an `ENTRYPOINT` instead of `CMD`, i.e.:
    ```
    ENTRYPOINT ["/app/bin/syncstorage", "--config=spanner_config.ini"]
    ```
    Then a one liner can be called to replace the default entrypoint.  The kubernetes job would   overwrite the entrypoint with the confusingly named key `command`:
    ```yaml
    ...
      command: ["/usr/bin/python3", "/tools/spanner/purge_ttl.py"]
    ...
    ```
    If called directly like above, without a bash script, the return code is passed back to the docker daemon and the kubernetes job status accurately reflects the script's exit status.

    [If the image continues the main app invocation via`CMD`, we can make that work as well by    overwriting via the `args` key in the purge ttl kubernetes job manifest.]

*  We can reuse the existing app google credentials, the key file will be mounted to the container at runtime in the kubernetes job spec and an environment variable will point the python library where to find it.

* One thing I noticed.  This Dockerfile executes the commands to create the app user and group but the app still runs as the root user.  Need to add something like this after the build artifacts are copied in:
  ```
  USER app:app
  ```
  We may also need to chown the files to `app:app` in the `/app` dir.

tublitzed:
@erkolson seems to work fine locally without chown-ing in the dockerfile so I'll leave that one alone unless you need it for some other reason in there.

updated to use `ENTRYPOINT` vs `CMD`.

docker-compose.e2e.yaml
```diff
@@ -6,7 +6,7 @@ services:
         # itself a few times or should include a wait-for-it.sh script
         # inside its docker that would do this for us. Same (probably
         # the latter solution) for server-syncstorage below
-        command: >
+        entrypoint: >
```
tublitzed: made this change to fix circle failures related to changing the CMD in dockerfile to an ENTRYPOINT

.gitignore
```diff
@@ -35,3 +35,5 @@ service-account.json
 
 config/local.toml
 venv
+
```
jrconlin: nit: probably can remove this

PR #390, opened 2019-12-31
feat: route reads through the streaming sql api

### LINKED ISSUE #205: Switch to execute_streaming_sql
Spanner provides a streaming sql request call, which should act more resource efficient for large result sets. 

The Python client lib (and likely other libs) default its sql calls to using this API. It uses the same `ExecuteSqlRequest` struct for building the request, but returns a `PartialResultSet` vs a regular `ResultSet`.

The result iterator would have to adapt to this different ResultSet.

grpcio provides a `ClientSStreamReceiver` for handling the stream. It supports an async interface. There's also a `wait` call for a sync request? Would we need to defer this to a thread pool to take advantage of it before we go full async?

pjenvey:
## Description

feat: route reads through the streaming sql api

writes continue to use plain execute sql (via execute_dml)

## Testing

I wish there were better tests for this (like a mock Stream).

I intend to load test this on stage in hopes to trigger some actual streaming requiring multiple reads. Error handling's all in place if anything goes wrong.

## Issue(s)

Closes #205


src/db/spanner/models.rs
```diff
@@ -566,15 +566,16 @@ impl SpannerDb {
                 "ids".to_owned(),
                 as_list_value(uncached.into_iter().map(|id| id.to_string())),
             );
-            let result = self
+            let rs = self
```
jrconlin: later on, you use `affected_rows`. We might want to use a better var name here as well, say `collection_row` or something.

pjenvey: I'm using rs for ResultSet (from execute) whereas the new execute_dml now returns the affected_rows

PR #392, opened 2020-01-02
feat: break apart middleware.rs

### LINKED ISSUE #391: Break apart "middleware.rs"
middleware.rs is a bit unweildy. We should break it apart into discrete components rather than pile them all into one file.

jrconlin:
## Description
Broke `middleware.rs` into sub components that are easier to understand,
compartmentalize, and potentially share with the larger org.

## Testing
There should be no impact to existing unit or run tests.

## Issue(s)
Closes #391


src/web/middleware/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+use actix_web::dev::ServiceRequest;
+use actix_web::Error;
+
+use crate::db::util::SyncTimestamp;
+use crate::error::{ApiError, ApiErrorKind};
+use crate::server::ServerState;
+use crate::web::{extractors::HawkIdentifier, tags::Tags, DOCKER_FLOW_ENDPOINTS};
+
+/// The resource in question's Timestamp
+pub struct ResourceTimestamp(SyncTimestamp);
```
pjenvey: apparently this is dead code, we can remove it later

jrconlin: Huh, wonder why it wasn't getting flagged? Guessing it's because it's in a `mod.rs` file. Good to know for the future.

Issue #396, opened 2020-01-03
Update examples to make it easier to point to spanner from local machine


tublitzed:
Ie, update README, Makefile.

PR #397, opened 2020-01-03
docs: update spanner examples

tublitzed:
## Description

I was exploring connecting to spanner from a local machine, and figured I'd update the docs while I was in there.

## Testing

Follow the instructions I've added to the README. Do they make sense?

## Issue(s)

Closes [#396](https://github.com/mozilla-services/syncstorage-rs/issues/396).


Makefile
```diff
@@ -3,6 +3,12 @@
 ##
 
 SYNC_DATABASE_URL = 'mysql://sample_user:sample_password@localhost/syncstorage_rs'
+PATH_TO_SYNC_SPANNER_KEYS = `pwd`service-account.json
+
+# TODO: replace with rust grpc alternative when ready
```
tublitzed: Is there a better one for me to link to than the one in `server-syncstorage` since that repo is going away?

.gitignore
```diff
@@ -25,8 +25,6 @@ pypy/
 ./src/
 .tox/
 .eggs/
-autopush_rs/target
```
tublitzed: unrelated, but I figured I'd kill these while in here.

Makefile
jrconlin: might want to insert a `/` before `services-account.json`. Platforms like Linux don't automatically add a path separator to pwd.

jrconlin: Also, the google key files could be anywhere. I think that google suggests a common `~/keys` directory or something. I think having one under a given project as a default may be good. If a given user has a known key dir, they can specify it as an env that make will pick up. We could add a comment to the Makefile indicating that.

PR #400, opened 2020-01-07
Add basic logging to stdout and statsd metrics for purge_ttl.py

erkolson:
## Description

Add start/stop/duration logging for the purge_ttl.py script as well as post the durations timings to a
statsd server (defaults to `localhost:8125`, overrides with env variables `STATSD_HOST:STATSD_PORT`

## Testing

set env variables for:
```
export GOOGLE_APPLICATION_CREDENTIALS=path/to/credentials.json
export SYNC_DATABASE_URL=<url-to-database>
```
[optional] Listen for statsd in another shell:
```
nc -u -l 8125
```
Then execute:
```
python3 ./tools/spanner/purge_ttl.py
```

## Issue(s)

Closes [link](link).

### DIFF
--- a/tools/spanner/purge_ttl.py
+++ b/tools/spanner/purge_ttl.py
@@ -5,12 +5,23 @@
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 import os
+import sys
+import logging
+from datetime import datetime
+from statsd.defaults.env import statsd
 from urllib import parse
 
 from google.cloud import spanner
 
-# Change these to match your install.
+# set up logger
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+handler = logging.StreamHandler(sys.stdout)
+formatter = logging.Formatter('{"datetime": "%(asctime)s", "message": "%(message)s"}')
+handler.setFormatter(formatter)
+logger.addHandler(handler)
 
+# Change these to match your install.
 client = spanner.Client()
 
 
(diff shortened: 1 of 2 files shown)


jrconlin:
Sort of nit:
for `clog` to note this change properly, we usually state the description of the PR with `feat:` for feature, `bug:` for fix, `chore:` for a general task, etc. (see the [Contributing](https://github.com/mozilla-services/syncstorage-rs/blob/master/CONTRIBUTING.md) doc)

If possible, could you modify the git commit message (not the title here) to flag this as 

`feat: add basic logging to stdout and statsd metrics for purge_ttl.py`

Thanks!

tools/spanner/purge_ttl.py
```diff
@@ -36,21 +47,36 @@ def spanner_read_data(request=None):
...
     # Delete Batches. Also deletes child batch_bsos rows (INTERLEAVE
     # IN PARENT batches ON DELETE CASCADE)
+    batches_start = datetime.now()
     query = 'DELETE FROM batches WHERE expiry < CURRENT_TIMESTAMP()'
     result = database.execute_partitioned_dml(query)
-    outputs.append("batches: removed {} rows".format(result))
+    batches_end = datetime.now()
+    batches_duration = batches_end - batches_start
+    logger.info("batches: removed {} rows, batches_duration: {}".format(result, batches_duration))
+    statsd.timing("sync.purge_ttl.batches_duration", batches_duration)
```
pjenvey: This should really be converted to milliseconds. There's a context manager that can handle all this for you:

```
with statsd.timer("syncstorage.purge_ttl.batches_duration"):
    query = 'DELETE FROM batches WHERE expiry < CURRENT_TIMESTAMP()'
    ...
```

pjenvey: syncstorage-rs uses the "syncstorage" prefix"
```suggestion
    statsd.timing("syncstorage.purge_ttl.batches_duration", batches_duration)
```

tools/spanner/purge_ttl.py
```diff
@@ -5,12 +5,23 @@
...
+import logging
+from datetime import datetime
+from statsd.defaults.env import statsd
 from urllib import parse
 
 from google.cloud import spanner
 
-# Change these to match your install.
+# set up logger
+logger = logging.getLogger()
```
pjenvey: nit: FYI alternatively there's `logging.basicConfig` which can do pretty much all this in one function call

tools/spanner/purge_ttl.py
erkolson: noted

erkolson: I think deltas from datetime don't require multiplying by 1000, but this context manager is much simpler, thanks!

erkolson:
We've come this far, might as well do it right.

PR #401, opened 2020-01-07
feat: User migration scripts

jrconlin:
Issue: #286

## Description
Several scripts to assist in user migration from MySQL to spanner. Consult the `README.md` file for details on use and configuration.

## Testing

1. Set up a [local syncserver](https://mozilla-services.readthedocs.io/en/latest/howtos/run-sync-1.5.html), connect, and store some BSO data.
2. Configure the `move_dsns.lst` config file to point to the correct mysql and spanner instances. (e.g. 
```
mysql://test:test@localhost/syncstorage
spanner://projects/sync-spanner-dev-225401/instances/spanner-test/databases/sync_schema3
```
3. Store the known local syncserver userID into `move_users.lst` (e.g. `1`)
4. After running setup (`python3 -m venv venv && . venv/bin/activate && pip install -r requirements.txt`) Run the `python3 migrate_user.py` script.
5. user data record should be copied from mysql to spanner, Note the 
`Processsing... # -> ###:###` 
which displays the original mysql `userid` and the spanner `fxa_uid:fxa_kid`

## Issue(s)

Issue #286, https://github.com/mozilla-services/services-engineering/issues/18


tublitzed:
@jrconlin just to clarify, this is related to [#18](https://github.com/mozilla-services/services-engineering/issues/18), not 186 or 286, y?

jrconlin:
Sigh, typo. I've updated to address the two issues I believe are related. (Hilariously, I got the issue
correct in the git PR.)

tublitzed:
np, just making sure I wasn't missing something

tools/user_migration/migrate_user.py
```diff
@@ -0,0 +1,288 @@
...
+            print("Processing... {} -> {}:{}".format(
+                user, fxa_uid, fxa_kid))
+            for (cid, bid, exp, mod, pay, sid) in cursor:
+                # columns from sync_schema3
+                mod_v = datetime.utcfromtimestamp(mod/1000)
+                exp_v = datetime.utcfromtimestamp(exp)
+                # User_Collection can only have unique values. Filter
+                # non-unique keys and take the most recent modified
+                # time. The join could be anything.
+                uc_key = "{}_{}_{}".format(fxa_uid, fxa_kid, cid)
```
pjenvey: this key could just be a tuple of the values (probably not much of an optimization though)

tools/user_migration/migrate_user.py
```diff
@@ -0,0 +1,288 @@
...
+    count = 0
+    with databases['spanner'].batch() as batch:
+        cursor = databases['mysql'].cursor()
+        try:
+            cursor.execute(sql.format(user))
+            print("Processing... {} -> {}:{}".format(
+                user, fxa_uid, fxa_kid))
+            for (cid, bid, exp, mod, pay, sid) in cursor:
+                # columns from sync_schema3
+                mod_v = datetime.utcfromtimestamp(mod/1000)
```
pjenvey:
```suggestion
                mod_v = datetime.utcfromtimestamp(mod/1000.0)
```

tools/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+* [hawk](hawk) - a tool for generating test HAWK authorization headers
+* [spanner](spanner) - Google Cloud Platform Spanner tools for maintenance and testing
+* [user_migration](user_migration) - scripts for dumping and moving user data from SQL to Spanner
+
+## Installation
+
+These tools are mostly written in python. It is recommended that you create a commonly shared virtual environment using something like:
+
+`python3 -m venv venv`
```
pjenvey: should make it clear here and probably in the other README that user_migration is python2

tools/user_migration/migrate_user.py
```diff
@@ -0,0 +1,288 @@
...
+    """.format(uid=user)
+    cursor = databases['mysql'].cursor()
+    cursor.execute(sql)
+    (email, generation, keys_changed_at, client_state) = cursor.next()
+    fxa_uid = email.split('@')[0]
+    fxa_kid = format_key_id(
+        keys_changed_at or generation,
+        bytes.fromhex(client_state),
+    )
+    cursor.close()
```
pjenvey: not worth changing now (unless we start using pypy for this) but I think these calls can be made via the higher level sqlalchemy connection object which should support the with statement (vs cursors)

tools/user_migration/migrate_user.py
```diff
@@ -0,0 +1,288 @@
...
+                    uc_values = [(
+                        fxa_kid,
+                        fxa_uid,
+                        cid,
+                        mod_v,
+                    )]
+                    if args.verbose:
+                        print("### uc:")
+                        dumper(uc_columns, uc_values)
+                    batch.insert(
```
pjenvey: we need some additional work here somewhat similar to what this block is doing: non std entries in the `collections` table should be lazily added there

tools/user_migration/migrate_user.py
```diff
@@ -0,0 +1,288 @@
...
+    cursor.execute(
+        """
+        UPDATE
+            users
+        SET
+            replaced_at = {timestamp},
+            nodeid = {nodeid}
+        WHERE
+           uid = {uid}
+        """.format(
```
pjenvey: I can't help but be worried by regular str interpolation vs bind parameters but I guess it's harmless here

tools/user_migration/migrate_user.py
```diff
@@ -0,0 +1,288 @@
...
+    ORDER BY modified DESC"""
+
+    count = 0
+    with databases['spanner'].batch() as batch:
+        cursor = databases['mysql'].cursor()
+        try:
+            cursor.execute(sql.format(user))
+            print("Processing... {} -> {}:{}".format(
+                user, fxa_uid, fxa_kid))
+            for (cid, bid, exp, mod, pay, sid) in cursor:
```
pjenvey: with a large enough set of bsos, the spanner `batch()` could run into the mutation limit :(

tools/README.md
jrconlin: `migrate_user.py` can run under Python3. the `dump_*.py` however can't because of a requirement in arvo.

tools/user_migration/migrate_user.py
jrconlin: Probably. `with` should call close automatically. I was skipping that here to aid in debugging.

tools/user_migration/migrate_user.py
jrconlin: Ah, good point.

tools/user_migration/migrate_user.py
jrconlin: Really hoping that since this is per-user based, a user doesn't have *that* many BSOs, but yeah. I'm debating switching this to use a `transaction()` internal call, but that seems equally gross.

tools/user_migration/migrate_user.py
jrconlin: Not a fan either, but I couldn't find an interpolation function inside of `mysql.connector`. The examples in the library use python string formatters. Since this is being executed by trusted individuals, I'm not quite as uneasy.

tools/user_migration/migrate_user.py
jrconlin: Wait, looking at the code, I'm lazily adding `user_collections`, but maybe I was presuming that `collections` is fairly static. Do I need to add unique user collections to that table as well?

pjenvey: Nope, this part is correct. Users get a `user_collections` entry for each collection they have `bsos` for.

And every unique collection gets its own entry in the top level `collections` table, storing its id and name shared between all users. So users end up creating arbitrary collections when storing bsos with a previously unknown collection for the firs time.

jrconlin: 🐉🐉🐉🐉🐉🐉🐉🐉
... 
Ok, so there may be a problem here. Since `user_collections` are "unique" to a node, there's the potential that two different nodes have different values for "unique", this could lead to collisions or duplication. 

Need to research this a bit to determine the impacts or how we want to handle this. 

🐉🐉🐉🐉🐉🐉🐉🐉

jrconlin: Looking into the problem:

`collection` is the only place where a given `collection_id` is mapped to a `collection_name`. `collection_name` is used as one of the primary locking mechanisms. Most users only have well known collections, so it should be possible to return a well known collection id. The problem is that some users may have "custom" collections which have a different value. 

Option 1:
Spanner makes modifications of table indexes problematic. We *could* modify user_collections to include "name" as an index field, but we would have to either replicate the data in the existing user_collections to a new table, or just create a new table that contains
`fxa_uid`, `collection_id`, `collection_name` (all indices) and use that as the fail over lookup table for collection_id if it's not one of the well-known ones.

Option 2:
Modify user_collection tables to include the name field and do a secondary index lookup on that if no well-known collections found.

There may be other options as well, but those are the two I've thought of tonight.

No idea how this might impact adding new users 2020-01-13

tublitzed:
@jrconlin following the setup instructions here, regarding step 2: where does the `move_dsns.lst` file live? I'm not seeing it here in this repo...

tublitzed:
^ ah, nvm. I see now. Following the [example here](https://github.com/mozilla-services/syncstorage-rs/pull/401/files#diff-4cd4f4923ed388113445bcedb5b6c437R11) going to assume I create this file and that it's used as the `dsns` arg. Trying that now....

tools/user_migration/migrate_user.py
jrconlin: Another option:

Client appears to only provide collection by name, which is a `String`. The migration script could load the known `user_collections` into a local cache, and remap the collection IDs per user. This should require the least amount of db modification, but potentially could impact the speed of migration.

tools/user_migration/migrate_user.py
```diff
@@ -0,0 +1,288 @@
...
+    data.
+
+    """
+    sql = """
+        SELECT
+            email, generation, keys_changed_at, client_state
+        FROM users
+            WHERE uid = {uid}
+    """.format(uid=user)
+    cursor = databases['mysql'].cursor()
```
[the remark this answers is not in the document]
jrconlin: Actually, no. This needs to come from the mysql db since we're pulling data from the `users` table.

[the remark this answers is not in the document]
pjenvey: But this is tokenserver's `users` table (w/ the `keys_changed_at` etc columns). Python syncstorage doesn't have a users table

tublitzed:
When I run `python3 migrate_user.py --dsns move_dsns.lst --users move_users.lst` (with the `.lst` files in the same dir as the script) I see the following error:

```
  File "migrate_user.py", line 380, in <module>
    main()
  File "migrate_user.py", line 365, in main
    databases[dsn.scheme] = conf_db(dsn)
  File "migrate_user.py", line 135, in conf_db
    return conf_mysql(dsn)
  File "migrate_user.py", line 112, in conf_mysql
    database=dsn.path[1:]
  File "/Users/tublitzed/GIT/syncstorage-rs/tools/user_migration/venv/lib/python3.7/site-packages/mysql/connector/__init__.py", line 179, in connect
    return MySQLConnection(*args, **kwargs)
  File "/Users/tublitzed/GIT/syncstorage-rs/tools/user_migration/venv/lib/python3.7/site-packages/mysql/connector/connection.py", line 95, in __init__
    self.connect(**kwargs)
  File "/Users/tublitzed/GIT/syncstorage-rs/tools/user_migration/venv/lib/python3.7/site-packages/mysql/connector/abstracts.py", line 716, in connect
    self._open_connection()
  File "/Users/tublitzed/GIT/syncstorage-rs/tools/user_migration/venv/lib/python3.7/site-packages/mysql/connector/connection.py", line 210, in _open_connection
    self._ssl)
  File "/Users/tublitzed/GIT/syncstorage-rs/tools/user_migration/venv/lib/python3.7/site-packages/mysql/connector/connection.py", line 142, in _do_auth
    auth_plugin=self._auth_plugin)
  File "/Users/tublitzed/GIT/syncstorage-rs/tools/user_migration/venv/lib/python3.7/site-packages/mysql/connector/protocol.py", line 102, in make_auth
    auth_data, ssl_enabled)
  File "/Users/tublitzed/GIT/syncstorage-rs/tools/user_migration/venv/lib/python3.7/site-packages/mysql/connector/protocol.py", line 58, in _auth_response
    auth = get_auth_plugin(auth_plugin)(
  File "/Users/tublitzed/GIT/syncstorage-rs/tools/user_migration/venv/lib/python3.7/site-packages/mysql/connector/authentication.py", line 191, in get_auth_plugin
    "Authentication plugin '{0}' is not supported".format(plugin_name))
mysql.connector.errors.NotSupportedError: Authentication plugin 'caching_sha2_password' is not supported
(venv)
```

Anyone else seeing this? I'm running from OSX if that helps.

tublitzed:
Looks like this might be a known issue with [the caching_sha2_password plugin against newer MySQL versions](https://stackoverflow.com/questions/49194719/authentication-plugin-caching-sha2-password-cannot-be-loaded)

I have tried the suggested alter workaround to no avail. @jrconlin -what version of MySQL are you testing this against? Looks like 8.x might just be too new for `caching_sha2_password`...in which case I can try against an older version. Figured I'd check though before going through that :) 

(FWIW, I don't think it matters if this script might require an older MySQL version, just noting this for anyone else trying to test this out)

jrconlin:
@tublitzed My server is reporting it's "5.7.28-0ubuntu0.18.04.4", which is what I get out of `apt`/debian, which might be stable locked for ubuntu.

jrconlin:
@pjenvey Fixed the `collection_id` to be in the same transaction, but tempted to leave the other two points as future bugs, mostly because i'll need to create a few datasets to try and replicate those so I understand what the breaking points are.

tools/user_migration/migrate_user.py
```diff
@@ -0,0 +1,374 @@
...
+            for row in rows:
+                self._by_name[row[0]] = row[1]
+
+    def get_id(self, name):
+        """ Get/Init the ID for a given collection """
+        if name in self._by_name:
+            return self._by_name.get(name)
+
+        # option 1: do what spanner-rs does
+        with self.databases['spanner'].batch() as cursor:
```
pjenvey: Does batch() support execute_sql? You may need to switch to transaction() instead

tools/user_migration/migrate_user.py
```diff
@@ -0,0 +1,374 @@
...
+            result = cursor.execute_sql("""
+                SELECT
+                    COALESCE(MAX(collection_id), 1)
+                FROM
+                    collections""")
+            collection_id = result.one()[0] + 1
```
pjenvey: I forgot this:
```suggestion
            collection_id = max(result.one()[0] + 1, 101)
```

jrconlin: Not sure about this. I'm guessing that we're going to pre-populate the `collections` table with the knowns. I'm guessing we'd also include a `100:reserved` marker so that the custom values start above that.

Granted, the `collection_id` is only used by the server. I'm not really sure what value the reserved range really is, but that's just my ignorance.

tools/user_migration/migrate_user.py
jrconlin: Batch does support `execute_sql`. I believe it's part of the base class for `transaction`, `snapshot` and `batch`.

pjenvey: Batch is different, _BatchBase only supports mutation operations: https://github.com/googleapis/google-cloud-python/blob/ee6214251fbe83cb83a9b5b5019c605a27a01354/spanner/google/cloud/spanner_v1/batch.py#L29 (_SessionWrapper has nothing)

tools/user_migration/migrate_user.py
pjenvey: The standard/reserved collections we hardcode should have been pre-populated in the collections table (but we should really double check that this did indeed happen on prod first).

I'm leaning towards implementing this max line for now, to match the syncstorage implementations. But we should probably have a conversation with rfk about how this reservation's done and what we want to do going forward

tublitzed:
@jrconlin I downgraded to MySQL 5.7, still seem same errors from OSX. It'd probably be good to confirm we can get at least a few people able to run this locally so we can effectively test client behaviour internally before handing off to QA. IIRC the plan was to test locally first, so I don't want that to fall entirely to you if you're the only one who can run this :)

Also curious if maybe it's just my machine. @pjenvey are you able to run this locally?

jrconlin:
Also kinda curious if this might be a mac thing vs. linux thing. (if it is, i am even LESS enamored with mysql). 

IIUC, the scripts would be run from debian sh/bash shells, since that's what GCP provides. I'm running ubuntu, which is a debian fork, so the libraries should match up reasonably well.

tools/user_migration/migrate_user.py
```diff
@@ -0,0 +1,372 @@
...
+
+    """
+    sql = """
+        SELECT
+            email, generation, keys_changed_at, client_state
+        FROM users
+            WHERE uid = {uid}
+    """.format(uid=user)
+    try:
+        cursor = databases['mysql'].cursor()
```
pjenvey: We forgot this or it got lost somewhere
```suggestion
        cursor = databases['token'].cursor()
```

jrconlin: Ah, I figured out what the confusion is from. For stand alone, these tables live on the same server. Adding logic to test for token first, then mysql.

tools/user_migration/migrate_user.py
```diff
@@ -139,7 +140,9 @@ def update_token(databases, user):
         logging.warn("Skipping token update...")
         return
     logging.info("Updating token server for user: {}".format(user))
-    with databases['token'].cursor() as cursor:
+    try:
+        cursor = databases['token'].cursor()
+        cursor = databases['token'].cursor()
```
pjenvey: dupe

tools/user_migration/migrate_user.py
```diff
@@ -193,6 +199,19 @@ def get_fxa_id(databases, user):
     return (fxa_kid, fxa_uid)
 
 
+def create_migration_table(database):
+    try:
+        cursor = database.cursor()
+        cursor.execute(
+            """create table if not exists migration (
+                fxa_uid varchar(255) NOT NULL PRIMARY KEY,
```
pjenvey: With this being for the python syncstorage side to reference, I think it will be much easier to use its "legacy" `userid` vs syncstorage-rs's `fxa_uid`

tools/user_migration/migrate_user.py
```diff
@@ -240,6 +276,9 @@ def move_user(databases, user, args):
 
     # Genereate the Spanner Keys we'll need.
     (fxa_kid, fxa_uid) = get_fxa_id(databases, user)
+    if not mark_user(databases, fxa_uid):
+        logging.error("User {} already being migratted?".format(fxa_uid))
```
pjenvey:
```suggestion
        logging.error("User {} already being migrated?".format(fxa_uid))
```

tools/user_migration/migrate_user.py
```diff
@@ -193,6 +199,19 @@ def get_fxa_id(databases, user):
     return (fxa_kid, fxa_uid)
 
 
+def create_migration_table(database):
+    try:
+        cursor = database.cursor()
+        cursor.execute(
+            """create table if not exists migration (
+                fxa_uid varchar(255) NOT NULL PRIMARY KEY,
+                started BIGINT
```
pjenvey: probably want NOT NULL?

tools/user_migration/migrate_user.py
jrconlin: thanks, merge weirdness.

tools/user_migration/migrate_user.py
jrconlin: I was debating this. To be honest, we only really need some ID for the user. I figured that "started" might be useful for some diagnostics. Still, adding `NOT NULL` is harmless since we're always setting it to `int(time.time())` since it doesn't need to be precise either.

tools/user_migration/migrate_user.py
jrconlin: Yeah, but "userid" is kinda db dependent since [it can come off of the db row](https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/assignment/sqlnode/sql.py#L368-L396). fxa_uid isn't. I'd rather there be zero chance of collision here, so I'd prefer to use fxa_uid.

pjenvey: Cool, I forgot it's really not difficult to get at either (it's already setup on the user dict/object "fxa_uid" field).

PR #405, opened 2020-01-09
Filter out "" values from sentry

### LINKED ISSUE #404: Don't send empty tags to sentry
Apparently sentry dislikes empty string values for tags, emitting in events "There were X errors encountered while processing this event" "Discarded invalid value for parameter 'tags' "

E.g. https://sentry.prod.mozaws.net/operations/syncstorage-stage/issues/7036251/?query=is:unresolved

Looking at the raw JSON it seems due to tag values being `""`.

jrconlin:
Closes #404

## Description

Sentry flags empty strings `""` as an error, so filter those out.

## Testing

unit tests include `no_empty_tags()` which tests that "ua.os.ver" is not set for a UA that does not contain an OS version.

## Issue(s)

Closes #404

### DIFF
--- a/src/web/tags.rs
+++ b/src/web/tags.rs
@@ -32,7 +32,7 @@ impl Serialize for Tags {
     {
         let mut seq = serializer.serialize_map(Some(self.tags.len()))?;
         for tag in self.tags.clone() {
-            if &tag.1 != "" {
+            if !tag.1.is_empty() {
                 seq.serialize_entry(&tag.0, &tag.1)?;
             }
         }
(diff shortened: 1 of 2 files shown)


src/web/tags.rs
```diff
@@ -32,12 +32,20 @@ impl Serialize for Tags {
     {
         let mut seq = serializer.serialize_map(Some(self.tags.len()))?;
         for tag in self.tags.clone() {
-            seq.serialize_entry(&tag.0, &tag.1)?;
+            if &tag.1 != "" {
```
pjenvey: nit: to be consistent 
```suggestion
            if !tag.1.is_empty() {
```

PR #415, opened 2020-01-15
Upgrade to actix web 2

### LINKED ISSUE #394: Upgrade to actix web 2.x
It's probably worth making time to upgrade to 2.x before we get too far behind. Looks like 2.0 went live [9 days ago](https://github.com/mozilla-services/syncstorage-rs/issues/348), so not a huge rush, but filing so we've got it.

Can we do the actix upgrade as a standalone task without also converting everything to async/await? (we have a [separate task](https://github.com/mozilla-services/syncstorage-rs/issues/348) filed for async/await changes).

fzzzy:
## Description

This pr updates to actix web 2 and future 0.3, but does not switch anything to async await.

## Testing

Assuming it compiles, cargo test, cargo test in db-tests, and the python tests all pass, it should be good to go.

## Issue(s)

Closes #394

### DIFF
--- a/src/db/spanner/support.rs
+++ b/src/db/spanner/support.rs
@@ -1,10 +1,13 @@
 use std::{
+    cell::RefCell,
     collections::{HashMap, VecDeque},
     fmt, mem,
     result::Result as StdResult,
 };
 
-use futures::stream::{Stream, Wait};
+use actix_rt::{System, SystemRunner};
+use futures::compat::{Compat01As03, Stream01CompatExt};
+use futures::stream::{StreamExt, StreamFuture};
 use googleapis_raw::spanner::v1::{
     result_set::{PartialResultSet, ResultSetMetadata, ResultSetStats},
     spanner::ExecuteSqlRequest,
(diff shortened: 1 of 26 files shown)


HenkPoley:
A slight problem is that the Actix project stopped existing at the moment.

fzzzy:
Yes... that is a slight problem...

src/db/spanner/support.rs
```diff
@@ -188,7 +189,13 @@ impl StreamedResultSet {
     ///
     /// Returns false when the stream is finished
     fn consume_next(&mut self) -> Result<bool> {
-        let mut partial_rs = if let Some(result) = self.stream.next() {
+        let (result, stream) = System::new("").block_on(
```
fzzzy: I'm pretty sure the overhead of calling System::new("") on every single bsos request is going to be untenable.

fzzzy: I changed it so that there is a threadlocal System for every thread in the db pool.

fzzzy:
Ok issues are filed for all comments. Merging.

PR #417, opened 2020-01-17
feat: recategorize logging messages into appropriate states

### LINKED ISSUE #416: Make sure the various messages we're logging are properly categorized

jrconlin:
Closes #416

## Description

* error: requires action or attention
* warn: non-critical event impacting performance or function
* info: notification of interest
* debug: crap

## Testing

no code impact (aside from logging)

## Issue(s)

Closes #416.

### DIFF
--- a/src/server/metrics.rs
+++ b/src/server/metrics.rs
@@ -48,7 +48,7 @@ impl Drop for Metrics {
                 match tagged.try_send() {
                     Err(e) => {
                         // eat the metric, but log the error
-                        debug!("⚠️ Metric {} error: {:?} ", &timer.label, e);
+                        warn!("⚠️ Metric {} error: {:?} ", &timer.label, e);
                     }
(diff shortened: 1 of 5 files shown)


pjenvey:
FYI I think this makes us more verbose overall.

The release build has `release_max_level_info` set in Cargo.toml, which'll render INFO/WARN/ERROR.

Most messages should lean towards debug and or maybe we should lower the `release_max_level_`.

jrconlin:
The approach I took was:
Serious actionable errors are `error!` this includes the "no app state" which is a significant error we'd need to address.
Errors that do not require action are `warn!` and informational messages are `info!`, these are messages that could be useful for early monitoring.
`debug!` is reserved for low level messages.

I'd be fine with setting `release_max_level` to error for production, or drop it to warn if we're seeing unusual behaviors we want to track.

PR #421, opened 2020-02-05
bug: filter out variable data from URI metric

### LINKED ISSUE #420: Don't track uri path as Sentry tag.
Ie, https://sentry.prod.mozaws.net/operations/syncstorage-prod/issues/7051423/tags/uri.path/

Given that it's unique per user, it's likely we'll blow up Sentry this way; can either remove this, or track in a way that doesn't create a new tag for each sync id?

jrconlin:
## Description

filter out UID and batch numbers from URI.

This drops the first two path elements from the URI as well as masks any large hexidecimal identifier that may appear in the URI parameters. This should retain enough diagnostic information to be useful without causing overload for sentry.

## Testing

While running normal testing note that the `uri.path` metric goes from something like:
`/1.5/135772380/storage/history?batch=5ef59eae10244edc90935bab2ffea0f0&commit=true`
to 
`storage/history?batch=###&commit=true`

## Issue(s)

Closes #420

### DIFF
--- a/src/web/middleware/sentry.rs
+++ b/src/web/middleware/sentry.rs
@@ -67,12 +67,14 @@ where
             // most values are either "=true" or a lower case hex string.
             static ref RE: regex::Regex = regex::Regex::new(r"=[a-z\d]{5,}").unwrap();
         }
+        // to satisify clippy and rustc.
+        let slash: &str = "/";
         let trimmed = uri
             .to_string()
-            .split('/')
+            .split(slash)
             .collect::<Vec<&str>>()
             .split_off(3)
-            .join('/');
+            .join(slash);
         RE.replace_all(&trimmed, "=###").to_owned().to_string()
     }
 }
--- a/src/web/middleware/sentry.rs
+++ b/src/web/middleware/sentry.rs
@@ -92,7 +92,7 @@ where
                     if let Some(t) = sresp.response().extensions().get::<Tags>() {
                         debug!("Found response tags: {:?}", &t.tags);
                         for (k, v) in t.tags.clone() {
-                            tags.extra.insert(k, v);
+                            tags.tags.insert(k, v);
                         }
                     };
                     // add the uri.path (which can cause influx to puke)
--- a/src/web/middleware/sentry.rs
+++ b/src/web/middleware/sentry.rs
@@ -119,18 +119,19 @@ where
                     if let Some(t) = sresp.response().extensions().get::<Tags>() {
                         debug!("Found response tags: {:?}", &t.tags);
                         for (k, v) in t.tags.clone() {
-                            tags.tags.insert(k, v);
+                            tags.extra.insert(k, v);
                         }
                     };
                     // add the uri.path (which can cause influx to puke)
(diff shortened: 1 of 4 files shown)


src/web/middleware/sentry.rs
```diff
@@ -52,6 +55,28 @@ pub struct SentryWrapperMiddleware<S> {
...
+where
+    S: Service<Request = ServiceRequest, Response = ServiceResponse<B>, Error = Error> + 'static,
+    S::Future: 'static,
+    B: 'static,
+{
+    fn clean_uri(&self, uri: &Uri) -> String {
+        // Remove the variable userID and batch tokens from the URI tags
+        lazy_static! {
+            // most values are either "=true" or a lower case hex string.
+            static ref RE: regex::Regex = regex::Regex::new(r"=[a-z\d]{5,}").unwrap();
```
fzzzy: Just to be sure I understand, this matches anything starting with = followed with any lowercase letter or number; and it has to match at least 5 letters or numbers in order to not have it match on the string `=true`. Is that right?

jrconlin: Yep, that's it exactly. I was debating about making it match case independent, but most of the ones I'm seeing are lower case.  If folks feel otherwise, it's an easy fix.

src/web/middleware/sentry.rs
```diff
@@ -52,6 +55,28 @@ pub struct SentryWrapperMiddleware<S> {
...
+        }
+        let trimmed = uri
+            .to_string()
+            .split('/')
```
fzzzy: I think these have to be "

src/web/middleware/sentry.rs
```diff
@@ -52,6 +55,28 @@ pub struct SentryWrapperMiddleware<S> {
...
+            .collect::<Vec<&str>>()
+            .split_off(3)
+            .join('/');
```
fzzzy: Same here

src/web/middleware/sentry.rs
jrconlin: They do, in spite of clippy saying otherwise. I've added a bit of a crap patch with comment to deal with that.

pjenvey:
I don't think it's terribly important for uri.path to be a sentry tag. It could be sent as an "extra" instead, which isn't indexed.

A similar argument could probably be made for other tags we currently send.

src/web/middleware/sentry.rs
```diff
@@ -95,8 +94,6 @@ where
                             tags.tags.insert(k, v);
                         }
                     };
```
pjenvey: let's continue sending it to Sentry but in its extras for now

jrconlin: Sadly, the lib we're using doesn't understand how to send "extras", only "tags".

pjenvey: Isn't this it here? https://docs.rs/sentry/0.17.0/sentry/protocol/struct.Event.html#structfield.extra

jrconlin: Ah, I might be confusing this with the cadence library. Sorry. That kind of morning.

Issue #389, opened 2019-12-31
Remove temporary sentry tags for the mutation limit issue


pjenvey:
Once #329 lands we should drop the added debugging code added for #333 (#373).

PR #423, opened 2020-02-06
refactor: kill unnecessary copies from protobuf Values

### LINKED ISSUE #422: [Spanner] Don't copy the payload String
The spanner impl has a few calls of `Value::get_string_value().to_owned()` which makes a pointless copy, in one case on the larger payload field (in `bso_from_row`).

This should call `take_string_value()` instead.

pjenvey:
## Description

and simplify get_bso_ids' processing

also revert the temp. sentry tags for the mutation limit issue

## Testing

db/integration tests

## Issue(s)

Closes #422, #389

### DIFF
--- a/src/db/spanner/batch.rs
+++ b/src/db/spanner/batch.rs
@@ -152,107 +152,6 @@ pub fn delete(db: &SpannerDb, params: params::DeleteBatch) -> Result<()> {
     Ok(())
 }
 
-pub fn _audit_commit_err<T>(
-    db: &SpannerDb,
-    params: &params::CommitBatch,
-    result: Result<T>,
-) -> Result<T> {
-    if let Err(ref e) = result {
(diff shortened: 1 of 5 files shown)


src/db/spanner/batch.rs
```diff
@@ -152,107 +152,6 @@ pub fn delete(db: &SpannerDb, params: params::DeleteBatch) -> Result<()> {
     Ok(())
 }
 
-pub fn _audit_commit_err<T>(
```
jrconlin: pretty sizable delete here. IIUC, we're no longer auditing if the GRPC call fails, correct?

pjenvey: We'll still get the normal sentry reporting for it, all this code predated the sentry capturing w/ tags (so some of it's now duplicated).

The mutation counts were nice to have to diagnose the original issue, but are no longer necessary

PR #429, opened 2020-02-12
bug: convert migration state to smallint

### LINKED ISSUE #428: bug: Convert migration state flag to small int.
See https://github.com/mozilla-services/server-syncstorage/issues/143

jrconlin:
## Description

Alter the migration table to use SmallInt rather than string. Also add state transitions during migrations.

## Testing

* Remove the `migration` table if it already exists (there is no migration path to update since this table is new).
* run a test migration as noted in https://github.com/mozilla-services/syncstorage-rs/pull/401

## Issue(s)

Closes #428 

(see also https://github.com/mozilla-services/server-syncstorage/pull/144)

### DIFF
--- a/tools/user_migration/migrate_user.py
+++ b/tools/user_migration/migrate_user.py
@@ -25,6 +25,11 @@ SPANNER_NODE_ID = 800
 class BadDSNException(Exception):
     pass
 
+# From server_syncstorage
+class MigrationState:
+    UKNOWN = 0
+    IN_PROGRESS = 1
+    COMPLETE = 2
 
 class Collections:
     """Cache spanner collection list.
@@ -219,7 +224,8 @@ def create_migration_table(database):
             """CREATE TABLE IF NOT EXISTS
                 migration (
                     fxa_uid VARCHAR(255) NOT NULL PRIMARY KEY,
-                    started BIGINT NOT NULL
+                    started BIGINT NOT NULL,
+                    state SMALLINT
                 )
             """)
         database.commit()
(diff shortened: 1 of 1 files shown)


tools/user_migration/migrate_user.py
```diff
@@ -236,18 +242,30 @@ def dumper(columns, values):
...
+        if state == MigrationState.COMPLETE:
+            try:
+                logging.info("Marking {} as migrating...".format(user))
+                mysql.execute(
+                    "UPDATE migration SET state = %s WHERE fxa_uid = %s",
+                    (state, user)
+                )
+                databases['mysql'].commit()
+            except IntegrityError:
+                return False
```
pjenvey: Should this be logged maybe or handled by the caller (it's ignored for `COMPLETE`)

pjenvey:
Looks good other than my question.

Reminder: didn't we have some other changes to apply to the migration script here that occurred w/ the last server-syncstorage PR?

tools/user_migration/migrate_user.py
jrconlin: Heh, I actually think that this is a copy/paste thing. I don't believe that an Update would trigger the IntegrityError since it doesn't alter the key index, just sets the value. Will remove that.

tools/user_migration/migrate_user.py
```diff
@@ -219,7 +226,8 @@ def create_migration_table(database):
             """CREATE TABLE IF NOT EXISTS
                 migration (
                     fxa_uid VARCHAR(255) NOT NULL PRIMARY KEY,
-                    started BIGINT NOT NULL
+                    started BIGINT NOT NULL,
```
pjenvey: This was the other change:

```suggestion
                    started_at BIGINT NOT NULL,
```

and elsewhere in the file

PR #445, opened 2020-02-24
Fix #432 Convert trivial db functions to async/await.

### LINKED ISSUE #432: Convert the trivial functions to async/await
Related to #431 

 - commit
 - rollback
 - check
 - Tags::from_request
 - BsoParam::from_request
 - CollectionParam::from_request
 - TestErrorRequest::from_request
 - Box::from_request
 - DbTransaction::new_transform
 - PreConditionCheck::new_transform
 - SentryWrapper::new_transform
 - WeaveTimestamp::new_transform

fzzzy:
## Description

Convert the trivial spanner db functions to async/await.

## Testing

All tests should pass.

## Issue(s)

Closes #432.

### DIFF
--- a/src/web/extractors.rs
+++ b/src/web/extractors.rs
@@ -15,7 +15,7 @@ use actix_web::{
     Error, FromRequest, HttpMessage, HttpRequest,
 };
 
-use futures::future::{self, FutureExt, LocalBoxFuture, Ready, TryFutureExt};
+use futures::future::{self, FutureExt, LocalBoxFuture, TryFutureExt};
 
 use lazy_static::lazy_static;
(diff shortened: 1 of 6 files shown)


pjenvey:
None of these blocks need `.await` calls currently so I'd argue they're simpler/maybe clearer without this conversion (e.g. seeing a `Ready` vs a `LocalBoxFuture` seems informative). It doesn't help that they can't become `async` fn/methods (because they're trait methods).

fzzzy:
That's true, and I did think about that already, but I thought consistency was better. But I'm fine with leaving them as Ready.

PR #449, opened 2020-02-26
Fix #434 Convert db collection calls to async await.

### LINKED ISSUE #434: Convert db collection calls to async/await
- ~~get_collection_timestamp~~
 - ~~get_collection_timestamps~~
 - ~~get_collection_counts~~
 - ~~get_collection_usage~~
 - ~~delete_collection~~

fzzzy:
## Description

This branch converts all the methods related to collections.

-    get_collection_timestamp
-    get_collection_timestamps
-    get_collection_counts
-    get_collection_usage
-    delete_collection
- map_collection_names
- load_collection_names

## Testing

All tests should pass.

## Issue(s)

Closes #434.

### DIFF
--- a/src/db/spanner/models.rs
+++ b/src/db/spanner/models.rs
@@ -624,7 +624,8 @@ impl SpannerDb {
             })
             .execute_async(&self.conn)?;
         let mut results = HashMap::new();
-        while let Some(row) = streaming.one_or_none().await? {
+        while let Some(row) = streaming.next_async().await {
+            let row = row?;
             let collection_id = row[0]
                 .get_string_value()
                 .parse::<i32>()
@@ -676,7 +677,8 @@ impl SpannerDb {
                 )?
                 .params(params)
                 .execute_async(&self.conn)?;
-            while let Some(mut row) = rs.one_or_none().await? {
+            while let Some(row) = rs.next_async().await {
+                let mut row = row?;
                 let id = row[0]
                     .get_string_value()
                     .parse::<i32>()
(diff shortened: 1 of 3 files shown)


src/db/spanner/models.rs
```diff
@@ -618,21 +622,21 @@ impl SpannerDb {
...
-                    .parse::<i32>()
-                    .map_err(|e| DbErrorKind::Integrity(e.to_string()))?;
-                let modified = SyncTimestamp::from_rfc3339(&row[1].get_string_value())?;
-                Ok((collection_id, modified))
-            })
-            .collect::<Result<HashMap<_, _>>>()?;
-        self.map_collection_names(modifieds)
+            .execute_async(&self.conn)?;
+        let mut results = HashMap::new();
+        while let Some(row) = streaming.one_or_none().await? {
```
pjenvey: These iterating blocks shouldn't use `one`/`one_or_none` (they error out unless there's *only* one result), did db-tests not catch this (they really should)?

You probably just want streaming.next_async().await.transpose()? (the need to transpose is a little annoying, I wonder what it would take to implement the `Stream` type for this).

fzzzy: I guess the async version doesn't error out if there is more than one, which is why it works. Shall I rename it next_or_none?

fzzzy: Huh, no, you're right, it should error if there is more than one. Weird. Yes it seems like the db tests should be catching this.

fzzzy: I guess they can just use next_async.

fzzzy:
Yeah I intentionally chose not to use macros for now... I'm hoping we can use something like https://github.com/dtolnay/async-trait eventually.

PR #452, opened 2020-02-27
refactor: Fix #435 Convert db batch calls to async await.

### LINKED ISSUE #435: Convert database batch calls to async/await
- ~~create_batch~~
 - ~~validate_batch~~
 - ~~append_to_batch~~
 - ~~get_batch~~
 - ~~commit_batch~~

fzzzy:
## Description

Convert db batch calls to async await.

-    create_batch
-    validate_batch
-    append_to_batch
-    get_batch
-    commit_batch
- delete_batch
- do_append
- pretouch_collection

## Testing

All tests should pass.

## Issue(s)

Closes #435.

### DIFF
--- a/src/db/spanner/models.rs
+++ b/src/db/spanner/models.rs
@@ -118,14 +118,6 @@ impl Deref for SpannerDb {
     }
 }
 
-macro_rules! batch_db_method {
-    ($name:ident, $batch_name:ident, $type:ident) => {
-        pub fn $name(&self, params: params::$type) -> Result<results::$type> {
-            batch::$batch_name(self, params)
-        }
-    }
-}
-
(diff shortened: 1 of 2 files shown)


src/db/spanner/models.rs
```diff
@@ -958,6 +950,8 @@ impl SpannerDb {
         }
     }
 
+    // I think we can remove this but I'm not 100% sure if the db tests use it or not.
```
pjenvey: db-tests use it via two places:

- The Db trait touch_collection (which could be converted to the async version anyway)
- The test-only versions of put/post_bso which haven't been converted to async yet, so I think we still need it for now

PR #460, opened 2020-02-28
fix: Fix #459 db-tests on master

### LINKED ISSUE #459: Fix db-test breakage on master
https://github.com/mozilla-services/syncstorage-rs/commit/9279782f607fa87577f49f86a6017515f7c5d2b0 broke db-tests against spanner on master. The db-tests have a different code flow so it may not have actually broken the real code, but it should still be fixed.

fzzzy:
## Description

The db tests became broken by a recent async await commit because there was a special, db-tests only function that didn't get converted to async. This uses the old synchronous implementation when running in db-tests.

## Testing

All tests should pass against spanner.

## Issue(s)

Closes #459.

### DIFF
--- a/src/db/spanner/models.rs
+++ b/src/db/spanner/models.rs
@@ -1855,15 +1855,34 @@ impl Db for SpannerDb {
         Box::pin(async move { db.get_bso_timestamp_async(param).map_err(Into::into).await })
     }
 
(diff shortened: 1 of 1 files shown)


jrconlin:
Looks like you need some `fmt` love, though

PR #464, opened 2020-03-03
refactor: Fix #442 Use map_ok in handlers to simplify the code and im…

### LINKED ISSUE #442: actix 2 cleanup: Change map to map_ok
based on feedback in https://github.com/mozilla-services/syncstorage-rs/pull/415#discussion_r382868153

fzzzy:
…prove error reporting.

## Description

Switching from future 0.1 to 0.3 changed the behavior of map. map_ok is closer to the old behavior and much more like what we really want.

## Testing

All tests should pass.

## Issue(s)

Closes #442.


src/web/handlers.rs
```diff
@@ -418,15 +394,12 @@ pub fn get_bso(bso_req: BsoRequest) -> impl Future<Output = Result<HttpResponse,
...
-        .map(
-            |result: std::result::Result<std::option::Option<GetBso>, Error>| {
-                let result = result.expect("Could not get result in get_bso");
-                result.map_or_else(
-                    || Ok(HttpResponse::NotFound().finish()),
-                    |bso| Ok(HttpResponse::Ok().json(bso)),
-                )
-            },
-        )
+        .map_ok(|result: std::option::Option<GetBso>| {
```
pjenvey:
```suggestion
        .map_ok(|result: Option<GetBso>| {
```

and also all the `std::collections::HashMap` -> `HashMap` (there's already a use statement for it)

pjenvey: On second thought: 

```suggestion
        .map_ok(|result| {
```

..I doubt we need any of these type annotations any longer (the original version didn't)

src/web/handlers.rs
```diff
@@ -187,12 +170,8 @@ where
...
         })
-        .map(
-            move |r: std::result::Result<
-                (Paginated<T>, std::result::Result<SyncTimestamp, Error>),
-                Error,
-            >| {
-                let (result, ts) = r.expect("Could not get r in finish_get_collection");
+        .map_ok(
+            move |(result, ts): (Paginated<T>, std::result::Result<SyncTimestamp, Error>)| {
                 let ts = ts.expect("Could not get ts in finish_get_collection");
```
pjenvey: Forgot this last `expect()`

fzzzy: No, I removed the expect for r but ts is still a result (both r and ts were results previously).

PR #469, opened 2020-03-06
fix: GETs with a limit and no sort never advance X-Weave-Next-Offset

### LINKED ISSUE #468: GETs with a `limit` and no `sort` never advance `X-Weave-Next-Offset`
I've been able to reproduce this consistently in Nightly, from the Browser Console and the About Sync add-on, when I fetch all my bookmark records (that is, no `older`, `newer`, or `ids` query params; I haven't tested if those change anything).

I have just over 4k bookmarks, so the client tries to fetch them in batches. All my other collections sync successfully, since they have under 1k records.

Here's what I see in the logs:

```
1583458802175	Sync.Collection	DEBUG	GET success 200 https://sync-1-us-west1-g.sync.services.mozilla.com/1.5/131779759/storage/bookmarks?full=1&limit=1000&offset=1575593490410%3A1
X-W-N-O = 1575593490410:1
X-L-M = 1583453382.92
X-W-R = 1000
[ ... ]
1583458802367	Sync.Collection	DEBUG	GET success 200 https://sync-1-us-west1-g.sync.services.mozilla.com/1.5/131779759/storage/bookmarks?full=1&limit=1000&offset=1575593490410%3A1
X-W-N-O = 1575593490410:1
X-L-M = 1583453382.92
X-W-R = 1000
[ ... ]
1583458802544	Sync.Collection	DEBUG	GET success 200 https://sync-1-us-west1-g.sync.services.mozilla.com/1.5/131779759/storage/bookmarks?full=1&limit=1000&offset=1575593490410%3A1
X-W-N-O = 1575593490410:1
X-L-M = 1583453382.92
X-W-R = 1000
[ ... ]
```

So it looks like `X-Last-Modified` stays the same (which is what we expect; otherwise, we'd see a 412), but `X-Weave-Next-Offset` never returns the next offset...so it gets stuck in an infinite loop fetching the first 1k records!

Please let me know if there's anything I can do to help debug! ❤️

fzzzy:
## Description

When the client does not pass a sort argument, the python server defaults to sort by oldest. The rust server defaults to sort by nothing, causing bug #468.

## Testing

The STR in the bug should show that X-Weave-Next-Offset when fetching collections with more than 1000 elements. (lina should probably test it)

## Issue(s)

Closes #468.

### DIFF
--- a/src/db/mysql/models.rs
+++ b/src/db/mysql/models.rs
@@ -484,10 +484,9 @@ impl MysqlDb {
         }
 
         query = match sort {
-            Sorting::Index => query.order(bso::sortindex.desc()),
-            Sorting::Newest => query.order(bso::modified.desc()),
-            Sorting::Oldest => query.order(bso::modified.asc()),
-            _ => query,
+            Sorting::Index => query.order(bso::id.desc()).order(bso::sortindex.desc()),
+            Sorting::Newest => query.order(bso::id.desc()).order(bso::modified.desc()),
+            Sorting::Oldest | Sorting::None => query.order(bso::id.asc()).order(bso::modified.asc()),
         };
 
         let limit = limit.map(i64::from).unwrap_or(-1);
--- a/src/db/mysql/models.rs
+++ b/src/db/mysql/models.rs
@@ -485,7 +485,9 @@ impl MysqlDb {
 
         query = match sort {
             Sorting::Index => query.order(bso::id.desc()).order(bso::sortindex.desc()),
-            Sorting::Newest | Sorting::None => query.order(bso::id.desc()).order(bso::modified.desc()),
+            Sorting::Newest | Sorting::None => {
+                query.order(bso::id.desc()).order(bso::modified.desc())
+            }
             Sorting::Oldest => query.order(bso::id.asc()).order(bso::modified.asc()),
         };
 
--- a/src/db/mysql/models.rs
+++ b/src/db/mysql/models.rs
@@ -485,8 +485,8 @@ impl MysqlDb {
 
         query = match sort {
             Sorting::Index => query.order(bso::id.desc()).order(bso::sortindex.desc()),
-            Sorting::Newest => query.order(bso::id.desc()).order(bso::modified.desc()),
-            Sorting::Oldest | Sorting::None => query.order(bso::id.asc()).order(bso::modified.asc()),
+            Sorting::Newest | Sorting::None => query.order(bso::id.desc()).order(bso::modified.desc()),
+            Sorting::Oldest => query.order(bso::id.asc()).order(bso::modified.asc()),
         };
 
         let limit = limit.map(i64::from).unwrap_or(-1);
--- a/src/db/spanner/models.rs
+++ b/src/db/spanner/models.rs
@@ -976,7 +976,9 @@ impl SpannerDb {
         }
         query = match sort {
             Sorting::Index => format!("{} ORDER BY sortindex DESC, bso_id DESC", query),
-            Sorting::Newest | Sorting::None => format!("{} ORDER BY modified DESC, bso_id DESC", query),
+            Sorting::Newest | Sorting::None => {
+                format!("{} ORDER BY modified DESC, bso_id DESC", query)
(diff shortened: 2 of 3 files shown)


src/db/spanner/models.rs
```diff
@@ -1204,7 +1204,7 @@ impl SpannerDb {
             Sorting::Index => format!("{} ORDER BY sortindex DESC", query),
             Sorting::Newest => format!("{} ORDER BY modified DESC", query),
             Sorting::Oldest => format!("{} ORDER BY modified ASC", query),
```
pjenvey:
```suggestion
            Sorting::Oldest | Sorting::None => format!("{} ORDER BY modified ASC", query),
```

but I'm wondering if we should just kill Sorting::None entirely (it would come in as Newest)

fzzzy: I'd rather have Sorting::None state still be there. This code is confusing enough as it is.

src/db/spanner/models.rs
```diff
@@ -1204,7 +1204,7 @@ impl SpannerDb {
...
-            _ => query,
```
jrconlin: Kinda wondering if this might be better as:
```
query = match sort {
        query = match sort {
            Sorting::Index | Sorting::Newest => format!("{} ORDER BY sortindex DESC", query),
            _ => format!("{} ORDER BY sortindex ASC", query),
};
```

linacambridge:
I can test! Does this get deployed to a dev FxA/Sync stack? Getting it running locally might be a bit much; I don't have access to `mozilla-rust-sdk`, for example (do I need it at all, or is it just for Docker? Do you recommend Docker at all, or just cloning it locally and installing deps via `brew` should be fine?)

fzzzy:
Getting it running locally against spanner might be a bit tough. You need mozilla-rust-sdk (you can get it from phil or me or jr) and you need access to the spanner database that we use for local testing (I think you have to ask phil about this)

I think we'll try to get it running on stage, and you can test it there.

jrconlin:
(shh, but I pushed a `/vendor` branch that has the library in it, since
we're getting more than a few folks asking about it.)

src/db/spanner/models.rs
fzzzy: Hm, Index needs to be ORDER BY sortindex and Newest needs to be ORDER BY modified though, so this won't work.

src/db/spanner/models.rs
```diff
@@ -975,10 +975,9 @@ impl SpannerDb {
             sqltypes.insert("newer".to_string(), as_type(TypeCode::TIMESTAMP));
         }
         query = match sort {
-            Sorting::Index => format!("{} ORDER BY sortindex DESC", query),
-            Sorting::Newest => format!("{} ORDER BY modified DESC", query),
-            Sorting::Oldest => format!("{} ORDER BY modified ASC", query),
-            _ => query,
+            Sorting::Index => format!("{} ORDER BY sortindex DESC, bso_id DESC", query),
+            Sorting::Newest => format!("{} ORDER BY modified DESC, bso_id DESC", query),
+            Sorting::Oldest | Sorting::None => format!("{} ORDER BY modified ASC, bso_id ASC", query),
```
pjenvey: Oops did we all miss this before? None should be equiv. to Newest

pjenvey:
You can `git cherry-pick 63943e08` to fix the rust-chanelog failure (it'll merge back to master fine)

src/db/mysql/models.rs
```diff
@@ -484,10 +484,11 @@ impl MysqlDb {
         }
 
         query = match sort {
-            Sorting::Index => query.order(bso::sortindex.desc()),
-            Sorting::Newest => query.order(bso::modified.desc()),
-            Sorting::Oldest => query.order(bso::modified.asc()),
-            _ => query,
+            Sorting::Index => query.order(bso::id.desc()).order(bso::sortindex.desc()),
```
pjenvey: A tuple also works FYI:

```suggestion
            Sorting::Index => query.order((bso::sortindex.desc(), bso::id.desc())),
```

Issue #424, opened 2020-02-11
Add minumum supported Rust version to cargo.toml


tublitzed:
Looks like it's [now supported](https://github.com/rust-lang/rfcs/pull/2495).

Probably makes sense to do this to avoid future issues with versions that are too old.

pjenvey:
Reopening because I just noticed the RFC hasn't been implemented yet: https://github.com/rust-lang/rust/issues/65262

So #475 emits a harmless warning:

```
$ cargo build
warning: unused manifest key: package.rust
```

Let's revert it and just add the minimum version to the README for now

tublitzed:
Ah shoot, thanks @pjenvey! Closing this now and will re-open if/when implemented.

PR #475, opened 2020-03-10
chore: add minumum supported rust version

tublitzed:
## Description

To help us avoid future issues with folks using the wrong rust version, I'm adding [this field](https://github.com/rust-lang/rfcs/pull/2495) into our cargo.toml.

## Testing

Well...the main thing to review is: did I get the version right? :) I picked 1.39.0 given that I at least know for sure that we're depending on async/await which was introduced then. Anything that rolled into Rust stable afterwards we're requiring?

## Issue(s)

Closes [#424](https://github.com/mozilla-services/syncstorage-rs/issues/424).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ authors = [
   "Phil Jenvey <pjenvey@underboss.org>",
 ]
 edition = "2018"
+rust = "1.39.0"
 
 [profile.release]
 # Enables line numbers in Sentry


jrconlin:
1.39 should be good. [1.40](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1400-2019-12-19) has a bunch of fixes, but I don't believe we're impacted by any of them. 

Local build --all-targets & tests ran without error under 1.39.

PR #479, opened 2020-03-11
chore: tag 0.2.5

pjenvey:


src/db/mysql/models.rs
```diff
@@ -479,10 +479,11 @@ impl MysqlDb {
         }
 
         query = match sort {
-            Sorting::Index => query.order(bso::sortindex.desc()),
-            Sorting::Newest => query.order(bso::modified.desc()),
-            Sorting::Oldest => query.order(bso::modified.asc()),
-            _ => query,
+            Sorting::Index => query.order(bso::id.desc()).order(bso::sortindex.desc()),
```
tublitzed: @pjenvey -  the description here says tagging a release, but this looks like we've got some extra changes along for the ride?

pjenvey:
Updated to the 0.2.5 tag w/ the ttl fix

src/db/mysql/models.rs
pjenvey: It's due to 0.2.4/5 being actual dot releases. their 2 fixes landed on the release branch first to avoid including the latest async changes on master. we'll test/deploy that separately in a 0.3

PR #486, opened 2020-03-12
fix: Convert erect_tombstone to async/await

fzzzy:
## Description

Convert erect_tombstone to async/await, nd in the process, remove all the spanner synchronous support code. The spanner backend is now truly 100% sync code free.

## Testing

All tests should pass.

## Issue(s)

Closes #463.

### DIFF
--- a/src/db/mysql/models.rs
+++ b/src/db/mysql/models.rs
@@ -379,7 +379,7 @@ impl MysqlDb {
         let timestamp = self.timestamp().as_i64();
 
         self.conn.transaction(|| {
-            let payload = bso.payload.as_ref().map(Deref::deref).unwrap_or_default();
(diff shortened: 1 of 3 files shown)


src/db/mysql/models.rs
```diff
@@ -379,7 +379,7 @@ impl MysqlDb {
         let timestamp = self.timestamp().as_i64();
 
         self.conn.transaction(|| {
-            let payload = bso.payload.as_ref().map(Deref::deref).unwrap_or_default();
+            let payload = bso.payload.as_deref().unwrap_or_default();
```
pjenvey: Note: this bumps us to minimum rust of 1.40 (which is fine)

PR #487, opened 2020-03-15
refactor: rewrite purge_ttl in rust

### LINKED ISSUE #385: Rewrite purge_ttl in Rust

mirefly:
Fixes #385.
I tried to follow steps mentioned by @pjenvey in https://github.com/mozilla-services/syncstorage-rs/issues/385#issuecomment-596830897


tools/spanner/purge_ttl/src/main.rs
```diff
@@ -0,0 +1,49 @@
...
+    meta.add_str("google-cloud-resource-prefix", &database)?;
+    meta.add_str("x-goog-api-client", "googleapis-rs")?;
+    let opt = CallOption::default().headers(meta.build());
+    let session = client.create_session_opt(&req, opt)?;
+
+    let mut req = ExecuteSqlRequest::new();
+    req.session = session.get_name().to_string();
+
+    req.sql = "DELETE FROM batches WHERE expiry < CURRENT_TIMESTAMP()".to_string();
+    client.execute_sql(&req)?;
```
pjenvey: One thing we'll need to to execute this as a [Spanner "partitioned dml"](https://cloud.google.com/spanner/docs/dml-partitioned) statement.

The python client lib provides an easy to use `execute_partitioned_dml` function whereas our rust client lib doesn't (as it's a lower level library).

We can mimic what the python client does:

https://github.com/googleapis/python-spanner/blob/master/google/cloud/spanner_v1/database.py#L395

Basically: 

- Begin a transaction with the `partitioned_dml` TransactionOption
- Call execute_sql within that transaction

We have examples of creating transactions in the spanner/models.rs and the utilizing them w/ execute_sql calls in spanner/support.rs

mirefly: Great. Thanks for pointing this out.

mirefly:
@pjenvey I added "execute_partitioned_dml" and metrics. This patch is ready for another review.
I do not know how to do some tests for this script locally (do we have ways to test it?), so I am not sure if this script works well as expected, even though the compiler is happy.

tools/spanner/purge_ttl/src/main.rs
```diff
@@ -0,0 +1,142 @@
...
+    let mut meta = MetadataBuilder::new();
+    meta.add_str("google-cloud-resource-prefix", &database)?;
+    meta.add_str("x-goog-api-client", "googleapis-rs")?;
+    let opt = CallOption::default().headers(meta.build());
+    let session = client.create_session_opt(&req, opt)?;
+
+    let statsd = statsd_from_env()?;
+
+    {
+        let _timer_total = start_timer(&statsd, "syncstorage.purge_ttl.total_duration");
```
pjenvey: For all the start_timers:

```suggestion
        let _timer_total = start_timer(&statsd, "purge_ttl.total_duration");
```

because cadence prepends the prefix specified to StatsdClient::builder

pjenvey:
Also a very minor nit about the commit message format ([same note I made here](https://github.com/mozilla-services/syncstorage-rs/pull/497#issuecomment-600880133))

mirefly:
Thanks @pjenvey . I addressed the comments. I tried to have all logging messages in `purge_ttl.py`, but not exactly same format, hoping that is fine.

I noticed that, in python code, [`execute_partitioned_dml` uses `execute_streaming_sql`](https://github.com/googleapis/python-spanner/blob/e378ba039fc5714c715386b3a1d9ba5b2c7ce306/google/cloud/spanner_v1/database.py#L450). Do we need that? I used `execute_sql` just for simplicity now.

pjenvey:
As for testing: an automated test for this will be difficult, we'll skip it for now.

We'll manually test this against our Spanner dev db (I forgot to try this today, we'll try it out tomorrow).

tools/spanner/purge_ttl/src/main.rs
```diff
@@ -0,0 +1,147 @@
...
+        label: label.to_owned(),
+        client: client.clone(),
+    }
+}
+
+pub fn statsd_from_env() -> Result<StatsdClient, Box<dyn Error>> {
+    let statsd_host = env::var("STATSD_HOST")
+        .unwrap_or("localhost".to_string())
+        .parse::<IpAddr>()
+        .unwrap();
```
pjenvey: let's kill this `unwrap()` for ?, and same for the one a couple lines down

tools/spanner/purge_ttl/src/main.rs
```diff
@@ -0,0 +1,147 @@
...
+    MetricTimer {
+        start: Instant::now(),
+        label: label.to_owned(),
+        client: client.clone(),
+    }
+}
+
+pub fn statsd_from_env() -> Result<StatsdClient, Box<dyn Error>> {
+    let statsd_host = env::var("STATSD_HOST")
+        .unwrap_or("localhost".to_string())
```
pjenvey: IpAddr won't accept a hostname

```suggestion
        .unwrap_or("127.0.0.1"to_string())
```

tools/spanner/purge_ttl/src/main.rs
```diff
@@ -0,0 +1,147 @@
...
+    let statsd = statsd_from_env()?;
+
+    {
+        let _timer_total = start_timer(&statsd, "purge_ttl.total_duration");
+        {
+            let _timer_batches = start_timer(&statsd, "purge_ttl.batches_duration");
+            let mut req = prepare_request(&client, &session)?;
+            req.set_sql("DELETE FROM batches WHERE expiry < CURRENT_TIMESTAMP()".to_string());
+            let result = client.execute_sql(&req)?;
+            info!("batches: removed {} rows", result.get_stats().get_row_count_exact())
```
pjenvey: Paritioned dml calls should use the lower count bound (as in the Python client code)

```suggestion
            info!("batches: removed {} rows", result.get_stats().get_row_count_lower_bound())
```

mirefly:
I addressed the comments. @pjenvey If you want me to squash two commits into one, please let me know.

tools/spanner/purge_ttl/src/main.rs
```diff
@@ -0,0 +1,155 @@
...
+            let _timer_bso = start_timer(&statsd, "purge_ttl.bso_duration");
+            let mut req = prepare_request(&client, &session)?;
+            req.set_sql("DELETE FROM bsos WHERE expiry < CURRENT_TIMESTAMP()".to_string());
+            let result = client.execute_sql(&req)?;
+            info!(
+                "bso: removed {} rows",
+                result.get_stats().get_row_count_lower_bound()
+            )
+        }
+        info!("Completed purge_ttl.py")
```
pjenvey:
```suggestion
        info!("Completed purge_ttl")
```

pjenvey:
Also I'll note: `env_logger` by default only logs `error!` level so the script is still silent by default. That's ok though, env_logger let's this be easily toggled via a `RUST_LOG=info` env var.

mirefly:
Thanks for comments. I fixed clippy warnings and added another `cargo install` in Dockerfile as suggested, hoping it works well.

PR #488, opened 2020-03-16
Build images when PRs are merged to master

erkolson:
## Description

Presumably, master only changes with a PR merge.  This circleci config change causes circle to build an image for every commit on master, it makes the docker tag equal to the git commit sha1.  The downstream automation will then deploy to the development environment when the docker tag looks like a commit sha1 (matches the regex `/[0-9a-f]{40}/`).

I think the `if echo "${CIRCLE_BRANCH}" | grep '^feature\..*' > /dev/null; the` block is not used and can be deleted.

When a release is tagged, the following block will cause the docker tag to match the git tag.  The downstream automation will deploy tags that don't match the sha1 type regex to stage:
```
            if [ -n "${CIRCLE_TAG}" ]; then
              DOCKER_TAG="$CIRCLE_TAG"
            fi
```

## Testing

Merge to master and see if our deployment is triggered.

## Issue(s)

Related to [Create sync dev environment](https://bugzilla.mozilla.org/show_bug.cgi?id=1622286).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -189,7 +189,7 @@ jobs:
           name: Deploy to Dockerhub
           command: |
             if [ "${CIRCLE_BRANCH}" == "master" ]; then
-              DOCKER_TAG="latest"
+              DOCKER_TAG="${CIRCLE_SHA1}"
             fi
 
             if echo "${CIRCLE_BRANCH}" | grep '^feature\..*' > /dev/null; then


pjenvey:
Could downstream be configured to pick up the `latest` tag as is? I know the autopush dev envs does this.

My concern is this littering docker hub with many images

erkolson:
Dockerhub doesn't actually clean up previously uploaded images so uploading a new image with the latest tag just causes the latest tag to move to the new image (which is identified internally by a sha256 digest), we still keep the same number of images in the registry.  It does clutter the `tags` page in dockerhub though.  Is that the actual concern, or an equally important concern?

tublitzed:
@pjenvey are you the [rustsyncstorage](https://hub.docker.com/u/rustsyncstorage) user? If so, I wonder if we just periodically go in and prune the tags over there (yearly/quarterly) for now, and then revisit if it becomes unmanageable.

pjenvey:
> @pjenvey are you the [rustsyncstorage](https://hub.docker.com/u/rustsyncstorage) user? If so, I wonder if we just periodically go in and prune the tags over there (yearly/quarterly) for now, and then revisit if it becomes unmanageable.

Nope, ops setup/owns this dockerhub user

erkolson:
I couldn't find any documentation about images being deleted.  I did find a lot of gists and blogs about people writing scripts to accomplish it though.

Issue #407, opened 2020-01-09
User Migration Meta bug


jrconlin:
A Meta Bug to cross track all the User Migration issues and PRs.

https://github.com/mozilla-services/tokenserver/issues/159
https://github.com/mozilla-services/syncstorage-rs/pull/401
https://github.com/mozilla-services/server-syncstorage/issues/125
https://github.com/mozilla-services/services-engineering/issues/31
https://github.com/mozilla-services/services-engineering/issues/6

jrconlin:
Additional info from [sync meetup](https://docs.google.com/document/d/1TKLpafQweKbVic4Na3pZl3Mzfgl4WItj32xaWVqPoFg/edit).
- [ ]   Need to [add a second table to syncstorage](https://github.com/mozilla-services/server-syncstorage/pull/136) containing the fxa_uids of migrating users. If id is present, kick back to tokenserver to eventually route to spanner
- [ ]  alter the `migration` table to indicate state of the user migration 
- [ ]  check that user node hasn't changed since start and finish of migration. 
- [ ]  ensure failed user migration ids are logged
- [ ] verify that there are no additional issues around mobile clients.
- [ ] generate fake m/g record for migrated user (before "done") that has new SyncIDs.

tublitzed:
Related [bugzilla link](https://bugzilla.mozilla.org/show_bug.cgi?id=1639645) for migrating the first node. Going to close this out as the migration script portion of this work is done. Will open + link any further refinements needed.

PR #491, opened 2020-03-16
feat: Add `--abort` and `--user_range` flags

jrconlin:
## Description

* --abort stops copying BSO records after N instances.
* --user_range limits copy to offset:limit users.
* optionally sorts users by fxa_uid via `--sort_users` flag
* collapses previous pushes and syncs with master branch

## Testing

* configure local environment using the steps provided in the `tools/README.md` file
* select a contiguous range of users specified in the given node starting from a given offset (e.g. if a node contains 100 users, you wish to copy 20 users, and you wish to skip the first 10 users, use a `user_range=10:20`. Verify users being copied fit the specified range.
* for a given category, select a subset of records you wish to copy over. Other records are dropped. (e.g. if a user has 100 history items and you wish to test that only 10 are copied over, use `abort=history:10`). App reports rows added or skipped.
* Can use `--dryrun` which will not write data to spanner, but act as if it did. 

## Issue(s)

Issue #407


tools/user_migration/migrate_node.py
```diff
@@ -451,34 +459,52 @@ def spanner_transact_bso(transaction, data, fxa_kid, fxa_uid, args):
...
+                try:
+                    (fxa_kid, fxa_uid) = fxa.get(user)
+                    users.append((user, fxa_kid, fxa_uid))
+                except TypeError:
+                    logging.error("⚠️User not found in tokenserver data: {} ".format(
+                        user
+                    ))
+            users.sort(key=lambda tup: tup[2])
+        except Exception as ex:
+            import pdb; pdb.set_trace()
```
tublitzed: ✂️ ?

jrconlin: well, it's in an exception, and the script is meant to be hand-run. I'd rather it dump to an examinable state at that point.

tublitzed:
Would it be possible to isolate the rust migration script from this PR so we can get this version merged and review that separately?

tublitzed:
@jrconlin can you update the testing instructions here? What specifically are we supposed to run to test this?

jrconlin:
@tublitzed comment updated.

PR #493, opened 2020-03-17
feat: Add KV trait for Tags

### LINKED ISSUE #387: Implement KV trait for Tags
KV's a trait that the `Tags` struct can implement. [slog-derive](https://github.com/slog-rs/derive) makes it easy.

Probably better to tackle that in a separate issue.

_Originally posted by @pjenvey in https://github.com/mozilla-services/syncstorage-rs/pull/372_

undef1nd:
## Description

Add `KV` trait for `Tags` struct to simplify logging its content.

## Testing

For example, in `extractors.rs` instead of 
```rust
warn!("⚠️ Missing BSO: {:?}", uri.path();
            "ua.os.family" => tags.get("ua.os.family"),
            "ua.browser.family" => tags.get("ua.browser.family"),
            "ua.name" => tags.get("ua.name"),
            "ua.os.ver" => tags.get("ua.os.ver"),
            "ua.browser.ver" => tags.get("ua.browser.ver"));
```
this snippet can be used:
```rust
warn!("⚠️ Unexpected BSO URI: {:?}", uri.path(); tags);
```

## Issue(s)

Closes #387 .

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,7 +52,7 @@ serde_json = { version = "1.0", features = ["arbitrary_precision"] }
 serde_urlencoded = "0.6.1"
 scheduled-thread-pool = "0.2"
 sha2 = "0.8.0"
-slog = { version = "2.5", features = ["max_level_trace", "release_max_level_error"] }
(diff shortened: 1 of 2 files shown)


undef1nd:
@jrconlin will also research how to properly test it.

PR #494, opened 2020-03-18
fix: Enforce stricter offset parsing

### LINKED ISSUE #440: Enforce stricter offset parsing
A followup to #368: Python syncstorage has stricter handling of offset values, raising an InvalidOffsetError when:

- a special "offset token" (`timestamp_bound:offset` format) is specified for `sort=index`
- a regular offset (just a plain numeric `offset` value) for other sort types

The change to db-tests' `get_bsos_limit_offset` in #368 somewhat came across case #1: the fact that passing a timestamp bound to `sort=index` is bogus.

Allowing case #2 made sense for the sake of deploying #368 (so a client w/ a plain offset retrieved from just before its deployment would continue working after the deployment). Now that it's deployed we could begin rejecting it also.

fzzzy:
## Description

Return errors if:

- Sort is index, but a timestamp:offset pair is passed as offset
- Sort is not index, but an integer is passed as offset

## Testing

All tests should pass.

## Issue(s)

Fix #440


src/web/extractors.rs
```diff
@@ -1353,7 +1353,25 @@ impl FromRequest for BsoQueryParams {
                             .into());
                         }
                     }
+                } else if let Some(offset) = params.offset.as_ref() {
+                    if offset.offset == 0 {
```
pjenvey: I don't understand this check, is it even necessary? I believe we're already sort != Index + no timestamp here. I think this could use maybe a couple new tests too.

Also I'll throw this out there: I suspect this whole large block might benefit from a refactoring now. It's destructuring params.offset w/ `if let Some(..)` + `and_then` multiple times. Maybe wrapping the whole thing in `if let Some(offset) = params.offset` and then destructuring everything up front (`let Offset { timestamp, offset } = offset;`) would be clearer (I'm having trouble following it myself).

fzzzy: The tests fail without it. It's because a request with no offset (the first request) will have neither a timestamp nor an offset, so offset.offset will be 0 here.

fzzzy: I refactored this whole block and it's much easier to understand, I think.

pjenvey: Thanks for that, much easier.

No offset specified should equate to a `None` for params.offset AFAICT (avoiding this whole block entirely). I can only see offset.offset becoming 0 when it was passed in as such from the client.

src/web/extractors.rs
```diff
@@ -1353,6 +1361,20 @@ impl FromRequest for BsoQueryParams {
                             .into());
                         }
                     }
+                } else {
```
pjenvey: Following my comment below -- so instead of the `offset.offset == 0` check I think this is what we want:

```suggestion
                } else if params.sort != Sorting::Index {
```

no timestamp + non-index. I still think a couple extra tests would be helpful for this.

But argh, this leads to another issue I hadn't considered: mysql doesn't implement an `encode_offset` equiv. yet. We either can't enforce this strictness until that's implemented or it would have to be Spanner only.

I think we forgot to log an issue for bringing this over to mysql, might be a _Good First Issue_.

fzzzy: Ok, great. This is much better. Thanks.

Well, should we wait to merge this pr until mysql has the encode_offset code then?

pjenvey: Do you think it'd be a good first issue? We could just implement it now too, I suspect it'd be pretty straightforward w Spanner already done (would probably use the same `encode_offset` so maybe just a matter of plugging it in)

fzzzy: I'll just do it in this pr and get it over with

fzzzy:
Closing this for now, but leaving the branch alive to be picked up again later.

PR #495, opened 2020-03-18
fix: do not populate mysql CollectionCache with invalid values

### LINKED ISSUE #239: MySql CollectionCache could populate w/ invalid values
Just like #224, but for the MySql implementation.

mirefly:
Fixes #239


src/db/mysql/models.rs
```diff
@@ -317,7 +317,9 @@ impl MysqlDb {
             .execute(&self.conn)?;
             collections::table.select(last_insert_id).first(&self.conn)
         })?;
-        self.coll_cache.put(id, name.to_owned())?;
+        if !self.session.borrow().in_transaction {
```
pjenvey: Let's get rid of this cache put entirely, since this is a write operation (`create_collection`)

src/db/mysql/models.rs
```diff
@@ -343,7 +345,9 @@ impl MysqlDb {
         .optional()?
         .ok_or(DbErrorKind::CollectionNotFound)?
         .id;
-        self.coll_cache.put(id, name.to_owned())?;
+        if !self.session.borrow().in_transaction {
```
pjenvey: Instead of guarding this with `in_transaction`, we need the equivalent of the Spanner `models.rs` where we denote whether or not we're in a write transaction.

We'll know we're writing if a `lock_for_write_sync` was called, so it should toggle a new `in_write_transaction` flag that we can check instead.

mirefly:
Thanks @pjenvey. This patch is ready for another look. I added the `in_write_transaction` flag, but did not add [`fn in_write_transaction` ](https://github.com/mozilla-services/syncstorage-rs/blob/532aa9d33ee6c4a15ab62e8122ced79d2ee0f212/src/db/spanner/models.rs#L440)as in spanner, since we only have two places using `self.session.borrow().in_write_transaction`.

fzzzy:
@mirefly Ok this is ready to merge, but we can't merge it because your commit is not signed. Can you please set up github commit signing, squash this branch into a new signed commit, and force push this branch to that new commit hash? Thanks!

mirefly:
@fzzzy Sorry, I forgot signing it. I updated it, hoping it is fine.

PR #500, opened 2020-03-19
Remove custom async_test proc_macro implementation

### LINKED ISSUE #461: Remove our custom async_test proc_macro (codegen crate) implementation
Our `#[async_test]` proc_macro is based on [tokio-async-await-test](https://github.com/LucioFranco/tokio-async-await-test) but we implemented our own custom version to avoid build issues with the crate and its tokio deps that existed at the time.

That crate hasn't been updated in a while but there's likely alternatives (e.g. [futures-await-test](https://crates.io/crates/futures-await-test)) that we could switch to.

Our impl isn't a lot of code, so this is a low priority issue, but its existence requires the `codegen` sub crate which would be nice to get rid of.

amrita72:
Closes #461


jrconlin:
nit: Please rename this PR to fit with our [Contributing guidelines](https://github.com/mozilla-services/syncstorage-rs/blob/master/CONTRIBUTING.md). This will ensure that the feature is included in our Change log. 

Please note, this will require the commit to be amended, not just this github PR edited.

Issue #470, opened 2020-03-06
[mysql] Make the userid column a BIGINT


pjenvey:
The mysql schema defines the `userid` column in all tables as a regular 32 bit `INT`. This is incorrect: the userid is handed to syncstorage-rs via an auth token by the tokenserver, which [uses a 64 bit `BIGINT`](https://github.com/mozilla-services/tokenserver/blob/e25d636/tokenserver/assignment/sqlnode/schemas.py).

syncstorage-rs correctly parses the auth token into the `HawkIdentifier` struct, w/ userid (HawkIdentifier refers to it as `legacy_id`) [being 64 bits](https://github.com/mozilla-services/syncstorage-rs/blob/master/src/web/extractors.rs#L1060).

The [mysql db code](https://github.com/mozilla-services/syncstorage-rs/blob/f699085/src/db/mysql/models.rs) then casts `legacy_id` down to the a u32 (which truncates!) per the column type.

We need to:

- Create a new [diesel migration](https://github.com/mozilla-services/syncstorage-rs/tree/46c650a/migrations) to convert the column to a `BIGINT`
- Modify the [diesel schema](https://github.com/mozilla-services/syncstorage-rs/blob/317d960/src/db/mysql/schema.rs) accordingly
- Remove the truncating casts (though we still need to cast from `legacy_id`'s u64 type to `i64` which diesel expects for `BIGINT`s)

(This bug extracted from the original #74)

Emmanuel-Melon:
Hi @pjenvey, this sounds like a good first issue. I'd love to give it a shot. I'll take a look at the codebase and then give you an update.

undef1nd:
Hey @Emmanuel-Melon, are you working on this one? I'd like to give it a try 🙂

Emmanuel-Melon:
Yes, I'm. I'm making a PR this week.

PR #530, opened 2020-03-20
fix remove truncating tasks in mysql/db code

Emmanuel-Melon:
## Description

Casting `user_id's` from `u32` types into `i64` types as required by token server.
This fix is done by:
- Creating a new migration to modify the diesel schema.
- Removing truncating casts in mysql code.

## Testing

Work-In-Progress

## Issue(s)

Closes [470](https://github.com/mozilla-services/syncstorage-rs/issues/470).


Emmanuel-Melon:
@pjenvey @jrconlin I've just resumed working on this project and this is what I've got so far.
I've created a new diesel.toml file and specified file as `src/db/mysql/schema.rs` and ran my migration which modifies `user_id` fields but it also modified other fields `collection_id` for example and it's causing other tests to fail. I'm currently looking at this issue.

jrconlin:
Thanks for working on this.

Emmanuel-Melon:
> Thanks for working on this.

Thank you. I've ran the migrations and casted most of user_id instances but the tests keep on failing due to an issue with mapping Rust identifiers to MySQL identifiers. 
In the previous schema.rs file, I've noticed the presence of such attributes `#[sql_name="userid"]` right above conflicting identifiers so I went and checked this [PR](https://github.com/diesel-rs/diesel/pull/1252/files) in the Diesel repo and it states that these identifiers are automatically generated by diesel.
I'll make a few modifications then get back to you if I hit a brick wall.

PR #535, opened 2020-03-22
refactor: Refactor Tags logging by utilising KV trait

### LINKED ISSUE #503: Utilize Tags' slog::KV trait
A Followup to #387:

Now that Tags implements the KV trait, let's convert all log statements to utilize it where appropriate. See the `Testing` section described in #493 for an example.

Note: let's be careful to ensure this change doesn't add excessive number of new tags to our log messages. I do believe this change will end up adding the `uri.method` tag to log messages, which is fine.

undef1nd:
## Description

Refactor how `Tags` content is logged. Due to KV trait being implemented for this struct, it can be passed to `slog` as is without need to log each of its key/value individually. See example in #493 `Testing` section.

## Testing

Nothing should change in log messages, except `uri.method` tag being added.

## Issue(s)

Closes #503 .

### DIFF
--- a/src/server/metrics.rs
+++ b/src/server/metrics.rs
@@ -31,12 +31,7 @@ impl Drop for Metrics {
         if let Some(client) = self.client.as_ref() {
             if let Some(timer) = self.timer.as_ref() {
                 let lapse = (Instant::now() - timer.start).as_millis() as u64;
-                trace!("⌚ Ending timer at nanos: {:?} : {:?}", &timer.label, lapse;
-                "ua.os.family" => tags.get("ua.os.family"),
-                "ua.browser.family" => tags.get("ua.browser.family"),
-                "ua.name" => tags.get("ua.name"),
-                "ua.os.ver" => tags.get("ua.os.ver"),
-                "ua.browser.ver" => tags.get("ua.browser.ver"));
+                trace!("⌚ Ending timer at nanos: {:?} : {:?}", &timer.label, lapse; &tags);
                 let mut tagged = client.time_with_tags(&timer.label, lapse);
                 // Include any "hard coded" tags.
                 // tagged = tagged.with_tag("version", env!("CARGO_PKG_VERSION"));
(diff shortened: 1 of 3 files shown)


undef1nd:
@pjenvey Could you plz take a look if I'm not missing anything?

 I looked through the code and it seems to me there won't be any tags added except `uri.method` after this change. 
But I would like to confirm that with you prior to marking PR as ready for review 🙂  Thanks.

undef1nd:
@pjenvey 
I've changed the docstring in `test_error`. But maybe it's worth to leave original example of logging KV pairs and add one more example of logging `Tags`?

Reg the `metrics.rs`: `mtags` is a Hashmap there. But I can wrap it into Tags in order to use KV trait. I guess, like this:
```rust
let mtags = Tags::default().extend(mtags);
warn!("⚠️ Metric {} error: {:?} ", label, e; mtags);
```
Should I do it?

jrconlin:
Part of the problem here is a lifetime issue around the `cadence::Counter` instance, and the tag value being written, but there's some cleanup that could be done:

```
    pub fn incr_with_tags(self, label: &str, tags: Option<Tags>) {
        if let Some(client) = self.client.as_ref() {
            let mut tagged = client.incr_with_tags(label);
            let mut mtags = self.tags.clone().unwrap_or_default();
            if let Some(tags) = tags {
                mtags.extend(tags.tags);
            }
            for key in mtags.tags.keys().clone() {
                if let Some(val) = mtags.tags.get(key) {
                    tagged = tagged.with_tag(&key, val.as_ref());
                }
            }
            // Include any "hard coded" tags.
            // incr = incr.with_tag("version", env!("CARGO_PKG_VERSION"));
            match tagged.try_send() {
                Err(e) => {
                    // eat the metric, but log the error
                    warn!("⚠️ Metric {} error: {:?} ", label, e; mtags
                    );
                }
                Ok(v) => trace!("☑️ {:?}", v.as_metric_str()),
            }
        }
    }
```

undef1nd:
@jrconlin thanks for the help!

PR #538, opened 2020-03-23
chore: Dependency update

### LINKED ISSUE #537: chore: Version update
There are a large number of updates from depend-bot. 

Merge them into one and update dependencies.

jrconlin:
## Description

* mass close depend-bot PRs 
* includes vendored googleapis-raw

## Testing

chore

## Issue(s)

Closes #537.


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "vendor/mozilla-rust-sdk/googleapis-raw/grpc"]
+	path = vendor/mozilla-rust-sdk/googleapis-raw/grpc
+	url = https://github.com/grpc/grpc.git
```
tublitzed: it [looks like](https://stackoverflow.com/questions/49490272/how-to-specify-a-tag-in-gitmodules-file) you may be able to specify a tag here. Would it be possible to do this here so we know which specific version of the grpc lib we're depending on?

jrconlin: Thanks, I'll see if I can dig into that. Part of the problem is also that dependbot will flag "old" versions as well. 

I honestly wonder if part of the problem is also home bound folks "catching up on bugs" and releasing a LOT of new revisions of libraries causing all sorts of fun cascades.

jrconlin: no longer an issue with the update of the vendored code.

.circleci/config.yml
```diff
@@ -28,8 +28,9 @@ commands:
       - run:
           name: Core Rust Checks
           command: |
-            cargo audit
             cargo fmt -- --check
+            echo "Skip audit for now since hawk 3.1.0 -> ring 0.16.11 -> abandoned `spin`"
```
pjenvey: The spin +1 other audit on this branch are just warnings, so we can keep audit on

Cargo.toml
```diff
@@ -17,54 +18,56 @@ actix-http = "1"
...
 serde = "1.0"
 serde_derive = "1.0"
 serde_json = { version = "1.0", features = ["arbitrary_precision"] }
 serde_urlencoded = "0.6.1"
 scheduled-thread-pool = "0.2"
-sha2 = "0.8.0"
-slog = { version = "2.5", features = ["max_level_info", "release_max_level_info", "dynamic-keys"] }
-slog-async = "2.3"
+sha2 = "0.8"
+slog = { version = "2.5", features = ["max_level_trace", "release_max_level_error", "dynamic-keys"] }
```
pjenvey: Bad merge w/ #502

.circleci/config.yml
jrconlin: Ok, was getting a fatal off of that, but will retry.

pjenvey: I recall a fatal on this branch at one point but I think it was something else now gone

PR #549, opened 2020-03-25
fix: allow hostnames for STATSD_HOST

### LINKED ISSUE #548: Rust purge_ttl should accept hostnames for STATSD_HOST

pjenvey:
## Description

and improve the missing db env error message

also don't package purge_ttl's target/ into the docker

(merging into the 0.2b release branch)

## Testing

run w/ e.g. STATSD_HOST=localhost

## Issue(s)

Closes #548

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,9 @@ RUN \
     cargo --version && \
     rustc --version && \
     cargo install --path . --locked --root /app && \
-    cargo install --path tools/spanner/purge_ttl --locked --root /app
+    cargo install --path tools/spanner/purge_ttl --locked --root /app && \
+    cd tools/spanner/purge_ttl && \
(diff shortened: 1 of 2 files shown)


pjenvey:
> I'd prefer if we just use Url to parse this, and then use the host + path rather than slicing.

This is copied from [spanner/manager.rs](https://github.com/mozilla-services/syncstorage-rs/blob/master/src/db/spanner/manager.rs#L36), originally I opted to slice just in case there's more than a host + path being specified as the db (query, fragment?) -- just pass the whole thing over, w/ the starts_with above guarding from any sketchy slicing.

(This is even after using Url parse in [pool_from_settings](https://github.com/mozilla-services/syncstorage-rs/blob/master/src/db/mod.rs#L254))

PR #558, opened 2020-03-31
refactor: use u64 instead of i64 for Offset.offset

### LINKED ISSUE #414: Use u64 for offset instead of i64
As mentioned in https://github.com/mozilla-services/syncstorage-rs/pull/368#discussion_r355782830 it would be nice if offset was u64 instead of i64. The code does use -1 as a sentinel though, so it would involve switching many types as well as probably using an Option instead of -1 as a sentinel where needed.

amrita72:
## Description

Change struct Offset to use u64 offset instead of i64.

## Testing

?

## Issue(s)

Closes #414

### DIFF
--- a/src/db/mysql/models.rs
+++ b/src/db/mysql/models.rs
@@ -503,7 +503,7 @@ impl MysqlDb {
         // match the query conditions
         query = query.limit(if limit >= 0 { limit + 1 } else { limit });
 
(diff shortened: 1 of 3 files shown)


amrita72:
Thanks @fzzzy! I'll rebase and push ASAP!

Issue #570, opened 2020-04-06
Debug purge_ttl script with Eric


fzzzy:
We have an open bugzilla bug (bug number?) for this but I am making this bug so I can track this for myself.

fzzzy:
We should also make sure that the rust script handles retry on ABORTED properly.

PR #575, opened 2020-04-07
refactor: Remove python dependency from the dockerfile.

### LINKED ISSUE #567: Remove python dependency
Now that [purge_ttl is in Rust](https://github.com/mozilla-services/syncstorage-rs/issues/385) we should be able to kill the python dep for this codebase.

fzzzy:
## Description

Now that purge_ttl has been ported to rust, we may remove the python dependency from the dockerfile.

## Testing

Tests should pass.

## Issue(s)

Fixes #567

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,11 +23,7 @@ RUN \
     groupadd --gid 10001 app && \
     useradd --uid 10001 --gid 10001 --home /app --create-home app && \
     apt-get -q update && \
-    apt-get -q install -y \
-    build-essential \
-    default-libmysqlclient-dev libssl-dev ca-certificates libcurl4 python3-venv python3-pip && \
-    python3 -m pip install setuptools wheel && \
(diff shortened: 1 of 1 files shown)


pjenvey:
We can't remove the Python version until the Rust version is at least on par with the Python version it lacking retrys on ABORTED should be addressed first: https://bugzilla.mozilla.org/show_bug.cgi?id=1626707

fzzzy:
@pjenvey We have both that bugzilla bug and this github bug to track improving the rust script until it works, is that ok? I plan on trying to get it done in the next few days. I can work on retry on ABORTED there if that's ok with you.

https://github.com/mozilla-services/syncstorage-rs/issues/570

fzzzy:
Ok, I'm going to work on retry on ABORTED in this pr.

fzzzy:
Well, we're already using the rust version now, so we'll work on ABORTED in the linked issue.

PR #587, opened 2020-04-10
Latest ops requests for migrate_node

jrconlin:
## Description
- Add `--hoard_limit` to filter out users with exceptionally high amounts of sync data
- Add `reason` for `failure_*.csv` 
- bug fixes

## Testing

run `migrate_node.py --hoard_limit=2 --dryrun` for any user with more than 2 BSO records and check  that `failure_{pid}.csv` contains a record that says 
`{userid}\thoarder #` where `#` matches the total number of BSO records for that user.

## Issue(s)

Issue #401


tools/user_migration/migrate_node.py
```diff
@@ -535,45 +564,39 @@ def get_users(args, databases, fxa, bso_num, report):
...
                         "User {} not found in "
                         "tokenserver data.".format(user))
-                    report.fail(user)
+                    report.fail(user, "not found")
         else:
             try:
-                sql = ("""select distinct userid from bso{}"""
-                       """ order by userid""".format(bso_num))
+                sql = ("""select userid, count(*) as count from bso{}"""
+                       """ group by userid order by userid""".format(bso_num))
```
pjenvey, line 571: nit: I don't think we need the `order by` we've had here?

PR #593, opened 2020-04-15
Fix purge ttl timeout

### LINKED ISSUE #570: Debug purge_ttl script with Eric

fzzzy:
## Description

This code attempts to fix the purge_ttl timeout by grabbing all the ids to delete and chunking them up into batches of 1000.

## Testing

For testing this, build it with:

`cargo install --path . --bin purge_ttl --locked --root output`

Then run it with:

`RUST_LOG=info GOOGLE_APPLICATION_CREDENTIALS=... SYNC_DATABASE_URL=... output/bin/purge_ttl`

You will need an appropriate credentials file and database url to a spanner database. First of all, it should be able to run and not have any errors deleting 0 items.

Then, if some records are added to the bsos and batches tables, with expired expires timestamps, running it should print info about how many records were deleted. Examining spanner should show that they were actually deleted.

To add records to the tables, I modified the load tester to have a 100% chance of doing a PUT, then ran it for 1000 iterations. I also modified the load tester to pass a ttl of 1, meaning the records expire in 1 second. Here is the patch against the load tester:

```
diff --git a/loadtest.py b/loadtest.py
index 867b579..dbbd143 100644
--- a/loadtest.py
+++ b/loadtest.py
@@ -20,7 +20,7 @@ _WEIGHTS = {'metaglobal': [40, 60, 0, 0, 0],
             'post_count_distribution': [67, 18, 9, 4, 2],
             'delete_count_distribution': [99, 1, 0, 0, 0]}
 
-_PROBS = {'get': .1, 'post': .2, 'deleteall': 0.01}
+_PROBS = {'get': 0, 'post': 1, 'deleteall': 0}
 _COLLS = ['bookmarks', 'forms', 'passwords', 'history', 'prefs']
 _BATCH_MAX_COUNT = 100
 
@@ -103,7 +103,7 @@ async def test(session):
     if should_do('post'):
         cid = str(get_num_requests('distribution'))
         url = "/storage/clients"
-        wbo = {'id': 'client' + cid, 'payload': cid * 300}
+        wbo = {'id': 'client' + cid, 'payload': cid * 300, 'ttl': 1}
         data = json.dumps([wbo])
         resp, result = await storage.post(url, data=data, statuses=(200,))
         assert len(result["success"]) == 1, "No success records"
@@ -153,7 +153,7 @@ async def test(session):
             token = storage.auth_token.decode('utf8')
             payload_chunks = int((payload_length / len(token)) + 1)
             payload = (token * payload_chunks)[:payload_length]
-            wbo = {'id': id, 'payload': payload}
+            wbo = {'id': id, 'payload': payload, 'ttl': 1}
             data.append(wbo)
 
         data = json.dumps(data)

```

By default, the purge_ttl script deletes everything. If instead you want to use the incremental delete, necessary in cases where the normal delete takes too long and is killed, pass `SYNC_INCREMENTAL=1` to purge_ttl. It will report every 1000 records on progress, and delete at most `1000000` rows. ⏹ Run it again to delete more.

It would be useful to test more than 1000 records, since some of the logic splits on that, and also more than 1000000 records. In incremental mode, the script will only delete one million 💯 0️⃣᠈ 0️⃣ 0️⃣ 0️⃣ rows, so if there are more than that before you run the script, there should be some left.

## Issue(s)

Closes #570.

### DIFF
--- a/src/bin/purge_ttl.rs
+++ b/src/bin/purge_ttl.rs
@@ -167,7 +167,7 @@ fn delete_incremental(
     chunk_size: u64,
     max_to_delete: u64,
 ) -> Result<(), Box<dyn Error>> {
-    let mut total: usize = 1;
+    let mut total: u64 = 1;
     let (mut req, mut txn) = begin_transaction(&client, &session, RequestType::ReadWrite)?;
     loop {
         let select_sql = format!("SELECT fxa_uid, fxa_kid, collection_id, {} FROM {} WHERE expiry < CURRENT_TIMESTAMP() LIMIT {}", column, table, chunk_size);
--- a/src/bin/purge_ttl.rs
+++ b/src/bin/purge_ttl.rs
@@ -7,17 +7,27 @@ use std::time::Instant;
 use cadence::{
     BufferedUdpMetricSink, Metric, QueuingMetricSink, StatsdClient, Timed, DEFAULT_PORT,
 };
-use env_logger;
+
 use googleapis_raw::spanner::v1::{
-    spanner::{BeginTransactionRequest, CreateSessionRequest, ExecuteSqlRequest, Session},
+    spanner::{
+        BeginTransactionRequest, CommitRequest, CreateSessionRequest, ExecuteSqlRequest, Session,
+    },
     spanner_grpc::SpannerClient,
-    transaction::{TransactionOptions, TransactionOptions_PartitionedDml, TransactionSelector},
+    transaction::{
+        TransactionOptions, TransactionOptions_PartitionedDml, TransactionOptions_ReadOnly,
+        TransactionOptions_ReadWrite, TransactionSelector,
+    },
 };
 use grpcio::{CallOption, ChannelBuilder, ChannelCredentials, EnvBuilder, MetadataBuilder};
 use log::{info, trace, warn};
 use url::{Host, Url};
 
 const SPANNER_ADDRESS: &str = "spanner.googleapis.com:443";
+
+type ResultSet = googleapis_raw::spanner::v1::result_set::ResultSet;
+type ResultSetStats = googleapis_raw::spanner::v1::result_set::ResultSetStats;
+
+use protobuf::well_known_types::Value;
 
 pub struct MetricTimer {
     pub client: StatsdClient,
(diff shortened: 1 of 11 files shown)


src/bin/purge_ttl.rs
```diff
@@ -17,8 +17,15 @@ use grpcio::{CallOption, ChannelBuilder, ChannelCredentials, EnvBuilder, Metadat
 use log::{info, trace, warn};
 use url::{Host, Url};
 
+const BATCH_SIZE: i64 = 1000;
```
jrconlin, line 20: nit: Insert: "If it's a static, it should be an CLI option" spiel here.

src/bin/purge_ttl.rs
```diff
@@ -132,21 +163,73 @@ fn main() -> Result<(), Box<dyn Error>> {
         {
             let _timer_batches = start_timer(&statsd, "purge_ttl.batches_duration");
             let mut req = prepare_request(&client, &session)?;
-            req.set_sql("DELETE FROM batches WHERE expiry < CURRENT_TIMESTAMP()".to_string());
-            let result = client.execute_sql(&req)?;
+            req.set_sql("SELECT (collection_id, fxa_kid, fxa_uid, batch_id) FROM batches WHERE expiry < CURRENT_TIMESTAMP()".to_string());
+            let result = SyncResultSet { result: client.execute_sql(&req)? };
+            let mut sql = "DELETE FROM batches WHERE (collection_id, fxa_kid, fxa_uid, batch_id) IN (".to_string();
```
jrconlin, line 168: Tempted to suggest just leaving this as a `&str` and then doing the `format!` on demand on line 179 & 192, but whatever.

src/bin/purge_ttl.rs
```diff
@@ -132,21 +163,73 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+            let mut i = 0;
+            for row in result {
+                i = i + 1;
+                let collection_id = row[0].get_string_value().parse::<i64>().unwrap();
+                let fxa_kid = row[1].get_string_value().to_owned();
+                let fxa_uid = row[2].get_string_value().to_owned();
+                let batch_id = row[3].get_string_value().to_owned();
+                if i % BATCH_SIZE == 0 {
+                    sql = format!("{}({}, '{}', '{}', '{}'))", sql, collection_id, fxa_kid, fxa_uid, batch_id);
```
jrconlin, line 177: use a different var name here? So you don't have to restore it later?

src/bin/purge_ttl.rs
```diff
@@ -132,21 +163,73 @@ fn main() -> Result<(), Box<dyn Error>> {
...
-            )
+                result.stats().unwrap().get_row_count_lower_bound()
+            );
         }
         {
             let _timer_bso = start_timer(&statsd, "purge_ttl.bso_duration");
             let mut req = prepare_request(&client, &session)?;
-            req.set_sql("DELETE FROM bsos WHERE expiry < CURRENT_TIMESTAMP()".to_string());
-            let result = client.execute_sql(&req)?;
+            req.set_sql("SELECT (collection_id, fxa_kid, fxa_uid, bso_id) FROM bsos WHERE expiry < CURRENT_TIMESTAMP()".to_string());
```
jrconlin, line 202: Fair bit of duplication here. I wonder if it might be worth turning these into a real function so that we don't miss steps in the future?

src/bin/purge_ttl.rs
```diff
@@ -132,21 +163,101 @@ fn main() -> Result<(), Box<dyn Error>> {
...
             let _timer_batches = start_timer(&statsd, "purge_ttl.batches_duration");
             let mut req = prepare_request(&client, &session)?;
-            req.set_sql("DELETE FROM batches WHERE expiry < CURRENT_TIMESTAMP()".to_string());
-            let result = client.execute_sql(&req)?;
+            req.set_sql("SELECT (collection_id, fxa_kid, fxa_uid, batch_id) FROM batches WHERE expiry < CURRENT_TIMESTAMP()".to_string());
+            let result = SyncResultSet {
+                result: client.execute_sql(&req)?,
+            };
+            let mut sql =
+                "DELETE FROM batches WHERE (collection_id, fxa_kid, fxa_uid, batch_id) IN ("
```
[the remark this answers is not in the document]
fzzzy: Yes, I had thought of that but this does seem simpler. Maybe we should just open a followup to do it, if someone has time it would be a nice cleanup.

src/bin/purge_ttl.rs
fzzzy: This is outdated because clippy made me switch to enumerate and get rid of the relevant code.

src/bin/purge_ttl.rs
fzzzy: Yeah, I should make it an env var.

fzzzy:
> Can we retain the old behavior (keeping one opt-in via e.g. an env var or cmd line arg)?

Can you explain further. The context is removed from the history due to a squash.

src/bin/purge_ttl.rs
```diff
@@ -132,21 +163,101 @@ fn main() -> Result<(), Box<dyn Error>> {
         {
             let _timer_batches = start_timer(&statsd, "purge_ttl.batches_duration");
             let mut req = prepare_request(&client, &session)?;
-            req.set_sql("DELETE FROM batches WHERE expiry < CURRENT_TIMESTAMP()".to_string());
-            let result = client.execute_sql(&req)?;
+            req.set_sql("SELECT (collection_id, fxa_kid, fxa_uid, batch_id) FROM batches WHERE expiry < CURRENT_TIMESTAMP()".to_string());
```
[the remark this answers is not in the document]
fzzzy: Done.

src/bin/purge_ttl.rs
fzzzy: Done.

tublitzed:
@fzzzy can you add some more detailed testing instructions here so it's clear how to test locally for reviewers?

Ie, what to run, what to look for, how to seed data, etc. Given that this has the potential to delete user data incorrectly, I think it's probably worth making sure we're taking time to test the functionality here as well as part of the review.

fzzzy:
For testing this, build it with:

`cargo install --path . --bin purge_ttl --locked --root output`

Then run it with:

`RUST_LOG=info GOOGLE_APPLICATION_CREDENTIALS=... SYNC_DATABASE_URL=... output/bin/purge_ttl`

You will need an appropriate credentials file and database url to a spanner database. First of all, it should be able to run and not have any errors deleting 0 items.

Then, if some records are added to the bsos and batches tables, with expired expires timestamps, running it should print info about how many records were deleted. Examining spanner should show that they were actually deleted.

It would be useful to test more than 1000 records, since some of the logic splits on that, and also more than 1000000 records. The script will only delete one million 💯 0️⃣ 0️⃣ 0️⃣ 0️⃣ rows, so if there are more than that before you run the script, there should be some left.

tublitzed:
@fzzzy - could you just update the testing instructions here for the PR so it's easy to find later?

Also, in terms of importing those records into spanner, can you point us to how you're doing that? It's not immediately clear to me from the GCP UI how to do that; I assume you're using the CLI? Do you have some datasets you've been using to test that you could share?

fzzzy:
This is hard to test, so here is proof that it works:

Screen Shot 2020-04-23 at 5 34 37 PM

fzzzy:
If you run the script with no env vars it runs the old behavior, where it deletes everything in one transaction. To activate the incremental functionality, use the env var SYNC_INCREMENTAL=1

src/bin/purge_ttl.rs
```diff
@@ -69,32 +79,184 @@ pub fn statsd_from_env() -> Result<StatsdClient, Box<dyn Error>> {
...
     let mut opt = TransactionOptions::new();
-    opt.set_partitioned_dml(TransactionOptions_PartitionedDml::new());
+    match request_type {
+        RequestType::ReadWrite => {
+            opt.set_read_write(TransactionOptions_ReadWrite::new());
+        }
+        RequestType::ReadOnly => {
+            opt.set_read_only(TransactionOptions_ReadOnly::new());
+        }
+        RequestType::PartitionedDml => {}
```
fzzzy, line 102: Oops, I should implement this properly before this lands.

src/bin/purge_ttl.rs
```diff
@@ -69,32 +79,184 @@ pub fn statsd_from_env() -> Result<StatsdClient, Box<dyn Error>> {
...
+
+fn delete_incremental(
+    client: &SpannerClient,
+    session: &Session,
+    table: String,
+    column: String,
+    batch_size: u64,
+) -> Result<(), Box<dyn Error>> {
+    let (mut req, _txn) = prepare_request(&client, &session, RequestType::ReadOnly)?;
+    let sql = format!("SELECT collection_id, fxa_kid, fxa_uid, {} FROM {} WHERE expiry < CURRENT_TIMESTAMP() LIMIT 1000000", column, table);
```
pjenvey, line 165: nit: I'd prefer the order of columns here matches what we use elsewhere (in the schema and spanner impl), uid, kid, cid, id. For what it's worth I think it's helpful for any new eyes looking at this stuff (and probably long term maintenance -- didn't bother with this on the migration scripts because they're temporary)

src/bin/purge_ttl.rs
```diff
@@ -69,32 +79,184 @@ pub fn statsd_from_env() -> Result<StatsdClient, Box<dyn Error>> {
...
+        trace!(
+            "Selected row for delete: i={} collection_id={} fxa_kid={} fxa_uid={} {}={}",
+            i,
+            collection_id,
+            fxa_kid,
+            fxa_uid,
+            column,
+            bso_id
+        );
+        if i as u64 % batch_size == 0 {
```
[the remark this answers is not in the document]
fzzzy: The latest code got restructured in a way that makes everything simpler and makes this check not required.

src/bin/purge_ttl.rs
fzzzy: Sure. Done

src/bin/purge_ttl.rs
```diff
@@ -69,32 +76,192 @@ pub fn statsd_from_env() -> Result<StatsdClient, Box<dyn Error>> {
...
+    let (mut req, mut txn) = begin_transaction(&client, &session, RequestType::ReadWrite)?;
+    let select_sql = format!("SELECT fxa_uid, fxa_kid, collection_id, {} FROM {} WHERE expiry < CURRENT_TIMESTAMP() LIMIT {}", column, table, chunk_size);
+    trace!("Selecting rows to delete: {}", select_sql);
+    req.set_sql(select_sql.clone());
+    let mut result = SyncResultSet {
+        result: client.execute_sql(&req)?,
+    };
+
+    let mut total: usize = 1;
+    while total < 1_000_000 {
```
pjenvey, line 178: This select could also probably live inside of a `loop {}` (with an extra break for total >= 1_000_000) to avoid duplicating it. Basically emulating a `do while` style loop

src/bin/purge_ttl.rs
```diff
@@ -69,32 +76,192 @@ pub fn statsd_from_env() -> Result<StatsdClient, Box<dyn Error>> {
...
+                id
+            );
+            delete_sql = format!(
+                "{}('{}', '{}', {}, '{}'), ",
+                delete_sql, fxa_uid, fxa_kid, collection_id, id
+            );
+
+            total += 1;
+        }
+        delete_sql = format!("{}('', '', 0, ''))", delete_sql);
```
pjenvey, line 209: I'm probably just paranoid but it'd be nice to avoid this end case in the delete. Could keep an eye on the index while iterating to decide whether or not to omit a comma or not

fzzzy: Another trick I used to use is to remove the trailing comma. I'll switch to that kind of impl.

src/bin/purge_ttl.rs
fzzzy: Great suggestion.

src/bin/purge_ttl.rs
```diff
@@ -69,32 +76,192 @@ pub fn statsd_from_env() -> Result<StatsdClient, Box<dyn Error>> {
...
+        }
+        delete_sql = format!(
+            "{})",
+            delete_sql.trim_end_matches(&", ".to_string()).to_string()
+        );
+        trace!("Deleting chunk with: {}", delete_sql);
+        let mut delete_req = continue_transaction(&session, txn.clone())?;
+        delete_req.set_sql(delete_sql);
+        client.execute_sql(&delete_req)?;
+        info!("{}: removed {} rows", table, total);
```
pjenvey, line 218: total here is wrong since it begins with a value of 1

fzzzy: I'm not using % 1000 any more, so I can start from 0.

PR #601, opened 2020-04-21
User migration5

jrconlin:
## Description
More operational fixes and features

* fixed typos in files
* added `--start_bso`, `--end_bso` to `gen_bso_users.py`
* added `bso_num` arg (same as `--start_bso=# --end_bso=#`) to `migrate_node.py`
* `gen_bso_users.py` takes same `bso_users_file` template as `migrate_node.py`

## Testing

TBD

## Issue(s)

Issue #401

### DIFF
--- a/tools/user_migration/gen_bso_users.py
+++ b/tools/user_migration/gen_bso_users.py
@@ -96,7 +96,7 @@ class BSO_Users:
             )
             self.report.fail(uid, "Unexpected error {}".format(ex))
 
-    def run(self, bso_num=0):
+    def run(self, bso_num):
         connection = self.conf_mysql(self.dsn)
         out_users = []
         bso_file = self.args.bso_users_file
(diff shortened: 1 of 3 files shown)


tools/user_migration/gen_bso_users.py
```diff
@@ -90,17 +96,20 @@ def get_users(self, args):
             )
             self.report.fail(uid, "Unexpected error {}".format(ex))
 
-    def run(self):
+    def run(self, bso_num=0):
```
pjenvey, line 99: nit: I think the bso_num 0 default here is potentially more trouble than it's worth (in case code around it changes later, it's easy to forget about it)

jrconlin: Yeah, probably. I prefer that something like that have a default value, but I'm not sure I can strongly defend it here.

PR #620, opened 2020-05-04
bug: Add retry and sleep to purge_ttl attempts

jrconlin:
## Description

* env: RETRY_COUNT sets the number of retries for a transaction (default
10)
* env: RETRY_SLEEP_MILLIS sets the sleep period between retry attempts
(default 0)

## Testing

How should reviewers test?

Setting `RUST_LOG=info` will display a confirmation message indicating the values read from the environment. (e.g.
```
lvbox2:~/src/mozilla/syncstorage-rs> RUST_LOG=info RETRY_COUNT=5 SYNC_DATABASE_URL=spanner://projects/foo target/debug/purge_ttl 
[2020-05-04T18:09:40Z INFO  purge_ttl] Retries: 5, sleep: 0ms
[2020-05-04T18:09:40Z INFO  purge_ttl] For projects/foo
...
```
## Issue(s)

Closes https://github.com/mozilla-services/syncstorage-rs/issues/626

### DIFF
--- a/src/bin/purge_ttl.rs
+++ b/src/bin/purge_ttl.rs
@@ -24,8 +24,8 @@ use log::{info, trace, warn};
 use url::{Host, Url};
 
 const SPANNER_ADDRESS: &str = "spanner.googleapis.com:443";
-const RETRY_ENV_VAR: &str = "RETRY_COUNT"; // Default value = 10
-const SLEEP_ENV_VAR: &str = "RETRY_SLEEP_MILLIS"; // Default value = 0
+const RETRY_ENV_VAR: &str = "PURGE_TTL_RETRY_COUNT"; // Default value = 10
+const SLEEP_ENV_VAR: &str = "PURGE_TTL_RETRY_SLEEP_MILLIS"; // Default value = 0
 
 use protobuf::well_known_types::Value;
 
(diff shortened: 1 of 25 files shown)


src/bin/purge_ttl.rs
```diff
@@ -298,34 +308,69 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+                } else {
+                    delete_all(&client, &session, "batches".to_owned())
+                } {
+                    Ok(_) => {
+                        success = true;
+                        break;
+                    }
+                    Err(e) => {
+                        warn!("Batch transaction error: {}: {:?}", i, e);
+                        if nap_time.as_secs() > 0 {
```
pjenvey, line 331: any millis value < 1sec would produce 0 for `as_secs`:

```suggestion
                        if nap_time.as_millis() > 0 {
```

(and same below too)

src/bin/purge_ttl.rs
```diff
@@ -298,34 +308,69 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+                        max_to_delete,
+                    )
+                } else {
+                    delete_all(&client, &session, "batches".to_owned())
+                } {
+                    Ok(_) => {
+                        success = true;
+                        break;
+                    }
+                    Err(e) => {
```
pjenvey, line 329: I think this should check against specific error codes. It looks like everything is currently `dyn Error` so we'd have to start returning grpcio Errors to do this.

PR #621, opened 2020-05-04
feat: add async to `delete_all`

### LINKED ISSUE #615: [spanner] delete_all blocks while beginning its transaction
Related to #581:

`delete_all` doesn't operate on a collection so the db middleware doesn't explicitly begin a transaction for it (via `lock_for_read/write`).

Instead it relies on the implicit calling of `begin` (via `.sql()`). This isn't async, so it's doing a blocking io call within an async method running in the executor.

We can fix this similarly to the mysql impl: by having it explicitly call `begin()` itself.

jrconlin:
## Description

* elevated `begin()` to an interface call to allow `delete_all` the
ability to lock the database.
* added `begin_async()` wrapper to mysql & mock to mirror spanner function
* added `begin()` call to `delete_all`

## Testing

cargo test should pass for both mysql and spanner.

## Issue(s)

Closes #615

### DIFF
--- a/src/db/mock.rs
+++ b/src/db/mock.rs
@@ -53,6 +53,10 @@ impl Db for MockDb {
         Box::pin(future::ok(()))
     }
 
+    fn begin(&self, _for_write: bool) -> DbFuture<()> {
+        Box::pin(future::ok(()))
+    }
+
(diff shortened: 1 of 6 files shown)


jrconlin:
Pending #622 which will have clippy fixes.

PR #623, opened 2020-05-05
feat: Don't report Conflict errors to sentry

### LINKED ISSUE #614: Don't emit 503 Conflict errors to sentry
We've been meaning to stop sending these, let's finally stop sending them. No longer reporting them will help w/ the issue outlined in #613. 

Let's consider emitting a metric for these instead if it's not too difficult.

`lock_for_write` can emit these in a different manner than the typical way they're emitted (due to Spanner returning an ABORTED). If we do a metric, we could optionally also track these different types separately (via a metric tag): but this is just a nice to have.

jrconlin:
## Description

* Added new metric "db.conflict"
* Added new error method `is_reportable()` for future error filtering.

Describe these changes.

## Testing

Unfortunately, testing requires  a db conflict write and, metrics and sentry reporting. A `debug!()` line does include the error that is not being directly reported to Sentry.

## Issue(s)
Closes #614

### DIFF
--- a/src/db/error.rs
+++ b/src/db/error.rs
@@ -99,6 +99,11 @@ from_error!(grpcio::Error, DbError, |inner: grpcio::Error| {
         {
             DbErrorKind::Conflict
         }
+        grpcio::Error::RpcFailure(ref status)
+            if status.status == grpcio::RpcStatusCode::ALREADY_EXISTS =>
+        {
+            DbErrorKind::Conflict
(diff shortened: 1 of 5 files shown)


fzzzy:
I think we can produce conflict errors on stage by running the load tester and maybe the delete script at the same time. I think we should test this on stage, as getting everything working locally seems pretty hard.

PR #630, opened 2020-05-13
chore: tag 0.3.3

### LINKED ISSUE #555: Emit db Pool state metrics
There should be metrics emitted about our db Pool periodically. E.g. the number of connections in use and possibly the number of idle connections.

`r2d2::Pool` & `cadence::StatsdClient` are both Sync, so we should be able to run a background thread holding references to both, polling the Pool for its current state and emitting a metric every e.g. 15-30 seconds.

pjenvey:
## Description

Including the pool metrics hostname tag from the release branch

## Issue(s)

Closes #555


pjenvey:
closing for now there'll be more to merge later

PR #594, opened 2020-04-17
refactor: Convert actix-web frontend *_bso calls to async await

Emmanuel-Melon:
## Description

Convert the following actix-web front-end functions from Futures style to async await
- delete_bso
- get_bso
- put_bso

## Testing

Tests should pass

## Issue(s)

Closes [543](https://github.com/mozilla-services/syncstorage-rs/issues/543).

Issue #599, opened 2020-04-21
delete_bso eats errors


pjenvey:
We made a minor breaking change to the `delete_bso` handler when updating to new async/std Future:

https://github.com/mozilla-services/syncstorage-rs/compare/0.2.9...0.3.0#diff-f30baa29f1302c6ed83d9717ab8d3647R355

`git diff 0.2.9..0.3.0 src/web/handlers.rs`:
```
-pub fn delete_bso(bso_req: BsoRequest) -> impl Future<Item = HttpResponse, Error = Error> {
+pub fn delete_bso(bso_req: BsoRequest) -> impl Future<Output = Result<HttpResponse, Error>> {
     bso_req.metrics.incr("request.delete_bso");
     bso_req
         .db
@@ -354,10 +362,13 @@ pub fn delete_bso(bso_req: BsoRequest) -> impl Future<Item = HttpResponse, Error
             id: bso_req.bso,
         })
         .map_err(From::from)
-        .map(|result| HttpResponse::Ok().json(json!({ "modified": result })))
+        .map(|result: Result<SyncTimestamp, Error>| match result {
+            Ok(result) => Ok(HttpResponse::Ok().json(json!({ "modified": result }))),
+            Err(_e) => Ok(HttpResponse::NotFound().finish()),
+        })
 }
```

Instead of propagating any errors from the db layer it now returns a 404.

PR #638, opened 2020-05-14
refactor: Convert actix-web frontend *_bso calls to async await

### LINKED ISSUE #543: Convert actix-web frontend *_bso calls to async await
The actix-web frontend *_bso calls:

-    delete_bso
-   get_bso
-  put_bso

Are in a Future style. They can now be converted over to use async/await syntax instead.

fzzzy:
Squash and rebase of this: https://github.com/mozilla-services/syncstorage-rs/pull/594

Fix #543

## Description

Describe these changes.

## Testing

How should reviewers test?

## Issue(s)

Closes [link](link).

### DIFF
--- a/src/web/handlers.rs
+++ b/src/web/handlers.rs
@@ -354,18 +354,15 @@ pub fn post_collection_batch(
 
 pub async fn delete_bso(bso_req: BsoRequest) -> Result<HttpResponse, Error> {
     bso_req.metrics.incr("request.delete_bso");
-    match bso_req
+    let result = bso_req
         .db
         .delete_bso(params::DeleteBso {
             user_id: bso_req.user_id,
             collection: bso_req.collection,
             id: bso_req.bso,
         })
-        .await?
-    {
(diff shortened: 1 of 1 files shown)


src/web/handlers.rs
```diff
@@ -352,40 +352,42 @@ pub fn post_collection_batch(
...
+    match bso_req
         .db
         .delete_bso(params::DeleteBso {
             user_id: bso_req.user_id,
             collection: bso_req.collection,
             id: bso_req.bso,
         })
-        .map_err(From::from)
-        .map_ok(|result| HttpResponse::Ok().json(json!({ "modified": result })))
+        .await
```
pjenvey, line 364:
```suggestion
        .await?
```

Below is the old 404 code we fixed for #599. Instead it should just return the `HttpResponse::Ok` (when the ? above doesn't error out)

pjenvey: The below needs to change to just return the `Ok` (and should fix the build)

PR #662, opened 2020-06-08
refactor: quiet clippy warnings

pjenvey:
## Description

per rustc 1.44


jrconlin:
Huh, kinda figured there would be more with the new version.

Issue #631, opened 2020-05-13
unable to complete purge_ttl run


tublitzed:
See bugzilla issue here: [1626707](https://bugzilla.mozilla.org/show_bug.cgi?id=1626707).

Going to keep this one open until we feel confident we can run purge_ttl at the scale needed in order to complete the migrated.

erkolson:
Updates here:
https://bugzilla.mozilla.org/show_bug.cgi?id=1626707#c9

tl;dr, We are purging rows effectively now and will probably be down to zero each night tonight.

Deleted rows can take up to 48 hours to be deleted off disk and the 30-day trend of spanner disk usage shows a flatline for several days as purged rows are being removed permanently
<img width="1648" alt="Screen Shot 2020-05-18 at 1 40 17 PM" src="https://user-images.githubusercontent.com/9535472/82243516-c65f9980-990d-11ea-90f0-b0f25eba8da8.png">

pjenvey:
The 60 minute limit we're stopping at may be related to a one of the timeout values set in the driver (36000000):

https://github.com/googleapis/python-spanner/blob/62edbe1/google/cloud/spanner_v1/gapic/spanner_client_config.py#L28

I wonder if we're able to overwrite these values with larger values.

pjenvey:
According to the client team ([May 18th Sync Teams Sync](https://docs.google.com/document/d/1AlYkT4CaQiT3nnTnVqVMq3LbAu293EufiVgz53VqmLY/edit#)) only 4 collections utilize TTLs: History (2 months), clients (3 weeks), forms (3 years), tabs (1 year).

We can sort of confirm this, running against stage:

```
SELECT bsos.collection_id, collections.name, COUNT(bsos.expiry)
  FROM bsos
  LEFT JOIN collections
    ON bsos.collection_id = collections.collection_id
 WHERE expiry < CURRENT_TIMESTAMP()
 GROUP BY bsos.collection_id, collections.name
 ORDER BY bsos.collection_id
```

collection_id | name | Unspecified
--- | --- | ---
1 | clients | 543
3 | forms | 724250
4 | history | 757858
7 | bookmarks | 749967
8 | prefs | 730457
9 | tabs | 35
10 | passwords | 731108


The 5 collections with the large numbers are due to the load tester: https://github.com/mozilla-services/syncstorage-loadtest/blob/535dd4b/loadtest.py#L24

[...]

PR #668, opened 2020-06-11
feat: add conditions, args to purge_ttl script

jrconlin:
## Description

attempt to try and provide a way to allow the purge_ttl script to complete.

Adds arguments (ENV VARS):

  --instance_id (INSTANCE_ID)  Spanner instance id
  --database_id (DATABASE_ID)  Spanner database id
  --sync_database_url (SYNC_DATABASE_URL) Spanner DSN
        `spanner://instance/database`
  --collection_ids (COLLECTION_IDS)
        JSON formatted list of collections to limit deletions
        e.g. `--collection_ids=123` limits to just collection 123
             `--collection_ids=[123,456]` limits to both 123 & 456
             default is all collections

## Testing

May be used with the stage spanner database instance

## Issue(s)

Issue #631

### DIFF
--- a/tools/spanner/purge_ttl.py
+++ b/tools/spanner/purge_ttl.py
@@ -4,6 +4,8 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+import argparse
+import json
 import os
 import sys
 import logging
@@ -23,9 +25,8 @@ logging.basicConfig(
 client = spanner.Client()
 
 
-def from_env():
+def from_env(url):
     try:
-        url = os.environ.get("SYNC_DATABASE_URL")
         if not url:
             raise Exception("no url")
         purl = parse.urlparse(url)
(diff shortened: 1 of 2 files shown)


erkolson:
TBC, the syntax for the env var `COLLECTION_IDS` is identical to the cli flag?  i.e.
```
export COLLECTION_IDS="[123,456]"
```

jrconlin:
Yep. It uses the same logic. Let me know if that's a problem and I'll come up with something different. I tried it out locally on bash and it works, provided the value is quoted like you're doing.

tools/spanner/purge_ttl.py
```diff
@@ -41,39 +42,103 @@ def from_env():
...
+    )
+    args = parser.parse_args()
+    collections = json.loads(args.collection_ids)
+    if not isinstance(collections, list):
+        collections = [collections]
+    args.collection_ids = collections
+    if args.sync_database_url and not (
+            args.instance_id and args.database_id):
+        (args.instance_id,
+         args.collection_id) = from_env(args.sync_database_url)
```
pjenvey, line 131:
```suggestion
         args.database_id) = from_env(args.sync_database_url)
```

tools/spanner/purge_ttl.py
```diff
@@ -41,39 +42,103 @@ def from_env():
...
+        default=os.environ.get("COLLECTION_IDS", "[]"),
+        help="JSON array of collection IDs to purge"
+    )
+    args = parser.parse_args()
+    collections = json.loads(args.collection_ids)
+    if not isinstance(collections, list):
+        collections = [collections]
+    args.collection_ids = collections
+    if args.sync_database_url and not (
+            args.instance_id and args.database_id):
```
pjenvey, line 129: `INSTANCE/DATABASE_ID` have a default so they'll pretty much always override a `SYNC_DATABASE_URL` here. I think we can kill these defaults (or maybe these args entirely unless @erkolson or anyone else uses them, I don't), [the cron job uses `SYNC_DATABASE_URL`](https://github.com/mozilla-services/cloudops-infra/blob/ea0adbe/projects/sync/k8s/charts/sync/templates/ttl-purge-cron.yaml#L49).

tools/spanner/purge_ttl.py
```diff
@@ -41,39 +42,103 @@ def from_env():
...
-    instance = client.instance(instance_id)
-    database = instance.database(database_id)
+def deleter(database, name, query):
+    with statsd.timer("syncstorage.purge_ttl.{}_duration".format(name)):
+        logging.info("Running: {}".format(query))
+        start = datetime.now()
+        result = database.execute_partitioned_dml(query)
+        end = datetime.now()
+        logging.info("{}: removed {} rows, batches_duration: {}".format(
+            name, result, end - start))
```
pjenvey, line 52:
```suggestion
        logging.info("{}: removed {} rows, {}_duration: {}".format(
            name, result, name, end - start))
```

tools/spanner/purge_ttl.py
```diff
@@ -41,39 +42,103 @@ def from_env():
...
+
+
+def add_conditions(args, query):
+    if args.collection_ids:
+        query += " AND collection_id"
+        if len(args.collection_ids) == 1:
+            query += " = {:d}".format(args.collection_ids[0])
+        else:
+            query += " in ({})".format(
+                ', '.join(map(str, args.collection_ids)))
```
pjenvey, line 62: I always prefer bind params :) but I guess this'll do for now

tools/spanner/purge_ttl.py
jrconlin: d'oh

tools/spanner/purge_ttl.py
erkolson: I have always used the `SYNC_DATABASE_URL` and was unaware of the other options.

erkolson:
@jrconlin , what is the behavior when this is passed `COLLECTION_IDS="[]"`?

tools/spanner/purge_ttl.py
jrconlin: Heh, I use the others because I can never remember what the spanner DSN is. I've reworked things a bit to be smarter and more clear what's going on. Thanks!

tools/spanner/purge_ttl.py
jrconlin: Heh, so do I, but it would have made things really complicated. Hopefully we trust @erkolson not to insert ";drop tables;"

tools/spanner/purge_ttl.py
```diff
@@ -23,57 +25,119 @@
...
+    return query
+
 
-    logging.info("For {}:{}".format(instance_id, database_id))
+def spanner_purge(args, request=None):
+    instance = client.instance(args.instance_id)
+    database = instance.database(args.database_id)
+
+    logging.info("For {}:{}".format(args.instance_id, args.database_id))
+    batch_query = add_conditions(
```
pjenvey, line 71: Oh sorry, almost forgot: let's not apply this to batches.

They hopefully won't ever have timeout problems, their expiry index is sane: a "global" index or whatever spanner calls it, not interleaved in `user_collections`. They're always going to have much less expired data to purge and their expiry logic is also a little different: *all* batches expire (and rather quickly) regardless of collection_id.

The batch delete job currently takes less than 10 min on prod.

PR #687, opened 2020-06-26
Switch to bb8

### LINKED ISSUE #669: [spanner] Make DbPool fully async
Spanner's pool `get` still goes through a threadpool. It could potentially be a source of "clogged" connections when the server's overloaded. We saw similar "clogging" under overloaded conditions when all Db calls went through a thread pool.

Take a scenario where there's slowness/hiccups between syncstorage and Spanner:

- actix-threadpool defaults to a size of num_cpus * 5 (e.g. 40)
- we have a larger DB_POOL_MAX_SIZE (e.g. a value of 300)
- syncstorage-rs is overloaded with a large number of pending pool `get` requests

The threadpool could potentially "backup" with pending requests, impart because we're overloaded w/ requests (> 300 at a time) and also impart due to Spanner taking a long time to respond to the Pool's `is_valid` calls.

The tasks "backed up" in the threadpool may become stale (that is their corresponding db connection may become stale) by the time they are actually gotten to and ran by the threadpool.

Mcat12:
## Description

Updates spanner DB pool to use bb8, an async version of r2d2.

## Testing

All tests pass. Should also run an instance under load test to verify.

## Issue(s)

Closes #669


src/db/mod.rs
```diff
@@ -59,10 +61,11 @@ pub const BATCH_LIFETIME: i64 = 2 * 60 * 60 * 1000; // 2 hours, in milliseconds
 /// DbPools' worker ThreadPool size
 pub const DB_THREAD_POOL_SIZE: usize = 50;
 
-type DbFuture<T> = LocalBoxFuture<'static, Result<T, ApiError>>;
+type DbFuture<'a, T> = LocalBoxFuture<'a, Result<T, ApiError>>;
 
+#[async_trait(?Send)]
 pub trait DbPool: Sync + Send + Debug {
```
Mcat12, line 67: We can revert this back to `#[async_trait]` (adds back Send to the returned future)

src/db/spanner/manager.rs
```diff
@@ -16,29 +18,42 @@ use crate::{
 
 const SPANNER_ADDRESS: &str = "spanner.googleapis.com:443";
 
-pub struct SpannerConnectionManager {
+pub struct SpannerConnectionManager<T> {
```
Mcat12, line 21: This generic isn't used; it can be safely removed.

src/db/transaction.rs
```diff
@@ -0,0 +1,146 @@
...
+
+pub struct DbTransactionPool {
+    pool: Box<dyn DbPool>,
+    lock_collection: Option<params::LockCollection>,
+    is_read: bool,
+    tags: Tags,
+}
+
+impl DbTransactionPool {
+    pub async fn transaction<'a, A, F>(&'a self, action: A) -> Result<HttpResponse, Error>
```
Mcat12, line 26: We should probably document this struct and function

src/db/transaction.rs
```diff
@@ -0,0 +1,146 @@
...
+        async move {
+            let no_agent = HeaderValue::from_str("NONE")
+                .expect("Could not get no_agent in DbTransactionPool::from_request");
+            let useragent = req
+                .headers()
+                .get("user-agent")
+                .unwrap_or(&no_agent)
+                .to_str()
+                .unwrap_or("NONE");
+            info!(">>> testing db middleware"; "user_agent" => useragent);
```
Mcat12, line 91: This code was moved from the DB middleware, so some of these messages should be updated.

src/web/middleware/db.rs
```diff
@@ -16,6 +16,7 @@ use crate::web::middleware::sentry::{queue_report, report};
 use crate::web::{
```
Mcat12, line 16: This middleware is no longer part of the project, so the file can be deleted. Most of the code has been moved into `DbTransactionPool`.

src/web/middleware/mod.rs
```diff
@@ -1,4 +1,4 @@
-pub mod db;
+// pub mod db;
```
Mcat12, line 1: We can fully delete the module now.

src/web/extractors.rs
```diff
@@ -758,15 +753,12 @@ impl FromRequest for CollectionPostRequest {
...
-                    CollectionParam,
-                    BsoQueryParams,
-                    BsoBodies,
-                )>::from_request(&req, &mut payload)
-                .await?;
+
+            let user_id = HawkIdentifier::from_request(&req, &mut payload).await?;
+            let collection = CollectionParam::from_request(&req, &mut payload).await?;
+            let query = BsoQueryParams::from_request(&req, &mut payload).await?;
+            let mut bsos = BsoBodies::from_request(&req, &mut payload).await?;
```
Mcat12, line 760: This change (and there's one other instance) can be mostly reverted, since the DB is no longer one of the extractors. This change was made because the DB extractor was not `'static`, while the impl for the tuple of extractors required `'static`.

src/web/middleware/mod.rs
```diff
@@ -38,3 +39,22 @@ impl SyncServerRequest for ServiceRequest {
         HawkIdentifier::extrude(self, &method.as_str(), &self.uri(), &ci, &state, Some(tags))
     }
 }
+
+impl SyncServerRequest for HttpRequest {
+    fn get_hawk_id(&self) -> Result<HawkIdentifier, Error> {
```
Mcat12, line 44: This should be de-duplicated with the impl for `ServiceRequest`, if possible.

src/web/middleware/precondition.rs
```diff
@@ -76,140 +76,123 @@ where
...
-                    sreq.into_response(
+                    ));
+                }
+            };
+
+            let (req, payload) = sreq.into_parts();
+            let db_pool = extrude_db_pool(&req).await?;
+            let edb = db_pool.get().await.map_err(Error::from);
+            let sreq = ServiceRequest::from_parts(req, payload)
+                .unwrap_or_else(|_| panic!("TODO: this should never happen"));
```
Mcat12, line 129: This is a bit iffy, probably better to manually extract the DB pool from state. `ServiceRequest::from_parts` panics if there is more than one copy of the request (shouldn't happen with this code, but is a liability).

fzzzy:
These are all great comments about things to clean up; what do you think about creating a new issue for each individual chunk of cleanup? We can mark them as Good First Bug and come through and take them out whenever we have some downtime and need a bug that is a `1` difficulty.

fzzzy:
Superseded by https://github.com/mozilla-services/syncstorage-rs/pull/692

Mcat12:
Opened #693, #694, #695, #696, #697, and #698 to handle the review comments.

PR #691, opened 2020-06-26
Chore/update

jrconlin:
## Description
Updates
* grpcio to "0.6.0"
* protobuf to "2.15.0"
* vendor'd mozilla-rust-sdk to "0.0.3"

## Testing

How should reviewers test?

## Issue(s)

Chore


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM rust:1.42.0-buster as builder
+FROM rust:1.44.0-buster as builder
```
fzzzy, line 1: 👍

jrconlin: Yeah, I think we need to add a reminder to our calendars to update the docker file one week after a new rust version rolls out.

PR #692, opened 2020-06-29
Switch to bb8

fzzzy:
## Description

r2d2 is a blocking database pool, but the rest of our code is now fully nonblocking. Switch to bb8, which has a very similar interface. The fact that bb8 needs to know the lifetime necessitates a large number of changes, but they are fairly mechanical.

Also, @Mcat12 switched some handlers to async/await, which helps give bb8 the lifetime it needs; database connections now are acquired in the handler instead of in an extractor, which requires the 'static lifetime.

## Testing

This should be a transparent change; all tests should pass, things should continue working as expected.

## Issue(s)

Closes #669.

Issue #695, opened 2020-06-29
Avoid splitting `ServiceRequest` into parts to access the DB pool


Mcat12:
Due to the bb8 refactor (#692), `PreConditionCheckMiddleware` splits the `ServiceRequest` into `HttpRequest` and the payload in order to use the DB pool extractor (`extrude_db_pool`):
https://github.com/mozilla-services/syncstorage-rs/blob/3a5fddb54b7641f7bb93d340d0728c64b70cf3bb/src/web/middleware/precondition.rs#L125-L129

The parts are then recombined to get the `ServiceRequest` back. The recombining code will fail if there is more than one reference to the request (i.e. if it has been cloned), as that is an invariant of `ServiceRequest`. This is a liability, so we should try to refactor that code to access the DB pool without splitting the service request into parts.

pjenvey:
The latest switch-to-bb8 updates likely negate this (killing the PreconditionCheck middleware)

fzzzy:
bb8 is merged and this is killed now.

PR #709, opened 2020-07-02
Fix performance issues in bb8 branch

Mcat12:
## Description

#692 was experiencing some very surprising performance regressions in some integration tests. These regressions have been narrowed down to a change in how DB transactions are handled in the branch. Previously the whole request was in one DB transaction. This was changed to only running the DB transaction in the handler. There were some middlewares/extractors directly accessing the DB pool, which is believed to be the cause of the performance regressions. This PR wraps those non-handler DB interactions in transactions, along with a few other fixes (ex. remove unsafe Send impl for SpannerDb).

## Testing

All requests made during the `test_batch_ttl_update` integration test should take < 4 seconds. Previously, some requests took 15-20 seconds. Make sure you are using Spanner for this test.

## Issue(s)

N/A

### DIFF
--- a/src/db/transaction.rs
+++ b/src/db/transaction.rs
@@ -15,6 +15,7 @@ use futures::future::LocalBoxFuture;
 use futures::FutureExt;
 use std::future::Future;
 
+#[derive(Clone)]
 pub struct DbTransactionPool {
     pool: Box<dyn DbPool>,
     lock_collection: Option<params::LockCollection>,
@@ -78,6 +79,11 @@ impl FromRequest for DbTransactionPool {
     type Config = ();
 
     fn from_request(req: &HttpRequest, _: &mut Payload<PayloadStream>) -> Self::Future {
+        // Cache in extensions to avoid parsing for the lock info multiple times
+        if let Some(pool) = req.extensions().get::<Self>() {
+            return futures::future::ok(pool.clone()).boxed_local();
+        }
+
         let req = req.clone();
         async move {
             let no_agent = HeaderValue::from_str("NONE")
@@ -134,12 +140,15 @@ impl FromRequest for DbTransactionPool {
                 (None, true)
             };
 
-            Ok(Self {
+            let pool = Self {
                 pool: state.db_pool.clone(),
                 lock_collection: lc,
                 is_read,
                 tags,
-            })
+            };
+
+            req.extensions_mut().insert(pool.clone());
+            Ok(pool)
         }
         .boxed_local()
     }
--- a/src/db/transaction.rs
+++ b/src/db/transaction.rs
@@ -24,49 +24,112 @@ pub struct DbTransactionPool {
 }
 
 impl DbTransactionPool {
-    pub async fn transaction<'a, A, F>(&'a self, action: A) -> Result<HttpResponse, Error>
+    /// Perform an action inside of a DB transaction. If the action fails, the
+    /// transaction is rolled back. If the action succeeds, the transaction is
+    /// NOT committed. Further processing is required before we are sure the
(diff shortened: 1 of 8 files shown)


src/web/extractors.rs
```diff
@@ -1474,18 +1475,23 @@ impl FromRequest for BatchRequestOpt {
...
-                            Some(ftags),
-                        )
-                        .into());
-                    }
-                    Some(batch)
+                    let transaction_pool: DbTransactionPool =
+                        DbTransactionPool::extract(&req).await?;
+
+                    transaction_pool
+                        .transaction(|db| async move {
```
fzzzy, line 1481: This is just a nit, but the batch could be valid for the duration of this transaction, but then not valid for the transaction that the handler is running in. This is mostly pedantic, because it would never happen in real life unless there was an attack happening or something, but this transaction really should be the same transaction as the handler that needs this check. I'm going to open a new bug to move from_request for BatchRequestOpt into a normal function.

fzzzy: https://github.com/mozilla-services/syncstorage-rs/issues/710

src/db/transaction.rs
```diff
@@ -23,49 +24,112 @@ pub struct DbTransactionPool {
...
+    /// if the HTTP response is an error.
+    pub async fn transaction_http<'a, A, F>(&'a self, action: A) -> Result<HttpResponse, Error>
+    where
+        A: FnOnce(Box<dyn Db<'a>>) -> F,
+        F: Future<Output = Result<HttpResponse, Error>> + 'a,
+    {
+        let (resp, db) = self.transaction_internal(action).await?;
+
+        // HttpResponse can contain an internal error
         let result = match resp.error() {
```
jrconlin, line 124: Sorry, talking through a bit of code here:

🦆 I think I'm missing something here. If I understand correctly, `transaction_internal` wraps the `action`, which will return a Result. That result could be a different Result than the `transaction_internal` wrapper. That's what we're trapping for below.  (So `transaction_internal` could return an `Ok(SomeErr(), db)` correct?) 

I see we've got [a handler](https://github.com/mozilla-services/syncstorage-rs/pull/709/files#diff-fb3d577e13883be5bf8aee1c9db0e19bR64) inside of the `transaction_internal` to handle releasing the locks on error.

src/web/middleware/precondition.rs
```diff
@@ -158,7 +139,18 @@ where
             let bso = BsoParam::extrude(sreq.head(), &mut sreq.extensions_mut()).ok();
             let bso_opt = bso.map(|b| b.bso);
 
-            let resource_ts = db.extract_resource(user_id, collection, bso_opt).await?;
+            let (req, payload) = sreq.into_parts();
```
fzzzy, line 142: Very nice way to work around the extractor not usually having access to the ServiceRequest, and then still turning it back into sreq later. Nice.

Opened a new bug to move this into a plain function at some point. https://github.com/mozilla-services/syncstorage-rs/issues/711

Mcat12: There's actually an issue open to avoid the possible panic when we recombine the parts: #695

src/db/transaction.rs
Mcat12: So, Actix responses can contain an internal error: https://docs.rs/actix-web/2.0.0/actix_web/struct.HttpResponse.html#method.error

`transaction_internal` handles if the action returns an error, but it doesn't know if the Ok value contains an error or not (ex. HttpResponse). `transaction_http` handles `HttpResponse` explicitly and can rollback if there's an internal error in the response.

jrconlin: Ah! Ok, that was the bit I was missing. Thanks!

PR #717, opened 2020-07-08
Update hkdf, hmac, and sha2, as they all need to be updated at once.

fzzzy:
## Description

There are version updates for hkdf, hmac, and sha2, but in order for the tests to pass they all need to be updated at once.

## Testing

Just dependency updates; nothing should change. All tests should pass.


tublitzed:
@fzzzy are you still actively working on this, or should we close?

fzzzy:
It would be a good idea to get it landed. There are some broken tests that need investigating.

tublitzed:
ok great. Can you open a bug to track this on the board then and assign to yourself?

PR #724, opened 2020-07-15
feat: re-enable offset optimization

### LINKED ISSUE #645: [mysql] Re-enable offset optimization
We implemented #53 (+ #469 related fix) but disabled it in #559. It should be re-enabled on the mysql backend without impacting Spanner: we probably don't want it enabled on Spanner at all at this point, so the extractor portion of this optimization needs to accommodate that.

jrconlin:
Closes #645

## Description
Restore mysql pagination optimizations.

NOTE:
there's some discussion that this may not be required due to how it could impact the query cache.

NOTE:
During work on this patch, the application faulted very early in
processing. This may have been due to a library issue around libssl
setting an `_atexit()` handler that conflicts with the rust threading
model. (see https://github.com/sfackler/rust-openssl/issues/1174 and
https://github.com/openssl/openssl/issues/6214 for possibly related
issues)

## Testing

Testing included.

## Issue(s)

Closes #645


src/main.rs
```diff
@@ -25,6 +25,7 @@ struct Args {
 
 #[actix_rt::main]
 async fn main() -> Result<(), Box<dyn Error>> {
+    println!("Go...");
```
fzzzy, line 28: I think this is good to have here at the very beginning, but let's make it "syncstorage-rs is starting up"

src/web/extractors.rs
```diff
@@ -1227,13 +1227,6 @@ impl ToString for Offset {
 impl FromStr for Offset {
     type Err = ParseIntError;
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        // issue559: Disable ':' support for now: simply parse as i64 as
```
fzzzy, line 1230: This is going to turn it on for spanner as well. Can you make turning it on a configuration parameter?

fzzzy: Default it to off

src/main.rs
jrconlin: actually, it's crap. I had added it to see where the seg fault was happening.

src/web/extractors.rs
jrconlin: This method cracks the Offset correctly, based on what it's handed. Spanner (at least currently) never reads (or writes) the timestamp. It's unlikely that the data store would switch while processing, but even if it did, I believe things would continue to work correctly. I will add a bit of logic to the mysql side to use the numeric offset if the sorting is by time and there's no timestamp specified, just to cover a possible edge case.

jrconlin:
After group discussion, we may belay this patch.

The discussion centered around two things. In the Python code, some of the offset [work was disabled](https://github.com/mozilla-services/server-syncstorage/blob/master/syncstorage/storage/sql/queries_generic.py#L230-L245) due to a problem with older mysql databases. It appears that this option was never enabled. 

In addition, re-writing this query would throw out results stored in the mysql query cache, which would be far faster than recreating the cached data set, so this patch may actually *hurt* performance (thanks to @fzzzy for pointing that out).

Finally, since the `bso_id` is non-numeric, significant work would need to be done to either make rust handle semi-numeric offsets, or have rust determine the number of bsos with matching `modified` timestamps to calculate the appropriate offset.

The current, numeric only offsetting method is fine as is.

Issue #732, opened 2020-07-17
[Spanner] google_default_credentials should probably be async


pjenvey:
When creating/connecting a Spanner client, there's potentially 2 blocking operations required:

1. Creating/connecting? creds `ChannelCredentials::google_default_credentials()`
2. Creating/connecting? a grpc Channel/client `ChannelBuilder::secure_connect`

One of these is likely a potentially blocking operation, probably 1. as [grpc's docs](https://grpc.io/docs/guides/auth/#authenticate-with-google-3) show C#'s `google_default_credentials()` equivalent as async:

```c#
using Grpc.Auth;  // from Grpc.Auth NuGet package
...
// Loads Google Application Default Credentials with publicly trusted roots.
var channelCredentials = await GoogleGrpcCredentials.GetApplicationDefaultAsync();

var channel = new Channel("greeter.googleapis.com", channelCredentials);
var client = new Greeter.GreeterClient(channel);
...
```

 The grpcio lib does not provide async versions of these methods. They may potentially be blocking our async `SpannerConnection::connect` method, which could be problematic.

We should probably contact Google's grpc client team for advise here and followup with grpc-rs project.

fzzzy:
I was able to make the block alarm work, so we could use it to detect any other places like this in the code. https://github.com/fzzzy/block-alarm

pjenvey:
I haven't been able to trigger `block_alarm` on the latest master even with the lack of async here.

Tracing the c# version it goes:

- GoogleGrpcCredentials.GetApplicationDefaultAsync()
  - GoogleCredentials.GetDefaultCredentialAsync() (google-api-dotnet-client lib)
    - loads from a cached credential or calls [looks around from different sources](https://github.com/googleapis/google-api-dotnet-client/blob/a6ac8db/Src/Support/Google.Apis.Auth/OAuth2/DefaultCredentialProvider.cs#L82)

Our rust code:
- google_default_credentials()
  - [rpc_sys::grpc_google_default_credentials_create()](https://github.com/grpc/grpc/blob/dcb5840/src/core/lib/security/credentials/google_default/google_default_credentials.cc#L335) (involving various calls deep into grpc creds code)

The composite credentials code refers to a `composite_call_metadata_cb` method, suggesting some kind of async callback. Without more digging I'm not sure if there's any caching similar to the C# going on here or any potential blocking code that'd be triggered on every call.

pjenvey:
[...]

PR #733, opened 2020-07-17
fix: switch create_session to async

### LINKED ISSUE #731: [spanner] switch create_session to async
We forgot to switch `create_session_opt` in Spanner's `manager.rs` to async. It's called from async methods but blocking and create session can be one of Spanner's slowest calls.

pjenvey:
## Description

note google_default_credentials blocking concerns (#732)

## Testing

Further stage load testing

## Issue(s)

Closes #731
Issue #732

### DIFF
--- a/src/db/spanner/manager.rs
+++ b/src/db/spanner/manager.rs
@@ -72,18 +72,23 @@ impl<T: std::marker::Send + std::marker::Sync + 'static> ManageConnection
     type Error = grpcio::Error;
 
     async fn connect(&self) -> Result<Self::Connection, Self::Error> {
-        // Requires GOOGLE_APPLICATION_CREDENTIALS=/path/to/service-account.json
-        let creds = ChannelCredentials::google_default_credentials()?;
+        let chan = {
+            // Requires
+            // GOOGLE_APPLICATION_CREDENTIALS=/path/to/service-account.json
+            // XXX: issue732: Could google_default_credentials (or
+            // ChannelBuilder::secure_connect) block?!
+            let creds = ChannelCredentials::google_default_credentials()?;
 
-        // Create a Spanner client.
-        let chan = ChannelBuilder::new(self.env.clone())
-            .max_send_message_len(100 << 20)
-            .max_receive_message_len(100 << 20)
-            .secure_connect(SPANNER_ADDRESS, creds);
+            // Create a Spanner client.
+            ChannelBuilder::new(self.env.clone())
+                .max_send_message_len(100 << 20)
+                .max_receive_message_len(100 << 20)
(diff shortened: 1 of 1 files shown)


src/db/spanner/manager.rs
```diff
@@ -72,18 +72,23 @@ impl<T: std::marker::Send + std::marker::Sync + 'static> ManageConnection
...
-        let chan = ChannelBuilder::new(self.env.clone())
-            .max_send_message_len(100 << 20)
-            .max_receive_message_len(100 << 20)
-            .secure_connect(SPANNER_ADDRESS, creds);
+        let chan = {
+            // Requires
+            // GOOGLE_APPLICATION_CREDENTIALS=/path/to/service-account.json
+            // XXX: issue732: Could google_default_credentials (or
+            // ChannelBuilder::secure_connect) block?!
+            let creds = ChannelCredentials::google_default_credentials()?;
```
pjenvey, line 80: Note the added nesting here is to drop `ChannelCredentials` (which is not `Send`) before the await call. Otherwise we end up with the issue described in detail in this [blog post](https://blog.rust-lang.org/inside-rust/2019/10/11/AsyncAwait-Not-Send-Error-Improvements.html).

```
error: future cannot be sent between threads safely
  --> src/db/spanner/manager.rs:74:70
   |
74 |       async fn connect(&self) -> Result<Self::Connection, Self::Error> {
   |  ______________________________________________________________________^
75 | |         // Requires GOOGLE_APPLICATION_CREDENTIALS=/path/to/service-account.json
76 | |         // XXX: async?
77 | |         let creds = ChannelCredentials::google_default_credentials()?;
...  |
94 | |         })
95 | |     }
   | |_____^ future returned by `__connect` is not `Send`
   |
   = help: within `impl core::future::future::Future`, the trait `std::marker::Send` is not implemented for `*mut grpcio_sys::bindings::grpc_channel_credentials`
note: future is not `Send` as this value is used across an await
  --> src/db/spanner/manager.rs:88:23
   |
77 |         let creds = ChannelCredentials::google_default_credentials()?;
   |             ----- has type `grpcio::security::credentials::ChannelCredentials` which is not `Send`
...
88 |         let session = create_session(&client, &self.database_name).await?;
   |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ await occurs here, with `creds` maybe used later
...
95 |     }
   |     - `creds` is later dropped here
   = note: required for the cast to the object type `dyn core::future::future::Future<Output = std::result::Result<db::spanner::manager::SpannerSession, grpcio::error::Error>> + std::marker::Send`

error: aborting due to previous error
```

PR #744, opened 2020-07-24
bug: normalize id elements to remove potential wrap characters

### LINKED ISSUE #680: unquote params extracted from paths
The `extractors` module produces `BsoParam`, `CollectionParam` and `HawkIdentifier` from 
extracted bits of the request path.

Our `<..>_from_path` methods aren't unquoting path segments. E.g. a bso_id of "`{password-id}`" comes in as `PUT /1.5/uid/storage/passwords/%7Bpassword-id%7D`: we're incorrectly considering this a `bso_id` of "`%7Bpassword-id%7D`" instead.

This likely only affects bso_ids. collection_ids are restricted to `[a-zA-Z0-9._-]{1,32}` and uids to `[0-9]{1,10}` (and we ensure uids match the hawk payload's, so there's no funny business there anyway). Whereas bso_ids allows any printable ascii character.

We've likely written garbage bso_ids via PUT. Should we attempt to fix the garbage bso_ids in the db? It would require extensive scanning of our request logs to do correctly.

DELETEs against migrated data would unexpectedly 404, causing:

https://bugzilla.mozilla.org/show_bug.cgi?id=1640123#c2

jrconlin:
Closes: #680

## Description

Remove potential "{}" wrapper from bsoids and collection ids. (See original issue for details)

## Testing

Test included. 

Try sending in a sync request with a bso_id wrapped and URLencoded. 

## Issue(s)

Closes #680.


jrconlin:
Note: this took longer because I wasn't sure if we needed to use serde's `with_deserializer`. Turns out, we don't.

At least now I have a file locally that does it. so, yay?

Mcat12:
Looks like the current diff contains a bunch of changes to `mozilla-rust-sdk`, is that expected?

jrconlin:
Must be legacy from when I packed up the version for google.

This will bring us up to date with whatever they publish out, so no no harm in it.

Mcat12:
Can you move those changes to a separate PR?

jrconlin:
Possibly easier to move the wrap characters to a different PR, but sure,
let me see what I can do.

jrconlin:
Closing in favor of #748

PR #747, opened 2020-07-28
chore: Update vendored SDK to use protobuf 2.16.2

jrconlin:
## Description

Update vendored mozilla-sdk to match hand-off. 

## Testing

N/A

## Issue(s)

chore


vendor/mozilla-rust-sdk/Cargo.toml
```diff
@@ -3,3 +3,5 @@ members = [
   "googleapis",
   "googleapis-raw",
 ]
+
+# ci kick -- can be deleted
```
Mcat12, line 7: There should be a restart build button on Travis and Circle.

jrconlin: There is, but for some stupid reason, I don't believe that Circle got the webhook notice that anything was pending. I looked on the circle task list and there was nothing. 

Fun of complex, interwoven systems. Lots of room for stuff to break in fun ways.

Mcat12: There is also a way to tell GitHub to resend the webhook, if you have access to the repo settings.

PR #748, opened 2020-07-28
bug: normalize id elements to remove potential wrap characters

### LINKED ISSUE #680: unquote params extracted from paths

jrconlin:
## Description:

Remove potential "{}" wrapper from bsoids and collection ids. (See original issue for details)

Blocks on #747

## Testing:
Test included.

Try sending in a sync request with a bso_id wrapped and URLencoded.

## Issues:

Closes #680


Mcat12:
Do these changes depend on `mozilla-rust-sdk`? It looks like this branch could be rebased against master and merged before the other PR.

src/error.rs
```diff
@@ -74,6 +74,9 @@ pub enum ApiErrorKind {
 
     #[fail(display = "{}", _0)]
     Validation(#[cause] ValidationError),
+
+    #[fail(display = "Invalid submission: {}", _0)]
+    Invalid(String),
```
Mcat12, line 79: I imagine there's multiple ways for a request to be invalid. Maybe renaming this to `InvalidSubmission` would make sense?

jrconlin:
This is probably just an aspect of my local build. Right now, I can't build or test this PR without #747 without doing a good deal of work to fish out the various interdependencies.

src/web/extractors.rs
```diff
@@ -2082,6 +2112,37 @@ mod tests {
...
+    #[test]
+    fn test_quoted_id() {
+        let payload = HawkPayload::test_default(*USER_ID);
+        lazy_static! {
+            // Some libraries encode the bso id as "{guid}" which, is technically allowed
+            // since the spec says that bso ids could be any character. However, this can
+            // screw some things up, and while we've not seen occurances of this in the
+            // wild, we're going to transcode these to "normal" GUIDs.
+            static ref ALTERED_ID: String = format!("\"{{{}}}\"", *USER_ID);
+        }
```
Mcat12, line 2124: Why is this a static item? It looks like it can be changed to:
```suggestion
        // Some libraries encode the bso id as "{guid}" which, is technically allowed
        // since the spec says that bso ids could be any character. However, this can
        // screw some things up, and while we've not seen occurances of this in the
        // wild, we're going to transcode these to "normal" GUIDs.
        let altered_id = format!("\"{{{}}}\"", *USER_ID);
```

jrconlin: I get a lifetime error when building the TestRequest later (see line 2134) `.param()` requires both key and value to be `\'static`. 

Since this is a test, and USER_ID is already a const, a `lazy_static` is a fairly simple fix.

Mcat12: Hm, that is an odd requirement, but now the static makes sense.

Mcat12: Alternatively, you may be able to include it in the URI instead.

src/error.rs
jrconlin: Even better! It's unused. Dropping it.

src/web/extractors.rs
```diff
@@ -66,6 +66,15 @@ pub struct UidParam {
     uid: u64,
 }
 
+pub fn clean_element(v: &str) -> Result<String, urlencoding::FromUrlEncodingError> {
+    // Some systems are wrapping elements with "{}". While not strictly prohibited,
+    // and there are no cases in the wild , it  may cause some problems for data
+    // storage and retrival. Normalize these by removing these charcters.
+    let decoded = urlencoding::decode(v)?;
+    let fixed = decoded.replace("\"", "").replace("{", "").replace("}", "");
```
pjenvey, line 74: Why the replace? I don't think we want to mess with the actual value that was specified -- that's the original problem here.

E.g. though `PUT` gets `bso_id` from the path (this would convert e.g. `{password-id}` to `password-id` , a `POST` doesn't, getting it from the payload: updating ```'id': '{password-id}'``` would touch a different `bso_id` of `{password-id}`

src/web/extractors.rs
```diff
@@ -530,14 +539,22 @@ impl CollectionParam {
...
-            let sv = String::from_str(v).map_err(|_e| {
+            let mut sv = String::from_str(v).map_err(|_e| {
                 ValidationErrorKind::FromDetails(
                     "Missing Collection".to_owned(),
                     RequestErrorLocation::Path,
                     Some("collection".to_owned()),
                     Some(tags.clone()),
                 )
             })?;
+            sv = clean_element(&sv).map_err(|_e| {
```
pjenvey, line 550: It occurs to me now that we can probably now kill significant `extrude` paths that we made to work against `ServiceRequest` now that we've removed the 2 larger middlewares. E.g. all the `x_from_path` methods could work on the regular actix-web `HttpRequest` again, referencing its `match_info()` call to grab these values from the path (and actix would urldecode them for us).

This is a pretty quick fix so that's probably a separate issue 😄

src/web/extractors.rs
jrconlin: Ah, I thought the goal was to do that, in which case I'd need to make sure POST does the same thing. If we're just going to flat out reject it, then yeah, returning an error makes more sense.

src/web/extractors.rs
jrconlin: Huh, yeah, you're right. 

I think it should be a different issue as well.

jrconlin:
So, reworked things to urldecode the bso_ids. 

Granted, now part of me is wondering if we should filter on those to prevent potential exploits by including high values.

src/web/extractors.rs
```diff
@@ -66,6 +66,24 @@ pub struct UidParam {
     uid: u64,
 }
 
+fn clean_entry(s: &str) -> Result<String, ApiError> {
+    println!("### GOT: {:?}", s);
```
Mcat12, line 70: I assume these printlns are for debugging and will be deleted before the PR merges. If not, they should be converted to trace statements.

src/web/extractors.rs
Mcat12: Now that the URI contains this data, we don't need to call `param` and therefore don't need to use `lazy_static`.

src/web/extractors.rs
```diff
@@ -2082,6 +2143,40 @@ mod tests {
...
+        );
+        let header = create_valid_hawk_header(&payload, &state, "GET", &uri, TEST_HOST, TEST_PORT);
+        let req = TestRequest::with_uri(&uri)
+            .data(state)
+            .header("authorization", header)
+            .header("accept", "application/json,text/plain:q=0.5")
+            .method(Method::GET)
+            .param("uid", &USER_ID_STR)
+            .param("collection", "tabs")
+            .param("bso", &ALTERED_BSO)
```
Mcat12, line 2170: These params are already set in the URI, so they do not need to be set here. Because the `param` calls are not needed, the lazy static for `ALTERNED_BSO` is not needed either (replace with a normal `format` call).

PR #749, opened 2020-07-29
feat: add metric for db conflicts

### LINKED ISSUE #742: [spanner] Emit metrics for Spanner ConflictErrors
Followup to #614 -- we're not quite emitting metrics for every ConflictError produced by Spanner.

We only emit a metric for the 503 produced by `lock_for_write` which we manually generate when the locking condition fails. Looking at prod metrics, those are very rarely emitted.

This metric will let us clearly know when Spanner is responsible for producing spikes of 503s. I believe the nginix frontend converts almost every 5xx into a 503 (for the sake of the client), so metrics from the load balancer won't tell us specifically what was caused by Spanner.

It might make sense to emit this metric from the SentryMiddleware (or some middleware).

jrconlin:
Closes: 742

## Description

This will log all `is_conflict()` errors as a `db.conflict` metric from inside the sentry logging middleware.

NOTE: due to local compile requirements, sentry has been updated to 0.19, which required changes in the initialization. 
Please pay close attention to those as I believe they are correct, but appreciate better eyes.

## Testing

How should reviewers test?

## Issue(s)

Closes #742.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -13,6 +13,7 @@ use actix_web::{
     error::ResponseError,
     http::StatusCode,
     middleware::errhandlers::ErrorHandlerResponse,
+    web::Data,
     HttpResponse, Result,
 };
 use failure::{Backtrace, Context, Fail};
@@ -22,6 +23,8 @@ use serde::{
 };
 
 use crate::db::error::{DbError, DbErrorKind};
+use crate::server::metrics::Metrics;
+use crate::server::ServerState;
 use crate::web::error::{HawkError, ValidationError, ValidationErrorKind};
 use crate::web::extractors::RequestErrorLocation;
 
@@ -125,6 +128,12 @@ impl ApiError {
             _ => (),
         }
         true
+    }
+
+    pub fn on_response(&self, state: &Data<ServerState>) {
+        if self.is_conflict() {
+            Metrics::from(state.as_ref()).incr("storage.confict")
+        }
     }
 
     fn weave_error_code(&self) -> WeaveError {
--- a/src/main.rs
+++ b/src/main.rs
@@ -34,20 +34,26 @@ async fn main() -> Result<(), Box<dyn Error>> {
     // Set SENTRY_DSN environment variable to enable Sentry.
     // Avoid its default reqwest transport for now due to issues w/
     // likely grpcio's boringssl
+    // Commented out with sentry 0.19
+    // looks like this may no longer be a problem?
+    /*
     let curl_transport_factory = |options: &sentry::ClientOptions| {
         // Note: set options.debug = true when diagnosing sentry issues.
         Box::new(sentry::transports::CurlHttpTransport::new(&options))
             as Box<dyn sentry::internals::Transport>
     };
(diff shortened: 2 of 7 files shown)


src/main.rs
```diff
@@ -34,20 +34,26 @@ async fn main() -> Result<(), Box<dyn Error>> {
     // Set SENTRY_DSN environment variable to enable Sentry.
     // Avoid its default reqwest transport for now due to issues w/
     // likely grpcio's boringssl
+    // Commented out with sentry 0.19
```
Mcat12, line 37: By the way, we might be impacted here by the bug I found with sentry 0.19:
https://github.com/getsentry/sentry-rust/issues/237

TL;DR workaround is to either use the `debug-logs` feature or call `Hub::with` on the main thread with a no-op before doing any sentry things.

src/web/middleware/sentry.rs
```diff
@@ -147,6 +149,11 @@ where
                 Some(e) => {
                     let apie: Option<&ApiError> = e.as_error();
                     if let Some(apie) = apie {
+                        if apie.is_conflict() {
+                            if let Some(state) = sresp.request().app_data::<Data<ServerState>>() {
+                                Metrics::from(state.as_ref()).incr("db.conflict");
+                            };
+                        }
```
Mcat12, line 156: I imagine we will have other errors which we want to always increment a metric for. We might want to generalize this so that this hook is implemented on the error. We could make an `on_response(&ServerState)` hook for `ApiError` and then implement this logic there instead of directly in the sentry middleware.

jrconlin: not a bad idea. thanks!

pjenvey:
I tested the panic handler -- [it still works](https://sentry.prod.mozaws.net/operations/syncstorage-local/issues/9285317)

src/web/middleware/sentry.rs
```diff
@@ -147,6 +149,11 @@ where
                 Some(e) => {
                     let apie: Option<&ApiError> = e.as_error();
                     if let Some(apie) = apie {
+                        if apie.is_conflict() {
+                            if let Some(state) = sresp.request().app_data::<Data<ServerState>>() {
+                                Metrics::from(state.as_ref()).incr("db.conflict");
```
pjenvey, line 154: Oh one nit: could we call this `"storage.conflict"`, everything else uses the storage prefix

jrconlin: yeah, just copied what was there, but no reason not to get a jump on normalizing this crap.

src/error.rs
```diff
@@ -127,6 +130,12 @@ impl ApiError {
         true
     }
 
+    pub fn on_response(&self, state: &Data<ServerState>) {
```
Mcat12, line 133: A post-merge nit: This should be `&ServerState` because `Data<T>` is an actix-specific type just used for extracting stuff from app data, and dereferences to `&T`.

jrconlin: sigh. yeah, good point. I'll do a fixup PR on this monday.

PR #751, opened 2020-07-30
Feat/746.1

### LINKED ISSUE #746: Set and respect storage quota per user collection.
We need to set a quota server side (2GB) and return a error to clients that will cause them to fail to sync (and log details) in this instance. This is to support spanner limitations.

Quota should be configurable server side to avoid need to do a train release to make changes there. Ideally it's some config var that will also prevent us from needing to do a server side release as well.

We also need to add tracking for this on the server side so we can see how many users this is affecting.

### Related docs:
* [Sync Quota Detection](https://docs.google.com/document/d/1R3I06Os0QJdoDY1VnO-Pr0rBlZfVLh1K7mVQS0Ccdqs/edit#heading=h.h2lsi0fznyih)
* [Sync users > 2GB mitigation](https://docs.google.com/document/d/1-vE423-65KlnAWWDH5yqhVE-Tr0eV-iVh_qrsJmAyRE/edit)
* [Sync Tombstone Learnings](https://docs.google.com/document/d/1MIl6aZXt4wMy6Tf6ZLSRShcPrk9v0DOgt6c1IHrs1U4/edit#)

jrconlin:
Closes #746

## Description

This PR defines a new config option `limits.debug_client`. this is a
comma separated list of UIDs which will receive an 'over_quota` error
whenever they attempt to put a BSO.

This is not meant for production.

e.g.
```
limits.debug_client=1,123
```

## Testing
Add a config option for:

```
limits.debug_client=1,123
```

You can then use the put.sh script in `/tools/examples` and verify that it returns a 400 / 17 error.

## Issue(s)

Closes #746

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2315,39 +2315,94 @@ checksum = "388a1df253eca08550bef6c72392cfe7c30914bf41df5269b68cbd6ff8f570a3"
 
 [[package]]
 name = "sentry"
-version = "0.18.1"
+version = "0.19.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8b01b723fc1b0a0f9394ca1a8451daec6e20206d47f96c3dceea7fd11ec9eec0"
+checksum = "9bf0d1227e6b2cf0a4787c6ba7deed5156ab19fba88e00ffc009bfca8f84812b"
+dependencies = [
+ "curl",
+ "httpdate",
+ "log",
+ "reqwest",
+ "sentry-backtrace",
+ "sentry-contexts",
+ "sentry-core",
+ "sentry-failure",
+ "sentry-panic",
+ "serde_json",
+]
+
+[[package]]
+name = "sentry-backtrace"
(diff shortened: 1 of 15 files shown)


pjenvey:
We'd need more than just PUT. The Python version has `check_storage_quota` in the default view decorators, so it's checking on almost every call.

Maybe we can just get away with PUT/POST but I'm not sure how that might affect the client, I wonder if it could get into an odd state if we allowed some calls but not others. You'd think it'd allow DELETEs but even that has the check..

jrconlin:
I'm hoping(?) that since this is basically a hack to check client behaviors, we don't have to be super rigorous about things. 

This now works on anything that wants to extract a BsoBody from a request. That should cover most cases where we're trying to write a BSO. That should give a fairly reasonable approximation of a working system, right? (Since the other calls wouldn't be posting, only getting stuff.)

pjenvey:
POST collection uses `BatchBsoBody` instead of `BsoBody` so we really need it there too

PR #754, opened 2020-07-31
dereference Data<> before calling `on_response()`

jrconlin:
## Description

* Set grpcio feature `"openssl"` on ubuntu platforms to address seg faulting on startup due to combination of openssl and boringssl.

## Testing

N/A

## Issue(s)

Issue #742

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,6 +67,9 @@ validator = "0.10"
 validator_derive = "0.10"
 woothee = "0.11"
 
+[target.'cfg(target_vendor="ubuntu")'.dependencies]
+grpcio = { version="0.6", features=["openssl"] }
+
 [dev-dependencies]
 tokio = { version = "0.2", features = ["macros"] }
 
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -34,7 +34,7 @@ env_logger = "0.7.1"
 failure = "0.1.8"
 futures = { version = "0.3", features = ["compat"] }
 googleapis-raw = { version = "0", path = "vendor/mozilla-rust-sdk/googleapis-raw" }
-grpcio = { version = "0.6.0" }
+grpcio = { version = "0.6.0"}
```
Mcat12, line 37: nit: formatting (and below)

src/web/extractors.rs
```diff
@@ -227,6 +227,34 @@ impl FromRequest for BsoBodies {
                 ));
             }
         };
+
+        // ### debug_client
+        if let Some(uids) = &state.limits.debug_client {
```
Mcat12, line 232: This got added in with the format commit, which is probably unexpected?

src/web/extractors.rs
```diff
@@ -228,6 +228,33 @@ impl FromRequest for BsoBodies {
             }
         };
 
+        // ### debug_client
+        if let Some(uids) = &state.limits.debug_client {
```
Mcat12: The "bad pasta" is back

PR #760, opened 2020-08-04
chore: tag 0.5.4 release

tublitzed:
## Description

Tagging release 0.5.4 ahead of the official Github tab.

## Testing

We've had a ton of changes since an official tag + clog change. The only difference here is the addition of https://github.com/mozilla-services/syncstorage-rs/pull/758 since 0.5.3.


tublitzed:
yup. no worries. I just called it out so it's clear why there's so many changes included in this patch release :)

PR #763, opened 2020-08-05
bug: set config env separator to double underscore.

### LINKED ISSUE #762: Sub-section environment configuration options not being read properly
From limited local testing:

~1) `Config` wants the prefix to match case, PREFIX should be UPPERCASE~
2) The default separator is ".", which can cause some shells to improperly recognize the env var.

(e.g. `SYNC_LIMITS.DEBUG_CLIENT=2 cargo run`) produces a shell error `SYNC_LIMITS.DEBUG_CLIENT=2: command not found`

Suggest that double underscore be used as the separator value.

jrconlin:
Closes #762

## Description
Passing the default separator of "." can cause problems with some shells. It's better to use something like a double underscrore, which is more widely accepted.

## Testing
run the application, specifying the `limits.max_post_records` set to some value, e.g.
`SYNC_LIMITS__MAX_POST_RECORDS=99 cargo run -- --config sync.ini`

You should now be able to see the "debug_client" value appear in the `/info/configuration` response:
```
> curl http://localhost:8000/1.5/1/info/configuration
{"max_post_bytes":2097152,"max_post_records":99,"max_record_payload_bytes":2097152,"max_request_bytes":2101248,"max_total_bytes":100000000,"max_total_records":10000,"debug_client":"1,123"}
```

## Issue(s)

Closes #762.

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -108,7 +108,12 @@ impl Settings {
         }
 
         // Merge the environment overrides
-        s.merge(Environment::with_prefix(PREFIX))?;
+        // While the prefix is currently case insensitive, it's traditional that
+        // environment vars be UPPERCASE, this ensures that will continue should
+        // Environment ever change their policy about case insensitivity.
+        // This will accept environment variables specified as
+        // `SYNC_FOO__BAR_VALUE="gorp"` as `foo.bar_value = "gorp"`
+        s.merge(Environment::with_prefix(&PREFIX.to_uppercase()).separator("__"))?;
 
         Ok(match s.try_into::<Self>() {
             Ok(s) => {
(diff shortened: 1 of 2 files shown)


jrconlin:
I'm not sure that's required because the actual fix is to allow for environment provided variables to alter settings that are contained in configuration subsets. I chose the `limits.max_post_records` because the actual usage case for this (`limits.debug_client`) is a temporary patch enabled for stage testing. 

Since the `/info/configuration` call returns the contents of the `limits` configuration subset, if the value matches what was provided from the environment variable (and not what was present as either the default, or what was provided in any configuration file) then we can rest assured that the value is being used appropriately in other places in the code. 

I will modify the testing steps to change this to match the `limits.debug_client` since that matches the more "real-world" usage case.

Issue #768, opened 2020-08-06
/info/ endpoints rely on implicit transactions


pjenvey:
These endpoints lack a collection so `transaction_http` doesn't call locking methods for them, leaving the implicit transaction code to create their transactions in a blocking manner.

We can't add a quick fix of begin(false).await? to their handlers to solve this as the precondition check's `extract_resource` is now called from inside `transaction_http`, which implicitly starts the transaction.

Probably the easiest alternative is to create a `read_transaction_http` and `write_transaction_http`. For the case of no collection these would call `begin(...)` w/ either false or true.

Somewhat related to #581

PR #777, opened 2020-08-11
fix: Avoid implicit transactions in DbTransactionPool

Mcat12:
## Description

 Always create a transaction when using `DbTransactionPool`. This avoids implicit transactions.

## Testing

Not sure how the issue is/was detected, so not sure on testing.

## Issue(s)

Potential fix for #768

### DIFF
--- a/src/db/transaction.rs
+++ b/src/db/transaction.rs
@@ -21,7 +21,6 @@ use std::future::Future;
 #[derive(Clone)]
 pub struct DbTransactionPool {
     pool: Box<dyn DbPool>,
-    lock_collection: Option<params::LockCollection>,
     is_read: bool,
     tags: Tags,
     user_id: HawkIdentifier,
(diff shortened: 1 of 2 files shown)


src/db/transaction.rs
```diff
@@ -51,7 +51,7 @@ impl DbTransactionPool {
         let result = match (self.lock_collection.clone(), self.is_read) {
             (Some(lc), true) => db.lock_for_read(lc).await,
             (Some(lc), false) => db.lock_for_write(lc).await,
-            _ => Ok(()),
+            (None, is_read) => db.begin(!is_read).await,
```
pjenvey, line 54: Nice, this fix is much simpler than I thought it'd be.

One issue though: we only correctly set `is_read` (in the extractor below) when a collection exists. The `/info` endpoints lack a collection and always get `is_read = true` which is good.

However the `delete_all` handler currently assumes nothing calls `begin` for it (which is no longer the case) so it calls `begin(true)` itself. We should fix the `is_read` value for it by always checking the http method whether a collection was found or not, then we can remove `delete_all`'s manual `begin(true)`

PR #780, opened 2020-08-12
Feat/0.4 w debug

### LINKED ISSUE #779: Prepare debug_client for production
Currently we can only test firefox ios against a production Sync environment, so we'll need to deploy the debug_client code there. This requires a couple of tasks:

- we should pre-parse the `debug_client` list into a hashset so there's 0% chance of disrupting prod. It can live on `ServerState`
- the 0.5.x branch isn't quite ready for prod yet, so let's cherry-pick the 2-3 commits for `debug_client` onto the 0.4 branch for a new release there

jrconlin:
## Description

* backport the `debug_client` feature to 0.4
* include latest rust clippy fixes
* update to latest protobuf/grpcio due to local build 

I've tried to break this PR into commits that address the above issues. The first updates the googleapis so that things will build locally, the second wraps in the `debug_client` fixes, clippy bits, etc. This should help avoid "Wall of Code" issues while reviewing.

I couldn't `git cherry-pick` because there were too many changes between things, so I had to hand roll the `debug_client` patch into 0.4 base.

## Testing

Should work just like the current stage version.

## Issue(s)

Closes #779


src/web/extractors.rs
```diff
@@ -224,6 +224,28 @@ impl FromRequest for BsoBodies {
...
             }
         };
+
+        // ### debug_client
+        if let Some(uids) = &state.limits.debug_clients {
+            for uid in uids {
+                debug!("### checking uaid: {:?}", &uid);
+                let tested = HawkIdentifier::uid_from_path(req.uri(), None).unwrap_or(0);
+                if uid == &tested {
+                    debug!("### returning quota exceeded.");
```
Mcat12, line 234: This doesn't use any of the benefits of a `HashSet`. This should be changed to call `uids.contains(tested)` instead of looping through all of the IDs in the set.

jrconlin: yep. crap.

pjenvey:
superseded by #781

PR #781, opened 2020-08-12
feat: Backport `debug_client` opt

### LINKED ISSUE #779: Prepare debug_client for production

pjenvey:
## Description

includes latest rust clippy fixes and a sentry bump related to build issues

Co-authored-by: JR Conlin <jrconlin@gmail.com>

## Testing

Should work just like the current prod version.

## Issue(s)

Closes #779

### DIFF
--- a/src/web/extractors.rs
+++ b/src/web/extractors.rs
@@ -224,6 +224,26 @@ impl FromRequest for BsoBodies {
                 ));
             }
         };
+
+        // ### debug_client
+        if let Some(uids) = &state.limits.debug_clients {
+            let tested =
+                HawkIdentifier::uid_from_path(req.uri(), Some(Tags::from_request_head(req.head())))
+                    .unwrap_or(0);
+            if uids.contains(&tested) {
+                error!("Returning over quota for {:?}", tested);
+                return Box::pin(future::err(
+                    ValidationErrorKind::FromDetails(
+                        "over-quota".to_owned(),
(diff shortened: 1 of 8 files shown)


src/web/extractors.rs
```diff
@@ -224,6 +224,26 @@ impl FromRequest for BsoBodies {
                 ));
             }
         };
+
+        // ### debug_client
+        if let Some(uids) = &state.limits.debug_clients {
```
Mcat12, line 229: This is duplicated below and should be moved to a function, assuming the code will exist for more than a trivial amount of time.

jrconlin: Yep. Fortunately, this hack is just to test to see if production mobile clients behave correctly when an OverQuota response is returned, and will be dropped shortly afterward.

It's the reason that the patch also has a really crappy "--debug_client" "--debug_clients" pair of options. That's horribly confusing and should never be allowed (except if it's part of some crappy hack to test clients that can't hit stage)

PR #792, opened 2020-08-18
feat: emit internal bb8 Pool errors to logs/sentry

### LINKED ISSUE #764: Don't urldecode bso_ids from JSON
Follow up to #744 (sorry I missed this before it was merged): we *don't* want to urldecode `bso_id`s pulled from JSON payloads, as there's no escaping done there so they should come in unmodified. 

We only want to urldecode values received from url path elements.

Another thought: I don't think we need the extra `is_ascii` check (and the associated `InvalidSubmission`) -- because all 3 values we pull from the path (`bso_id`, `collection`, and `uid`) do their own validation. The first 2 via a REGEX and uid via `u64::from_str`.

To further enforce validation, if anything, we could also:

- Make sure these results are always validated. Forcing all callers through `extrude` ensures so. We mistakenly left `BsoParam::bsoparam_from_path` as a pub function. If mistakenly called instead of `extrude` the caller could get an unvalidated `BsoParam`.
- [u64::from](https://doc.rust-lang.org/std/primitive.u64.html#method.from_str_radix) does recognize a leading '+', any source of funny business here with the auth token's contents (probably not)?

### LINKED ISSUE #785: Add a keepalive setting
Add a setting for [actix-web's HttpServer::keep_alive](https://docs.rs/actix-web/2.0.0/actix_web/struct.HttpServer.html#method.keep_alive).

The default of 5 maybe be a little low per our nginx keepalive settings

### LINKED ISSUE #786: Emit bb8 Pool errors to logs
r2d2 by default [logs internal "Pool" errors](https://docs.rs/r2d2/0.8.8/r2d2/struct.Builder.html#method.error_handler) whereas [bb8 does not](https://docs.rs/bb8/0.4.2/bb8/struct.Builder.html#method.error_sink).

So let's setup an ErrorSink that logs them to `error!`.

### LINKED ISSUE #787: Pass along mobile client info to sentry
It looks like somewhere along the way (possible not in this repo, but filing here for now) we're dropping the user agent details for iOS and Android clients before it gets to Sentry.

See [here](https://sentry.prod.mozaws.net/operations/syncstorage-prod/issues/8179022/events/50353610/) for an error that's from an iOS client (which I know because it was me) and [here](https://sentry.prod.mozaws.net/operations/syncstorage-prod/issues/8179022/events/50354172/) for Android Nightly (same). We should track down where we're losing the UA data, and make sure it gets passed along to Sentry to help with debugging possible mobile client related issues in the future.

pjenvey:
- add a keepalive setting
- fix: don't urldecode bso_ids from JSON
- pass the user-agent to sentry as an extra

Closes #786
Closes #785
Closes #764
Closes #787

## Description

Describe these changes.

## Testing

How should reviewers test?

## Issue(s)

Closes [link](link).

### DIFF
--- a/src/web/tags.rs
+++ b/src/web/tags.rs
@@ -57,6 +57,7 @@ impl Tags {
         // Return an Option<> type because the later consumers (ApiErrors) presume that
         // tags are optional and wrapped by an Option<> type.
         let mut tags = HashMap::new();
+        let mut extra = HashMap::new();
         if let Some(ua) = req_head.headers().get(USER_AGENT) {
             if let Ok(uas) = ua.to_str() {
                 let (ua_result, metrics_os, metrics_browser) = parse_user_agent(uas);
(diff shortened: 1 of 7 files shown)


src/web/tags.rs
```diff
@@ -65,14 +66,14 @@ impl Tags {
...
             }
         }
-        // `uri.path` causes too much cardinality for influx.
         tags.insert("uri.method".to_owned(), req_head.method.to_string());
-        Tags {
-            tags,
-            extra: HashMap::new(),
-        }
+        // `uri.path` causes too much cardinality for influx but keep it in
+        // extra for sentry
```
jrconlin, line 74: Kinda wonder if we should add a comment to the `Tags` struct to note that `extra` is ONLY used by sentry. That or change it to be something like `sentry_tags`

Issue #801, opened 2020-08-24
Add metric information around quotas


jrconlin:
Record:
* for batches
   * number of inserts, deletes (updates if possible?)
* add metrics around timer operations so that we know not only how long things take, but how often.

jrconlin:
closed via #806

Issue #797, opened 2020-08-21
Add metric when Quota is met or exceeded


jrconlin:
We should record a metric when a sync request meets or exceeds a quota limit. Metric should include a tag for the collection_id

jrconlin:
closed via #806

Issue #789, opened 2020-08-18
Modify user_collections db schema to include new quota columns


jrconlin:
Add the following columns to db schemas for 

+ [ ] Spanner
+ [ ] MySQL

`total_bytes, INT64, default value=0 -- total bytes per collection per user`
`count, INT64, default_value=0 -- total count of items in collection per user`

*NOTE* Spanner does not use "default_value". Code should check and presume `Null` == 0.

jrconlin:
closed via #806

PR #806, opened 2020-08-27
Add Quota Monitoring

### LINKED ISSUE #791: Perform Quota check before sync operations that may alter storage
Blocks on https://github.com/mozilla-services/syncstorage-rs/issues/790

When performing operations that may alter or potentially increase sync storage use, check the current quota limits against the meta values stored in `user_collections` database.

### LINKED ISSUE #793: remove testing "debug_client" feature
This was only needed to test if iOS devices handle over quota errors cleanly. This was accomplished. This function is no longer required for the server and should be removed.

jrconlin:
Closes #791, #793, #797, #789, #801

## Description

* removed `debug_client` code #793
* added metric info `storage.quota.*` around quota events #797, #801 
* modify spanner/mysql to include total_bytes & count for user_collections #789 
* changed `touch_collection` to `update_collection` since it's doing more that just touching.
* added setting for `max_quota_limit` (set to `0` to disable quotas) and `enable_quota` (defaults to `true`) #791

##  NOTE:

A schema change for Spanner will need to be performed before this version can to go Production.
```
ALTER TABLE user_collections ADD COLUMN total_bytes INT64;
ALTER TABLE user_collections ADD COLUMN count INT64;
```
A migration is already provided for MySQL.

## Testing

Tests included. TBD: trying to work out how to send a bookmark set with `SYNC_LIMITS__MAX_QUOTA_LIMIT=10` set.

## Issue(s)

Closes #791, #793, #797, #789, #801


jrconlin:
Realized I need to do running tallies of the quota during BSO loads rather than just a check at the start with the bso length.

jrconlin:
Having thought about this a bit more, I'm not sure that's a good idea.

To be "proper" we should keep a running tally of the data being loaded up and cut a user off once they cross the quota max. The problem is that this may mean only a partial batch is loaded. I'm not sure how all of the clients would deal with that. Instead, since we have the buffer, I think it's fine if we do a quota check at the start and then allow the client to load data, even if that means exceeding the quota by some amount. This will ensure that the client doesn't get a partial load success, and we don't have to do a lot of extra work to return which BSOs loads and which were denied.

src/db/spanner/models.rs
```diff
@@ -996,7 +1139,14 @@ impl SpannerDb {
         .params(sqlparams)
         .execute_dml_async(&self.conn)
         .await?;
-        self.touch_collection_async(&params.user_id, collection_id)
+        let mut tags = Tags::default();
+        tags.tags
+            .insert("collection_id".to_string(), collection_id.to_string());
+        self.metrics
+            .incr_with_tags("self.storage.delete_bsos", Some(tags));
+        self.calc_quota_usage_async(&params.user_id, collection_id)
```
pjenvey, line 1147: Not necessarily a request but I think preferably the tombstone collection wouldn't be involved in calculating a quota for itself, I think it's a bit confusing.

src/db/spanner/models.rs
```diff
@@ -869,13 +996,17 @@ impl SpannerDb {
...
+            self.metrics
+                .incr_with_tags("storage.spanner.delete_collection", Some(tags));
             self.erect_tombstone(&params.user_id).await
         } else {
             self.get_storage_timestamp(params.user_id).await
         }
     }
 
-    pub(super) async fn touch_collection_async(
+    pub(super) async fn update_collection_async(
```
pjenvey, line 1009: The writes always call touch_collection first to ensure a parent record exists in user_collections (because of interleaving). Switching touch_collection to calculating the quota means we're calculating it before writing new data.

pjenvey: I think your recent reply on the bug is in reference to this?

pjenvey:
> The problem is that this may mean only a partial batch is loaded.

If you mean actual "batches" here, they wouldn't hit the quota. Their temporary data lives in batch_bsos which isn't calculated. You will have to deal with them when they do a final commit

src/db/mysql/models.rs
```diff
@@ -47,6 +48,8 @@ pub const USER_ID: &str = "userid";
 pub const MODIFIED: &str = "modified";
 pub const EXPIRY: &str = "ttl";
 pub const LAST_MODIFIED: &str = "last_modified";
+pub const COUNT: &str = "count";
+pub const TOTAL_BYTES: &str = "total_bytes";
```
pjenvey, line 52: is "total_bytes" accurate? Since it's a STRING length

src/db/spanner/models.rs
jrconlin: Ah, thanks! I missed moving that metric call to `calc_quota_usage_async` which does the quota measurement. `update_collection_async` in spanner still just deals with modified only.

src/db/spanner/models.rs
jrconlin: I mostly didn't want multiple writes, if possible.

pjenvey:
We discussed this already but mentioning it here: I think merging this to master will make it difficult to cut any releases from there going forward (as this can't hit prod yet, it requires the schema changes). That might become an issue because I think we'll need some time to play around with this (load testing, etc) -- this is still in the investigatory phase.

Maybe we could keep this on a branch and tag from there? (that's might also be a pain)

Other concerns:

- This isn't the ongoing tally. I know some initial testing of reading all a user's bso sizes (basically the calc quota method) ran pretty quickly, but I'm concerned about much different results on a larger, more active db. Especially with users with larger sized collections.
- This really emphasizes the need for an exhaustive test plan for ensuring this performs well. Is the load tester a good enough case? (We need a good test plan even for the ongoing tallying version, anyway)

I'm fine with trying this method out first but I want to throw up the caution flag here. Ops signed off on the ongoing tally so I know they're going to voice similar concerns.

src/db/mysql/models.rs
jrconlin: Ah, thanks. It should use `TOTAL_BYTES()`

jrconlin:
I agree that this needs to be tested. The concern I have about an "ongoing tally" approach is mostly that it would be more error prone. We'd still have to do the full quota calculation for each record the first time we encountered it, then we'd need to predict things like the differences between records that are updates vs insertions. (e.g. if you update to a smaller record, that reduces the total record size, right?)

Totally fine sitting on this one for a bit.

pjenvey:
Agreed, neither method is straightforward. I have the exact same concerns for the ongoing's tallys initial calculation as this method.

src/db/tests/db.rs
```diff
@@ -471,6 +472,8 @@ async fn delete_collection() -> Result<()> {
             collection: coll.to_owned(),
         })
         .await?;
+    // ### returns a timestamp of 0 because the collection is gone.
```
pjenvey, line 475: is this comment outdated? this test looks like it would fail if it's still the case

pjenvey (changes requested):

config/local.example.toml
```diff
@@ -8,4 +8,9 @@ database_url = "mysql://sample_user:sample_password@localhost/syncstorage_rs"
 master_secret = "INSERT_SECRET_KEY_HERE"
 
 # removing this line will default to moz_json formatted logs (which is preferred for production envs)
-human_logs = 1 
+human_logs = 1
+
+# enable quota limits
+enable_quota = 1
```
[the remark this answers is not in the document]
tublitzed: do we have prior art for running tests for both sides of feature flags like this? If not, this feels like a topic that's worth further discussion.

src/db/spanner/models.rs
```diff
@@ -694,7 +707,7 @@ impl SpannerDb {
     ) -> Result<results::GetCollectionUsage> {
         let mut streaming = self
             .sql(
-                "SELECT collection_id, SUM(LENGTH(payload))
+                "SELECT collection_id, SUM(BYTE_LENGTH))
```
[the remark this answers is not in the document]
jrconlin: yeah, crap. s/ f'up.

src/db/spanner/models.rs
```diff
@@ -805,15 +941,8 @@ impl SpannerDb {
...
-        self.sql(
-            "INSERT INTO user_collections (fxa_uid, fxa_kid, collection_id, modified)
-             VALUES (@fxa_uid, @fxa_kid, @collection_id, @modified)",
-        )?
-        .params(params)
-        .param_types(types)
-        .execute_dml_async(&self.conn)
-        .await?;
+        // This is a write op, so update the user_collection quota counts for the user.
+        self.calc_quota_usage_async(&user_id, TOMBSTONE).await?;
```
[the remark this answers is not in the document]
jrconlin: Yeah, fair enough. Since it was a write I wanted to make sure we stayed consistent.

src/db/spanner/models.rs
```diff
@@ -1352,10 +1552,12 @@ impl SpannerDb {
                 let (columns, values) = bso_to_update_row(&user_id, collection_id, bso, timestamp)?;
                 load_size += values.compute_size() as usize;
                 updates.entry(columns).or_insert_with(Vec::new).push(values);
+                self.metrics.incr("storage.spanner.update");
```
pjenvey, line 1555: I don't think a counter for updates/inserts will tell us much -- I think what we want is a histogram metric with the whole counts, e.g. this request contained x inserts, and x updates.

These stats are much more important within batch update (plain POSTs are capped at 100 bsos). The counts there would be the results of the update/insert calls. The question we wanted answered with this was: on average, how much do bulk writes (batches) insert vs update

src/db/spanner/models.rs
```diff
@@ -805,15 +946,6 @@ impl SpannerDb {
         .param_types(types.clone())
         .execute_dml_async(&self.conn)
         .await?;
-
```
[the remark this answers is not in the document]
jrconlin: this is done in `touch_user_collections()`

jrconlin:
Meta issue: https://github.com/mozilla-services/services-engineering/issues/41

src/db/spanner/models.rs
```diff
@@ -783,6 +796,157 @@ impl SpannerDb {
...
+        &self,
+        params: params::GetQuotaUsage,
+    ) -> Result<results::GetQuotaUsage> {
+        let check_sql = if self.quota_enabled {
+            "SELECT IFNULL(total_bytes,0), IFNULL(count,0)
+            FROM user_collections
+           WHERE fxa_uid = @fxa_uid
+             AND fxa_kid = @fxa_kid
+             AND collection_id = @collection_id"
+        } else {
```
pjenvey, line 809: I think this clause is dead code at this point? We only call it when quota is enabled, even in the proposed (commented out) test. If so it could just `assert!(self.quota_enabled)`

pjenvey (commented):
Followup from yesterday: do we have a strategy for calculating quota around batch commit (or intentionally *not* doing it?).

It may possibly add additional mutations which is not great. It may be good enough to not calc quota around it, but I get the impression sync clients do the majority of their writes via batches (at least most engines). It might be worth checking in w/ the client team on that point.

src/db/spanner/models.rs
```diff
@@ -783,6 +796,152 @@ impl SpannerDb {
...
+                    AND collection_id = @collection_id";
+            } else {
+                set_sql = "UPDATE user_collections
+                SET modified = @modified
+                WHERE fxa_uid = @fxa_uid
+                AND fxa_kid = @fxa_kid
+                AND collection_id = @collection_id";
+            };
+        } else {
+            // we're explicitly setting a tombstone.
```
pjenvey, line 909: Couple edge cases here that I think are problematic:

- Given a user_collection that already exists but does not have any bsos, the SELECT would return no results (None), triggering an INSERT into user_collections. Which would error out w/ ALREADY_EXISTS
- Given a user_collection that doesn't already exist, it seems like this no longer inserts (touches) user_collections unless TOMBSTONE was specified

If that's the case I'm surprised no tests caught this -- might be worth adding some that would?

pjenvey (changes requested):

src/db/spanner/models.rs
jrconlin: Ok, refactored this function a fair bit. as part of it:
1) it no longer looks for explicit tombstones, since tombstones are just another collection type and are handled as such.
2) will ALWAYS have a `set_sql` so will ALWAYS perform some action.

I proved that tests are hitting all cases both with and without `SYNC_ENABLE_QUOTA`.

src/db/spanner/models.rs
```diff
@@ -783,6 +799,172 @@ impl SpannerDb {
...
+                AND collection_id = @collection_id"
+            }
+        } else {
+            // Otherwise, there are no BSOs that match, check to see if there are
+            // any collections.
+            let result = self
+                .sql(
+                    "SELECT 1 FROM user_collections
+                WHERE fxa_uid=@fxa_uid AND fxa_kid=@fxa_kid AND collection_id=@collection_id
+                LIMIT 1",
```
pjenvey, line 932: we're directly querying the primary key (fxa_uid/kid/collection_id) so the LIMIT 1's superfluous
```suggestion
                WHERE fxa_uid=@fxa_uid AND fxa_kid=@fxa_kid AND collection_id=@collection_id",
```

src/db/spanner/models.rs
```diff
@@ -783,6 +799,172 @@ impl SpannerDb {
...
+            .execute_async(&self.conn)?
+            .one_or_none()
+            .await?;
+        let set_sql = if let Some(result) = result {
+            // Update the user_collections table to reflect current numbers.
+            // If there are BSOs, there are user_collections (or else something
+            // really bad already happened.)
+            if self.quota_enabled {
+                let mut v = Value::new();
+                v.set_string_value(result[0].get_string_value().to_owned());
```
pjenvey, line 898: as_value can help here. you should be able to take the string from the result too:

```suggestion
                let mut v = as_value(result[0].take_string_value());
```

src/db/spanner/models.rs
```diff
@@ -783,6 +799,172 @@ impl SpannerDb {
...
+                    "INSERT INTO user_collections (fxa_uid, fxa_kid, collection_id, modified, total_bytes, count)
+                    VALUES (@fxa_uid, @fxa_kid, @collection_id, @modified, 0, 0)"
+                } else {
+                    "INSERT INTO user_collections (fxa_uid, fxa_kid, collection_id, modified)
+                    VALUES (@fxa_uid, @fxa_kid, @collection_id, @modified)"
+                }
+            } else {
+                // there are collections, best modify what's there.
+                // NOTE, tombstone is a single collection id, it would have been created above.
+                if self.quota_enabled {
```
pjenvey, line 951: I'll argue we don't need to set total_bytes/count here because it should already be accurate -- simplifying this into the same update whether or not quota's enabled

src/db/spanner/models.rs
pjenvey: Do we need any additional tests for these cases?

src/db/spanner/models.rs
jrconlin: Kinda depends on paranoia about whether or not things default to 0 correctly, I suppose. You could make the same argument about the `INSERT` call above it. I prefer to be explicit in these cases.

src/db/spanner/models.rs
jrconlin: Ah! `take_string_value()` that's why it wasn't resolving correctly for me.

src/db/spanner/models.rs
jrconlin: We shouldn't no, because the existing tests are already exercising the paths and checking the results.

src/db/spanner/models.rs
```diff
@@ -1341,6 +1574,8 @@ impl SpannerDb {
             let mut row = row?;
             existing.insert(row[0].take_string_value());
         }
+        self.check_quota(&user_id, &params.collection, collection_id)
```
pjenvey, line 1577: btw why do we `check_quota` here? seems a bit random of a location for it

pjenvey (approved):
just a couple last questions: did you see [this one](https://github.com/mozilla-services/syncstorage-rs/pull/806#pullrequestreview-491057391)?

src/db/spanner/models.rs
jrconlin: Ah, yeah, this was a mistake. Probably a bad paste.

jrconlin:
As for the batch commit strategy, that's one of the reasons that I took the non-running count approach for now. We do a calc at the end of the batch run which totals up the tallies. This may mean that a batch will put someone over quota, but since batches are reasonably small in comparison to the quota, overage shouldn't be a concern. That should also make sure the mutation counts are low.

src/db/params.rs
```diff
@@ -76,7 +77,7 @@ collection_data! {
         id: String,
     },
     AppendToBatch {
-        id: String,
+        id: results::CreateBatch,
```
jrconlin, line 80: really wanna refactor this name to "batch" or something.

src/db/tests/db.rs
```diff
@@ -632,14 +639,65 @@ async fn get_collection_usage() -> Result<()> {
...
+            &(quota.total_bytes as i64),
+            expected.get("bookmarks").unwrap()
+        );
+        assert_eq!(quota.count, 5); // 3 collections, 5 records
+    }
+    Ok(())
+}
+
+#[tokio::test]
+async fn test_quota() -> Result<()> {
```
pjenvey, line 665: looks like this test should skip like the quota tests in batch.rs do (it fails for me locally against the regular dev db)

src/db/results.rs
```diff
@@ -21,7 +21,21 @@ pub type DeleteBsos = SyncTimestamp;
 pub type DeleteBso = SyncTimestamp;
 pub type PutBso = SyncTimestamp;
 
-pub type CreateBatch = String;
+#[derive(Debug, Clone)]
+pub struct CreateBatch {
+    pub id: String,
+    pub size: Option<usize>,
```
pjenvey, line 27: nit: it seems like size isn't needed in this struct. Unless I'm mistaken it seems to only be shared between `append`/`create` and `do_append`, I think those methods could just pass the value along to `do_append`

src/db/results.rs
```diff
@@ -21,7 +21,21 @@ pub type DeleteBsos = SyncTimestamp;
 pub type DeleteBso = SyncTimestamp;
 pub type PutBso = SyncTimestamp;
 
-pub type CreateBatch = String;
+#[derive(Debug, Clone)]
```
pjenvey, line 24: nit: the defaults for both CreateBatch/GetQuotaUsage are their basic inner type's defaults, so you don't even have to implement it yourself (let derive Default do it)

```suggestion
#[derive(Debug, Default, Clone)]
```

src/db/results.rs
```diff
@@ -30,6 +44,21 @@ pub type CommitBatch = PostBsos;
 pub type ValidateBatchId = ();
 pub type Check = bool;
 
+#[derive(Debug)]
```
pjenvey, line 47:
```suggestion
#[derive(Debug, Default)]
```

pjenvey (dismissed):

src/db/tests/db.rs
jrconlin: ah, good idea!

src/db/results.rs
jrconlin: I suppose. Part of my decision to include it here was to not keep adding arguments.

src/db/tests/db.rs
```diff
@@ -632,14 +639,79 @@ async fn get_collection_usage() -> Result<()> {
...
+    assert!(result.is_err());
     Ok(())
 }
 
 #[tokio::test]
 async fn get_collection_counts() -> Result<()> {
-    let pool = db_pool().await?;
+    let settings = crate::settings::test_settings();
+
+    if !settings.enable_quota {
```
pjenvey, line 709: I don't think we want to skip this one, though, it's existed for a while now

jrconlin: ah, doh. sorry. got carried away.

pjenvey (approved):

PR #807, opened 2020-08-28
Make the rust mysql batch implementation match the python one.

### LINKED ISSUE #640: [mysql] Match the Python batch implementation
To support existing mysql databases, we should match its batch implementation.

fzzzy:
## Description

When the rust syncstorage server was prototyped, some of the mysql field names were changed unnecessarily. This would require those that were running their own syncstorage python mysql databases to make changes to the db for no good reason. Also, because the rust implementation is also semantically not the same as the python one, the server may not behave exactly the same.

This branch fixes both of those things.

This version implements the mysql upsert using diesel's raw sql escape hatch, because diesel isn't expressive enough to implement that query in a straightforward way, apparently.

## Testing

Tests should pass, both when testing with a mysql database and a spanner db url.

## Issue(s)

Closes #640.


src/db/mysql/batch.rs
```diff
@@ -34,92 +50,105 @@ pub fn create(db: &MysqlDb, params: params::CreateBatch) -> Result<results::Crea
...
             }
         })?;
-    Ok(encode_id(timestamp))
+
+    do_append(db, batch_id, params.user_id, collection_id, params.bsos)?;
+
+    Ok(encode_id(batch_id))
 }
 
 pub fn validate(db: &MysqlDb, params: params::ValidateBatch) -> Result<bool> {
```
jrconlin, line 59: I was going to ask if it made more sense to do a fetch of all the `batch_uploads.batch_id`s for a given user and then compare them against the incoming list vs. doing a ton of `select 1 where...` calls, but I see that this is a long standing thing.

src/db/params.rs
```diff
@@ -102,8 +102,8 @@ bso_data! {
 #[derive(Clone, Debug, Default, Queryable)]
 pub struct Batch {
     pub id: String,
-    pub bsos: String,
-    pub expiry: i64,
+    //    pub bsos: String,
```
jrconlin, line 105: remove?

src/db/mysql/batch.rs
pjenvey: this just validates the batch itself: so I wouldn't say there's a ton of them

src/db/mysql/batch.rs
```diff
@@ -3,28 +3,44 @@ use diesel::{
...
+    // Sync timestamps are in seconds and quantized to two decimal places, so
+    // when we convert one to a bigint in milliseconds, the final digit is
+    // always zero. But we want to use the lower digits of the batchid for
+    // sharding writes via (batchid % num_tables), and leaving it as zero would
+    // skew the sharding distribution.
+    //
+    // So we mix in the lowest digit of the uid to improve the distribution
+    // while still letting us treat these ids as millisecond timestamps.  It's
+    // yuck, but it works and it keeps the weirdness contained to this single
+    // line of code.
```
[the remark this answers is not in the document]
pjenvey: don't give me much credit I just copied it from the Python :grin:

jrconlin (approved):

PR #809, opened 2020-08-28
feat: Convert some of the validation storage errors into metrics

### LINKED ISSUE #808: Convert some validation/processing errors to metrics
Seeing an increase in the number of validation errors like https://mail.google.com/mail/u/0/#inbox/FMfcgxwJXfkFBVcwmnBcPSHwHlzzwHJF.

These can be turned into metrics so we don't flood ourselves.

jrconlin:
Closes #808

## Description

Added new field to Errors to indicate the metric name. If present (and in a non-reportable error) will not report a given error to Sentry, but will instead increment a metric associated with that metric name.

## Testing

When testing against stage or dev, notice the reduced number of reported error into sentry and the increased number of metric reports. See [spreadsheet](https://docs.google.com/spreadsheets/d/18RL-Brz-Q-jR-iClNPCwDMVPYS0P2CvkHSJ2WZK-ceA/edit#gid=419074517) under "Silence Errors" for new metric names and categories.

## Issue(s)

Closes #808.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -57,6 +57,7 @@ pub const RETRY_AFTER: u8 = 10;
 pub struct ApiError {
     inner: Context<ApiErrorKind>,
     status: StatusCode,
+    metric_label: Option<String>,
 }
 
 /// Top-level ErrorKind.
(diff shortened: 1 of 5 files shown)


jrconlin:
Duplicate of #810

PR #818, opened 2020-09-02
Fix #811 Document use_test_transactions and expose it using an environment variable

### LINKED ISSUE #811: Document use_test_transactions and expose it using an environment variable
use_test_transactions in support.rs is a really useful thing to know about. We should document it and expose it to an env var so it is easier to use.

fzzzy:
## Description

Normally, when running cargo test, the test transaction data is discarded. This makes debugging test state difficult. There is a use_test_transactions setting that can be overridden, but the tests then re-override your decision to force them on.

This branch exposes the variable using the SYNC_DATABASE_USE_TEST_TRANSACTIONS environment variable.

Also, it documents how to use this feature.

## Testing

All tests should pass.

Running `SYNC_DATABASE_USE_TEST_TRANSACTIONS=false cargo test commit` should leave data in the mysql database.

Running `cargo test commit` should not leave anything in the database, as usual.

## Issue(s)

Closes https://github.com/mozilla-services/syncstorage-rs/issues/811

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,23 @@ We use [env_logger](https://crates.io/crates/env_logger): set the `RUST_LOG` env
 
 `make test` - open the Makefile to adjust your `SYNC_DATABASE_URL` as needed.
 
+#### Debugging unit test state
+
+In some cases, it is useful to inspect the mysql state of a failed test. By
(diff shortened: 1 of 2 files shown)


tublitzed (dismissed):
👍

jrconlin (approved):

PR #819, opened 2020-09-08
Chore/update 202009

jrconlin:
## Description
Update dependencies to address cargo audit issues

## Testing

N/A

## Issue(s)

N/A

### DIFF
--- a/src/web/auth.rs
+++ b/src/web/auth.rs
@@ -11,7 +11,7 @@ use std::convert::TryInto;
 use chrono::offset::Utc;
 use hawk::{self, Header as HawkHeader, Key, RequestBuilder};
 use hkdf::Hkdf;
-use hmac::{Hmac, Mac};
+use hmac::{Hmac, Mac, NewMac};
 use serde::{Deserialize, Serialize};
 use sha2::Sha256;
 use time::Duration;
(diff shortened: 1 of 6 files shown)


jrconlin:
awesome: https://github.com/bodil/sized-chunks/issues/11

pravic:
> # skipping audit pending https://github.com/bodil/sized-chunks/issues/11

There's `cargo audit --ignore ADVISORY_ID` that might help.

src/web/auth.rs
```diff
@@ -206,8 +206,10 @@ pub fn hkdf_expand_32(info: &[u8], salt: Option<&[u8]>, key: &[u8]) -> ApiResult
 
 /// Helper function for [HMAC](https://tools.ietf.org/html/rfc2104) verification.
 fn verify_hmac(info: &[u8], key: &[u8], expected: &[u8]) -> ApiResult<()> {
-    let mut hmac: Hmac<Sha256> = Hmac::new_varkey(key)?;
-    hmac.input(info);
+    type HmacSha256 = Hmac<Sha256>;
+
+    let mut hmac: HmacSha256 = HmacSha256::new_varkey(key)?;
```
pjenvey, line 211: Let's kill the HmacSha256 type alias in these 3 places.

With us having to specify the type parameter at the `new_varkey` callsite it's superfluous to do it on the left side.

```suggestion
    let mut hmac = Hmac::<Sha256>::new_varkey(key)?;
```

pjenvey (changes requested):

src/web/auth.rs
pjenvey: ..and I just noticed the hmac docs' examples use the type alias. I really don't see a good reason why though 🤷‍♂️

jrconlin: Probably just legibility. I kinda liked the type for that reason, but yeah, better to be explicit.

pjenvey: Forgot the other 2 places

tublitzed:
@jrconlin do we have a ticket associated with this PR?

pjenvey (approved):

Issue #619, opened 2020-05-01
Handle AlreadyExists in bsos errors


pjenvey:
Separate from #618 (which is about `batch_bsos`, this is about `bsos`) and also Related to #613 & https://bugzilla.mozilla.org/show_bug.cgi?id=1623671:

We produce ALREADY_EXISTS errors on `bsos`:
`ApiError: A database error occurred: RpcFailure: 6-ALREADY_EXISTS Row...`

(From issue1623671):
```
Some <row> in table bsos already exists
```

We suspect these are due to a subtle clause in batch commit's `INSERT` & `UPDATE` operations: ` AND expiry > CURRENT_TIMESTAMP()`

https://github.com/mozilla-services/syncstorage-rs/blob/beddaf6/src/db/spanner/batch_commit_update.sql#L51

This only update bsos that have not expired. This makes sense: expired bsos should be deleted (eventually). Updating them would result in resurrecting potentially expired values from the (un)dead.

https://github.com/mozilla-services/syncstorage-rs/blob/beddaf6/src/db/spanner/batch_commit_insert.sql#L26

Insert works similarly: only `INSERT` rows that we didn't just update (the inverse). However, given a `bso_id` in a batch that matches an already existing in `bsos` `bso_id` that has expired, an AlreadyExists could occur. We'd be attempting to `INSERT` a new bso when one already exists (in the undead expired state).

We differ here from python syncstorage: it *does not* include this logic, it could potentially resurrect expired data.

Our `expiry` check is more correct but can produce more errors to the client. These errors would only be temporary until `purge_ttl` cleans up expired data (in likely less than 24 hours).

This should be a rare occurrence, issue1623671 counted only two occurrences of it > 1 month ago. Since then it's occurred much more often, likely due to purge_ttl being out of service (see #613 for digging these events out of sentry).

To handle this, let's:
- detect when this happens
- emit a metric for counting how often it does
- it's painfully obscure: document it well in the codebase where we detect it
- Additionally (and what probably deserves its own entirely different issue): address the fact that `post_bsos` doesn't work like this! :tired_face:

pjenvey:
The history collection seems much more susceptible to this issue. I don't think I've yet seen any instance of this other than w/ history (collection id = 4 in the error message).

pjenvey:
This was fixed in #632 

[...]

PR #824, opened 2020-09-17
feat: rework error logging/metric reporting; fix BSO batch updates for spanner

### LINKED ISSUE #174: Improve and standardize trace statements and logging
The app currently does not provide sufficient debugging statements or clarity around some error types. These are messages that do not need to be propagated to production logs, but are immensely valuable during debugging and diagnostics.

### LINKED ISSUE #618: [spanner] Handle AlreadyExists in batch_bsos errors
Related to #613 & https://bugzilla.mozilla.org/show_bug.cgi?id=1623671:

We produce an ALREADY_EXISTS errors on `batch_bsos`:
`ApiError: A database error occurred: RpcFailure: 6-ALREADY_EXISTS Row...`

(From issue1623671):
```
Some <row> in table batch_bsos already exists
```
I suspect this happens when appending a `bso_id` to a batch that already exists in the batch. We always insert new items into a batch. Python syncstorage instead does an insert_or_update.

We should double check this w/ a new test, likely in the `test_storage.py` e2e suite, so we can easily compare handling of this case to Python syncstorage.

jrconlin:
Closes #618

## Description

~We're reporting too many errors to sentry, (HAWK errors, DB conflicts, etc). Many of these can be switched to metrics.
This PR cleans up a bit of the error/metric reporting mechanism, pushing the self reporting decision to the original ErrorKind enums where possible.~

This PR may be a monument to sunk cost and misunderstanding.

It addresses some shortcomings with how errors were being handled and reported (because of an initial misunderstanding about an issue). To that end,it moves some of the logic regarding whether an error is displayed in sentry or a metric stored into the actual `Error` mods. 

To rectify the original misunderstanding, this PR also includes logic to perform updates to pending BSO batch updates, mostly for spanner, with some initial work done for MySQL.

May it be used by future generations to scare their kids into eating their vegetables.

#### For the metric/sentry reporting:
Testers can monitor stage and see a decrease in errors similar to https://sentry.prod.mozaws.net/operations/syncstorage-prod/issues/9481936/?query=is%3Aunresolved 

#### For the spanner batch update merge function:
Internal testing included.

## Issue(s)

Closes #174, #619, #618, #827


jrconlin:
Going to block this on https://github.com/mozilla-services/syncstorage-rs/pull/807, since that one introduces a few useful patterns for doing batch item checks.

jrconlin:
See also https://github.com/56quarters/cadence/issues/103 for testing

src/db/mysql/batch.rs
```diff
@@ -170,24 +172,69 @@ pub fn do_append(
...
+        #[sql_type = "Integer"]
+        batch: i32,
+        #[sql_type = "Integer"]
+        id: i32,
+    };
+
+    let mut existing = HashSet::new();
+    let mut inserts: Vec<_> = Vec::new();
+
+    for item in sql_query("SELECT batch, id FROM batch_upload_items where userid=@userid;")
```
pjenvey, line 198: The Python version constructs a fancy INSERT ON DUPLICATE UPDATE that does both inserts and updates in one go.

This version is probably good enough for all our needs, but we might wanna log an issue that there's room for improvement

pjenvey (commented):
(FYI I'm still look at spanner's commit)

src/db/mysql/batch.rs
jrconlin: Yep. Part of my approach here was to mirror the way that Spanner works. I'll agree it's less efficient, but it's a bit easier to keep in mind.

src/db/error.rs
```diff
@@ -65,6 +65,37 @@ impl DbError {
...
     }
+
+    pub fn is_reportable(&self) -> bool {
+        // Yeah, this could be an "if", but this gives easy extensibility.
+        match self.inner.get_context() {
+            DbErrorKind::SpannerGrpc(ref err) => match err {
+                grpcio::Error::RpcFailure(status) => match status.status {
+                    grpcio::RpcStatusCode::ALREADY_EXISTS => false,
+                    _ => true,
+                },
```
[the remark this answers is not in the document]
jrconlin: fair enough.

I'll comment out the block with a comment that it is an example of how to exclude a specific error, since it may not be obvious for others.

src/db/spanner/batch.rs
```diff
@@ -269,22 +253,74 @@ pub async fn do_append_async(
...
-            if payload != null_value() {
-                running_size += payload.get_string_value().len();
-            }
-            let ttl = bso
-                .ttl
-                .map(|ttl| as_value(ttl.to_string()))
-                .unwrap_or_else(null_value);
 
+    // problem: Append may try to insert a duplicate record into the batch_bsos table.
+    // this is because spanner doesn't do upserts. at all. Batch_bso is a temp table and
```
pjenvey, line 258: It does support a form of upsert but only via mutations (mentioned [here](https://github.com/mozilla-services/syncstorage-rs/issues/618#issuecomment-680227710))

src/db/spanner/batch.rs
```diff
@@ -335,23 +370,63 @@ pub async fn do_append_async(
...
+                                    sortindex, payload, ttl)
+            SELECT * FROM UNNEST(@values)",
+        )?
+        .params(sqlparams)
+        .param_types(sqlparam_types)
+        .execute_dml_async(&db.conn)
+        .await?;
+    }
+
+    // assuming that "update" is rarer than an insert, we can try using the standard API for that.
```
pjenvey, line 407: In the original issue I mentioned concern about the performance of issuing individual updates like this but you're probably right, updates in the case of append should be rare (AlreadyExists doesn't happen that often and I would guess there's very few UPDATE bsos when it does happen)

They seem like a possible client bug.

src/db/spanner/batch.rs
```diff
@@ -269,22 +253,74 @@ pub async fn do_append_async(
...
+        sortindex: Value,
+        ttl: Value,
+        bso_id: String,
+        payload: String,
+    };
+
+    //prefetch the existing batch_bsos for this user.
+    let mut existing = HashSet::new();
+    let mut existing_stream = db
+    .sql("SELECT collection_id, batch_id, batch_bso_id from batch_bsos where fxa_uid=@fxa_uid and fxa_kid=@fxa_kid;")?
```
pjenvey, line 280: I think specifying the bso_ids on hand here (like `post_bsos` does, i.e. `AND batch_bso_id IN UNNEST(@ids`) is worth the effort.

We'd specify a max number of 100 bso_ids (on hand) vs selecting all might potentially return 1666 bso_ids. That doesn't sound like the greatest payoff but isolating the SELECT to fewer rows might be better for the Spanner r/w transaction (the less rows/columns it has to lock the better)

src/db/spanner/batch.rs
```diff
@@ -269,22 +253,74 @@ pub async fn do_append_async(
...
+
+        let exist_idx = exist_idx(&collection_id.to_string(), &batch.id, &bso.id);
+
+        if existing.contains(&exist_idx) {
+            // need to update this record
+            update.push(UpdateRecord {
+                sortindex,
+                ttl,
+                bso_id: bso.id,
+                payload: payload.get_string_value().to_owned(),
```
pjenvey, line 320: You don't want to copy here, especially payload
```suggestion
                payload: payload.take_string_value(),
```

src/db/spanner/batch.rs
```diff
@@ -335,23 +370,63 @@ pub async fn do_append_async(
...
+        .param_types(sqlparam_types)
+        .execute_dml_async(&db.conn)
+        .await?;
+    }
+
+    // assuming that "update" is rarer than an insert, we can try using the standard API for that.
+    if !update.is_empty() {
+        for val in update {
+            db.sql(
+                "UPDATE batch_bsos SET sortindex=@sortindex, payload=@payload, ttl=@ttl
```
pjenvey, line 411: I think this works slightly different than the Python INSERT ON CONFLICT UPDATE version. That version only issues UPDATES for batch columns that are NOT NULL.

If fixing this be painful we could ping @rfk about the behavior first.

pjenvey (changes requested):

src/db/spanner/batch.rs
rfk: My gut reaction (IIUC) is that yeah, clients really shouldn't be submitting the same BSO more than once in a batch, and if we're seeing that in practice it's worth investigating as a client bug. I'd go so far as to suggest that we should update the spec to forbid it, unless we find a specific client behaviour that needs it.

src/db/spanner/batch.rs
```diff
@@ -335,23 +370,63 @@ pub async fn do_append_async(
...
+            SELECT * FROM UNNEST(@values)",
+        )?
+        .params(sqlparams)
+        .param_types(sqlparam_types)
+        .execute_dml_async(&db.conn)
+        .await?;
+    }
+
+    // assuming that "update" is rarer than an insert, we can try using the standard API for that.
+    if !update.is_empty() {
```
rfk, line 408: If possible can we please log something here to help us understand how often we do such updates in practice?

src/db/spanner/batch.rs
rfk: Hrm, yeah, I guess you can imagine a weird-but-technically-allowed client sending two partial updates in a batch, like `{"id": "foo", "payload": "bar"}` and then later `{"id": "foo", "ttl": 27}` in the same batch. The correct semantics would be to apply them as two partial updates, resulting in `{"id": "foo", "payload": "bar", "ttl": 27}`, but it would be even better if we could just convince ourselves that current clients don't do that, and then forbid it.

pjenvey (changes requested):
Also: I don't think this closes #174

tublitzed:
@jrconlin can you update the PR description here to reflect what is actually changing and how reviewers should test? This looks like quite a bit more than metric/reporting changes.

src/db/tests/batch.rs
```diff
@@ -242,3 +246,50 @@ async fn quota_test_append_batch() -> Result<()> {
...
+    ];
+
+    db.append_to_batch(ab(uid, coll, new_batch.clone(), updated_bsos))
+        .await?;
+
+    assert!(db
+        .get_batch(gb(uid, coll, new_batch.id.clone()))
+        .await?
+        .is_some());
+    // ideally, extract the updated batch_bso record out of SpannerDB and compare it against what we just
```
pjenvey, line 284: I'd modify / add to this test a couple things (I don't think it's *too* much extra code): 

call `append_to_batch` with updates to "b0" twice. Maybe ttl the first time, sortindex the second.

Then copy some of the code from something like `append_commit` above: call `commit_batch`, then `get_bso(.. "b0")`. Finally assert payload is still "payload 0", but ttl and sortindex were updated

src/db/tests/batch.rs
```diff
@@ -242,3 +246,50 @@ async fn quota_test_append_batch() -> Result<()> {
...
+        "b0",
+        Some("payload 0"),
+        Some(10),
+        Some(ttl + 10_000),
+    )];
+    let new_batch = db.create_batch(cb(uid, coll, first_bsos)).await?;
+
+    // update the single bso twice, once changing
+    let updated_bsos = vec![
+        postbso("b0", None, Some(10), None),
```
pjenvey, line 273: let's change the ttl here, to say `Some(20)` instead of the original value `Some(10)` (see below for the rest)

src/db/spanner/batch.rs
```diff
@@ -433,20 +434,48 @@ pub async fn do_append_async(
...
             .incr("storage.spanner.adding_updates_to_batch_bsos");
         for val in update {
-            db.sql(
-                "UPDATE batch_bsos SET sortindex=@sortindex, payload=@payload, ttl=@ttl
+            let mut fields = Vec::new();
+            let mut sort_index = val.sortindex.get_string_value().to_owned();
+            if !sort_index.is_empty() {
+                fields.push("sortindex");
+            } else {
+                // This is not written, but the sql builder requires a numeric string
```
pjenvey, line 442: What requires this value here (I don't understand about the sql builder requiring it). I suspect it's params and maybe param_types? We could build those dynamically (and or pop from param_types since it's already set up front)

src/db/tests/batch.rs
jrconlin: Crap, yeah, I had dropped the second call because of the mysql thing. Let me add it back in.

Sadly, looks like I'll also need to land #853 first because of protobuf fun.
Thankfully, that didn't impact any code.

src/db/spanner/batch.rs
jrconlin: As far as I can tell, before the SQL is actually processed, the query builder resolves the types of the arguments. The problem is that the NoneType resolves to "" which fails to resolve to a numeric. That's why I pass in the "0", which does.

If I was SUPER clever, I would probably just exclude the values from the params using the same crafting technique I'm doing to build the sql, but that feels like making things really complicated rather than just do this, particluarly with the way that params have to be added.

jrconlin:
As for this not addressing #174, I'll point out 
https://github.com/mozilla-services/syncstorage-rs/pull/824/files#diff-43d942e74dc90492bfd15b8a7e310d37R68-R78, 

https://github.com/mozilla-services/syncstorage-rs/pull/824/files#diff-2c5b3fa267b18e8f400e96cc130d0f70R31-R56,
 and
https://github.com/mozilla-services/syncstorage-rs/pull/824/files#diff-9a269209bed069ffb36d5565a91c9840R130-R138

which are easy to miss in the fog of all the other changes.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM rust:1.45-buster as builder
+FROM rust:1.46-buster as builder
```
tublitzed, line 1: this is an already huge PR. Do we need to add dep upgrades into it, or can that be done separately?

jrconlin: The protobuf update was required due to a local build issue. This was addressed in https://github.com/mozilla-services/syncstorage-rs/pull/853. The rust image update brings our image closer to the currently stable rust branch.

tublitzed: I get that the image update gets us closer to stable rust, but what I don't understand is why it's required as part of this work?

The protobuf update seems...completely unlreated? Unless I'm missing something obvious here.

tublitzed:
@jrconlin could you also update the title of this PR so it's easier to track what's happening once it merges? I think this is no longer just `feat: rework error logging/metric reporting.`? I know it's a bit of a nit, but stuff like this really helps a lot when grokking code changes after they merge.

Dockerfile
jrconlin: Basically, I can't develop, run or test on my machine unless I update protobuf. This is because protobuf is often a system level requirement and is not directly tied to a given project release. A single update of protobuf from anything will cause all versions of protobuf to update. It's possible to spin that out as [a dependency](https://github.com/mozilla-services/syncstorage-rs/pull/853), however it's often just more direct and quicker to deploy to roll it into an existing PR. 

As for the rust update, this makes sure that our current testing environment again matches the local development environment. This ensures that things like clippy, fmt and other tools operate the same and reduces the need for using CI as a live debugging environment to address issues that may only present themselves on one version.

pjenvey (approved):

Issue #120, opened 2019-03-22
Port test_overquota


pjenvey:
Port the equivalent of the check_storage_quota decorator: if a quota is set, ensure writes won't exceed it.

(This doesn't appear to have tests in test_storage.py)

pjenvey:
There is a test_overquota though it doesn't run by default. The test suite sets `self.distant` to True which causes the test to skip

pjenvey:
Since test_overquota doesn't run by default we should likely port it over to rust (likely as a test in server/test.rs) to replicate what it does.

fzzzy:
There's already a rust test that covers this, test_quota, that's landed on master.

pjenvey:
Reopening as test_quota only covers the db layer w/ a unit test.

The e2e test_overquota is worth porting as it covers write operation responses, e.g. that an `X-Weave-Quota-Remaining` header is returned along with a `quota-exceeded` status body.

I don't think we're doing either of these yet.

tublitzed:
@fzzzy - heads up I just slotted this back into the queue. Let's knock this out so we can wrap up the quota work.

fzzzy:
The rust server does not currently implement X-Weave-Quota-Remaining so we can't test it, but we also think the client never does anything with it. If the client never does anything with it, we may be safe to leave it out. Regardless, we should add a new bug for implementing it and testing it if we decide we do need it.

fzzzy:
There's nothing to do here. The rust server does not implement X-Weave-Quota-Remaining so we can't test it, and db/tests/db.rs test_quota already tests is_err in case of an overquota situation. All this test would cover is whether Actix can turn this Err into a 302.

However, we don't have the appropriate test harness on the http side to change the Settings object to lower the quota to test the 302 situation. We could spend a lot of time rewriting the http test harness such that the Settings object can be passed through, but it touches every http test.

Given that test_quota already tests that the overquota and the test harness for changing the Settings object already exists in the db tests, this test would add less than one line of additional test coverage.

This test would best be implemented in a Python 3 integration test.

fzzzy:
wip is here https://github.com/mozilla-services/syncstorage-rs/compare/overquota-test

pjenvey:
This can't really be a Python e2e test easily as it requires changing syncstorage-rs's quota settings.

I've made `init_app!`'s settings configurable now, so we should be able to finish this up easily. A couple things I noticed:

[...]

PR #837, opened 2020-09-30
bug: return correct error code and value for OverQuota users

### LINKED ISSUE #836: Return 403 for over quota users.
Return status code of 403 with content of 14 for users that are overquota.

jrconlin:
Closes #836

## Description
Return status 403 (weave error "14") for users that trigger the Quota limits.

## Testing

Start a server with 

```
SYNC_LIMITS__MAX_QUOTA_LIMIT=10 SYNC_ENABLE_QUOTA=1  ...
```
you can then use a script similar to the one in `tools/examples/put.bash` (mind that the value of `SYNC_MASTER_SECRET` matches whatever you're using for the sync server). Since that script attempts to put more than 10 bytes* it should fail with the correct statuscode and body.

* this script may need to be run a second time if the database is empty.

## Issue(s)

Closes #836 .

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -142,7 +142,6 @@ impl ApiError {
     }
 
     fn weave_error_code(&self) -> WeaveError {
-        dbg!("### self.kind", self.kind());
         match self.kind() {
             ApiErrorKind::Validation(ver) => match ver.kind() {
                 ValidationErrorKind::FromDetails(
--- a/src/error.rs
+++ b/src/error.rs
@@ -142,6 +142,7 @@ impl ApiError {
     }
 
     fn weave_error_code(&self) -> WeaveError {
+        dbg!("### self.kind", self.kind());
         match self.kind() {
             ApiErrorKind::Validation(ver) => match ver.kind() {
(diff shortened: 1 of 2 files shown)


pjenvey:
I think adding an additional integration test now would be worthwhile, see https://github.com/mozilla-services/syncstorage-rs/issues/120#issuecomment-701569093

jrconlin:
True. My thought was to keep that a separate issue and get this version on stage for QA testing tomorrow AM.

src/error.rs
```diff
@@ -142,6 +142,7 @@ impl ApiError {
     }
 
     fn weave_error_code(&self) -> WeaveError {
+        dbg!("### self.kind", self.kind());
```
pjenvey, line 145: nit: probably don't want to keep this

pjenvey (dismissed):

src/error.rs
jrconlin: crap, yeah. let me kill it quick.

pjenvey (approved):

PR #848, opened 2020-10-06
bug: Return FORBIDDEN if a user's batch is Over Quota

### LINKED ISSUE #418: Deprecate POST "failed" bsos
@mhammond outlines this here: https://github.com/mozilla-services/syncstorage-rs/issues/376#issuecomment-571851614

Individual bsos can fail during a multi-write (regular POST or a batch) when others succeed. This poses complications for Sync clients: they would prefer multi-writes either completely succeeded or completely failed.

In Python syncstorage and our Spanner backend the only bsos added to failed are those that failed validation. Our mysql backend also adds bsos that failed to write due to DbErrors (along with a TODO to remove this).

To deprecate this, we could stop adding bsos to the failed collection entirely and let validation (40x) or db errors (50x) encountered fail the entire operation.

jrconlin:
Issue #418

## Description

If a user is over quota during a batch operation, the server would return OK with a list of `failed` ids. The client would retry them, leading to potentially a lot of useless network traffic.

## Testing

Enable the QUOTA system.
Create a batch that contains more data than should be allowed by the QUOTA.
Confirm that the returned result is a `403 FORBIDDEN` status with the body of `14`

(Note, I used the spanner load test script for this.)

## Issue(s)

Issue #418


fzzzy (approved):
👍 Wonderful 🎉

src/web/handlers.rs
```diff
@@ -355,7 +355,15 @@ pub async fn post_collection_batch(
     match result {
         Ok(_) => success.extend(bso_ids),
         Err(e) if e.is_conflict() => return Err(e.into()),
-        Err(_) => failed.extend(bso_ids.into_iter().map(|id| (id, "db error".to_owned()))),
+        Err(apperr) => match apperr.kind() {
+            // If we're over quota, return immediately with a 403 to let the client know.
+            // Otherwise the client will simply keep retrying records.
+            ApiErrorKind::Db(dberr) => match dberr.kind() {
+                DbErrorKind::Quota => return Ok(apperr.into()),
```
fzzzy, line 362: lol @ "return Ok(apperr)"

Thanks actix-web!

fzzzy (approved):

src/web/handlers.rs
jrconlin: heh, not actix-web, that was me.Trying to make things a bit clearer.

src/web/handlers.rs
```diff
@@ -355,7 +355,16 @@ pub async fn post_collection_batch(
     match result {
         Ok(_) => success.extend(bso_ids),
         Err(e) if e.is_conflict() => return Err(e.into()),
-        Err(_) => failed.extend(bso_ids.into_iter().map(|id| (id, "db error".to_owned()))),
+        Err(apperr) => {
+            if let ApiErrorKind::Db(dberr) = apperr.kind() {
+                // If we're over quota, return immediately with a 403 to let the client know.
+                // Otherwise the client will simply keep retrying records.
+                if let DbErrorKind::Quota = dberr.kind() {
+                    return Ok(apperr.into());
```
pjenvey, line 363: Shouldn't this return Err instead of Ok? Though I'm not sure it makes a difference.

We might wanna make sure this is tested which can be added later during #120

pjenvey (changes requested):
I was hoping we could take the opportunity to fix #418 here but I'm fine with a quicker fix such as this. This does not fix #418 though, so let's log a "propagate quota errors in batch commit" issue for this PR (I really should have logged it for [my comment describing the issue](https://github.com/mozilla-services/services-engineering/issues/41#issuecomment-703010374) ) and fixup the commit message to point to it

jrconlin:
updated commit to reflect that this is an "issue" resolution to #418 rather than "closes"

pjenvey (approved):

PR #855, opened 2020-10-09
chore: Update circleci to use docker auth

### LINKED ISSUE #854: Modify CircleCI config to avoid rate limiting.
There's a [change](https://github.com/mozilla-services/Dockerflow/pull/58) in the DockerFlow spec to move the Dockerhub login further up the build config.  Dockerhub is enabling rate limiting early November, so this should be implemented before then to avoid running into build problems pulling containers or otherwise interacting with Dockerhub.

jrconlin:
## Description

Adding authorization per steps outlined in https://github.com/mozilla-services/Dockerflow/pull/58/files

## Testing

Should pass CircleCI.

## Issue(s)

Closes #854


Micheletto (approved):
From my understanding, this should work.  If you wanted to avoid multiple **docker login** commands boiler plate, you could do something like [this](https://github.com/mozilla-bteam/bmo/blob/master/.circleci/config.yml#L8-L17).  But, I would leave this as is.

jrconlin:
Oh, thanks! Didn't know about that. 

I'll definitely include that in the update for the other packages.

PR #862, opened 2020-10-16
feat: remove Tags handoffs

### LINKED ISSUE #403: Remove the Tags Error and extractor passthru
With the sentry middleware now gathering tags via the request/response extensions, we can remove the original API for this where Tags were passed around throughout extractors and ValidationErrors (to ultimately propagate them out to sentry). This should be a nice simplification/cleanup.

https://github.com/mozilla-services/syncstorage-rs/pull/372#discussion_r359140320

jrconlin:
Closes #403

## Description

This removes the "Tags" struct from being passed around in various
calls. Instead the `sentry` mod pulls the `Tags` from the `extensions`
associated with either the request or response.

## Testing

Inline testing should handle most of this.
In addition, there were no tags being registered outside of the ones generated from the request header in the `sentry` mod. 

Tags for metrics and errors should remain unchanged.

## Issue(s)

Closes #403

### DIFF
--- a/src/web/handlers.rs
+++ b/src/web/handlers.rs
@@ -538,22 +538,12 @@ pub async fn heartbeat(hb: HeartbeatRequest) -> Result<HttpResponse, Error> {
 // try returning an API error
 pub async fn test_error(
     _req: HttpRequest,
-    ter: TestErrorRequest,
+    _ter: TestErrorRequest,
 ) -> Result<HttpResponse, ApiError> {
     // generate an error for sentry.
 
-    /*  The various error log macros only can take a string.
-        Content of Tags struct can be logged as KV (key value) pairs after a `;`.
-        e.g.
-        ```
-        error!("Something Bad {:?}", err; wtags)
-        ```
-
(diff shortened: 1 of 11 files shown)


src/web/handlers.rs
```diff
@@ -538,21 +538,10 @@ pub async fn heartbeat(hb: HeartbeatRequest) -> Result<HttpResponse, Error> {
...
-    /*  The various error log macros only can take a string.
-        Content of Tags struct can be logged as KV (key value) pairs after a `;`.
-        e.g.
-        ```
-        error!("Something Bad {:?}", err; wtags)
-        ```
-
-        TODO: find some way to transform Tags into error::KV
-    */
-    error!("Test Error: {:?}", &ter.tags);
```
pjenvey, line 554: maybe worth keeping the error! statement? I think actix dumps the error out somewhere (but not to logs) and we always tend to log something to error during `/__error__`

pjenvey (approved):

pjenvey (approved):

PR #871, opened 2020-10-21
feat: Implement rudimentary tokenserver route in syncstorage-rs

### LINKED ISSUE #864: Implement mysql backend for tokenserver route
When we implement the tokenserver route in rust, we should support the same mysql backend schema as the old server. This will make migrations easier.

fzzzy:
## Description

Add the hacky initial tokenserver route. Uses an existing tokenserver mysql database.

## Testing

First, you need python3 and to run `pip3 install tokenlib`.

Need a tokenserver mysql database:

```
create table services (id int(11) auto_increment not null, service varchar(30), pattern varchar(128), primary key(id));
create table nodes (
  id bigint(20) auto_increment not null, service int(11) not null,
  node varchar(64) not null, available int(11) not null, current_load int(11) not null,
  capacity int(11) not null, downed int(11) not null, backoff int(11) not null,
  primary key(id), foreign key(service) references services(id)
);
create table users (
  uid bigint(20) auto_increment not null, service int(11) not null,
  email varchar(255) not null, generation bigint(20) not null,
  client_state varchar(32) not null, created_at bigint(20) not null,
  replaced_at bigint(20), nodeid bigint(20) not null,
  keys_changed_at bigint(20), primary key(uid),
  foreign key(nodeid) references nodes(id)
);
```

Run syncstorage-rs.

In firefox, create a new profile. In about:config, set identity.sync.tokenserver.uri to http://localhost:8000/1.0/sync/1.5

Log in to sync with your normal prod fxa credentials. It should be able to put your stuff in your localhost syncstorage-rs.

## Issue(s)

Fix #864


src/web/tokenserver.rs
```diff
@@ -1,26 +1,165 @@
...
     })
 }
 
-pub fn get_sync(_request: TokenServerRequest) -> Result<TokenServerResult, ApiError> {
-    // TODO Perform any blocking calls needed to respond to the tokenserver request.
-    Ok(TokenServerResult {})
+pub fn get_sync(auth: &BearerAuth) -> Result<TokenServerResult, ApiError> {
+    // the public rsa components come from
+    // https://oauth.accounts.firefox.com/v1/jwks
+    // TODO we should fetch it from an environment var instead of hardcoding it.
```
jrconlin, line 83: yes. this.

src/web/tokenserver.rs
```diff
@@ -1,26 +1,165 @@
...
+    let token_data = decode::<Claims>(
+        &auth.token(),
+        &DecodingKey::from_rsa_components("2lDphW0lNZ4w1m9CfmIhC1AxYG9iwihxBdQZo7_6e0TBAi8_TNaoHHI90G9n5d8BQQnNcF4j2vOs006zlXcqGrP27b49KkN3FmbcOMovvfesMseghaqXqqFLALL9us3Wstt_fV_qV7ceRcJq5Hd_Mq85qUgYSfb9qp0vyePb26KEGy4cwO7c9nCna1a_i5rzUEJu6bAtcLS5obSvmsOOpTLHXojKKOnC4LRC3osdR6AU6v3UObKgJlkk_-8LmPhQZqOXiI_TdBpNiw6G_-eishg8V_poPlAnLNd8mfZBam-_7CdUS4-YoOvJZfYjIoboOuVmUrBjogFyDo72EPTReQ", "AQAB"),
+        &Validation::new(Algorithm::RS256),
+    ).map_err(|ee| {
+        ApiError::from(ApiErrorKind::Internal(format!("Unable to decode token_data: {:}", ee)))
+    })?;
+    println!("token data! {:?}", token_data);
```
jrconlin, line 91: maybe make a `dbg!`?

src/web/tokenserver.rs
```diff
@@ -1,26 +1,165 @@
...
+
+    // TODO pull out of settings instead
+    let shared_secret = env::var("SYNC_MASTER_SECRET").expect("SYNC_MASTER_SECRET must be set");
+    let database_url =
+        env::var("TOKENSERVER_DATABASE_URL").expect("TOKENSERVER_DATABASE_URL must be set");
+
+    let connection = MysqlConnection::establish(&database_url)
+        .unwrap_or_else(|_| panic!("Error connecting to {}", database_url));
+    let user_record = diesel::sql_query(
+        "select users.uid, services.pattern, users.email, users.generation, users.client_state, users.created_at, users.replaced_at, nodes.node, users.keys_changed_at from users, services, nodes where users.email = ? and services.id = users.service and nodes.id = users.nodeid and nodes.service = services.id")
```
jrconlin, line 102: nit: Might want to format this a bit. I know pjenvey is an advocate of https://www.sqlstyle.guide/, and I'll admit it makes parsing what's going on here a bit easier.

src/web/tokenserver.rs
```diff
@@ -1,26 +1,165 @@
...
+        .bind::<Text, _>(email)
+        .load::<TokenserverUser>(&connection).unwrap();
+    println!("user record!! {:?}", user_record);
+    let (python_result, python_derived_result) = Python::with_gil(|py| {
+        let tokenlib = PyModule::from_code(
+            py,
+            r#"
+import tokenlib
+def make_token(plaintext, shared_secret):
```
jrconlin, line 111: nit: I think flake8 would complain about this code, but I'll leave it up to you if you want to style it.

src/web/tokenserver.rs
```diff
@@ -1,26 +1,165 @@
...
+                e.print_and_set_sys_last_vars(py);
+                return Err(e);
+            }
+            Ok(x) => x.extract::<String>().unwrap(),
+        };
+        //assert_eq!(result, false);
+        Ok((result, derived_result))
+    })
+    .unwrap();
+    println!("python result {:}", python_result);
```
jrconlin, line 156: again `dbg!`?

src/web/tokenserver.rs
```diff
@@ -1,26 +1,165 @@
...
+            ("fxa_kid", "asdf"), // userid component of authorization email
+            ("fxa_uid", "qwer"),
+            ("hashed_device_id", "..."),
+            ("hashed_fxa_uid", "..."),
+        ]
+        .into_py_dict(py);
+        // todo don't hardcode
+        // we're supposed to check the "duration" query
+        // param and use that if present (for testing)
+        thedict.set_item("expires", 300).unwrap();
```
jrconlin, line 137: I'd at least make `300` a const, since it's used later as well.

jrconlin (commented):
It looks ok, but still seems a bit WIP.

I'd feel better with less `println!` and hard coded values, particularly if we're going to land this in master.

fzzzy:
There's obviously a tension between stuff being wip and landing on master. As the new tokenserver route is not used yet, I feel it is prudent to land rougher code since it breaks the larger task up into smaller chunks, which we have been having a difficult time doing.

src/web/tokenserver.rs
```diff
@@ -1,26 +1,162 @@
...
+                e.print_and_set_sys_last_vars(py);
+                return Err(e);
+            }
+            Ok(x) => x.extract::<String>().unwrap(),
+        };
+        //assert_eq!(result, false);
+        Ok((result, derived_result))
+    })
+    .unwrap();
+    let api_endpoint = format!("{:}/1.5/{:}/", user_record[0].node, token_data.claims.sub);
```
pjenvey, line 154: We want the uid from the db here instead of the fxa id. +nit: i don't think this usually renders a trailing slash

```suggestion
    let api_endpoint = format!("{:}/1.5/{:}", user_record[0].node, user_record[0].uid);
```

src/server/mod.rs
```diff
@@ -128,7 +128,7 @@ macro_rules! build_app {
             )
             // Tokenserver
             .service(
-                web::resource(&cfg_path("/1.0/sync/1.5")).route(web::get().to(tokenserver::get)),
+                web::resource("/1.0/sync/1.5".to_string()).route(web::get().to(tokenserver::get)),
```
pjenvey, line 131: Now that this route does something other than nothing, we should probably disable this during development somehow on the live service. A setting or maybe something as simple as detecting a debug build (but note the docker-compose tests in ci run against release mode)

pjenvey (changes requested):

src/server/mod.rs
fzzzy: The docker-compose tests don't hit this tokenserver yet, so that should be fine.

fzzzy: You have to go far out of your way to configure your firefox to use this new tokenserver route and it's not documented anywhere, so does this really buy anything?

src/web/tokenserver.rs
fzzzy: I'm going to leave this for now and do it in bug #898

src/web/tokenserver.rs
fzzzy: I'm going to leave this for now and do it in bug #898

src/server/mod.rs
pjenvey: I'll admit I'm being pedantic but I think it makes sense at some point to disable the route when it's not needed anyway (thinking about it more, presence of `TOKENSERVER_DATABASE_URL` would be a good toggle).

Definitely a long shot anyone hits this endpoint but I was more concerned about it being untested/load tested, particularly w/ this handler potentially loading the Python runtime. So I'd rather just disable it sooner rather than later so there's one less thing on prod to worry about while it's still under development.

pjenvey (dismissed):

src/web/tokenserver.rs
```diff
@@ -1,26 +1,173 @@
...
+        &auth.token(),
+        &DecodingKey::from_rsa_components("2lDphW0lNZ4w1m9CfmIhC1AxYG9iwihxBdQZo7_6e0TBAi8_TNaoHHI90G9n5d8BQQnNcF4j2vOs006zlXcqGrP27b49KkN3FmbcOMovvfesMseghaqXqqFLALL9us3Wstt_fV_qV7ceRcJq5Hd_Mq85qUgYSfb9qp0vyePb26KEGy4cwO7c9nCna1a_i5rzUEJu6bAtcLS5obSvmsOOpTLHXojKKOnC4LRC3osdR6AU6v3UObKgJlkk_-8LmPhQZqOXiI_TdBpNiw6G_-eishg8V_poPlAnLNd8mfZBam-_7CdUS4-YoOvJZfYjIoboOuVmUrBjogFyDo72EPTReQ", "AQAB"),
+        &Validation::new(Algorithm::RS256),
+    ).map_err(|ee| {
+        ApiError::from(ApiErrorKind::Internal(format!("Unable to decode token_data: {:}", ee)))
+    })?;
+    let email = format!("{:}@api.accounts.firefox.com", token_data.claims.sub);
+
+    // TODO pull out of settings instead
+    let shared_secret = env::var("SYNC_MASTER_SECRET").expect("SYNC_MASTER_SECRET must be set");
```
jrconlin, line 94: nit: Tempted to move these up a bit in a future release. They'd be the trigger that @pjenvey is looking for and would save the cycles of building the `token_data` above.

jrconlin (dismissed):

pjenvey (approved):

PR #873, opened 2020-10-23
bug: handle duplicate keys in batch_upload_items for mysql

### LINKED ISSUE #827: [mysql] Handle AlreadyExists in batch_bsos errors
The mysql backend suffers from the same issue as Spanner does (see #618 for details).

jrconlin:
Closes #827

## Description

Handle duplicate keys while appending items into batch_upload_items

## Testing

Test included.

## Issue(s)

Closes #827

### DIFF
--- a/src/db/mysql/batch.rs
+++ b/src/db/mysql/batch.rs
@@ -200,7 +200,7 @@ pub fn do_append(
 
     let mut existing = HashSet::new();
 
-    // pre-load the "existing" table with any batched uploads that are already in the table.
+    // pre-load the "existing" hashset with any batched uploads that are already in the table.
     for item in sql_query(
         "SELECT userid as user_id, batch as batch_id, id FROM batch_upload_items where userid=?;",
     )
--- a/src/db/mysql/batch.rs
+++ b/src/db/mysql/batch.rs
@@ -172,49 +172,51 @@ pub fn do_append(
     db: &MysqlDb,
     batch_id: i64,
     user_id: HawkIdentifier,
-    collection_id: i32,
+    _collection_id: i32,
     bsos: Vec<params::PostCollectionBso>,
 ) -> Result<()> {
-    fn exist_idx(collection_id: i32, batch_id: i64, bso_id: &str) -> String {
+    fn exist_idx(user_id: u64, batch_id: i64, bso_id: &str) -> String {
+        // Construct something that matches the key for batch_upload_items
         format!(
-            "{collection_id}::{batch_id}::{bso_id}",
-            collection_id = collection_id,
+            "{batch_id}-{user_id}-{bso_id}",
             batch_id = batch_id,
+            user_id = user_id,
             bso_id = bso_id,
         )
     };
 
+    // It's possible for the list of items to contain a duplicate key entry.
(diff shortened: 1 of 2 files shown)


src/db/mysql/batch.rs
```diff
@@ -172,49 +172,51 @@ pub fn do_append(
...
-       in mind.
-    */
-
     let mut existing = HashSet::new();
-    let mut inserts: Vec<_> = Vec::new();
 
-    for item in sql_query("SELECT batch, id FROM batch_upload_items where userid=@userid;")
-        .bind::<BigInt, _>(user_id.legacy_id as i64)
-        .get_results::<ExistsResult>(&db.conn)?
+    // pre-load the "existing" table with any batched uploads that are already in the table.
```
fzzzy, line 203: Nit, using the word table twice is confusing because the first time it's talking about a hashset and the second time it's talking about a sql table. Change the first use of table to hashset.

fzzzy (approved):
Simpler implementation, too. Great. Just one nit.

src/db/mysql/batch.rs
```diff
@@ -172,49 +172,51 @@ pub fn do_append(
...
         ));
     }
 
     for bso in bsos {
         let payload_size = bso.payload.as_ref().map(|p| p.len() as i64);
-        let exist_idx = exist_idx(collection_id, batch_id, &bso.id);
+        let exist_idx = exist_idx(user_id.legacy_id, batch_id, &bso.id);
 
         if existing.contains(&exist_idx) {
             diesel::update(
```
pjenvey, line 222: The update clause is slightly different than Spanner's: it *always* updates all non primary key 3 columns. It should conditionally update columns only if they are not NULL (e.g. given [rfk's example](https://github.com/mozilla-services/syncstorage-rs/pull/824#discussion_r497120589), the second update would write a ttl of 27 but also overwrite the first payload with NULL). 

I'd say adding another test case to `test_append_async_w_null` should be easy enough to catch this (also write a 'b1' with update #1 touching payload, #2 only ttl, and ensuring #1's was preserved after #2's)

src/db/mysql/batch.rs
```diff
@@ -172,49 +172,51 @@ pub fn do_append(
...
-
     let mut existing = HashSet::new();
-    let mut inserts: Vec<_> = Vec::new();
 
-    for item in sql_query("SELECT batch, id FROM batch_upload_items where userid=@userid;")
-        .bind::<BigInt, _>(user_id.legacy_id as i64)
-        .get_results::<ExistsResult>(&db.conn)?
+    // pre-load the "existing" hashset with any batched uploads that are already in the table.
+    for item in sql_query(
+        "SELECT userid as user_id, batch as batch_id, id FROM batch_upload_items where userid=?;",
```
pjenvey, line 205: not as important as in spanner but the batch_id should really be specified here as part of the WHERE clause

pjenvey (changes requested):

src/db/mysql/batch.rs
pjenvey: sweet, I forgot about this

pjenvey (approved):

Issue #874, opened 2020-10-23
Investigate pagination improvements


pjenvey:
Our basic pagination using LIMIT + OFFSET causes performance issues under Spanner. We should look into implementing alternatives, e.g.:

https://medium.com/swlh/why-you-shouldnt-use-offset-and-limit-for-your-pagination-4440e421ba87
https://cloud.google.com/spanner/docs/schema-design#primary-key-prevent-hotspots

Could the [previous offset optimization](https://github.com/mozilla-services/syncstorage-rs/issues/53) improve performance? Or could we overload the extra metadata it uses via the offset parameter (e.g. `<value>:offset`) for implementing cursor based pagination?

fzzzy:
I think if we use the previous offset optimization, except have every timestamp always be unique and have :<offset> always be 0, this will fix the problem and avoid the unnecessary complicated logic for dealing with records that have identical timestamps.

jrconlin:
So, we have a few things going on here that are of note. 

1) posts have a "modification_time" assigned to them when they're loaded in. For any BSO loaded in batch, they get the same timestamp. This leads to a lot of duplication as noted above.

2) spanner has a `[commit_timestamp()](https://cloud.google.com/spanner/docs/commit-timestamp)`, which is great. Apparently we can add it to an existing `TIMESTAMP` column, I presume it become effective after it is applied, which may be a problem for the TB or so of existing data. MySQL does not have a similar feature. 

3) of all the known BSO data, BSO_ID is known to be unique and could be used to disambiguate items or provide some order when processing records to avoid breaks.

It should be possible, as @fzzzy pointed out, to effectively batch on `modified` with a sub-index of the id. This would remove most of the table scan, but will require some additional index work on the remainder. I don't know how that may impact performance at this time.

jrconlin:
Pending #911 for tests fixes.

PR #875, opened 2020-10-26
feat: Add `SYNC_ENFORCE_QUOTA` flag

### LINKED ISSUE #870: Update quota to "non-enforce" mode.
A quota violation is logged, but no quota specific error is returned to the user.

Make sure that there is a metric reflecting the time of quota calculation per user and that the over quota error is logged to sentry.

jrconlin:
## Description

Adds `SYNC_ENFORCE_QUOTA` / `--enforce_quota` flag which will force an
error if a user exceeds quota.

So for the record:
    `SYNC_LIMITS__MAX_QUOTA_LIMIT` => Sets the quota limit (default 2GB)
    `SYNC_ENABLE_QUOTA` => Determine if quota calcs should be done
    `SYNC_ENFORCE_QUOTA` => Determine if over quota errors are returned

## Testing

Includes tests. Add `SYNC_ENABLE_QUOTA` to enable quota checks, and `SYNC_ENFORCE_QUOTA` to make sure errors are returned for over quota events

## Issue(s)

Closes #870

### DIFF
--- a/src/db/mysql/models.rs
+++ b/src/db/mysql/models.rs
@@ -31,7 +31,7 @@ use crate::db::{
     Db, DbFuture, Sorting,
 };
 use crate::server::metrics::Metrics;
-use crate::settings::Quota;
+use crate::settings::{Quota, DEFAULT_MAX_TOTAL_RECORDS};
 use crate::web::extractors::{BsoQueryParams, HawkIdentifier};
 use crate::web::tags::Tags;
 
@@ -42,7 +42,8 @@ type Conn = PooledConnection<ConnectionManager<MysqlConnection>>;
 /// We store the TTL as a SyncTimestamp, which is milliseconds, so remember
 /// to multiply this by 1000.
 pub const DEFAULT_BSO_TTL: u32 = 2_100_000_000;
-pub const DEFAULT_LIMIT: i64 = 10000;
+// this is the max number of records we will return.
+pub static DEFAULT_LIMIT: u32 = DEFAULT_MAX_TOTAL_RECORDS;
 
 pub const TOMBSTONE: i32 = 0;
 /// SQL Variable remapping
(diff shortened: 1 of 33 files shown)


src/db/mysql/models.rs
```diff
@@ -514,6 +518,10 @@ impl MysqlDb {
             query = query.filter(bso::id.eq_any(ids));
         }
 
+        // it's possible for two BSOs to be inserted with the same `modified` date,
+        // since there's no guarantee of order when doing a get, pagination can return
+        // an error. We "fudge" a bit here by taking the id order as a secondary, since
+        // that is guaranteed to be unique by the client.
```
fzzzy, line 524: Yikes, I finally see what you mean about the race condition this caused in the tests here.

src/db/mysql/models.rs
```diff
@@ -524,12 +532,12 @@ impl MysqlDb {
             Sorting::Oldest => query.order(bso::id.asc()).order(bso::modified.asc()),
             */
             Sorting::Index => query.order(bso::sortindex.desc()),
-            Sorting::Newest => query.order(bso::modified.desc()),
-            Sorting::Oldest => query.order(bso::modified.asc()),
+            Sorting::Newest => query.order((bso::modified.desc(), bso::id.desc())),
```
fzzzy, line 535: Yes. It's all so clear to me now. So to do the LIMIT/OFFSET removal optimization, we'll need to pass the timestamp:id pair instead of timestamp:offset pair, and skip over any records matching that timestamp that don't have the same id.

fzzzy (dismissed):

src/db/mysql/models.rs
```diff
@@ -524,12 +532,12 @@ impl MysqlDb {
             Sorting::Oldest => query.order(bso::id.asc()).order(bso::modified.asc()),
             */
             Sorting::Index => query.order(bso::sortindex.desc()),
```
pjenvey, line 534: maybe we should revert to the commented out issue559 code above? (does the same but also secondary sorts for Sorting::Index + None)

jrconlin: Yeah, I was wondering about that and was going to ask you. Just forgot will all the other crap. Thanks!

jrconlin: I think I may roll this change into the patch for #874, since it's impacting a test there.

pjenvey (approved):

PR #879, opened 2020-10-29
feat: Add `count` and `count_with_tags` metric for batch histogram

### LINKED ISSUE #878: Add count metric for histogram around batch updates
We currently only run `incr`, which doesn't let us do more than single bumps.

jrconlin:
## Description

Adding a metric handler for `counts`. This will allow us to do histograms to determine loads for things like batch processing.

## Testing

are batches metricked?

## Issue(s)

Closes #878

### DIFF
--- a/src/db/spanner/batch.rs
+++ b/src/db/spanner/batch.rs
@@ -284,10 +284,13 @@ pub async fn do_append_async(
 
     //prefetch the existing batch_bsos for this user's batch.
     let mut existing = HashSet::new();
+    let mut collisions = HashSet::new();
+    let mut count_collisions = 0;
     let mut tags = Tags::default();
     tags.tags.insert(
         "collection".to_owned(),
-        db.get_collection_name(collection_id),
+        db.get_collection_name(collection_id)?
+            .unwrap_or_else(|| "UNKNOWN".to_string()),
     );
 
     let bso_ids = bsos.iter().map(|pbso| pbso.id.clone());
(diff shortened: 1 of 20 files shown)


src/db/spanner/batch.rs
```diff
@@ -472,6 +491,11 @@ pub async fn do_append_async(
             .param_types(param_types.clone())
             .execute_dml_async(&db.conn)
             .await?;
+            db.metrics.count_with_tags(
```
pjenvey, line 494: this emits within the update loop (one per update) and always emits the same count_update (the number of total updates). and it's the same count emitted above (existing.len() as `storage.spanner.batch.pre-existing`). so I think this should just be killed?

src/db/spanner/models.rs
```diff
@@ -137,6 +137,13 @@ impl SpannerDb {
         }
     }
 
+    pub(super) fn get_collection_name(&self, id: i32) -> String {
+        self.coll_cache
+            .get_name(id)
```
pjenvey, line 142: this returns `<Result<Option>>`

we probably wanna propagate it, there's a lock underneath which could have been poisoned. feels bad hiding that kind of error

```suggestion
            .get_name(id)?
```

pjenvey: (FYI I actually implemented a _get_collection_name for mysql. apparently it is never used)

src/db/spanner/batch.rs
jrconlin: I should move it out of the update loop ('doh) but I was thinking more that `exists` is whatever was already in the table and this was the number of elements that resulted in updates, because a batch could self update as  well  (rare, but potential).

jrconlin: If anything I was debating killing the `metrics.incr` above, since it's fairly low information

src/db/spanner/models.rs
jrconlin: so you're saying that it's a TODO to make this a trait? ;)

pjenvey (approved):

PR #880, opened 2020-10-29
feat: optimize POST w/ ?batch=true&commit=true

### LINKED ISSUE #876: Optimize POST batch w/ ?batch=true&commit=true
Clients frequently send batch requests with both `batch=true` (create a batch) and `commit=true` (commit the batch and delete it). This is the equivalent of a simple POST sans a batch.

This is often done on collections with frequent updates, in particular `tabs` and `clients`. The `tabs` collection was previously kept solely in memcached vs on disk in the legacy cluster due to its very frequent changes.

We should detect these cases and convert them to simple regular post_collection calls, avoiding the overhead of creating a batch, calling a full commit on it and then deleting it.

pjenvey:
## Description

this is equivalent to a simpler post_bsos call: so do that instead of
incurring post_collection_batch's overhead

## Testing

e2e tests executes this path a couple of times (and there's a specific test_batch_with_immediate_commit test for it)

## Issue(s)

Closes #876

### DIFF
--- a/src/web/handlers.rs
+++ b/src/web/handlers.rs
@@ -236,8 +236,14 @@ pub async fn post_collection(
 
             // batches are a conceptual, singular update, so we should handle
             // them separately.
-            if coll.batch.is_some() {
-                return post_collection_batch(coll, db).await;
+            if let Some(ref batch) = coll.batch {
+                // Optimization: specifying ?batch=true&commit=true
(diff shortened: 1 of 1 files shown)


jrconlin (approved):

jrconlin:
I think it's a good fix. It's basically the same thing we were doing before, with the added check if the `commit` is set.

PR #897, opened 2020-11-02
feat: Enable cargo build `--release` flag

### LINKED ISSUE #896: Consider using `--release` build flag
We are currently building a rust dev (default) image. The cargo build `--release` flag enables a number of running optimizations. We should consider it as an option to improve performance.

jrconlin:
Closes #896

## Description

We are currently building a default "dev" environment. `cargo build --release` enables a number of optimizations which may improve performance.

## Testing

We should preform a comparative load test between a current and `--release` version to see if there are any notable benefits of using this flag.

## Issue(s)

Closes #896.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,7 +63,7 @@ commands:
 
   run-tests:
     steps:
-        - run:
+      - run:
           name: cargo test
           command: cargo test --all --verbose
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,7 +41,7 @@ commands:
     steps:
       - run:
           name: cargo build
-          command: cargo build
+          command: cargo build --release
   setup-gcp-grpc:
     steps:
       - run:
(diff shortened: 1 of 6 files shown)


tublitzed (approved):
Good call! It was added [here](https://github.com/mozilla-services/syncstorage-rs/blob/master/README.md#creating-releases) to the release instructions, but I guess we missed CI. thanks for catching and fixing.

.circleci/config.yml
```diff
@@ -41,7 +41,7 @@ commands:
     steps:
       - run:
           name: cargo build
-          command: cargo build
+          command: cargo build --release
```
pjenvey, line 44: This shouldn't be necessary. We build syncstorage twice in CI (sigh): here and within the Docker. They don't share target/ directories (at one point they did, but grpc support code became out of sync between the environments, so we stopped  that).

The build profile in the Docker is what counts for our perf and it's already building in release mode (via cargo install which defaults to release)

jrconlin: Ah, cool. 
I'll close this PR then.

jrconlin:
Not needed.

PR #900, opened 2020-11-03
chore: update dependencies

jrconlin:
## Description

* cargo audit flagged a deep dependency,
* which required updating libraries
* which broke a few tests due to formatting
* which also required a few function changes due to obsolete

## Testing

do CI tests pass and does the tree go green? 
👍 👍

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11,7 +11,7 @@ dependencies = [
  "futures-core",
  "futures-sink",
  "log",
- "pin-project",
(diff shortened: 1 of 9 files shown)


fzzzy (approved):
Thanks for jumping on this grenade.

jrconlin:
heh, I was going to add the correct issue. (I filed #901 yesterday) Will manually close that one.

PR #909, opened 2020-11-10
feat: Return somewhat more correct values inside the tokenlib signed …

### LINKED ISSUE #865: Return the correct values inside the tokenlib signed token
In the branch tokenserver-experiments, I have signing into sync with the rust tokenserver endpoint working. However, all the values returned are fake and hardcoded. We'll need to make sure all these values, including the generation and keys changed at logic and the hashed user id, are implemented and provide the correct values.

fzzzy:
## Description

Generate more-correct contents inside the tokenlib token.

There's some work to be done to match the old hashing algo exactly, but I'll open a new issue for that.

## Testing

First, you need python3 and to run pip3 install tokenlib.

Need a tokenserver mysql database:

    create table services (id int(11) auto_increment not null, service varchar(30), pattern varchar(128), primary key(id));
    create table nodes (
      id bigint(20) auto_increment not null, service int(11) not null,
      node varchar(64) not null, available int(11) not null, current_load int(11) not null,
      capacity int(11) not null, downed int(11) not null, backoff int(11) not null,
      primary key(id), foreign key(service) references services(id)
    );
    create table users (
      uid bigint(20) auto_increment not null, service int(11) not null,
      email varchar(255) not null, generation bigint(20) not null,
      client_state varchar(32) not null, created_at bigint(20) not null,
      replaced_at bigint(20), nodeid bigint(20) not null,
      keys_changed_at bigint(20), primary key(uid),
      foreign key(nodeid) references nodes(id)
    );

Run syncstorage-rs.

In firefox, create a new profile. In about:config, set identity.sync.tokenserver.uri to http://localhost:8000/1.0/sync/1.5

Log in to sync with your normal prod fxa credentials. It should be able to put your stuff in your localhost syncstorage-rs.

## Issue(s)

Closes #865 .

### DIFF
--- a/src/web/tokenserver.rs
+++ b/src/web/tokenserver.rs
@@ -107,23 +107,54 @@ pub fn get_sync(auth: &BearerAuth) -> Result<TokenServerResult, ApiError> {
                    AND nodes.id = users.nodeid
                    AND nodes.service = services.id"#,
     )
-    .bind::<Text, _>(email)
+    .bind::<Text, _>(&email)
     .load::<TokenserverUser>(&connection)
     .unwrap();
     let (python_result, python_derived_result) = Python::with_gil(|py| {
         let tokenlib = PyModule::from_code(
             py,
-            r#"
+            r###"
+import base64
+from hashlib import sha256
+import hmac
 import tokenlib
 
 
 def make_token(plaintext, shared_secret):
+    print("MAKE_TOKEN", plaintext)
     return tokenlib.make_token(plaintext, secret=shared_secret)
 
 
 def get_derived_secret(plaintext, shared_secret):
     return tokenlib.get_derived_secret(plaintext, secret=shared_secret)
-"#,
+
+
+def encode_bytes(value):
+    """Encode BrowserID's base64 encoding format.
+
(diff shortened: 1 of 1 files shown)


src/web/tokenserver.rs
```diff
@@ -107,43 +107,114 @@ pub fn get_sync(auth: &BearerAuth) -> Result<TokenServerResult, ApiError> {
...
+    return fxa_metrics_hash(fxa_uid[:32] + device, secret)[:32]
+    "###,
             "main.py",
             "main",
         )
         .map_err(|e| {
             e.print_and_set_sys_last_vars(py);
             e
         })?;
+        let client_state_b64 = match tokenlib.call1("encode_bytes", (&user_record[0].client_state,))
```
fzzzy, line 165: This would be the perfect place for a macro!

tublitzed:
@fzzzy can you add more detailed testing instructions? If it's the same as another PR, maybe just copy/paste those so reviewers don't have to go through and find them?

jrconlin (approved):

pjenvey (approved):

PR #913, opened 2020-11-12
Tokenserver jr

jrconlin:
* adds dbg!()
* adds route for `/1.0/sync/1.5/{uid}` (not sure if that's needed.

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -130,6 +130,10 @@ macro_rules! build_app {
             .service(
                 web::resource("/1.0/sync/1.5".to_string()).route(web::get().to(tokenserver::get)),
             )
+            .service(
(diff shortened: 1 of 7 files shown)


src/server/mod.rs
```diff
@@ -130,6 +130,10 @@ macro_rules! build_app {
             .service(
                 web::resource("/1.0/sync/1.5".to_string()).route(web::get().to(tokenserver::get)),
             )
+            .service(
```
fzzzy, line 133: Is this to route numeric non-uuid user ids to get? I really don't understand why the functional tests passed before this.

jrconlin: Yeah, I'm not convinced we need it, but I was getting 404s trying to pass anything to that URL and wasn't sure that it was required. I think the larger problem is that the tests are using the wrong path.

PR #917, opened 2020-11-13
Py3 integration tests

### LINKED ISSUE #911: Port integration tests to python3
The integration tests are still python2 and are becoming more difficult to run as dependencies slowly drop support for python2. The way the circle config runs the tests could also be far more efficient if we ran the tests under python3 in the same docker image without using docker compose.

fzzzy:
## Description

Convert the integration tests to python 3, and run them in the same docker container as the rust server, hopefully greatly speeding up the build process. Converting the tests to python 3 is good because it is getting harder to acquire the python 2 compatible dependencies, and we finally need to change the tests to fix an id ordering bug.

## Testing

The tests should continue passing.

## Issue(s)

Closes #911

### DIFF
--- a/src/server/metrics.rs
+++ b/src/server/metrics.rs
@@ -127,6 +127,7 @@ impl Metrics {
 
     pub fn incr_with_tags(&self, label: &str, tags: Option<Tags>) {
         if let Some(client) = self.client.as_ref() {
+            dbg!(&label, &tags);
             let mut tagged = client.incr_with_tags(label);
             let mut mtags = self.tags.clone().unwrap_or_default();
             if let Some(tags) = tags {
--- a/src/server/metrics.rs
+++ b/src/server/metrics.rs
@@ -127,7 +127,6 @@ impl Metrics {
 
     pub fn incr_with_tags(&self, label: &str, tags: Option<Tags>) {
         if let Some(client) = self.client.as_ref() {
-            dbg!(&label, &tags);
             let mut tagged = client.incr_with_tags(label);
             let mut mtags = self.tags.clone().unwrap_or_default();
             if let Some(tags) = tags {
--- a/tools/integration_tests/test_storage.py
+++ b/tools/integration_tests/test_storage.py
@@ -1,0 +1,2085 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+"""
+Functional tests for the SyncStorage server protocol.
+
+This file runs tests to ensure the correct operation of the server
+as specified in:
+
+    http://docs.services.mozilla.com/storage/apis-1.5.html
+
+If there's an aspect of that spec that's not covered by a test in this file,
+consider it a bug.
+
+"""
+
+import unittest2
+
+import re
+import json
+import time
+import random
+import string
+import urllib
+import webtest
+import contextlib
+# import math
+import simplejson
+
+from pyramid.interfaces import IAuthenticationPolicy
+from test_support import StorageFunctionalTestCase
+
+import tokenlib
+
+
+class ConflictError(Exception):
+    pass
+
+
+class BackendError(Exception):
+    pass
+
+
+WEAVE_INVALID_WBO = 8               # Invalid Weave Basic Object
+WEAVE_SIZE_LIMIT_EXCEEDED = 17      # Size limit exceeded
+
+BATCH_MAX_IDS = 100
+
+
+def get_limit_config(request, limit):
+    """Get the configured value for the named size limit."""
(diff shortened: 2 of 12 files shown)


src/server/mod.rs
```diff
@@ -130,6 +130,10 @@ macro_rules! build_app {
             .service(
                 web::resource("/1.0/sync/1.5".to_string()).route(web::get().to(tokenserver::get)),
             )
+            .service(
```
jrconlin, line 133: We can remove this since we fixed the tests.

src/error.rs
```diff
@@ -187,6 +187,7 @@ impl ApiError {
 
     pub fn render_404<B>(res: ServiceResponse<B>) -> Result<ErrorHandlerResponse<B>> {
         // Replace the outbound error message with our own.
+        dbg!("Returning 404...");
```
jrconlin, line 190: Probably can remove a lot of these dbg!, since we've isolated why the tests weren't connecting up. They'll probably just be noise later.

tools/integration_tests/test_storage.py
```diff
@@ -0,0 +1,2077 @@
...
+        res.sort()
+        self.assertEquals(res, ['00', '01', '02', '03', '04'])
+
+        # application/newlines
+        res = self.app.get(self.root + '/storage/xxx_col2',
+                           headers=[('Accept', 'application/newlines')])
+        self.assertEquals(res.content_type, 'application/newlines')
+
+        self.assertTrue(res.body.endswith(b'\n'))
+        res = [json_loads(line) for line in res.body.decode('utf-8').strip().split('\n')]
```
jrconlin, line 404: heh, yeah, we should run flake8 on this, just so we're consistent about things...

jrconlin:
Hrm, running the tests using `circleci local` and the e2e stuff is failing because it can't find the local server.

tools/integration_tests/test_storage.py
tublitzed: and maybe also add to CI?

fzzzy:
jr pulled this stuff into his branch and is landing it there.

fzzzy:
Reopening this so we can see the diffs required to the python tests by the offset changes more easily.

tools/integration_tests/test_storage.py
fzzzy: It's in ci already @tublitzed.

fzzzy:
Unfortunately, I can't see an easy way to put the tests back to the way they were before jr applied the logic fixes to them, as we cleaned them up and ported them to py3 as part of this. @jrconlin can you think of an easy way to revert your logic changes to the tests in this branch so they will show up as diffs in your offset branch?

tools/integration_tests/test_storage.py
tublitzed: python linting is? great! I must have missed it then. @fzzzy can you point me to where we're doing that? I'm probably overlooking it but not seeing anything [here](https://github.com/mozilla-services/syncstorage-rs/blob/master/.circleci/config.yml)

fzzzy: Ok, so apparently it's not in syncstorage-rs! I guess this is the first time we have any python files. I'll add it. (We relied on it being enabled in server-syncstorage)

jrconlin: The tool that we tend to use for this is Flake8. It's possible to add it into the CI process, but considering the general complexity that we already have, I don't really think we need to compound that. Also, we shouldn't be changing this file frequently, so running flake8 on this file every build just burns a not insignificant amount of CPU. Plus, flake8 isn't like `cargo fmt`, you have to manually edit the file to resolve issues it may find. 

We can run it manually if we ever touch these files and generally check for visual clues that it might be needed. (e.g. lines longer than 80 characters, some spacing issues, etc.)

src/server/metrics.rs
```diff
@@ -127,6 +127,7 @@ impl Metrics {
 
     pub fn incr_with_tags(&self, label: &str, tags: Option<Tags>) {
         if let Some(client) = self.client.as_ref() {
+            dbg!(&label, &tags);
```
jrconlin, line 130: Kind of ok with keeping this, but that's probably the recent patches talking.

We should trim out extra `dbg!()`s

jrconlin (approved):

jrconlin (approved):

tools/integration_tests/test_storage.py
fzzzy: > python linting is? great! I must have missed it then. @fzzzy can you point me to where we're doing that? I'm probably overlooking it but not seeing anything [here](https://github.com/mozilla-services/syncstorage-rs/blob/master/.circleci/config.yml)

Ah, right, it wasn't enabled for syncstorage-rs yet because we haven't yet had any python files. I added a new bug to track adding flake8 coverage to the new python3 test suite.

PR #919, opened 2020-11-14
Optimize Pagination

### LINKED ISSUE #874: Investigate pagination improvements

jrconlin:
## Description

* Adds pagination improvements (using modified timestamp + bso_id for pagination offsets)
* merges testing from @fzzzy's tokenserver branch so that sorting and skipping weird `offset` isn't a problem.

## Testing

Testing included.

## Issue(s)

Closes #874.


jrconlin:
Yeah, I might squash this down to something more reasonable.

jrconlin:
... and no idea why the e2e-test aren't firing off.

pjenvey:
I know there's some work to do around cleaning up the large integration test related changes.

cc: @fzzzy can we include separate commits (diffs) from the original in our repo? It would be helpful clearly see the original vs python 3 changes vs changes for pagination

jrconlin:
I wanted to pre-merge [@fzzzy's changes](https://github.com/mozilla-services/syncstorage-rs/pull/917) so I could get the fixed python3 tests. 

I will need to rebase and collapse them into a smaller set.

jrconlin:
Rebased. May still need to pare out some of the `ini` files.

fzzzy:
JR the test that's failing just needs to be <= instead of < just like the other test you fixed like that.

jrconlin:
@fzzzy thanks! yeah, I had to do a hand merge to get the changes in. I missed that one.

src/db/mysql/models.rs
```diff
@@ -606,23 +623,28 @@ impl MysqlDb {
...
         // fetch an extra row to detect if there are more rows that
-        // match the query conditions. Negative limits will cause an error.
-        query = query.limit(if limit == 0 { limit } else { limit + 1 });
-        let numeric_offset = offset.map_or(0, |offset| offset.offset as i64);
-        if numeric_offset != 0 {
-            // XXX: copy over this optimization:
-            // https://github.com/mozilla-services/server-syncstorage/blob/a0f8117/syncstorage/storage/sql/__init__.py#L404
-            query = query.offset(numeric_offset);
+        // match the query conditions. Negative limits will cause an error in mysql.
+        if limit >= 0 {
```
fzzzy, line 646: Was the negative limit equivalent to specifying no limit?

src/db/spanner/models.rs
```diff
@@ -1250,41 +1233,40 @@ impl SpannerDb {
             sqlparams.insert("newer".to_string(), as_value(newer.as_rfc3339()?));
             sqltypes.insert("newer".to_string(), as_type(TypeCode::TIMESTAMP));
         }
+
+        // if the offset has not yet been applied (we set it to None if we have), then apply it using the
```
fzzzy, line 1237: In the future we may want to remove the ability to even do limit/offset queries. It would greatly simplify the code.

src/db/spanner/models.rs
```diff
@@ -1294,54 +1276,16 @@ impl SpannerDb {
...
             Offset {
-                offset: offset + modifieds.len() as u64,
-                timestamp: None,
+                offset,
+                timestamp: timestamp.map(|v| SyncTimestamp::from_i64(v).unwrap_or_default()),
             }
             .to_string(),
         )
-        /*
-        let mut calc_offset = 1;
```
fzzzy, line 1312: Good riddance to this cursed code!

src/db/spanner/models.rs
```diff
@@ -1371,9 +1314,11 @@ impl SpannerDb {
         // https://bugzilla.mozilla.org/show_bug.cgi?id=963332
 
         let next_offset = if limit >= 0 && bsos.len() > limit as usize {
-            bsos.pop();
-            let modifieds: Vec<i64> = bsos.iter().map(|r| r.modified.as_i64()).collect();
-            self.encode_next_offset(sort, offset, timestamp.map(|t| t.as_i64()), modifieds)
+            let bso = bsos.pop();
+            let (offset, timestamp) = bso
+                .map(|b| (Some(b.id), Some(b.modified.as_i64())))
+                .unwrap_or((None, None));
```
fzzzy, line 1320: Does this unwrap multiple values in a tuple? Cool.

src/server/test.rs
```diff
@@ -669,6 +669,7 @@ async fn overquota() {
     .to_request();
     let response = app.call(req).await.unwrap();
     let status = response.status();
+    dbg!(&response);
```
fzzzy, line 672: Should this be removed? Or is it generally useful enough to leave in

tools/integration_tests/old_storage_checks.py
```diff
@@ -0,0 +1,272 @@
+class TestStorageMemcached(TestStorage):
```
fzzzy, line 1: This code is still in the old repos, let's just remove it from here.

tools/integration_tests/requirements.txt
```diff
@@ -0,0 +1,10 @@
+unittest2
```
fzzzy, line 1: I feel like this file should be alphabetized but feel free to completely ignore this.

tools/integration_tests/run.py
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env python3
+
+import atexit
+import subprocess
+import sys
+from test_storage import TestStorage
+from test_support import run_live_functional_tests
+import time
+
```
fzzzy, line 9: I wrote this code, so this is my fault, but there should be two blank lines here.

fzzzy (dismissed):
Looking good. A few nits. There are also ini files related to memcache that I think can be deleted now.

src/db/mysql/models.rs
jrconlin: Yeah, it was. The code originally passed it to the limit, which I'm not quite sure did the right thing in any case. I did a first run band-aid where I set the limit to something huge, but then realized a way to just not add the limit if it's not specified.

src/db/spanner/models.rs
jrconlin: probably. I can still kinda/sorta see the value of a limit, but agree that not having an offset would be good.

The only problem I had when I first did that was that while we have a "newer" and "older", they're more strict than we want a given offset to be. (e.g. you should never have a records = to a newer or older, while you absolutely will have one when doing an offset.)

src/db/spanner/models.rs
jrconlin: Yeah, it's a cheesy hack, but it works.

tools/integration_tests/requirements.txt
jrconlin: well, really ought to freeze to versions too, but I built it on the fly.

fzzzy:
We really should split this into two prs again, one with the python3 tests and one with pagination, so it is easier to review the changes to the tests required by pagination, as Phil mentioned in an earlier comment.

jrconlin:
Absolutely. I'd like to see #917 land first, then this PR, since I only really need to modify one line in a test.

fzzzy:
@jrconlin This has conflicts now unfortunately. I think you can probably take all of the changes on master.

pjenvey (commented):
I'm seeing `test_pagination_with_newer_and_sort_by_oldest` failing over here.

```
======================================================================
FAIL: test_pagination_with_newer_and_sort_by_oldest (syncstorage.tests.functional.support.LiveTestCases)
----------------------------------------------------------------------
  File "syncstorage/tests/functional/test_storage.py", line 1449, in test_pagination_with_newer_and_sort_by_oldest
    range(start + 1, NUM_ITEMS))
AssertionError: Lists differ: [4, 5, 6, 7, 8, 9] != [4, 5, 6, 7, 8, 9, 10, 11]

Second list contains 2 additional elements.
First extra element 6:
10

- [4, 5, 6, 7, 8, 9]
+ [4, 5, 6, 7, 8, 9, 10, 11]
?                  ++++++++


----------------------------------------------------------------------
```

I'm seeing this happen due to the sorting of bso_ids, e.g.:

```
SELECT bso_id, sortindex, payload, modified, expiry
              FROM bsos
             WHERE fxa_uid = "df871969-d8a8-414d-988f-35d66f0594d0"
               AND fxa_kid = "1ec82d3b-968a-470b-80ed-c95c4cd443b7"
               AND collection_id = 102
               AND expiry > CURRENT_TIMESTAMP()
               ORDER BY modified ASC, bso_id ASC

0 |   | x | 2020-12-10T00:48:10.72Z | 2087-06-27T14:08:10.72Z
1 |   | x | 2020-12-10T00:48:10.72Z | 2087-06-27T14:08:10.72Z
2 |   | x | 2020-12-10T00:48:10.72Z | 2087-06-27T14:08:10.72Z
3 |   | x | 2020-12-10T00:48:10.72Z | 2087-06-27T14:08:10.72Z
4 |   | x | 2020-12-10T00:48:12.54Z | 2087-06-27T14:08:12.54Z
5 |   | x | 2020-12-10T00:48:12.54Z | 2087-06-27T14:08:12.54Z
6 |   | x | 2020-12-10T00:48:12.54Z | 2087-06-27T14:08:12.54Z
7 |   | x | 2020-12-10T00:48:12.54Z | 2087-06-27T14:08:12.54Z
10 |   | x | 2020-12-10T00:48:13.98Z | 2087-06-27T14:08:13.98Z
11 |   | x | 2020-12-10T00:48:13.98Z | 2087-06-27T14:08:13.98Z
8 |   | x | 2020-12-10T00:48:13.98Z | 2087-06-27T14:08:13.98Z
9 |   | x | 2020-12-10T00:48:13.98Z | 2087-06-27T14:08:13.98Z


```

whereas the failing request results in:

```
SELECT bso_id, sortindex, payload, modified, expiry
              FROM bsos
             WHERE fxa_uid = "df871969-d8a8-414d-988f-35d66f0594d0"
               AND fxa_kid = "1ec82d3b-968a-470b-80ed-c95c4cd443b7"
               AND collection_id = 102
               AND expiry > CURRENT_TIMESTAMP()
               AND modified > "2020-12-10T00:48:10.720000000Z"
               AND modified >= "2020-12-10T00:48:13.980000000Z"
               AND bso_id >= "8"
               ORDER BY modified ASC, bso_id ASC


8 |   | x | 2020-12-10T00:48:13.98Z | 2087-06-27T14:08:13.98Z
9 |   | x | 2020-12-10T00:48:13.98Z | 2087-06-27T14:08:13.98Z
```

Is this due to the fact that we're not doing a "tuple comparison" e.g. `(modified, bso_id) >= (<modified>, "8")`?

e2e-tests/test_storage.py
```diff
@@ -0,0 +1,2364 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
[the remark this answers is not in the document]
jrconlin: ah, yeah. thanks. this was a work before I pulled in @fzzzy's code.

src/db/mysql/models.rs
```diff
@@ -507,72 +507,89 @@ impl MysqlDb {
             .filter(bso::expiry.gt(now))
             .into_boxed();
 
+        // Set the harder limits for "older" and "newer"
         if let Some(older) = older {
-            query = query.filter(bso::modified.lt(older.as_i64()));
+            let ts = older.as_i64();
+            if ts > 0 {
```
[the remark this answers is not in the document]
jrconlin: IIRC, we only want to filter if "older" is actually defined, it defaults to `Some(0)` in some cases, which can cause problems. Same with newer below.

src/web/extractors.rs
```diff
@@ -1125,45 +1125,54 @@ impl From<u32> for HawkIdentifier {
 #[serde(default)]
 pub struct Offset {
     pub timestamp: Option<SyncTimestamp>,
-    pub offset: u64,
+    pub offset: Option<String>,
```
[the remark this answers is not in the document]
jrconlin: Because offset is an optional argument and I'd rather it be an Option than "".

src/db/spanner/models.rs
```diff
@@ -1294,54 +1276,16 @@ impl SpannerDb {
...
     ) -> Option<String> {
-        // issue559: Use a simple numeric offset everwhere as previously for
-        // now: was previously a value of "limit + offset", modifieds.len()
-        // always equals limit
         Some(
             Offset {
-                offset: offset + modifieds.len() as u64,
-                timestamp: None,
+                offset,
+                timestamp: timestamp.map(|v| SyncTimestamp::from_i64(v).unwrap_or_default()),
```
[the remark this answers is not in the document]
jrconlin: Kind of on the fence about this one, but sure. 

timestamp is still an optional value in this context and the resulting value is used mostly as an option, but I can see the value of passing any parse error up.

fzzzy:
This is r+ from me but I'd like Phil to give it the final approval.

jrconlin:
@pjenvey, so did a bit more testing here and I think I figured out the discrepancy you're seeing regarding the old test vs the new one. 

In short, the problem is that we have sub-batches of randomized bso_ids associated with blocks of modified timestamps.  an example might be:

|id|timestamp|
|--|--|
|17| 00:01|
|37| 00:01|
|66| 00:01|
|06| 00:02|
|22| 00:02|
|10| 00:03|
|94| 00:03|

The problem is that we have to use modified as the primary key, but things get "complicated" when we have to cross a timestamp/ID boundary. We have to closely associate the id/timestamp combination because otherwise saying "AND id > ##" may miss some records that have an arbitrary ID less than that with the new timestamp. 

I'm thinking that we might need to manufacture an artificial pagination id that combines these two values.

PR #922, opened 2020-11-16
fix: add a short delay to avoid 503s

### LINKED ISSUE #920: Fix flakey overquota/quota tests on ci
These are currently failing sometimes on circleci

pjenvey:
##  Description

not ideal but a test w/ multiple write calls needs to handle 503s (similar to
the e2e test's retry delay)

## Testing

tests should reliably pass on ci

## Issue(s)

Closes #920

### DIFF
--- a/src/server/test.rs
+++ b/src/server/test.rs
@@ -655,6 +655,9 @@ async fn overquota() {
     let status = response.status();
     assert_eq!(status, StatusCode::OK);
 
+    // avoid the request calls running so quickly that they trigger a 503
+    actix_rt::time::delay_for(Duration::from_millis(10)).await;
+
     let req = create_request(
         http::Method::PUT,
         "/1.5/42/storage/xxx_col2/12345",


jrconlin (approved):

jrconlin:
WFM. I'm still not 100% sure why the huge difference in returned results for the `quota` test sometimes. Still trying to see if I can replicate that one locally.

jrconlin:
One stupid thought:

I added a hopefully meaningless assert. Basically, check if the response we're checking matches up with what we thought we made. I will be horrified if this ever gets triggered, but it might explain why sometimes `server::test::quota` fails with a vastly different response than what we're expecting.

```
diff --git a/src/server/test.rs b/src/server/test.rs
index 2b03c3a9e9..21359cecba 100644
--- a/src/server/test.rs
+++ b/src/server/test.rs
@@ -190,6 +190,11 @@ async fn test_endpoint(
         None => assert!(sresp.response().status().is_success()),
         Some(status) => assert!(sresp.response().status() == status),
     };
+    // We get occasional errors in `server::test::quota` where the response is
+    // radically different than what we're expecting. Do a really stupid sanity
+    // check to ensure the response we're getting matches up with the request we
+    // made.
+    assert_eq!(sresp.request().uri().path(), path);
     if let Some(x_body) = expected_body {
         let body = test::read_body(sresp).await;
         assert_eq!(body, x_body.as_bytes());
```

PR #929, opened 2020-11-19
bug: make actix-cors more permissive

### LINKED ISSUE #928: Fix CORS issues
Actix 0.8+ has gotten more strict about CORS. We need to ensure that the rules we publish are either compatible with the ones on NGINX.

This is causing connections to fail, per
```
1605700991933   FirefoxAccounts TRACE   not checking freshness of profile as it remains recent
1605700992125   Sync.Resource   DEBUG   DELETE fail 400 https://stage.sync.nonprod.cloudops.mozgcp.net/1.5/xxx/storage
1605700992125   Sync.Resource   WARN    DELETE request to https://stage.sync.nonprod.cloudops.mozgcp.net/1.5/xxx/storage failed with status 400
1605700992125   Sync.Service    DEBUG   Aborting wipeServer. Server responded with 400 response for https://stage.sync.nonprod.cloudops.mozgcp.net/1.5/xxx/storage/
1605700992125   Sync.Resource   WARN    Got exception parsing response body: SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data(resource://services-sync/resource.js:229:21) JS Stack trace: _processResponse/<@resource.js:229:21

```

jrconlin:
Closes: #928

## Description

Use more permissive CORS settings.

## Testing

Using the `tools/example` scripts, try submitting a request with a blank or invalid ORIGIN header. 

## Issue(s)

Closes #928.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -93,7 +93,7 @@ dependencies = [
  "serde_urlencoded",
  "sha-1",
  "slab",
- "time 0.2.22",
+ "time 0.2.23",
 ]
 
 [[package]]
@@ -259,7 +259,7 @@ dependencies = [
  "serde_json",
  "serde_urlencoded",
  "socket2",
- "time 0.2.22",
+ "time 0.2.23",
  "tinyvec 1.0.1",
  "url 2.1.1",
 ]
@@ -661,7 +661,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1373a16a4937bc34efec7b391f9c1500c30b8478a701a4f44c9165cc0475a6e0"
 dependencies = [
  "percent-encoding 2.1.0",
- "time 0.2.22",
+ "time 0.2.23",
  "version_check 0.9.2",
 ]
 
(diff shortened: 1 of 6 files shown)


pjenvey (approved):
I'm fine with merging this but I'm getting the impression `permissive()` makes the cors middleware essentially do nothing (i.e. we might as well not even use it).

Maybe let's merge this but keep an issue open for revisiting it later?

and from Rob (on discord):

"is the new 0.5 actix-cors permissive() equivalent to the 0.4 default()?"

-> not really, the default before was a mix of both but you had to look at the implementation to see what it was really doing
-> the new system is either everything allowed or as little as possible allowed so there's no ambiguity

pjenvey:
Also I'm surprised this survived cargo audit, there's a newer actix-cors out to avoid a new RUSTSEC-2020-0059. My local run of cargo audit was indeed triggered on it by this branch

jrconlin:
1) Yeah `permissive()` is very permissive. Since we're impacting users in the field, i figured that falling back to the nginx CORS wrappers would be a reasonable substitute for now. We can address things better later.

2) I force up `time` in Cargo.toml to address that, since the local build wasn't picking up the change (possibly related to a sub-dependency).

jrconlin:
see https://github.com/mozilla-services/syncstorage-rs/issues/930

PR #966, opened 2021-01-04
Chore/update

jrconlin:
## Description

Jan `21 Update:
New year, new rust, grpcio, etc.

## Testing

update (passing CI sufficient)

## Issue(s)

NA


vendor/mozilla-rust-sdk/googleapis/Cargo.toml
```diff
@@ -5,9 +5,9 @@ authors = ["Ferrous Systems"]
...
 
 [dependencies]
-futures = "0.3.5"
-grpcio = "0.6.0"
+futures = "0.3.8"
+grpcio = "0.7.1"
 
 [dependencies.googleapis-raw]
-version = "0.0.4"
+version = "0.0.5"
```
tublitzed, line 12: noticing now that we've got the `mozilla-rust-sdk` vendored in here and are making minor upgrades...what do you want to do about this separate repo? https://github.com/mozilla-services/mozilla-rust-sdk - should we archive it? Or start pointing to it, and updating those deps instead to avoid having to do this in two places?

jrconlin: Honestly, not really sure. Technically, our https://github.com/mozilla-services/mozilla-rust-sdk repo goes away once google's goes live, and then we'd just point to that one. We could just make ours "live", but we'd have to export the crate if we want to do it correctly, and that raises a bunch of issues and flags I'm not 100% sure we want (we fought pretty hard to NOT take ownership of this library). Of course, we could make it more a community support thing as well, which might help out. 

For now, I'm fine doing the occasional update. There hasn't been a major hit yet, and the scripts to update things are fairly automatic.

tublitzed (approved):
LGTM - thanks for grabbing this. Mind making sure we have a ticket on file for tracking these kind of things in future? (Helps with points/etc and reviewing the board down the road). In the meantime though, for this one no worries.

🚢

pjenvey (approved):

PR #970, opened 2021-01-06
feat: use a "running count" for payload quotas

### LINKED ISSUE #969: Quota: calculate payload sizes more dynamically
The spanner mutation system does not read your writes, meaning that calculating the quota after processing inserts and updates will not return an accurate size. @pjenvey had suggested doing a "running size" method, which is a bit trickier to calculate, due to the differing sizes of payloads, but would lead to a more accurate total payload assessment and would limit the amount of potential overage.

jrconlin:
## Description

This uses a "running count" of payload lengths in order to more accurately determine quota overages for spanner, since spanner mutations do not allow RYW. 

## Testing

start a sync server with quota enabled (e.g. 

```bash
#!/bin/bash
# Run this script in the project root
args=$*
# RUST_LOG=debug
SECRET=secret0
FEATURES="--features grpcio/openssl"
    RUST_LOG="debug" \
    SYNC_MASTER_SECRET=$SECRET \
    \
    SYNC_ENABLE_QUOTA=true \
    SYNC_ENFORCE_QUOTA=true \
    SYNC_LIMITS__MAX_QUOTA_LIMIT=10000 \
    SYNC_LIMITS__MAX_POST_BYTES=10000 \
    SYNC_LIMITS__MAX_RECORD_PAYLOAD_BYTES=10000 \
    \
    GOOGLE_APPLICATION_CREDENTIALS=keys/sync-spanner-dev.json \
    cargo run $FEATURES -- --config spanner_sync_jr.ini $args
```
)

create a payload `data_file` that is at the quota (e.g. a JSON file that contains:
```json
{"id": "womble", "sortindex":0, "ttl":86400, "payload":"`a*10_000`"}
```

Send the data file via curl (e.g. in the `./tools/examples` directory:

test_quota.sh
```bash
#!/bin/bash
NODE="http://localhost:8000"
ID="6384991"
URI="/1.5/$ID/storage/quota/DEADBEEF$1"
METHOD="PUT"
SYNC_MASTER_SECRET=secret0
AUTH=`../venv/bin/python ../hawk/make_hawk_token.py --node $NODE --uid $ID --uri $URI --method $METHOD --secret=$SYNC_MASTER_SECRET --as_header`
curl -vv -X PUT "$NODE$URI" \
    -H "$AUTH" \
    -H 'Content-Type: application/json' \
    -H 'Accept: application/json' \
    --data-binary @data_file
```

calling `test_quota.sh 1` should return a 200.
(note, calling `test_quota.sh 1` a second time may return a 200, as the bso body is replaced.)
calling `test_quota.sh 2` (or subsequent) should return a 403 with a body of `14`.

## Issue(s)

Closes #969.


tublitzed:
Closing for now until we decide to re-prioritize the quota work.

Issue #945, opened 2020-12-03
Investigate 0.8.2 node hitting max db connections


pjenvey:
Following the 0.8.2 deploy yesterday a node became "stuck" with the max (30) number of db connections.

<img width="289" alt="Screen Shot 2020-12-03 at 3 34 30 PM" src="https://user-images.githubusercontent.com/14639/101101285-16860c80-357d-11eb-91c6-be17157b3a04.png">

The node continued serving requests until ops manually killed it, but its response times were incredibly slow (e.g. mostly in double digits seconds). Its affect on the overall 95th percentile:

<img width="593" alt="Screen Shot 2020-12-03 at 3 35 05 PM" src="https://user-images.githubusercontent.com/14639/101101445-75e41c80-357d-11eb-8cc1-7c9da5281418.png">

(and it triggered the nginix GET dashboard alarm).

pjenvey:
Documenting more detail about these incidents [here](https://docs.google.com/document/d/1DOFcJ64HgSBEnCdUp4UyrDspNDqVA6D_7PIVYt5fLXk/edit#)

tublitzed:
We haven't hit this in prod since https://github.com/mozilla-services/syncstorage-rs/pull/985 rolled out - closing.

PR #974, opened 2021-01-15
feat: default to timing out deadpool checkouts (30 seconds)

### LINKED ISSUE #973: Timeout deadpool connection checkouts
deadpool does not default to any timeout for checking out a connection from the Pool.

I suspect that during spikes of traffic this could contribute to the [lockup w/ max db connections issue](https://github.com/mozilla-services/syncstorage-rs/issues/945).

A large number of requests become queued, waiting for a db checkout with none available. Without any timeouts over any duration of time, it could be difficult for a busy service to clear this queue. Requests continue to succeed during the lock up but with high request times. With a timeout this queue should have a better chance of eventually clearing.

We can probably start with the common in many libs default of at least 30 with an accompanying setting. The alarm triggered by this issue occurs after a 10 minute incident.

pjenvey:
## Description

and add a database_pool_connection_timout setting for it

## Testing

On prod (#945)

## Issue(s)

Closes #973

### DIFF
--- a/src/db/mysql/pool.rs
+++ b/src/db/mysql/pool.rs
@@ -6,6 +6,7 @@ use std::{
     collections::HashMap,
     fmt,
     sync::{Arc, RwLock},
+    time::Duration,
 };
 
 use diesel::{
@@ -70,6 +71,9 @@ impl MysqlDbPool {
         let manager = ConnectionManager::<MysqlConnection>::new(settings.database_url.clone());
         let builder = Pool::builder()
             .max_size(settings.database_pool_max_size.unwrap_or(10))
+            .connection_timeout(Duration::from_secs(
+                settings.database_pool_max_size.unwrap_or(30) as u64,
+            ))
             .min_idle(settings.database_pool_min_idle);
 
         #[cfg(test)]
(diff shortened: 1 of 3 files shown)


pjenvey:
deadpool also has a timeout for establishing a new connection and recycling one. I could see a timeout on the former being useful too but I doubt establishing a spanner connection ever significantly lags.

I took down some related notes [here](https://docs.google.com/document/d/1DOFcJ64HgSBEnCdUp4UyrDspNDqVA6D_7PIVYt5fLXk/edit#)

src/db/mysql/pool.rs
```diff
@@ -70,6 +71,9 @@ impl MysqlDbPool {
         let manager = ConnectionManager::<MysqlConnection>::new(settings.database_url.clone());
         let builder = Pool::builder()
             .max_size(settings.database_pool_max_size.unwrap_or(10))
+            .connection_timeout(Duration::from_secs(
+                settings.database_pool_max_size.unwrap_or(30) as u64,
```
jrconlin, line 75: Kinda think we should use a different setting for this rather than overload what pool max size.

jrconlin (changes requested):

src/db/mysql/pool.rs
jrconlin: Ah, looks like this is a typo. I see `database_pool_max_size` defined later.

jrconlin (approved):

PR #980, opened 2021-01-23
 bug: write bsos contained within a commit after the batch has been commited.

### LINKED ISSUE #822: Add tool to generate bookmarks
The quota system needs bookmarks. Add a tool to (reasonably) quickly generate gobs of them.

### LINKED ISSUE #882: Batch commit causes INVALID_ARGUMENT Row.. was already created in table bsos
Batch commit can trigger an INSERT of the same row twice in the same transaction:

https://sentry.prod.mozaws.net/operations/syncstorage-prod/issues/9910134/

`A database error occurred: RpcFailure: 3-INVALID_ARGUMENT Row [<fxa_uid>,<fxa_kid>,<collection_id>,<bso_id>] in table bsos was already created in this transaction.`

The cause seems to be the older [post_bsos]( #353) optimization: when pending bsos are included in a batch commit request, we write them to the `bsos` table immediately via `post_bsos`, bypassing the batch table.

The problem being `post_bsos` writes in bulk mutations whereas batch commit uses plain sql. [Mutations results are buffered and thus not visible in the transaction](https://cloud.google.com/spanner/docs/dml-best-practices#dml-and-mutation).

E.g. `post_bsos` issues an insert mutation for a `bso_id` that doesn't exist, then batch commit inserts its own version of the same `bso_id` (a duplicate). The buffered insert mutation is finally applied at the end of the request with the batch commit version of the bso having already been created, invalidating the insert mutation.

The given test case (which should be added to the e2e tests with a fix for this, as the unit tests don't use mutations), we would fail with this error:

- Create a batch with bso_id 'b0'
- Issue a batch commit request including a duplicate bso_id 'b0' in the same payload

A good solution here might just be to issue the `post_bsos` *after* the batch commit instead of where the batch append would have been issued.

jrconlin:
## Description

The client may sometimes include bsos in the batch
commit message. The problem is that due to the way
that data is written to spanner, mutations do not
retain the ability to see data previously written
in the same transaction. This causes collisions.

To solve this, treat the bsos included in the commit
as another batch update, then commit all the data.
This does run the risk of bumping up against the
mutation limit, but it ensures the best chance of
data consistency.

Writing the commit bsos prior to batch commit will
result in the "newer" records being overwritten by
"older" ones in the batch.

Writing the commit bsos after the batch commit runs
the same mutation index conflict problem.

## Testing

Unit test included

## Issue(s)

Closes #882

### DIFF
--- a/src/web/handlers.rs
+++ b/src/web/handlers.rs
@@ -323,27 +323,14 @@ pub async fn post_collection_batch(
 
     let mut success = vec![];
     let mut failed = coll.bsos.invalid;
-    // Option #2, mark each bso.id included in the "commit" as "failing" so that they're
-    // resubmitted.
     let bso_ids: Vec<_> = coll.bsos.valid.iter().map(|bso| bso.id.clone()).collect();
 
     let mut resp: Value = json!({});
 
-    if !breq.commit {
-        if !coll.bsos.valid.is_empty() {
-            let result = {
-                dbg!("Batch: Appending to {}", &new_batch.id);
-                db.append_to_batch(params::AppendToBatch {
-                    user_id: coll.user_id.clone(),
-                    collection: coll.collection.clone(),
-                    batch: new_batch.clone(),
-                    bsos: coll.bsos.valid.into_iter().map(From::from).collect(),
-                })
-                .await
-            };
-
+    macro_rules! handle_result {
             // collect up the successful and failed bso_ids into a response.
-            match result {
+            ( $r: expr) => {
+            match $r {
                 Ok(_) => success.extend(bso_ids.clone()),
                 Err(e) if e.is_conflict() => return Err(e.into()),
                 Err(apperr) => {
@@ -363,7 +350,28 @@ pub async fn post_collection_batch(
                 }
             };
         }
+    }
 
+    // If we're not committing the current set of records yet.
+    if !breq.commit {
+        // and there are bsos included in this message.
+        if !coll.bsos.valid.is_empty() {
+            // Append the data to the requested batch.
+            let result = {
+                dbg!("Batch: Appending to {}", &new_batch.id);
+                db.append_to_batch(params::AppendToBatch {
+                    user_id: coll.user_id.clone(),
+                    collection: coll.collection.clone(),
+                    batch: new_batch.clone(),
+                    bsos: coll.bsos.valid.into_iter().map(From::from).collect(),
+                })
+                .await
+            };
+            handle_result!(result);
+        }
+
+        // Return the batch append response without committing the current
+        // batch to the BSO table.
         resp["success"] = json!(success);
         resp["failed"] = json!(failed);
 
@@ -382,6 +390,8 @@ pub async fn post_collection_batch(
 
     // TODO: validate *actual* sizes of the batch items
     // (max_total_records, max_total_bytes)
+    //
+    // First, write the pending batch BSO data into the BSO table.
     let modified = if let Some(batch) = batch {
         db.commit_batch(params::CommitBatch {
             user_id: user_id.clone(),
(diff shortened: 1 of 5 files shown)


src/web/handlers.rs
```diff
@@ -318,77 +318,55 @@ pub async fn post_collection_batch(
         .await?
     };
 
-    let commit = breq.commit;
     let user_id = coll.user_id.clone();
     let collection = coll.collection.clone();
 
     let mut success = vec![];
     let mut failed = coll.bsos.invalid;
+    // Option #2, mark each bso.id included in the "commit" as "failing" so that they're
```
pjenvey, line 326: kill this stale comment, it's about the old version of this patch. the commit message also needs an update

src/web/handlers.rs
```diff
@@ -416,6 +394,54 @@ pub async fn post_collection_batch(
         return Err(ApiError::from(err).into());
     };
 
+    // write the BSOs contained in the commit request.
+    if !coll.bsos.valid.is_empty() {
```
pjenvey, line 398: should only happen on commit:

```suggestion
    if commit && !coll.bsos.valid.is_empty() {
```

src/web/handlers.rs
```diff
@@ -416,6 +394,54 @@ pub async fn post_collection_batch(
...
+                        ttl: batch_bso.ttl,
+                    })
+                    .collect(),
+                for_batch: false,
+                failed: Default::default(),
+            })
+            .await
+            .map(|_| ());
+
+        match result {
```
pjenvey, line 421: let's throw this block into a method or closure to avoid duplicating it

pjenvey (dismissed):

src/web/handlers.rs
```diff
@@ -318,77 +318,55 @@ pub async fn post_collection_batch(
...
-                .into_iter()
-                .map(|batch_bso| params::PostCollectionBso {
-                    id: batch_bso.id,
-                    sortindex: batch_bso.sortindex,
-                    payload: batch_bso.payload,
-                    ttl: batch_bso.ttl,
+    let mut resp: Value = json!({});
+
+    if !breq.commit {
+        if !coll.bsos.valid.is_empty() {
```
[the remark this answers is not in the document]
jrconlin: I'm concerned about this. `!breq.commit` includes both the append call at line 333 , and building the possible response at line 367. Collapsing the if skips the possible early return for requests that do not specify a commit, and run the risk of committing early.

src/web/handlers.rs
```diff
@@ -318,77 +318,55 @@ pub async fn post_collection_batch(
...
-                .valid
-                .into_iter()
-                .map(|batch_bso| params::PostCollectionBso {
-                    id: batch_bso.id,
-                    sortindex: batch_bso.sortindex,
-                    payload: batch_bso.payload,
-                    ttl: batch_bso.ttl,
+    let mut resp: Value = json!({});
+
+    if !breq.commit {
```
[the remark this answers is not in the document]
jrconlin: I agree. Going to over comment this section since I agree that it's easy to get lost with what's going on.

src/web/handlers.rs
jrconlin: I agree, that's why I was asking about the change to line 321 which handles the "not commit" messages. That should capture all non-commit messages, no? Meaning that anything that got as far as here should be a commit. I could make things a bit clearer by wrapping this whole section in an `else` for `!breq.commit`, but I'm not sure that matches the style we've used before.

tublitzed:
@jrconlin - is this related to #822 - I see it's linked above?

jrconlin:
Yes, slightly. A byproduct of that issue generated a number of the sort of conflicts that this issue addresses.

tublitzed:
Hmmm. That issue was a feature request that was never merged? Are we talking about the same https://github.com/mozilla-services/syncstorage-rs/issues/822 ?

jrconlin:
Ah, no, you're right. It should refer to https://github.com/mozilla-services/syncstorage-rs/issues/882

src/web/handlers.rs
```diff
@@ -418,6 +406,41 @@ pub async fn post_collection_batch(
...
+                    .collect(),
+                for_batch: false,
+                failed: Default::default(),
+            })
+            .await
+            .map(|_| ());
+
+        handle_result!(result);
+
+        resp["success"] = json!(success);
```
pjenvey, line 440: One last thing I noticed: let's move these 2 lines setting success/failure after the `is_empty` block. They should be always included in the response, even on a commit where they'd be empty

pjenvey (changes requested):

pjenvey (approved):

pjenvey (approved):

PR #981, opened 2021-01-26
refactor: remove duplicate code for incrementing counters

56quarters:
Delegate to count method and use glob import for metric traits

### DIFF
--- a/src/server/metrics.rs
+++ b/src/server/metrics.rs
@@ -2,8 +2,9 @@ use std::net::UdpSocket;
 use std::time::Instant;
 
 use actix_web::{error::ErrorInternalServerError, web::Data, Error, HttpRequest};
+use cadence::prelude::*;
 use cadence::{
-    BufferedUdpMetricSink, Counted, Metric, NopMetricSink, QueuingMetricSink, StatsdClient, Timed,
+    BufferedUdpMetricSink, Metric, NopMetricSink, QueuingMetricSink, StatsdClient,
 };
 
 use crate::error::ApiError;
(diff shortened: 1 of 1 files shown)


src/server/metrics.rs
```diff
@@ -2,9 +2,8 @@ use std::net::UdpSocket;
 use std::time::Instant;
 
 use actix_web::{error::ErrorInternalServerError, web::Data, Error, HttpRequest};
-use cadence::{
-    BufferedUdpMetricSink, Counted, Metric, NopMetricSink, QueuingMetricSink, StatsdClient, Timed,
-};
+use cadence::prelude::*;
```
jrconlin, line 5: We prefer not to do global includes like this for a number of reasons (mostly, it's easier for folk to find the source of a given method or structure if we note where it comes from, avoids some accidental collisions or usages, etc.).
Could you list the structs and methods you're actually using from inside of `cadence::prelude`?

jrconlin (changes requested):
Hi!
Thanks for this. It looks good, with one style nit.

src/server/metrics.rs
56quarters: Sure, will do.

jrconlin (approved):

56quarters:
Working on fixing commit signing. Sorry about that.

jrconlin:
Sigh...
For security reasons, we need to have every commit associated with the PR signed. The original commit was not. 

My apologies, but could you either rebase this PR to a signed commit or resubmit with all the commits signed? If this is too annoying, let me know and I'll resubmit for you, (with appropriate credit).

Sorry for the hassle.

jrconlin:
Oh, man, NO NEED to apologize for us making things harder.

Seriously, thank you for taking the time to submit a PR. It means a lot.

jrconlin (approved):

jrconlin:
@56quarters Just in case you don't want to deal with github being annoying, I resubmitted your patch in #983. 

Feel free to review, comment, and point out what I may have gotten wrong.

56quarters:
Closed in favor of #983

PR #997, opened 2021-02-02
feat: `__lbheartbeat__` will return 500 if the connection pool is exhausted

### LINKED ISSUE #996: `__lbheartbeat__` should return 500 if connection pool is exhausted
Per Ops request,

`__lbheartbeat__` should return 500 if the connection pool is exhausted.

jrconlin:
Closes #996

## Description

Return a 500 error if the connection pool is exhausted

## Testing

Includes unit test

## Issue(s)

Closes #996

### DIFF
--- a/src/web/handlers.rs
+++ b/src/web/handlers.rs
@@ -555,14 +555,41 @@ pub async fn lbheartbeat(req: HttpRequest) -> Result<HttpResponse, Error> {
 
     let deadarc = state.deadman.clone();
     let mut deadman = *deadarc.read().await;
-    let db_state = state.db_pool.clone().state();
+    let db_state = if cfg!(test) {
+        use crate::db::results::PoolState;
+        use actix_web::http::header::HeaderValue;
+        use std::str::FromStr;
+
+        let test_pool = PoolState {
+            connections: u32::from_str(
+                req.headers()
+                    .get("TEST_CONNECTIONS")
+                    .unwrap_or(&HeaderValue::from_static("0"))
+                    .to_str()
+                    .unwrap_or("0"),
+            )
+            .unwrap_or_default(),
+            idle_connections: u32::from_str(
+                req.headers()
+                    .get("TEST_IDLES")
+                    .unwrap_or(&HeaderValue::from_static("0"))
+                    .to_str()
+                    .unwrap_or("0"),
+            )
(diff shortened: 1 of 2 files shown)


src/web/handlers.rs
```diff
@@ -553,15 +555,44 @@ pub async fn lbheartbeat(req: HttpRequest) -> Result<HttpResponse, Error> {
...
     let active = db_state.connections - db_state.idle_connections;
+    let mut status_code = StatusCode::OK;
 
     if let Some(max_size) = deadman.max_size {
         if active >= max_size && db_state.idle_connections == 0 {
-            if deadman.previous_count > 0 {
+            if deadman.clock_start.is_none() {
                 deadman.clock_start = Some(time::Instant::now());
             }
+            status_code = StatusCode::INTERNAL_SERVER_ERROR;
```
pjenvey, line 595: Maybe only set the 500 status in an else block? Might be safer avoiding healthy nodes who happen to hit the max conns only once reporting 500 here.

```suggestion
            } else {
                status_code = StatusCode::INTERNAL_SERVER_ERROR;
            }                
```

pjenvey (commented):
Would a combination of this and the previous method (checking intermittently in the pool periodic reporter but then ultimately reporting that status in the health check) potentially be more accurate?

Though I'm fine to try this assuming the probe can essentially do the same -- decide a pod is bad when this it sees __lbheartbeat__ repeatedly report unhealthyness over some certain amount of time

src/web/handlers.rs
jrconlin: Yeah, that was the reasoning behind adding the `duration_ms`. I'm kinda/sorta ok if a saturated node returns a 500 early, because it can't really process more. Things reset fairly quickly once things become desaturated, and I think it takes multiple 500s returns before a node is terminated. We might want to talk with @Micheletto to see if that matches what he wants. I'd also be fine returning the 500 only after the saturation passes some `Duration`.

jrconlin: Based off of conversations with ops, returning 500 immediately is preferred. Monitoring will only kill a node once multiple 500 reports in a row are detected.

pjenvey (approved):

Micheletto (commented):
I know this is merged, but for the record: r+ from ops.

Issue #861, opened 2020-10-14
Spanner emits 14-UNAVAILABLE GOAWAY


pjenvey:
We began seeing `14-UNAVAILABLE GOAWAY received` errors after the 0.5.x release to prod, e.g.:

https://sentry.prod.mozaws.net/operations/syncstorage-stage/issues/9186276/

These result in 500 errors (which in turn are 503s from nginx).

pjenvey:
We have an issue open with google, inquiring on the best way to deal with this and if GOAWAYS are some indication of misbehavior on our end.

pjenvey:
Some information on GOAWAYS from gRPC's perspective:

- Should we retry on the existing connection? Or disconnect and retry on a separate connection (the go client seems to do the latter)?

GOAWAY indicates that the server wants the client to send no new RPCs on this connection. If the client has more RPCs to send, it should open a new connection. It is OK to wait for existing RPCs to finish before closing the connection.


- Any idea why we haven't seen these before? Does it make sense that we're seeing them this frequently? (is this a bug? there's a recent grpc fix related to GOAWAYs, related?: https://github.com/grpc/grpc/pull/23145

The GOAWAY is server behavior independent of the client, but client behavior such as keeping connections around for longer can trigger it. Many servers send GOAWAY after a fixed duration (say 1 hour) to force clients to reconnect. It is also common for servers to send GOAWAY after a period of inactivity (say 10 minutes) on the connection.


- What does GOAWAY mean besides "retry"? Does this signal something specific e.g. are we misusing our connections: using too many or holding too many idle connections?

See https://tools.ietf.org/html/rfc7540#section-6.8
```

pjenvey:
To reduce the number of GOAWAYS we should try:

- [ ] [in progress] Reducing our total number of connections. It's possible it would help (though it may not do much w/ deadpool's FIFO behavior). It may at least reduce a few occurrences of GOAWAYS due to inactivity, which might cause 503 errors QA has seen on an idle stage cluster (#846)
- [ ] Implement equivalent to r2d2/bb8's `Builder::max_lifetime` (unsupported by deadpool)
- [ ] Implement equivalent to r2d2/bb8's `Builder::idle_timeout` (unsupported by deadpool)

pjenvey:
Reducing the total number of connections has helped the number of GOAWAYS, one extreme example:

300 max conns:
https://sentry.prod.mozaws.net/operations/syncstorage-prod/issues/9482220/

[...]

PR #1006, opened 2021-02-12
feat: kill old or excessively idled connections

jrconlin:
## Description

* Two new optional arguments:

 _database_pool_connection_lifespan_ = max connection lifespan (in seconds)
 _database_pool_connection_max_idle_ = max idle time (in seconds)

I've also added two elements to the `extra` Tags which record the `age` and `idle` times for a connection that triggers an RPC error. This should identify if there's a pattern present for these, and what periods we should consider for either the `database_pool_connection_lifespan` and `database_pool_connection_max_idle` settings.

## Testing

A bit difficult, since it involves triggering a GOAWAY event, and we're not quite sure what does that.

## Issue(s)

Issue: #861


src/db/transaction.rs
```diff
@@ -144,12 +147,25 @@ impl DbTransactionPool {
...
         };
 
-        let (resp, db) = self.transaction_internal(check_precondition).await?;
+        let (mut resp, db) = self.transaction_internal(check_precondition).await?;
 
         // HttpResponse can contain an internal error
         match resp.error() {
             None => db.commit().await?,
-            Some(_) => db.rollback().await?,
+            Some(_) => {
```
pjenvey, line 155: Unfortunately this `Some(_)` case doesn't cover many errors, only "internal" ones in HttpResponse (not even sure what creates those, maybe calls like `HttpResponse:: InternalServerError`? maybe nothing in this code base).

Most errors will short circuit through line 150 (via the `?`) -- I'm not sure how we can easily leave `Tags::extra` for those cases, I don't think we have a req/resp on hand 🤷‍♂️  -- I'd have to think about this more

src/settings.rs
```diff
@@ -52,6 +52,10 @@ pub struct Settings {
     pub database_pool_min_idle: Option<u32>,
     /// Pool timeout when waiting for a slot to become available, in seconds
     pub database_pool_connection_timeout: Option<u32>,
+    /// Max age a given connection should live.
```
pjenvey, line 55: I think these are in seconds, let's mention that here in these comments

src/db/spanner/manager/session.rs
```diff
@@ -63,7 +67,23 @@ pub async fn create_spanner_session(
...
+    max_lifetime: Option<u32>,
 ) -> Result<(), DbError> {
+    if let Some(max_life) = max_lifetime {
+        // get the current UTC seconds
+        let now = SystemTime::now()
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .unwrap_or_default()
+            .as_secs();
+        if let Some(birth) = conn.session.create_time.clone().into_option() {
+            let birth = now - birth.seconds as u64;
```
pjenvey, line 79: let's call this var age instead of birth

src/db/spanner/manager/deadpool.rs
```diff
@@ -52,24 +55,47 @@ impl SpannerSessionManager {
...
+                .approximate_last_use_time
+                .clone()
+                .into_option()
+            {
+                // get current UTC seconds
+                let now = SystemTime::now()
+                    .duration_since(SystemTime::UNIX_EPOCH)
+                    .unwrap_or_default()
+                    .as_secs();
+                let idle = std::cmp::max(0, now as i64 - idle.seconds);
```
pjenvey, line 87: Is there a reason for i64 here vs u64 in the max_lifetime check code?

Keeping everything as a higher level `Duration` type might be a bit nicer and avoid some of this casting -- except the grpc `Timeout` uses signed values whereas `Duration` uses unsigned, sigh.

src/db/spanner/manager/session.rs
```diff
@@ -63,7 +67,23 @@ pub async fn create_spanner_session(
 pub async fn recycle_spanner_session(
     conn: &mut SpannerSession,
     database_name: &str,
+    max_lifetime: Option<u32>,
 ) -> Result<(), DbError> {
+    if let Some(max_life) = max_lifetime {
```
pjenvey, line 72: Adding a metric for both expiring cases might be handy. The Manager already has a `metrics` instance.

pjenvey (changes requested):
The Session metadata was a good find but in retrospect I'd be more comfortable tracking these timestamps ourselves, since we're relying on spanner's clock being roughly close to our system time (obviously more concerned about the latter 🤣 ). But I'll admit it probably isn't going to be an issue, so the metadata's fine.

That definitely does make me want to solve the `extra` metadata for more feedback

src/db/spanner/manager/deadpool.rs
jrconlin: yeah, pretty much that's the reason.

src/db/spanner/manager/session.rs
jrconlin: oh, cool! yeah, will do.

src/db/transaction.rs
jrconlin: I'm kinda/sorta playing around with the idea of adding the tags to the error so that they get propagated up that way, but it's getting a bit messy right now. If I come up with something tomorrow, I'll add it.

jrconlin: Added the tags to the passed-in request so they should automagically be added to the next sentry error call.

src/db/transaction.rs
```diff
@@ -177,6 +207,7 @@ impl FromRequest for DbTransactionPool {
         }
 
         let req = req.clone();
+        // let treq = req.clone();
```
pjenvey, line 210: kill and line 258

src/web/handlers.rs
```diff
@@ -31,148 +32,177 @@ pub const ONE_KB: f64 = 1024.0;
...
     db_pool
-        .transaction_http(|db| async move {
-            meta.metrics.incr("request.get_collections");
-            let result = db.get_collection_timestamps(meta.user_id).await?;
-
-            Ok(HttpResponse::build(StatusCode::OK)
-                .header(X_WEAVE_RECORDS, result.len().to_string())
-                .json(result))
-        })
+        .transaction_http(
```
pjenvey, line 38: can we make `request` the first argument to `transaction_http` instead? It kind of gets lost to my eyes since we pass the large closure inline everywhere in this file. I think it'll significantly reduce the diff from master too.

src/db/transaction.rs
```diff
@@ -144,12 +168,18 @@ impl DbTransactionPool {
             }
         };
 
-        let (resp, db) = self.transaction_internal(check_precondition).await?;
+        let (mut resp, db) = self
+            .transaction_internal(check_precondition, request.clone())
+            .await?;
+        // match on error and return a composed HttpResponse (so we can use the tags?)
```
pjenvey, line 174: Let's either a) only call `set_extra` on the error cases (which would be here as this comment describes and also below when there's a `resp.error()` OR b) call it on every single request (which is what the line 152 call is pretty much doing).

We're kinda doing both here. a) is probably nicer but b) might be ok, it's pretty cheap I think -- but in that case maybe call it at the beginning of the check_precondition closure as e.g. `extract_resource` could potentially error out before it was called

pjenvey (changes requested):

src/web/handlers.rs
jrconlin: (something, something, add arguments to the end of the list to avoid unintentional value setting, something, something...)

src/db/transaction.rs
jrconlin: Adding some commentary around the `set_extra` calls.

pjenvey (approved):

Issue #1029, opened 2021-03-19
Invalid value in BSOQueryParams


jrconlin:
An error is being generated during deserialization of the BSO Query parameters. I'm not sure which value is triggering it, but it appears to be one of the numerics. 

https://sentry.prod.mozaws.net/operations/syncstorage-prod/issues/10887369/?referrer=github_plugin

```
ApiError: Query deserialize error: Invalid value
  File "error.rs:331", line 17, in <syncstorage::error::ApiError as core::convert::From<syncstorage::error::ApiErrorKind>>::from
  File "mod.rs:546", line 9, in <T as core::convert::Into<U>>::into
  File "error.rs:343", line 17, in <syncstorage::error::ApiError as core::convert::From<syncstorage::web::error::ValidationError>>::from
  File "mod.rs:546", line 9, in <T as core::convert::Into<U>>::into
  File "error.rs:184", line 9, in syncstorage::web::error::<impl core::convert::From<syncstorage::web::error::ValidationErrorKind> for syncstorage::error::ApiError>::from
...
(112 additional frame(s) were not displayed)
```

pjenvey:
Possibly from `SyncTimestamp::from_header`

jrconlin:
Yeah, possibly. It could be related to a bad date format or something.

To be honest, I'm starting to not be a fan of the serde habit of just puking when it hits an error, particularly when we don't have a reproduceable state. I may add an `extra` field for this error that shows what got fed to serde to at least help isolate some of it.

pjenvey:
I'm no longer seeing any "Invalid Value" errors in sentry, so closing this out

PR #1030, opened 2021-03-19
bug: report query parameters with Invalid Value error

jrconlin:
## Description

This is not a resolution. Instead, this will report the query parameters
that lead to a serde parse failure so we can better isolate potential
problems.

(also fixes a typo in a logging message extra parameter)

## Testing

Invalid values in query params that trigger serde parse errors should result in sentry messages containing the query parameters

## Issue(s)

Issue #1029

### DIFF
--- a/src/web/extractors.rs
+++ b/src/web/extractors.rs
@@ -34,7 +34,6 @@ use crate::settings::Secrets;
 use crate::web::{
     auth::HawkPayload,
     error::{HawkErrorKind, ValidationErrorKind},
-    tags::Tags,
     X_WEAVE_RECORDS,
 };
 const BATCH_MAX_IDS: usize = 100;
@@ -2302,15 +2301,15 @@ mod tests {
 
     #[actix_rt::test]
     async fn test_offset() {
-        let sample_offset = Offset{
+        let sample_offset = Offset {
             timestamp: Some(SyncTimestamp::default()),
-            offset: 1234
+            offset: 1234,
         };
 
         //Issue559: only use offset, don't use timestamp, even if set.
-        let test_offset = Offset{
+        let test_offset = Offset {
             timestamp: None,
-            offset: sample_offset.offset
+            offset: sample_offset.offset,
         };
 
         let offset_str = sample_offset.to_string();
--- a/src/web/extractors.rs
+++ b/src/web/extractors.rs
@@ -1122,7 +1122,7 @@ impl From<u32> for HawkIdentifier {
     }
 }
 
-#[derive(Debug, Default, Clone, Deserialize, Validate)]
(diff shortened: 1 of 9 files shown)


src/web/extractors.rs
```diff
@@ -1216,6 +1217,10 @@ impl FromRequest for BsoQueryParams {
         Box::pin(async move {
             let params = Query::<BsoQueryParams>::from_request(&req, &mut payload)
                 .map_err(|e| {
+                    let mut exts = req.extensions_mut();
+                    let mut tags = Tags::default();
+                    tags.add_extra("query_params", req.query_string());
```
pjenvey, line 1222: We already have this in sentry, see `uri.path` [in extras](https://sentry.prod.mozaws.net/operations/syncstorage-prod/issues/10887369/?referrer=github_plugin) which includes the params.

So the `BsoQueryParams` error's happening due to a client passing in an encoded `offset` (with a semicolon) and is probably happening mostly on ios. Is the client attempting to use encoded offsets despite the server not supplying them as encoded?

pjenvey (changes requested):

src/web/extractors.rs
jrconlin: Yeah, that has to be the reason. I see that there's [a comment](https://github.com/mozilla-services/syncstorage-rs/blob/master/src/web/extractors.rs#L1143-L1148) where we don't process the encoded offset . I'm guessing we must still be leaking that form of offset.

jrconlin: Yep, we're still sending out the encoded format: https://github.com/mozilla-services/syncstorage-rs/blob/master/src/web/extractors.rs#L1135

pjenvey: but AFAICT we never set a timestamp _anywhere_ after we disabled these offsets, so I don't think we're actually emitting them.

jrconlin: Well, as far as I understand, we're the only thing that generates the offset, so I'm guessing it's getting leaked somehow.

pjenvey: We can certainly remove timestamp and this code path just in case (and wait for it to deploy before looking into ios further) but I don't see it changing anything.

IIRC we've seen this error since going back to last year when we first deployed. I had a theory it was a value "leftover" from the old python server but I don't think that's a possibility: the client resets its state when assigned a new node.

Granted the old sentry event isn't around any longer and we never tracked this with an issue so I'm fine disregarding my recollection of all this, removing the code path and going from there 😄

jrconlin: For now, I commented out the timestamp aspect of the Offset. I'm not going to do the full surgery since eventually we'd need it, and that would touch A LOT of code.

pjenvey (approved):
We might wanna track this

Issue #872, opened 2020-10-21
Errors with a metric_label not reported to metrics


pjenvey:
The code block responsible for emitting metrics for errors w/ a `metric_label` is in the Db middleware (which is no longer used #693). It should probably live in either the `transaction` module or in the sentry middleware.

pjenvey:
Closed in #1033

PR #1033, opened 2021-03-26
bug: Restore hawk error metrics

### LINKED ISSUE #812: ApiError: HAWK authentication error: expired payload
Seen since the 0.5.8 deploy last week.

https://sentry.prod.mozaws.net/operations/syncstorage-prod/issues/9482067/

```
ApiError: HAWK authentication error: expired payload
  File "error.rs", line 320, in <syncstorage::error::ApiError as core::convert::From<syncstorage::error::ApiErrorKind>>::from
  File "mod.rs", line 559, in <T as core::convert::Into<U>>::into
  File "error.rs", line 332, in <syncstorage::error::ApiError as core::convert::From<syncstorage::web::error::HawkError>>::from
  File "mod.rs", line 559, in <T as core::convert::Into<U>>::into
  File "error.rs", line 147, in syncstorage::web::error::<impl core::convert::From<syncstorage::web::error::HawkErrorKind> for syncstorage::error::ApiError>::from
...
(115 additional frame(s) were not displayed)
```

jrconlin:
## Description

(Includes clippy changes for rust 1.51)

Hawk errors should be returned as metrics. During the middleware purge, this broke. This PR includes metric reporting into the sentry handler. 

## Testing

Verify that metrics are incremented for HAWK header errors.

## Issue(s)

Closes #812.

### DIFF
--- a/src/web/middleware/rejectua.rs
+++ b/src/web/middleware/rejectua.rs
@@ -53,6 +53,7 @@ where
         future::ok(RejectUAMiddleware { service })
     }
 }
+#[allow(clippy::clippy::upper_case_acronyms)]
 
 #[allow(clippy::clippy::upper_case_acronyms)]
 pub struct RejectUAMiddleware<S> {
--- a/src/web/middleware/rejectua.rs
+++ b/src/web/middleware/rejectua.rs
@@ -32,6 +32,7 @@ $
     .unwrap();
 }
 
+#[allow(clippy::clippy::upper_case_acronyms)]
 #[derive(Debug, Default)]
 pub struct RejectUA;
 
@@ -53,6 +54,7 @@ where
     }
 }
 
+#[allow(clippy::clippy::upper_case_acronyms)]
 pub struct RejectUAMiddleware<S> {
     service: S,
 }
--- a/src/web/middleware/rejectua.rs
+++ b/src/web/middleware/rejectua.rs
@@ -54,7 +54,6 @@ where
     }
 }
 #[allow(clippy::clippy::upper_case_acronyms)]
-#[allow(clippy::clippy::upper_case_acronyms)]
 pub struct RejectUAMiddleware<S> {
     service: S,
 }
--- a/src/web/middleware/rejectua.rs
+++ b/src/web/middleware/rejectua.rs
@@ -54,7 +54,6 @@ where
     }
 }
 #[allow(clippy::clippy::upper_case_acronyms)]
-
 #[allow(clippy::clippy::upper_case_acronyms)]
 pub struct RejectUAMiddleware<S> {
     service: S,
--- a/src/web/middleware/sentry.rs
+++ b/src/web/middleware/sentry.rs
@@ -165,7 +165,6 @@ where
                         }
                         if let Some(metrics) = metrics {
                             if let Some(label) = apie.kind().metric_label() {
-                                dbg!(&label);
                                 metrics.incr(&label);
                             }
                         }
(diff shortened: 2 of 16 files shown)


src/web/middleware/sentry.rs
```diff
@@ -159,6 +163,12 @@ where
                             trace!("Sentry: Not reporting error: {:?}", apie);
                             return future::ok(sresp);
                         }
+                        if let Some(metrics) = metrics {
+                            if let Some(label) = apie.kind().metric_label() {
+                                dbg!(&label);
```
mythmon, line 168: Should this `dbg!` still be here? I noticed you removed others in this PR.

mythmon (commented):
* Would it be possible to add a test that would have caught this problem, since it sounds like it was a regression? I know pretty much nothing about the test suite, so that might be unreasonable.
* Could you split the formatting changes out from the bug fixes? Having them in a separate commit in this PR would make it a lot easier to sort out what's a bug fix and what's just satisfying clippy. If it helps, I have a branch where I split the two out, but I can't sign your commits, so we'd have to work together to use that work.

jrconlin:
The problem with the test suite is that it's not that easy to build out mocks for things like metrics (we're still trying to figure out good ways of doing that, but often those tend to require significant code loads). Instead, we tend to use integration tests, and in this case, yeah, we didn't include anything to monitor outbound metrics. 

I'll see about spitting out the clippy stuff into a separate commit, sorry about that.

pjenvey:
> The problem with the test suite is that it's not that easy to build out mocks for things like metrics (we're still trying to figure out good ways of doing that, but often those tend to require significant code loads). Instead, we tend to use integration tests, and in this case, yeah, we didn't include anything to monitor outbound metrics.

I'll note it should be easier to create metrics tests now that cadence [exposed a Sink for this purpose](https://github.com/56quarters/cadence/issues/103) (thanks to JR logging that)

src/web/middleware/sentry.rs
```diff
@@ -158,6 +163,12 @@ where
                             trace!("Sentry: Not reporting error: {:?}", apie);
                             return future::ok(sresp);
                         }
+                        if let Some(metrics) = metrics {
```
pjenvey, line 166: I think we want this in the above `!is_reportable()` block: reportable means it goes to sentry. If it's not reportable, we emit a metric instead (maybe "reportable" is not the best phrasing?).

Here's the [original from DbMiddleware](https://github.com/mozilla-services/syncstorage-rs/blob/0.9.0/src/web/middleware/db.rs#L178).

pjenvey (changes requested):

src/web/middleware/sentry.rs
jrconlin: Ah, right! thanks.

pjenvey (dismissed):
looks good, this also closes #872 

One more thing I noticed, we can fix it here or in a separate issue:

We added `on_response` to ApiError a while back, it's sole use being to emit a metric for `DbError::Conflict`. There's also a metric_label defined for it so w/ #872 now fixed, we're emitting 2 Conflict metrics (and each has a different label)

pjenvey:
We should probably stick with the existing label because it's already dashboard'd

pjenvey (approved):
thanks JR

src/web/middleware/rejectua.rs
```diff
@@ -54,7 +54,7 @@ where
     }
 }
 #[allow(clippy::clippy::upper_case_acronyms)]
-
+#[allow(clippy::clippy::upper_case_acronyms)]
```
pjenvey, line 57: kill (bad merge)

pjenvey (changes requested):

pjenvey (approved):

PR #1035, opened 2021-04-02
feat: Add "auto-split" arg to auto-gen UID prefixes

### LINKED ISSUE #1034: tool: Add Auto-split fxa_uid feature to purge_ttl.py
Ops would like a feature to have purge_ttl.py automagically split the fxa_uids.

jrconlin:
Closes: #1034

## Description
adds `--auto_split` / `PURGE_AUTO_SPLIT` arg to auto-gen `uid_prefix` fields for _N_ digits, (e.g. `--auto_split=3` will generate prefixes of `["000","001","002",...,"fff"]`

## Testing

You can use `--dry_run` and verify that queries for each prefix are generated

## Issue(s)

Closes #1034.

### DIFF
--- a/tools/spanner/purge_ttl.py
+++ b/tools/spanner/purge_ttl.py
@@ -112,6 +112,10 @@ def spanner_purge(args):
     instance = client.instance(args.instance_id)
     database = instance.database(args.database_id)
     expiry_condition = get_expiry_condition(args)
+    if args.auto_split:
+        args.uid_prefixes = [
+                hex(i).lstrip("0x").zfill(args.auto_split) for i in range(
+                    0, 16 ** args.auto_split)]
     prefixes = args.uid_prefixes if args.uid_prefixes else [None]
 
     for prefix in prefixes:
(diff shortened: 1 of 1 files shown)


tools/spanner/purge_ttl.py
```diff
@@ -190,6 +194,14 @@ def get_args():
         help="Array of strings used to limit purges based on UID. "
              "Each entry is a separate purge run."
     )
+    parser.add_argument(
```
pjenvey, line 197: Is it worth adding a guard to error out if both auto_split and uid_prefixes are specified?

pjenvey (approved):

tools/spanner/purge_ttl.py
jrconlin: If this was a general use tool, absolutely. Since it's in-house, I'm willing to let it slide and just answer any questions SRE might come up with. 

If they hit it as a problem, I'll add the guards.

PR #1043, opened 2021-04-14
refactor: Use generic tuple extractor in web extractors

### LINKED ISSUE #698: Add back tuple-extractor pattern to extractors
Due to the bb8 refactor (#692), some extractors which used the generic tuple extractor were changed to extract each item individually (due to a lifetime issue):
```rust
// From...
let (user_id, collection, query, mut bsos) =
    <(
        HawkIdentifier,
        CollectionParam,
        BsoQueryParams,
        BsoBodies,
    )>::from_request(&req, &mut payload)
    .await?;
// ...to...
let user_id = HawkIdentifier::from_request(&req, &mut payload).await?;
let collection = CollectionParam::from_request(&req, &mut payload).await?;
let query = BsoQueryParams::from_request(&req, &mut payload).await?;
let mut bsos = BsoBodies::from_request(&req, &mut payload).await?;
```

The issue which prompted this change has been fixed, so we can revert back to the tuple-extractor pattern. However, it's not clear which method is more efficient. If the tuple-extractor version is better, we should revert back to that pattern.

ethowitz:
## Description

Refactors parts of the web extractors to use a generic tuple extractor pattern in Rust. There was some discussion in #698 as to which of the two variants is more performant:

Generic tuple extraction:
```rust
let (user_id, query, collection, bso, body) =
    <(
        HawkIdentifier,
        BsoQueryParams,
        CollectionParam,
        BsoParam,
        BsoBody,
    )>::from_request(&req, &mut payload)
    .await?;
```

Extracting the values separately:
```rust
let user_id = HawkIdentifier::from_request(&req, &mut payload).await?;
let collection = CollectionParam::from_request(&req, &mut payload).await?;
let query = BsoQueryParams::from_request(&req, &mut payload).await?;
let bso = BsoParam::from_request(&req, &mut payload).await?;
let body = BsoBody::from_request(&req, &mut payload).await?;
```

The idea is that in the second example above, once the request future resolves, all of the `*::from_request` function calls will happen serially and immediately, resulting in a lot of pushes/pops to/from the call stack. Since these extractors are used in the context of a web request that grabs things from the database, I'd expect any performance difference between the two patterns to be negligible. That said, I may be off-base, and I'd love to learn why if that's the case!

## Testing

This is just a refactor, and there should be no difference in functionality introduced by this change. To ensure that nothing has changed, you could either:
1. Use the sync feature in Firefox such that each of the four extractors is used without an issue
2. Test endpoints that use these extractors using curl/Postman 

## Issue(s)

Closes #698

### DIFF
--- a/src/web/extractors.rs
+++ b/src/web/extractors.rs
@@ -665,11 +665,13 @@ impl FromRequest for CollectionRequest {
         let req = req.clone();
         let mut payload = Payload::None;
         async move {
-            let user_id = HawkIdentifier::from_request(&req, &mut payload).await?;
-            let query = BsoQueryParams::from_request(&req, &mut payload).await?;
-            let collection = CollectionParam::from_request(&req, &mut payload)
-                .await?
-                .collection;
+            let (user_id, query, collection) =
+                <(HawkIdentifier, BsoQueryParams, CollectionParam)>::from_request(
+                    &req,
+                    &mut payload,
+                )
+                .await?;
+            let collection = collection.collection;
 
             let accept = get_accepted(&req, &ACCEPTED_CONTENT_TYPES, "application/json");
             let reply = match accept.as_str() {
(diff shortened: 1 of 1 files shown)


jrconlin (approved):
Using a standard format for this stuff is good, I suppose, even if it does make things a bit magical feeling for non-rust folk.

pjenvey (approved):

pjenvey:
> I'd expect any performance difference between the two patterns to be negligible.

Agreed, they probably perform similarly.

PR #1046, opened 2021-04-16
refactor: Add ToSpannerValue trait

### LINKED ISSUE #260: Improve support code (as_value/as_list_value)
Looking forward a bit into an improved high level Spanner client, our `as_value` and `as_list_value` methods could be improved.

They provided an easy way to support both client sdks simultaneously, but we could improve how types are converted to protobuf Values via a trait, similar to diesel's `ToSql` trait.

Say a `ToSpannerValue` trait with impls for basic types like String, i32 (and in our specific case, SyncTimestamp). And an impl for either `Vec<V>` or possibly a fancy one like `T: where T: Iterator<Item=V>` (both `where V: ToSpannerValue`) if possible

ethowitz:
## Description

Improves our conversion of `String`s, `i32`s, and `u32`s to protobuf `Value`s by using a `ToSpannerValue` trait. This makes adding support for the conversion of a new type to a `Value` as easy as adding a `ToSpannerValue` implementation for that type.

## Testing

I tested this using the integration tests in `tools/integration_tests/`. To do this, I:
- Changed [this line](https://github.com/mozilla-services/syncstorage-rs/blob/master/tools/integration_tests/run.py#L25) to read:
```python
    the_server_subprocess = subprocess.Popen('SYNC_MASTER_SECRET=secret0 ' + target_binary + ' --config ./config/local.toml', shell=True)
```
- Ran `python3 tools/integration_tests/run.py 'http://localhost:8000#secret0'`

There was one integration test failure:
```
======================================================================
FAIL: test_batch_size_limits (test_storage.TestStorage)
----------------------------------------------------------------------
  File "/home/edonowitz/syncstorage-rs/tools/integration_tests/test_storage.py", line 1697, in test_batch_size_limits
    self.assertFalse(res.json['failed'])
AssertionError: {'big': 'db error', 'little': 'db error'} is not false
```

I confirmed that this test was also failing on the main branch, so I don't think the changes here introduced the failure.

## Issue(s)

Closes #260

### DIFF
--- a/src/db/spanner/batch.rs
+++ b/src/db/spanner/batch.rs
@@ -355,7 +355,8 @@ pub async fn do_append_async(
         } else {
             let sortindex = bso
                 .sortindex
-                .map(|sortindex| sortindex.to_spanner_value())
+                .as_ref()
+                .map(ToSpannerValue::to_spanner_value)
                 .unwrap_or_else(null_value);
             let payload = bso
                 .payload
--- a/src/db/spanner/batch.rs
+++ b/src/db/spanner/batch.rs
@@ -359,10 +359,12 @@ pub async fn do_append_async(
                 .unwrap_or_else(null_value);
             let payload = bso
                 .payload
+                .as_ref()
                 .map(ToSpannerValue::to_spanner_value)
                 .unwrap_or_else(null_value);
             let ttl = bso
                 .ttl
+                .as_ref()
                 .map(ToSpannerValue::to_spanner_value)
                 .unwrap_or_else(null_value);
 
@@ -483,10 +485,7 @@ pub async fn do_append_async(
             };
             if let Some(sortindex) = val.sortindex {
                 fields.push("sortindex");
-                params.insert(
-                    "sortindex".to_owned(),
-                    ToSpannerValue::to_spanner_value(sortindex.to_string()),
-                );
+                params.insert("sortindex".to_owned(), sortindex.to_spanner_value());
             }
             if let Some(payload) = val.payload {
                 fields.push("payload");
--- a/src/db/spanner/batch.rs
+++ b/src/db/spanner/batch.rs
@@ -10,11 +10,8 @@ use protobuf::{
 };
 use uuid::Uuid;
 
-use super::support::{null_value, struct_type_field};
-use super::{
-    models::{Result, SpannerDb, DEFAULT_BSO_TTL, PRETOUCH_TS},
-    support::{as_list_value, as_value},
-};
+use super::models::{Result, SpannerDb, DEFAULT_BSO_TTL, PRETOUCH_TS};
+use super::support::{null_value, struct_type_field, ToSpannerValue};
 use crate::{
     db::{params, results, util::to_rfc3339, DbError, DbErrorKind, BATCH_LIFETIME},
     web::{extractors::HawkIdentifier, tags::Tags},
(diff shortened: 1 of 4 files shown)


src/db/spanner/batch.rs
```diff
@@ -355,23 +355,28 @@ pub async fn do_append_async(
...
+                .map(ToSpannerValue::to_spanner_value)
                 .unwrap_or_else(null_value);
 
             // convert to a protobuf structure for direct insertion to
             // avoid some mutation limits.
             let mut row = ListValue::new();
             row.set_values(RepeatedField::from_vec(vec![
-                as_value(user_id.fxa_uid.clone()),
-                as_value(user_id.fxa_kid.clone()),
-                as_value(collection_id.to_string()),
```
ethowitz, line 372: There were a few cases like this where I removed the call to `to_string()` and just converted directly to a Value from `i32`/`f32`. From what I could tell, this seemed to be okay/desirable, but I wasn't 100% sure, so I thought I'd mention it here

src/db/spanner/support.rs
```diff
@@ -26,10 +26,47 @@ use crate::{
...
+
+impl ToSpannerValue for String {
+    fn to_spanner_value(&self) -> Value {
+        let mut value = Value::new();
+        value.set_string_value(self.clone());
+        value
+    }
+}
+
+impl ToSpannerValue for i32 {
```
ethowitz, line 41: What I really wanted to do here was:
```rust
trait ToSpannerValue for T
where T: Into<f64> {
    ...
}
```
but I ran into an error that read "error: conflicting implementations for trait ToSpannerValue", since it's possible that someone upstream could implement the `Into<f64>` trait for `String`, which would make for an ambiguous choice between the `impl ToSpannerValue for String` and `impl ToSpannerValue for Into<f64>` traits for `String`.

src/db/spanner/batch.rs
jrconlin: We might want to check on this, because javascript is kinda notorious at screwing up parsing of numerics. Converting to string at least normalizes them.

src/db/spanner/support.rs
```diff
@@ -26,10 +26,47 @@ use crate::{
...
+
+impl ToSpannerValue for u32 {
+    fn to_spanner_value(&self) -> Value {
+        let mut value = Value::new();
+        value.set_number_value(*self as f64);
+        value
+    }
+}
+
+impl<T> ToSpannerValue for Vec<T>
```
ethowitz, line 57: Same issue as above :( What I really wanted to do was:
```rust
impl<T, V> ToSpannerValue for T
where T: Iterator<Item=V>
      V: ToSpannerValue {
    ...
}
```
but it creates ambiguity for the compiler if someone "upstream" were to implement `Iterator<Item=V>` for `String`.

src/db/spanner/batch.rs
```diff
@@ -355,23 +355,28 @@ pub async fn do_append_async(
         } else {
             let sortindex = bso
                 .sortindex
-                .map(|sortindex| as_value(sortindex.to_string()))
+                .map(|sortindex| sortindex.to_spanner_value())
```
jrconlin, line 358: Why not use the same `ToSpannerValue::to_spanner_value` method you do below?

src/db/spanner/models.rs
```diff
@@ -1527,7 +1534,12 @@ impl SpannerDb {
         };
         sqlparams.insert(
             "ids".to_owned(),
-            as_list_value(params.bsos.iter().map(|pbso| pbso.id.clone())),
+            params
+                .bsos
+                .iter()
+                .map(|pbso| pbso.id.clone())
+                .collect::<Vec<String>>()
```
ethowitz, line 1541: Since we can't call `to_spanner_value()` on an iterator, we have to collect the iterator first. I think this may result in a performance hit, but I'm not sure how significant it would be

jrconlin (approved):
We definitely want to test this on stage to make sure that the client handles the collection_id as numeric, ok, but otherwise 👍🏻

src/db/spanner/batch.rs
ethowitz: An excellent point!

jrconlin (approved):

PR #1047, opened 2021-04-19
docs: Update spanner configuration documentation

### LINKED ISSUE #1045: Update Spanner configuration documentation in README.md
The current [documentation that relates to setting up the connection to Spanner](https://github.com/mozilla-services/syncstorage-rs/tree/master#spanner) prompts users to create a new access key via the GCP service account page. Instead, let's add documentation that explains the following:

- How to install the `gcloud` CLI tool
- How to authenticate using `gcloud auth login --update-adc` to generate a JSON file with an OAuth token in the proper location
- What to name this file (`service-account.json`) to be picked up by `.gitignore`
- How to authenticate with Spanner using this OAuth token (we may want to suggest symlinking the credential file to make potential revocation easier)

ethowitz:
## Description

Updates the documentation to include the following information:
- How to install the `gcloud` CLI tool
- How to authenticate using `gcloud auth application-default login` to generate a JSON file with an OAuth token in the proper location
- How to symlink this file to be picked up by `.gitignore`
- How to revoke these credentials in case of a leak

A question for the reviewer: is it worth including the URL that points to the Spanner instance that should be used for development?

## Testing

I tested this by revoking my previous OAuth keys, running through the new workflow outlined in the docs, and ensuring that I was able to access the Spanner dev instance with the newly-generated credentials.

## Issue(s)

Closes #1045

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,15 +74,7 @@ First, install the Google Cloud command-line interface by following the instruct
 ```sh
 gcloud auth application-default login
 ```
-The above command will prompt you to visit a webpage in your browser to complete the login process. Once completed, ensure that a file called `application_default_credentials.json` has been created in the appropriate directory (on Linux, this directory is `$HOME/.config/gcloud/`). Now, create a symbolic link from this file to a file called `service-account.json` in the root of the repository:
-```sh
-ln -s /path/to/application_default_credentials.json /path/to/syncstorage-rs/service-account.json
-```
-**Make sure that the filename matches `service-account.json` exactly!** The `.gitignore` file is set to ignore that filename specifically, and a misspelling or typo could result in committing a private OAuth key to the public repository. To avoid such errors, consider installing a tool like [git secrets](https://github.com/awslabs/git-secrets), which will add a git commit hook that checks for sensitive information whenever you add a new commit.
-
-Using a symbolic link (as opposed to copying the file) has a couple of benefits:
-1. If the OAuth token is updated, the update will be reflected in `service-account.json` without additional steps.
-2. Key revocation is simpler. `gcloud auth application-default revoke` requires that the key to be revoked is stored in the correct place. *Copying* the key to another location could lead to situations in which multiple active application default credentials are present on your computer; using a symbolic link helps to ensure that the only active credentials are stored in `application_default_credentials.json`. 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -68,44 +68,40 @@ GRANT ALL PRIVILEGES on syncstorage_rs.* to sample_user@localhost;
...
-}
+The correct way to authenticate with Spanner is by generating an OAuth token and pointing your local application server to the token. In order for this to work, your Google Cloud must have the correct permissions; contact the Ops team to ensure the correct permissions are added to your account.
+
+First, install the Google Cloud command-line interface by following the instructions for your operating system [here](https://cloud.google.com/sdk/docs/install). Next, run the following to log in with your Google account (this should be the Google account associated with your Mozilla LDAP credentials):
+```sh
+gcloud auth application-default login
+```
+The above command will prompt you to visit a webpage in your browser to complete the login process. Once completed, ensure that a file called `application_default_credentials.json` has been created in the appropriate directory (on Linux, this directory is `$HOME/.config/gcloud/`). Now, create a symbolic link from this file to a file called `service-account.json` in the root of the repository:
+```sh
+ln -s /path/to/application_default_credentials.json /path/to/syncstorage-rs/service-account.json
```
pjenvey, line 79: I don't think this symlink is needed? Google's auth seems to pick up mine from the `~/.config/` location.

It looks like the original docs referenced `service-account.json` in the root directory to match the setup in `Makefile`

ethowitz: :face I'll update the README to reflect this. Do you think it's worth updating `Makefile` to remove `GOOGLE_APPLICATION_CREDENTIALS`? If there's anybody authenticating with a service account, their auth would break until they set up app default credentials

jrconlin: It might be worth modifying the README to at least note the default path, but I think it's still good to include this bit too. Folk "in a hurry" tend not to read that deep, so answering questions now means not having to answer them later when they file issues.

ethowitz: For some reason, the email notification for this thread went into my spam folder so I'm just seeing it now 🤦🏻 Just to make sure I understand, you're saying it might make sense to include info about authenticating via application default creds in addition to authenticating with a service account?

jrconlin: I know this is a month old, but has this been addressed? Just wondering if we can clear the backlog a bit.

ethowitz: Sorry for the delay! I updated the docs to include information about authenticating via both OAuth and a service account

jrconlin (approved):

Issue #1052, opened 2021-04-26
Tokenserver: Refactor and clean up tokenserver.rs


ethowitz:
Let's clean up `tokenserver.rs` so that it's more readable and production-ready. We should investigate how to better integrate the Tokenserver work with the existing abstractions in the codebase (extractors, handlers, database abstractions, etc.).

ethowitz:
It seems like the best path forward is to create a separate `tokenserver` module. This module would probably live at the same level as `server` and `web`. Some things to keep in mind:

- Tokenserver will probably want to reuse code that lives in `server` and `web`
- We'll eventually need to add a separate database adapter for the Tokenserver MySQL database (#1054)
- Ops may want to run Tokenserver as a separate service from syncstorage-rs (this _could_ be done by routing traffic in nginx)

PR #1053, opened 2021-04-26
refactor: Tokenserver: Rewrite inlined Python code in Rust

### LINKED ISSUE #1049: Tokenserver: Rewrite inlined Python functions in Rust
There is a block of inlined Python code in `tokenserver.rs` that includes functions relating to generating tokens. Most of these functions can be easily rewritten in Rust, except for the two calls to the `tokenlib` library. Let's rewrite these functions in Rust and build a wrapper around the two `tokenlib` functions like so:

```rust
pub struct Tokenlib<'a> {
    py: Python<'a>,
    inner: &'a PyModule,
}

impl<'a> Tokenlib<'a> {
    pub fn new(py: Python<'a>) -> Result<Self, PyErr> {
        let inner = PyModule::import(py, "tokenlib").map_err(|e| {
            e.print_and_set_sys_last_vars(py);
            e
        })?;

        Ok(Self { py, inner })
    }

    pub fn make_token(&self, plaintext: &PyDict, shared_secret: &str) -> Result<String, PyErr> {
        match self.inner.call1("make_token", (plaintext, shared_secret)) {
            Err(e) => {
                e.print_and_set_sys_last_vars(self.py);
                Err(e)
            }
            Ok(x) => Ok(x.extract::<String>().unwrap()),
        }
    }

    pub fn get_derived_secret(
        &self,
        plaintext: &str,
        shared_secret: &str,
    ) -> Result<String, PyErr> {
        match self
            .inner
            .call1("get_derived_secret", (plaintext, shared_secret))
        {
            Err(e) => {
                e.print_and_set_sys_last_vars(self.py);
                Err(e)
            }
            Ok(x) => Ok(x.extract::<String>().unwrap()),
        }
    }
}
```

(If we end up doing this, #955 could be closed)

ethowitz:
## Description
Replaces the inlined Python code with native Rust code. I replaced all of the Python functions with equivalent Rust functions, except for `tokenlib#make_token` and `tokenlib#get_derived_secret`, which I added Rust wrappers for.

## Testing

I tested this by running the Python and Rust versions with the same input data and ensuring that the output was the same.
Python test file: https://gist.github.com/ethowitz/7d6b3590f4aae0c51b2df71297c1d303
Rust test code (I added this to syncstorage-rs's main function for expediency): https://gist.github.com/ethowitz/2ec06b230b2db66de6088ac30a8c9819

Python output:
```
dGVzdGluZzEyMw # URL-safe base64 encoded string with stripped padding
51c982a4554222dce10f7fd77145a92f87a6f55934403ff8d2546d6c0a4da872 # FxA metrics hash
402e41f93a922d7517f43668d2392e72 # hashed device ID
eyJ1c2VyX2lkIjogMTIzLCAic2FsdCI6ICIzMjkxYzQiLCAiZXhwaXJlcyI6IDE2MTk0NzM2MzQuMjQ1Mjc5Nn3kixJDJKmsgAc6GMizJZjB7IMwVBBexo8idSGHZUNlbg== # token
Sdqw4UlM3JIPKZ2tN2YU9N3ZrvA4dfCEEbI35fu7-ac= # derived secret
```
Rust output:
```
dGVzdGluZzEyMw # URL-safe base64 encoded string with stripped padding
51c982a4554222dce10f7fd77145a92f87a6f55934403ff8d2546d6c0a4da872 # FxA metrics hash
402e41f93a922d7517f43668d2392e72 # hashed device ID
eyJ1c2VyX2lkIjogMTIzLCAic2FsdCI6ICJlNGQ0Y2UiLCAiZXhwaXJlcyI6IDE2MTk0NzM1NzEuOTU4MTE5NH19_O2gNmQr7SzM-660260QllLQXysgrFzA7A4Da47mRA== # token
Sywsc6GJcQhSQoiT7ctlS2ltKNSdPbdvT2GbjaRXiwY= #derived secret
```
As you can see, the tokens and derived secrets differ, but these values differ across runs of the same version (e.g. when running the Python version twice, the values differ across the runs). I assume the tokenlib library adds nondeterminism (e.g. a nonce) so the values are different each time.

## Issue(s)

Closes #1049

### DIFF
--- a/src/web/tokenserver.rs
+++ b/src/web/tokenserver.rs
@@ -203,28 +203,29 @@ pub fn get_sync(
         user_record[0].keys_changed_at.unwrap_or(0),
         client_state_b64
     );
-    let (python_result, python_derived_result) = Python::with_gil(|py| -> Result<(String, String), PyErr>{
-        let thedict = [
-            ("node", &user_record[0].node),
-            ("fxa_kid", &fxa_kid), // userid component of authorization email
-            ("fxa_uid", &token_data.claims.sub),
-            ("hashed_device_id", &hashed_device_id),
-            ("hashed_fxa_uid", &hashed_fxa_uid),
-        ]
-        .into_py_dict(py);
-        // todo don't hardcode
-        // we're supposed to check the "duration" query
-        // param and use that if present (for testing)
-        thedict.set_item("expires", 300).unwrap(); // todo this needs to be converted to timestamp int (now + value * 1000)
-        thedict.set_item("uid", user_record[0].uid).unwrap();
+    let (python_result, python_derived_result) =
+        Python::with_gil(|py| -> Result<(String, String), PyErr> {
+            let thedict = [
+                ("node", &user_record[0].node),
+                ("fxa_kid", &fxa_kid), // userid component of authorization email
+                ("fxa_uid", &token_data.claims.sub),
+                ("hashed_device_id", &hashed_device_id),
+                ("hashed_fxa_uid", &hashed_fxa_uid),
+            ]
+            .into_py_dict(py);
+            // todo don't hardcode
+            // we're supposed to check the "duration" query
+            // param and use that if present (for testing)
+            thedict.set_item("expires", 300).unwrap(); // todo this needs to be converted to timestamp int (now + value * 1000)
+            thedict.set_item("uid", user_record[0].uid).unwrap();
 
-        let tokenlib = Tokenlib::new(py)?;
-        let result = tokenlib.make_token(thedict, &shared_secret)?;
-        let derived_result = tokenlib.get_derived_secret(&result, &shared_secret)?;
-        //assert_eq!(result, false);
-        Ok((result, derived_result))
-    })
-    .unwrap();
+            let tokenlib = Tokenlib::new(py)?;
+            let result = tokenlib.make_token(thedict, &shared_secret)?;
+            let derived_result = tokenlib.get_derived_secret(&result, &shared_secret)?;
+            //assert_eq!(result, false);
+            Ok((result, derived_result))
+        })
+        .unwrap();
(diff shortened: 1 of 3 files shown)


jrconlin:
Yep. The `token` is a base64 JSON blob that contains the user_id, salt, and expires timestamp, along with some padding noise. Looking at the tokens, I see that the user_ids match, with the salt and expires being different as would be expected.

src/web/tokenserver.rs
```diff
@@ -138,128 +189,43 @@ pub fn get_sync(
...
+                ("fxa_kid", &fxa_kid), // userid component of authorization email
+                ("fxa_uid", &token_data.claims.sub),
+                ("hashed_device_id", &hashed_device_id),
+                ("hashed_fxa_uid", &hashed_fxa_uid),
+            ]
+            .into_py_dict(py);
+            // todo don't hardcode
+            // we're supposed to check the "duration" query
+            // param and use that if present (for testing)
+            thedict.set_item("expires", 300).unwrap(); // todo this needs to be converted to timestamp int (now + value * 1000)
```
jrconlin, line 219: You could have this pull from `env` like you're doing above with an `unwrap_or_else` that uses 300 as the default.

It'd also be nice to have a section of `Settings` that defined all of these so that there's no risk of these getting lost in the code.

jrconlin (approved):

src/web/tokenserver.rs
ethowitz: @jrconlin I think this was done by someone else; I think it ended up in the diff since I changed the indentation or something 🤔 Do you think it's worth making this change here? Might be worth waiting for when I do the work for #1052, but happy to make the change here if you'd prefer that!

jrconlin: Well, not for this iteration. I was just noting the "todo". If it was something critical or important, I wouldn't have approved the changes 😉

PR #1056, opened 2021-04-28
fix: Convert integral values to String before converting to Value

### LINKED ISSUE #1055: Invalid value for column collection_id in table bsos: Expected INT64.
Per sentry:

https://sentry.prod.mozaws.net/operations/syncstorage-dev/issues/11102490/?environment=release

`
A database error occurred: RpcFailure: 9-FAILED_PRECONDITION Invalid value for column collection_id in table bsos: Expected INT64.
`

This began 9 days ago, so it's very likely related to the #1046 refactor.

ethowitz:
## Description

In the `ToSpannerValue` implementations for `i32` and `u32`, we were using `Value::set_number_value()` to construct a `Value` from an integer. In doing so, we coerced the integers into floats, which is what created the errors on Spanner. To resolve this, I convert the integral values to strings prior to creating a `Value`.

## Testing

The issue can be reproduced on the main branch by adding a new bookmark and syncing (your local server should throw the [same error that appeared in Sentry](https://sentry.prod.mozaws.net/operations/syncstorage-dev/issues/11102490/?query=is%3Aunresolved). Switching to this feature branch, adding a new bookmark, (recompiling,) and syncing should result in a successful sync, with no errors reported.

## Issue(s)

Closes #1055

### DIFF
--- a/src/db/spanner/batch.rs
+++ b/src/db/spanner/batch.rs
@@ -11,7 +11,7 @@ use protobuf::{
 use uuid::Uuid;
 
 use super::models::{Result, SpannerDb, DEFAULT_BSO_TTL, PRETOUCH_TS};
-use super::support::{null_value, struct_type_field, ToSpannerValue};
+use super::support::{as_type, null_value, struct_type_field, ToSpannerValue};
 use crate::{
     db::{params, results, util::to_rfc3339, DbError, DbErrorKind, BATCH_LIFETIME},
     web::{extractors::HawkIdentifier, tags::Tags},
@@ -35,20 +35,20 @@ pub async fn create_async(
         id: batch_id,
     };
 
+    let (sqlparams, mut sqlparam_types) = params! {
+        "fxa_uid" => params.user_id.fxa_uid.clone(),
+        "fxa_kid" => params.user_id.fxa_kid.clone(),
+        "collection_id" => collection_id,
+        "batch_id" => new_batch.id.clone(),
+        "expiry" => to_rfc3339(timestamp + BATCH_LIFETIME)?,
+    };
+    sqlparam_types.insert("expiry".to_owned(), as_type(TypeCode::TIMESTAMP));
     db.sql(
         "INSERT INTO batches (fxa_uid, fxa_kid, collection_id, batch_id, expiry)
          VALUES (@fxa_uid, @fxa_kid, @collection_id, @batch_id, @expiry)",
     )?
-    .params(params! {
-        "fxa_uid" => params.user_id.fxa_uid.clone(),
-        "fxa_kid" => params.user_id.fxa_kid.clone(),
-        "collection_id" => collection_id.to_string(),
-        "batch_id" => new_batch.id.clone(),
-        "expiry" => to_rfc3339(timestamp + BATCH_LIFETIME)?,
-    })
-    .param_types(param_types! {
-        "expiry" => TypeCode::TIMESTAMP,
-    })
+    .params(sqlparams)
+    .param_types(sqlparam_types)
     .execute_dml_async(&db.conn)
     .await?;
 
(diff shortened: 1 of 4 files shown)


src/db/spanner/support.rs
```diff
@@ -40,17 +40,13 @@ impl ToSpannerValue for String {
 
 impl ToSpannerValue for i32 {
     fn to_spanner_value(&self) -> Value {
-        let mut value = Value::new();
-        value.set_number_value(*self as f64);
-        value
+        self.to_string().to_spanner_value()
```
jrconlin, line 43: IIRC, our `SpannerDb` implementation in sync uses type hints `TypeCode` for queries at least.
e.g. https://github.com/mozilla-services/syncstorage-rs/blob/master/src/db/spanner/models.rs#L977

We might want to verify that for query instances where we do this conversion have the type set.

jrconlin: I think that's what this is doing? https://github.com/mozilla-services/syncstorage-rs/blob/c5da264a37196f85b116df9a66c681c0b65a2dc2/src/db/spanner/support.rs#L68

jrconlin: Now I'm wondering if we should just iterate through the params list to build the types much like how we iterate through the params to convert to spanner values. It would set a fair pattern for handling parameters. Bonus would be some impl we could call, pass the parameters to, and it would automagically gen the `.parameters()` and `parameter_types()`...

ethowitz: I like your idea about generating the type hints from the parameter name 🤔 At the very least, I'll be sure to add the type hints for the queries that involve the `ToSpannerValue` trait, but I'll see if I can figure something fancier out

src/db/spanner/batch.rs
```diff
@@ -181,39 +185,38 @@ pub async fn commit_async(
...
         let mut timer2 = db.metrics.clone();
         timer2.start_timer("storage.spanner.apply_batch_update", None);
+        let (sqlparams, mut sqlparam_types) = params! {
+            "fxa_uid" => params.user_id.fxa_uid.clone(),
+            "fxa_kid" => params.user_id.fxa_kid.clone(),
+            "collection_id" => collection_id,
+            "batch_id" => params.batch.id.clone(),
+            "timestamp" => as_rfc3339.clone(),
+        };
+        sqlparam_types.insert("timestamp".to_owned(), as_type(TypeCode::TIMESTAMP));
```
jrconlin, line 195: NOT THAT WE NEED TO DO THIS NOW

But we really could take advantage of they SyncTimestamp type better for stuff like this. 
(heh, Good first bug material.)

jrconlin (dismissed):

ethowitz:
@jrconlin I took a stab at having SQL parameter types automatically generated by the same macro that builds the param hashmaps. In short, I updated the `params!` macro to return a tuple consisting of `(params, param_types)`, which strongly suggests to the caller that the types should be used. I have a few concerns, and I'd be curious to hear your thoughts about them:

- There are certain params whose automatically generated types are incorrect. For example, we store timestamps in rust as strings (the macro will thus return `TypeCode::STRING` as the type), but Spanner stores timestamps as `TypeCode::TIMESTAMP`s (a possible solution to this would be to change how we represent timestamps in memory). If a developer forgets to update the timestamp type before passing `sqlparam_types` to the queryer, a database error would occur at runtime.
- I'm not sure how to comprehensively test these changes 😬 I made sure that I could perform a successful sync with spanner after adding a new bookmark, but I don't know how to comprehensively test each of the database transactions that uses the new param types functionality. Do you have any ideas?

jrconlin:
As for the comments:
* Yeah, the problem is that we break "type" by converting values to strings before we pass them to `params!` so there's no carry over type information. I noted it in the comment, but a solution to that would be "not convert these to strings" and use `From<foo> to String` impl's with associated types for the odder ones. We could probably also create SyncTimeStamp consts for `PRETOUCH_TS`. Timestamps seem to be the most problematic for these.
* I suppose the testing strategy here is to test the function of what you're doing, not necessarily how it's being used? This change really introduces a new behavior for `params!` which takes a HashMap and returns two separate HashMaps. What you're testing for is that values and types are properly recognized and converted. (e.g. values are converted to `Values` with correct `.kind`, and `param_types` are proper `Type` with a proper `.code`. I'd probably construct a "one of each" sort of test HashMap and validate that everything got converted correctly.

ethowitz:
I ran the spanner tests locally and they all passed 🙌🏻

jrconlin (approved):

PR #1078, opened 2021-05-06
chore: Updates for May 2021

### LINKED ISSUE #1077: Update libraries for Q2
Update the various dependent libraries for Q2

jrconlin:
## Description

Updates for grpcio, cadence, config, hkdf/hmac, and validator
updates vendored Google API calls

## Testing

Code should compile and tests should run.

## Issue(s)

Closes #1077


vendor/mozilla-rust-sdk/googleapis/src/spanner.rs
```diff
@@ -36,7 +49,7 @@ impl Client {
         let creds = ChannelCredentials::google_default_credentials()?;
 
         // Create a Spanner client.
-        let chan = ChannelBuilder::new(env)
+        let chan = ChannelBuilder::new(env.clone())
```
ethowitz, line 52: Does clippy complain if we don't have this `clone()` here?

pjenvey: If it's now needed env is an `Arc` so this would be preferred:

```suggestion
        let chan = ChannelBuilder::new(Arc::clone(&env))
```

jrconlin: This probably should be corrected in the mozilla-rust-sdk.

vendor/mozilla-rust-sdk/googleapis-raw/generate.sh
```diff
@@ -27,7 +41,8 @@ if ! [[ -x "$(command -v grpc_rust_plugin)" ]]; then
 fi
 
 echo "Pulling git submodules"
-git submodule update --init --recursive
+# comment out to work on master...
+#git submodule update --init --recursive
```
ethowitz, line 45: What if we just uncomment this line so the submodule updates happen automagically? Is there a reason we don't want it to happen in the script?

vendor/mozilla-rust-sdk/googleapis-raw/generate.sh
```diff
@@ -45,12 +60,24 @@ for proto in $proto_files; do
...
+big_table_dirs="
 bigtable/admin/cluster/v1
 bigtable/admin/table/v1
 bigtable/admin/v2
 bigtable/v1
 bigtable/v2
+"
+
+proto_dirs="
+api
```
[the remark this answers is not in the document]
jrconlin: Yeah, we should remove hit here too. It's not used, so we shouldn't try bringing it in.

vendor/mozilla-rust-sdk/googleapis-raw/generate.sh
jrconlin: Yeah, I'd rather it be a mindful thing, partly because it could introduce a lot of breakage.

ethowitz (approved):

PR #1079, opened 2021-05-10
feat: Integrate Spanner emulator with CI

### LINKED ISSUE #566: Test against and add docs for working with the cloud spanner emulator, integrate with CI
Looks like Google recently released a [cloud spanner emulator](https://github.com/GoogleCloudPlatform/cloud-spanner-emulator) - Lets:

1. Test against this locally and then update our README with instructions on how specifically to use it with this service.
2. Get it hooked up to CI

ethowitz:
## Description
- Updated docker compose files to support end-to-end tests using both Spanner and MySQL
- Added a script that automatically sets up a locally-running Spanner emulator using its REST API
- Updated the README to include info about using the emulator for development purposes
- Added the Spanner end-to-end tests to CI
- Bumped the version for the "builder" docker image to use the latest version of Rust (1.52.1)
- Updated the Spanner schema file to reflect the schema currently in use in production
- Fixed some of the end-to-end tests that were failing due to database nondeterminism

## Testing
I tested that:
- The new docker commands in Makefile operated as expected
- The `prepare-spanner.sh` script correctly sets up the Spanner emulator
- The end-to-end tests all pass using Spanner
- Everything passes on CI

## Issue(s)

Closes #566


jrconlin (approved):

jrconlin:
Not sure if you want to bump rust up to 1.52.1 in the `Dockerfile`

src/db/spanner/manager/session.rs
```diff
@@ -51,6 +50,7 @@ pub async fn create_spanner_session(
         } else {
             // Requires
             // GOOGLE_APPLICATION_CREDENTIALS=/path/to/service-account.json
+            metrics.start_timer("storage.pool.grpc_auth", None);
```
ethowitz, line 53: I _think_ this belongs here (I moved it in a prior PR)

src/db/spanner/schema.ddl
```diff
@@ -11,25 +11,28 @@
 --       ALSO, CONSOLE WANTS ONE SPACE BETWEEN DDL COMMANDS
 
 CREATE TABLE user_collections (
-  fxa_uid STRING(MAX)  NOT NULL,
-  fxa_kid STRING(MAX)  NOT NULL,
-  collection_id INT64  NOT NULL,
-  modified TIMESTAMP   NOT NULL,
+  fxa_uid STRING(MAX) NOT NULL,
```
ethowitz, line 14: The changes to the schema reflect the most up-to-date schema pulled from the GCP console by @jrconlin

src/db/spanner/manager/session.rs
jrconlin: `start_timer` records the time on drop. This metric is in place to monitor how long it takes for the Google Auth to happen (we were seeing a few instances of this taking a long time and want to track to spot anomalies). Keeping the metric within the same block as the credential call should provide the most accurate timing.

src/db/spanner/schema.ddl
```diff
@@ -41,58 +44,40 @@ INTERLEAVE IN user_collections;
...
+    payload STRING(MAX),
+    ttl INT64,
 )    PRIMARY KEY(fxa_uid, fxa_kid, collection_id, batch_id, batch_bso_id),
   INTERLEAVE IN PARENT batches ON DELETE CASCADE;
 
 -- batch_bsos' bso fields are nullable as the batch upload may or may
 -- not set each individual field of each item. Also note that there's
 -- no "modified" column because the modification timestamp gets set on
 -- batch commit.
-
```
jrconlin, line 80: In the interest of over-documenting, we might want to add a comment here like:

```
-- *NOTE*: 
-- Newly created Spanner instances should pre-populate the `collections` table by 
-- running the content of `insert_standard_collections.sql `
```

(Again, trying to be kind to the future person, who's in a panic.)

jrconlin (approved):

jrconlin (approved):

src/db/spanner/schema.ddl
```diff
@@ -11,25 +11,28 @@
...
 
 CREATE TABLE bsos (
-  fxa_uid STRING(MAX)  NOT NULL,
-  fxa_kid STRING(MAX)  NOT NULL,
-  collection_id INT64  NOT NULL,
-  bso_id STRING(MAX)   NOT NULL,
+  fxa_uid STRING(MAX) NOT NULL,
+  fxa_kid STRING(MAX) NOT NULL,
+  collection_id INT64 NOT NULL,
+  bso_id STRING(64)   NOT NULL,
```
pjenvey, line 27: @jrconlin was this from the dev env? This should be `MAX` which is what stage's syncdb4 has.

```suggestion
  bso_id STRING(MAX)   NOT NULL,
```

src/db/spanner/schema.ddl
```diff
@@ -57,15 +60,16 @@ CREATE TABLE batches (
...
-  fxa_uid STRING(MAX)      NOT NULL,
-  fxa_kid STRING(MAX)      NOT NULL,
-  collection_id INT64      NOT NULL,
-  batch_id STRING(MAX)     NOT NULL,
-  batch_bso_id STRING(MAX) NOT NULL,
+  fxa_uid STRING(MAX)     NOT NULL,
+  fxa_kid STRING(MAX)     NOT NULL,
+  collection_id INT64     NOT NULL,
+  batch_id STRING(MAX)    NOT NULL,
+  batch_bso_id STRING(64) NOT NULL,
```
pjenvey, line 72:
```suggestion
  batch_bso_id STRING(MAX) NOT NULL,
```

scripts/prepare-spanner.sh
```diff
@@ -0,0 +1,31 @@
...
+
+sleep 5
+
+set -e
+
+PROJECT_ID=test-project
+INSTANCE_ID=test-instance
+DATABASE_ID=test-database
+
+DDL_STATEMENTS=$(
```
pjenvey, line 11: thanks for doing this 👍

tools/integration_tests/test_storage.py
```diff
@@ -266,11 +266,9 @@ def test_get_collection_only(self):
         # Sets the maximum number of ids that will be returned
         self.retry_delete(self.root + '/storage/xxx_col2')
 
-        bsos = []
         for i in range(10):
             bso = {'id': str(i).zfill(2), 'payload': 'x', 'sortindex': i}
-            bsos.append(bso)
-        self.retry_post_json(self.root + '/storage/xxx_col2', bsos)
+            self.retry_post_json(self.root + '/storage/xxx_col2', [bso])
```
pjenvey, line 271: I still don't understand why this change was needed, I'll run these tests and take a look.

Were you seeing it fail solely on the Spanner emulator?

pjenvey (changes requested):

tools/integration_tests/test_storage.py
ethowitz: We just chatted about this, but posting this for posterity:

My understanding is that the tests starting on L298 expect the BSOs to be sorted based on timestamp, but since we insert all of the records at once, all of the BSOs have the same timestamp. This leads to intermittent test failures -- the `modified` timestamps are all equal, so when sorting on that field, the order of the BSOs returned by Spanner is nondeterministic. I've only seen the failure on the emulator, but I'm guessing that's due to implementation differences between the emulator and the hosted database. It just so happens that the hosted version of Spanner has been returning the BSOs in a consistent order between the `GET` calls on lines 299 and 303, but I don't think there's any guarantee that that will continue to be the case

jrconlin (dismissed):

tools/integration_tests/test_storage.py
pjenvey: Sorry I just got back around to looking at this.

Though the changes to the tests are subtle and small I think it would be safe to leave them as is (the assertEquals changes are probably even more significant than the bulk post change).

I think we can solve this another way by adding something like a "consistent_sort_order" flag to `SpannerDb` that enables the commented out issue559 section. This adds both a secondary `bso_id` stablilization sort (that spanner does implicitly since it's the primary key) and defaults no sort specified to act like sort=newest.

We might want to look into enabling this again on production Spanner but IIRC it caused a slow down.

Interestingly enough the Python version has this exact same flag for kind of the opposite reason -- the extra sorting block slowed down older Mysql's but not modern ones.

pjenvey (changes requested):
Added another suggestion for the test changes above

tools/integration_tests/test_storage.py
ethowitz: We chatted via slack about this; I'll go ahead and key the stabilization sort off of `spanner_emulator_host` 👍🏻

ethowitz:
Let me know if this approach makes sense to you!

src/db/spanner/models.rs
```diff
@@ -1270,20 +1273,23 @@ impl SpannerDb {
...
-            }
-            Sorting::Oldest => format!("{} ORDER BY modified ASC, bso_id ASC", query),
-            */
-            Sorting::Index => format!("{} ORDER BY sortindex DESC", query),
-            Sorting::Newest => format!("{} ORDER BY modified DESC", query),
-            Sorting::Oldest => format!("{} ORDER BY modified ASC", query),
-            _ => query,
-        };
+
+        if self.stabilize_bsos_sort_order() {
```
pjenvey, line 1277: Sorry Ethan, I probably wasn't clear enough about this: let's solely re-enable _only this_ section -- `Sorting::None` now acts like `Newest` and enabling `bso_id` as a secondary sort.

The rest of the related bits for issue559 should stay commented out. They were related to enabling this sorting but are really a different feature, they enable "encoded offsets" -- a different format of offset that potentially optimized the get queries during pagination.

The sorting changes were required to enable the encoded offset but we don't need (or want) the encoded offsets for solely this sorting change. `encode_next_offset`, the `extractor` changes and the related "timestamp always None" change above in this file should stay the same (commented out).

pjenvey (changes requested):

src/db/spanner/models.rs
ethowitz: Ah, my bad! I misunderstood. For some reason, I assumed the secondary sort required the other offset change. I'll go ahead and fix it

src/db/spanner/models.rs
```diff
@@ -1247,13 +1247,19 @@ impl SpannerDb {
         if let Some(timestamp) = offset.clone().unwrap_or_default().timestamp {
             query = match sort {
                 Sorting::Newest => {
-                    sqlparams.insert("older_eq".to_string(), as_value(timestamp.as_rfc3339()?));
-                    sqltypes.insert("older_eq".to_string(), as_type(TypeCode::TIMESTAMP));
+                    sqlparams.insert(
```
ethowitz, line 1250: I figured it made sense to leave these changes. The commented-out code was still using the now non-existent `as_value` function

pjenvey (approved):

Issue #968, opened 2021-01-05
tokenserver: Panic 'index out of bounds'


jrconlin:
seeing:

```
thread 'actix-web' panicked at 'index out of bounds: the len is 0 but the index is 0', src/web/tokenserver.rs:244:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
I suspect this is because I am creating a new sync record for a new sync db and there's no previous record to index from. We might want to add some handling for that case.

ethowitz:
Closed by #1052

PR #1087, opened 2021-05-17
refactor: Clean up Tokenserver code

### LINKED ISSUE #1052: Tokenserver: Refactor and clean up tokenserver.rs

ethowitz:
## Description

Adds a `tokenserver` module, with functionality split into separate files. 

## Testing

First, you'll need to create a new MySQL database and add the appropriate tables. DDL and SQL to accomplish this can be found in [this gist](https://gist.github.com/ethowitz/242edbc356688ba313c55a0bfdc4b3e9).

I was able to test this by creating a valid JWT (I generated a set of RSA keys to test with) with the `jwt` Python package and sending off a request like so (I used Postman):
```
curl --location --request GET 'localhost:8000/1.0/sync/1.5' \
  --header 'Authorization: Bearer <YOUR JWT HERE>'
```

I tested the following cases:
- Valid JWT --> 200 status code with expected JSON
- Valid JWT with FxA UID for nonexistent user --> 404 status code
- Invalid JWT --> 400 status code

## Issue(s)

Closes #1052, #968


src/tokenserver/extractors.rs
```diff
@@ -0,0 +1,228 @@
...
+
+    use std::sync::Arc;
+    use std::time::{Duration, SystemTime, UNIX_EPOCH};
+
+    lazy_static! {
+        static ref SECRETS: Arc<Secrets> = Arc::new(Secrets::new("Ted Koppel is a robot").unwrap());
+        static ref SERVER_LIMITS: Arc<ServerLimits> = Arc::new(ServerLimits::default());
+    }
+
+    const RSA_PRIVATE_KEY: &str = "-----BEGIN RSA PRIVATE KEY-----\n\
```
ethowitz, line 120: This is a key I generated specifically for the purposes of testing

mythmon: In this past, our security teams have advised me not to do even this. They suggested generating the keys at test time / somewhere in the repo, instead of checking them in.

ethowitz: Ah, noted - thanks for the tip. I'll go ahead and do that then

mythmon: Oh, and to expand on my previous message: The reasoning was that if you see an RSA key in a repo, it should set off alarm bells. It's easier to just never set of the alarms, instead of continually noting that it's fine.

ethowitz: Makes sense. And in this case, it literally set off alarm bells (in the form of a GitGuardian email)

src/tokenserver/extractors.rs
```diff
@@ -0,0 +1,205 @@
...
+    pub iat: i64,
+    pub exp: i64,
+}
+pub struct TokenserverRequest {
+    pub fxa_uid: String,
+}
+
+impl TokenserverRequest {
+    fn get_fxa_uid(jwt: &str, rsa_modulus: String, rsa_exponent: String) -> Result<String, Error> {
+        decode::<Claims>(
```
jrconlin, line 28: Stupid question, do we know what happens if we send in a JWT with `alg="None"`? That's a fairly common attack vector.

jrconlin (approved):

src/tokenserver/extractors.rs
ethowitz: Not a stupid question! It looks like the crate we use doesn't support the `None` algorithm, so validation will fail: https://github.com/Keats/jsonwebtoken/blob/2f25cbed0a906e091a278c10eeb6cc1cf30dc24a/src/algorithms.rs#L48-L66

I'll verify this manually to be sure

ethowitz: Confirmed that `"alg": "none"` in the JWT headers results in a 400 error

Issue #1102, opened 2021-06-14
Tokenserver: Improve error handling


ethowitz:
We shouldn't be calling `.unwrap()` unless it's obviously/trivially true that it won't result in a panic. Let's replace instances of `.unwrap()` with more idiomatic error handling (using `Result`s).

PR #1103, opened 2021-06-16
refactor: Use PyFxA to verify Tokenserver tokens

### LINKED ISSUE #1098: Tokenserver: Use PyFxA to verify OAuth tokens
Currently, we are manually verifying OAuth tokens using the `jsonwebtoken` crate. Instead, let's call into the PyFxA library using `pyo3` to make this process simpler and less prone to error.

ethowitz:
## Description

- Uses PyO3 to verify tokens with PyFxA
- Improves the ergonomics of calling into Python code
    * Clients of the `Tokenlib` and `OAuthVerifier` structs now require no knowledge of PyO3
    * The GIL is now held for the minimum possible duration for each operation
    * Clients are not responsible for managing the GIL, making usage of the Python code less error-prone
- Resolves an issue with how we were parsing `hashed_fxa_uid` (it needed to be truncated)
- Adds linting for the new Python file

## Testing
First, seed the database:
```sql
INSERT INTO services (service) VALUES("sync-1.5");
INSERT INTO nodes (service, node, available, capacity) VALUES(<service ID>, "http://localhost:8000/", 100, 100);
INSERT INTO users (service, email, generation, client_state, created_at, nodeid) VALUES(<service ID>, "<FxA email address>", 0, "yes", 1620924095, <node ID>);
```
`<service ID>` is the database ID of the service added in the first step, `<node ID>` is the database ID of the node added in the second step, and `<FxA email address>` is the email address associated with your FxA account (it should be of the format `<FxA UID>@api.accounts.firefox.com`.

Next, point Firefox to the syncstorage Tokenserver by setting the pref `identity.sync.tokenserver.uri` to `http://localhost:8000/1.0/sync/1.5`. Finally, attempt a sync. If the sync is successful, it means that PyFxA successfully verified the OAuth token passed by Firefox.

## Issue(s)

Closes #1098, #1102


pjenvey:
circleci's build-and-test is now failing

src/tokenserver/extractors.rs
```diff
@@ -163,13 +133,26 @@ mod tests {
...
+            let token_data = TokenData {
+                user: fxa_uid.to_owned(),
+                client_id: "client id".to_owned(),
+                scope: vec!["scope".to_owned()],
+                generation: current_time.as_secs() as i64,
+                profile_changed_at: current_time.as_secs() as i64,
+            };
+            let valid = false;
+
+            MockOAuthVerifier { valid, token_data }
```
ethowitz, line 149: We need to stub out the verifier, since we can't forge the JWT that comes from FxA

pjenvey (approved):
looks good so far Ethan

ethowitz:
Thanks Phil!

PR #1119, opened 2021-07-21
refactor: Tokenserver: Add mature MySQL adapter

### LINKED ISSUE #1054: Tokenserver: Build more mature database adapter
We'll need to add support for a second, dedicated MySQL database instance for Tokenserver. The necessary infrastructure will likely live in`src/tokenserver/db/mysql`, but more investigation may be needed.

ethowitz:
## Description

Adds mature infrastructure for a Tokenserver-specific MySQL database.

## How should reviewers test?
* Create an account on FxA stage: https://accounts.stage.mozaws.net/signup
* Set `identity.fxaccounts.remote.oauth.uri` to https://oauth.stage.mozaws.net
* Set `identity.fxaccounts.remote.root` to https://accounts.stage.mozaws.net
* Set `identity.fxaccounts.auth.uri` to https://api-accounts.stage.mozaws.net/v1
* Set `identity.fxaccounts.remote.profile.uri` to https://profile.stage.mozaws.net/v1
* Set `identity.sync.tokenserver.uri` to http://localhost:8000/1.0/sync/1.5
* Create a Tokenserver database:
```sql
CREATE DATABASE tokenserver_rs;
```
* Add a service and a node to your database:
```sql
INSERT INTO services (service) VALUES("sync-1.5");
INSERT INTO nodes (service, node, available, capacity) VALUES([service ID from previous step], "http://localhost:8000/", 100, 100);
```
* Set the following values in local.toml:
```toml
enable_tokenserver = true
tokenserver.database_url = "mysql://sample_user:sample_password@localhost/tokenserver_rs"
tokenserver.fxa_oauth_server_url = "https://oauth.stage.mozaws.net"
```
* Add `println!("{}", user_email)` to L47 in `src/tokenserver/handlers.rs`
* Start syncstorage via `make run`
* Restart Firefox, log in to sync, and attempt a sync
* Using the email address from the server logs, add a user to the database:
```sql
INSERT INTO users (service, email, generation, client_state, created_at, nodeid) VALUES([service ID from prior step], "[email address from logs]", 0, "yes", 1620924095, [node ID from prior step]);
```
* Remove the `println!` and recompile/restart the server
* Attempt a sync
* Ensure that the sync was successful via about:sync-log

## Issue(s)

Closes #1054

### DIFF
--- a/src/tokenserver/db/models.rs
+++ b/src/tokenserver/db/models.rs
@@ -64,10 +64,11 @@ impl TokenserverDb {
     fn get_user_sync(&self, email: String) -> DbResult<results::GetUser> {
         let query = r#"
             SELECT users.uid, users.email, users.client_state, users.generation,
-                users.keys_changed_at, users.created_at, nodes.node
-            FROM users
-            JOIN nodes ON nodes.id = users.nodeid
-            WHERE users.email = ?
+                   users.keys_changed_at, users.created_at, nodes.node
+              FROM users
+              JOIN nodes
+                ON nodes.id = users.nodeid
+             WHERE users.email = ?
         "#;
         let mut user_records = diesel::sql_query(query)
             .bind::<Text, _>(email)
@@ -87,9 +88,8 @@ impl TokenserverDb {
     fn post_node_sync(&self, node: params::PostNode) -> DbResult<results::PostNode> {
         let query = r#"
             INSERT INTO nodes (service, node, available, current_load, capacity, downed, backoff)
-               VALUES (?, ?, ?, ?, ?, ?, ?)
+            VALUES (?, ?, ?, ?, ?, ?, ?)
         "#;
-
         diesel::sql_query(query)
             .bind::<Integer, _>(node.service_id)
             .bind::<Text, _>(&node.node)
@@ -101,16 +101,16 @@ impl TokenserverDb {
             .execute(&self.inner.conn)?;
 
         let query = r#"
-            SELECT id FROM nodes
-            WHERE service = ? AND
-                  node = ? AND
-                  available = ? AND
-                  current_load = ? AND
-                  capacity = ? AND
-                  downed = ? AND
-                  backoff = ?
+            SELECT id
+              FROM nodes
+             WHERE service = ?
+               AND node = ?
+               AND available = ?
+               AND current_load = ?
+               AND capacity = ?
+               AND downed = ?
+               AND backoff = ?
         "#;
-
         diesel::sql_query(query)
             .bind::<Integer, _>(node.service_id)
             .bind::<Text, _>(&node.node)
@@ -125,13 +125,21 @@ impl TokenserverDb {
 
     #[cfg(test)]
     fn post_service_sync(&self, service: params::PostService) -> DbResult<results::PostService> {
-        let query = "INSERT INTO services (service, pattern) VALUES (?, ?)";
+        let query = r#"
+            INSERT INTO services (service, pattern)
+            VALUES (?, ?)
+        "#;
         diesel::sql_query(query)
             .bind::<Text, _>(&service.service)
             .bind::<Text, _>(service.pattern)
             .execute(&self.inner.conn)?;
 
-        let query = "SELECT id FROM services WHERE service = ? AND pattern = ?";
+        let query = r#"
+            SELECT id
+              FROM services
+             WHERE service = ?
+               AND pattern = ?
+        "#;
         diesel::sql_query(query)
             .bind::<Text, _>(&service.service)
             .bind::<Text, _>(&service.service)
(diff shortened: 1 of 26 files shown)


jrconlin (approved):
Nice work on the MockDB stuff.

ethowitz:
Thank you! I'm gonna hold off on merging this for now: there's an issue with how we're parsing Tokenserver settings in `local.toml` that I want to resolve before it lands on master

ethowitz:
@jrconlin I added a default for `tokenserver.database_url` so developers/SRE won't need to configure any Tokenserver settings to develop on/deploy syncstorage-rs (you only need to set Tokenserver settings if you intend to use Tokenserver). I also disabled the Tokenserver route just to be safe.

src/tokenserver/migrations/2021-07-16-001122_init/up.sql
```diff
@@ -0,0 +1,36 @@
+CREATE TABLE `services` (
```
pjenvey, line 1: I'll note that the services table may be able to be killed entirely.

Tokenserver has solely been used for one service, syncstorage. Either there's always been one entry in it with a pattern of "1.5" or possibly 2 entries (an additional one for pattern "1.1"). cc @rfk if he recalls this

It may make sense to keep it around for now to simplify things, particularly for the eventual migration, but FYI.

src/tokenserver/db/models.rs
```diff
@@ -1,37 +1,262 @@
...
+        "#;
+        let mut user_records = diesel::sql_query(query)
+            .bind::<Text, _>(email)
+            .load::<results::GetUser>(&self.inner.conn)?;
+
+        if user_records.is_empty() {
+            return Err(DbErrorKind::TokenserverUserNotFound.into());
+        }
+
+        user_records.sort_by_key(|user_record| (user_record.generation, user_record.created_at));
```
pjenvey, line 80: We probably want to copy the python version's SQL over directly which will push this sorting into the db layer

src/tokenserver/db/models.rs
```diff
@@ -1,37 +1,262 @@
...
+        };
+
+        Self {
+            inner: Arc::new(inner),
+        }
+    }
+
+    fn get_user_sync(&self, email: String) -> DbResult<results::GetUser> {
+        let query = r#"
+            SELECT users.uid, users.email, users.client_state, users.generation,
```
pjenvey, line 66: FYI: We've tried to standardize the SQL syntax in the rest of the syncstorage impl towards https://www.sqlstyle.guide/ but I won't hold you to it.

src/tokenserver/migrations/2021-07-16-001122_init/up.sql
rfk: > Either there's always been one entry in it with a pattern of "1.5" or possibly 2 entries (an additional one for pattern "1.1")

I believe there has only ever been a single entry in production use in this table; references to `1.1` are hangovers from testing and development.

rfk: Just to be very clear: +100 to simplifying the `services` table out of existence entirely

ethowitz: Cool -- I'll add a ticket to remove the `services` table as part of a separate ticket 👍🏻

src/tokenserver/db/models.rs
ethowitz: I think the legacy Tokenserver actually sorts the rows in memory to avoid making the DB indexes too big: https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/assignment/sqlnode/sql.py#L295

Do you think we should change our approach here?

src/tokenserver/db/models.rs
ethowitz: Good to know! I'll update the syntax I added in this PR.

src/tokenserver/db/models.rs
rfk: Two high-level comments on the legacy tokenserver's approach:

* The cost/benefit tradeoff there is entirely theoretical, I don't think we ever measured the potential bloat of indexes or established that it was too expensive.
* The alleged benefit depends on python's `sort` implementation having almost-linear performance on almost-sorted lists.

FWIW I recommend optimizing for simplicity of implementation, whatever that means in the new system.

ethowitz: I'm inclined to say we should just do the sorting in the code. It seems likely that, in the vast majority of cases, a user will only have a handful of old client states, which is what causes "duplicate" user records to be created. Keeping it in Rust seems to me to be less error-prone than writing it in SQL (though I don't think it matters too much). @pjenvey do you have a strong opinion either way?

pjenvey: My fault, I was looking solely at the `ORDER BY` clause of the python's get user records call and did not see that it also sorts in code (probably making the former pointless). I'm ok w/ sorting in memory.

pjenvey (dismissed):

ethowitz:
Had to rebase

jrconlin (approved):

PR #1122, opened 2021-07-27
chore: switch failure crate with thiserror

### LINKED ISSUE #1017: Move error handling off of `failure` crate
Failure is deprecated and is holding us back from using later versions of Sentry. 

We should convert to more modern error handling.

tiftran:
## Description

switch failure crate with thiserror and code changes that were required

## Testing

How should reviewers test?

## Issue(s)

Closes #1017.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ actix-web-httpauth = "0.5"
 actix-rt = "1"  # Pin to 1.0, due to dependencies on Tokio
 actix-cors = "0.5"
 async-trait = "0.1.40"
-backtrace = "0.3"
+backtrace = "0.3.61"
 base64 = "0.13"
 bb8 = "0.4.1"  # pin to 0.4 due to dependencies on Tokio
 bytes = "1.0"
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -58,7 +58,8 @@ protobuf = "2.20.0"
 rand = "0.8"
 regex = "1.4"
 # pin to 0.19: https://github.com/getsentry/sentry-rust/issues/277
-sentry = { version = "0.19", features = ["with_curl_transport"] }  # pin to 0.19 due to curl transport, failure
+sentry = { version = "0.19", features = ["with_curl_transport"] }# pin to 0.19 due to curl transport, failure
```
jrconlin, line 61: huh, we can probably drop the `failure` comment thing here, because we no longer need it.

jrconlin (approved):
Sorry for the delay. Thanks for this!

jrconlin:
... and looks like this picked up the new rust `1.54` clippy complaints. Most should be straight forward (removing an extra `&` that the code previously wanted), but let me know if there's anything super hairy.

PR #1134, opened 2021-08-04
remove generic from SpannerSessionManager

### LINKED ISSUE #697: Simplify DbPool async_trait usage and remove unnecessary generics after bb8 refactor
Two minor refactors related to the DB pool code after the bb8 refactor (#692):

1. A generic was added to `SpannerConnectionManager`. It is not used and should be removed.
2. When the `DbPool` trait was made async, it did not require the returned futures to be `Send`:
    https://github.com/mozilla-services/syncstorage-rs/blob/3a5fddb54b7641f7bb93d340d0728c64b70cf3bb/src/db/mod.rs#L66
    We can tighten this bound to require `Send` by changing the annotation to `#[async_trait]`. This is the default for `async_trait`, so this change will simplify the code.

tiftran:
## Description

Remove Generic from SpannerSessionManager

## Testing

How should reviewers test?

## Issue(s)

Closes #697.

### DIFF
--- a/src/db/spanner/manager/bb8.rs
+++ b/src/db/spanner/manager/bb8.rs
@@ -19,7 +19,7 @@ use super::session::{create_spanner_session, recycle_spanner_session, SpannerSes
 #[allow(dead_code)]
 pub type Conn<'a> = PooledConnection<'a, SpannerSessionManager<SpannerSession>>;
 
-pub struct SpannerSessionManager<T> {
+pub struct SpannerSessionManager {
     database_name: String,
     /// The gRPC environment
     env: Arc<Environment>,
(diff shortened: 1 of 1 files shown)


jrconlin:
I ran this through my local Spanner test suite and it worked correctly as well.

jrconlin:
A quick note on #697. 

Since that bug was filed, there was some refactoring that went on with DbPool. The `Send` trait is still required because `DbPool` is exchanged between threads by `HttpServer`. I don't believe that it's necessary to remove the `Send` trait from `Box<DbPool>`.

jrconlin (approved):

jrconlin:
Thanks!

Issue #1133, opened 2021-08-03
generation, keys_changed_at, and client_state: Errors and extractors


ethowitz:
Add types for Tokenserver-specific errors. `Serialize` and/or `Display` should be implemented such that the errors are presented in the exact same format as the errors from the legacy Tokenserver.

Implement extractors for the `X-KeyId` header, the database user record, `client_state`, the token-signing secret, the hashed FxA UID, and the hashed device ID.

Issue #866, opened 2020-10-19
Implement generation and keys_changed_at in the rust tokenserver route


fzzzy:
Implement generation_id and keys_changed_at in rust

there's a decent python testsuite for edgecases [here](https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/tests/test_service.py), which could be copied to rust

ethowitz:
[JIRA ticket](https://jira.mozilla.com/browse/CONSVC-173)

PR #1136, opened 2021-08-05
feat: generation, keys_changed_at, and client_state: Add database methods

### LINKED ISSUE #1132: generation, keys_changed_at, and client_state: Database methods
Add the following database methods to support implementing the required consistency checks for `generation` and `keys_changed_at`:

- `post_user`:
```sql
 INSERT INTO users (service, email, generation, client_state, created_at, nodeid, keys_changed_at, replaced_at)
      VALUES (?, ?, ?, ?, ?, ?, ?, NULL);
```
- `bulk_update_user`:
```sql
UPDATE users
   SET replaced_at = ?
 WHERE service = ?
   AND email = ?
   AND replaced_at IS NULL
   AND created_at < ?
```
- `put_user`:
```sql
UPDATE users
   SET generation = COALESCE(?, generation),
       keys_changed_at = COALESCE(?, keys_changed_at)
 WHERE service = ?
   AND email = ?
   AND generation <= COALESCE(?, generation)
   AND COALESCE(keys_changed_at, 0) <= COALESCE(?, keys_changed_at, 0)
   AND replaced_at IS NULL
```

Note that these queries come directly from this file from the legacy Tokenserver.

ethowitz:
## Description

Adds the relevant database methods necessary to implement the consistency checks between `generation`, `keys_changed_at`, and `client_state` and update the user, implemented [here](https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/views.py#L403) in the legacy Tokenserver.

## Testing

These methods are not actually used yet, so there's no way to test externally. The unit tests should suffice for now, until we can properly end-to-end test after #866 and #1133 are completed.

## Issue(s)

Closes #1132

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -141,7 +141,7 @@ macro_rules! build_app {
             // XXX: This route will be enabled when we are ready to roll out Tokenserver
             // Tokenserver
             // .service(
-            //     web::resource("/1.0/sync/1.5".to_string())
+            //     web::resource("/1.0/{application}/{version}".to_string())
             //         .route(web::get().to(tokenserver::handlers::get_tokenserver_result)),
             // )
             // Dockerflow
--- a/src/tokenserver/db/mod.rs
+++ b/src/tokenserver/db/mod.rs
@@ -1,5 +1,8 @@
 pub mod mock;
 pub mod models;
-mod params;
+pub mod params;
 pub mod pool;
 pub mod results;
+
+pub const SYNC_1_1_SERVICE_ID: i32 = 1;
+pub const SYNC_1_5_SERVICE_ID: i32 = 2;
--- a/src/tokenserver/db/models.rs
+++ b/src/tokenserver/db/models.rs
@@ -1,10 +1,8 @@
 use actix_web::web::block;
-#[cfg(test)]
-use diesel::sql_types::{Bigint, Integer, Nullable};
 use diesel::{
     mysql::MysqlConnection,
     r2d2::{ConnectionManager, PooledConnection},
-    sql_types::Text,
+    sql_types::{Bigint, Integer, Nullable, Text},
     RunQueryDsl,
 };
 #[cfg(test)]
@@ -12,7 +10,11 @@ use diesel_logger::LoggingConnection;
 use futures::future::LocalBoxFuture;
 use futures::TryFutureExt;
 
-use std::{self, sync::Arc};
+use std::{
+    result,
+    sync::Arc,
+    time::{SystemTime, UNIX_EPOCH},
+};
 
 use super::{params, results};
 use crate::db::error::{DbError, DbErrorKind};
@@ -20,7 +22,7 @@ use crate::error::ApiError;
 use crate::sync_db_method;
 
 pub type DbFuture<'a, T> = LocalBoxFuture<'a, Result<T, ApiError>>;
-pub type DbResult<T> = std::result::Result<T, DbError>;
+pub type DbResult<T> = result::Result<T, DbError>;
 type Conn = PooledConnection<ConnectionManager<MysqlConnection>>;
 
 #[derive(Clone)]
@@ -48,6 +50,13 @@ pub struct DbInner {
 }
 
 impl TokenserverDb {
+    // Note that this only works because an instance of `TokenserverDb` has *exclusive access* to
+    // a connection from the r2d2 pool for its lifetime. `LAST_INSERT_ID()` returns the ID of the
+    // most recently-inserted record *for a given connection*. If connections were shared across
+    // requests, using this function would introduce a race condition, as we could potentially
+    // get IDs from records created during other requests.
+    const LAST_INSERT_ID_QUERY: &'static str = "SELECT LAST_INSERT_ID() AS id";
+
     pub fn new(conn: Conn) -> Self {
         let inner = DbInner {
             #[cfg(not(test))]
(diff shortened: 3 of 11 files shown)


ethowitz:
Not ready for review yet -- adding some annotations to the PR to make it clearer

src/server/mod.rs
```diff
@@ -141,7 +141,7 @@ macro_rules! build_app {
             // XXX: This route will be enabled when we are ready to roll out Tokenserver
             // Tokenserver
             // .service(
-            //     web::resource("/1.0/sync/1.5".to_string())
```
ethowitz, line 144: I added support for both `sync-1.1` and `sync-1.5` for now. We may remove the `services` table entirely, at which point we would be able to remove this

src/tokenserver/db/mod.rs
```diff
@@ -1,5 +1,8 @@
 pub mod mock;
 pub mod models;
-mod params;
+pub mod params;
 pub mod pool;
 pub mod results;
+
+pub const SYNC_1_1_SERVICE_ID: i32 = 1;
```
ethowitz, line 7: I simplified the logic from the legacy Tokenserver by hard-coding the only two services that will ever exist with static DB IDs. This simplifies some of the database querying.

src/tokenserver/handlers.rs
```diff
@@ -82,7 +90,7 @@ pub async fn get_tokenserver_result(
                 let current_time = start.duration_since(UNIX_EPOCH).unwrap();
                 let expires = current_time + Duration::new(DEFAULT_TOKEN_DURATION, 0);
 
-                expires.as_secs_f64()
+                expires.as_secs()
```
ethowitz, line 93: On the legacy Tokenserver, the expires timestamp is an integer

src/tokenserver/db/models.rs
```diff
@@ -48,6 +50,13 @@ pub struct DbInner {
 }
 
 impl TokenserverDb {
+    // Note that this only works because an instance of `TokenserverDb` has *exclusive access* to
+    // a connection from the r2d2 pool for its lifetime. `LAST_INSERT_ID()` returns the ID of the
+    // most recently-inserted record *for a given connection*. If connections were shared across
+    // requests, using this function would introduce a race condition, as we could potentially
+    // get IDs from records created during other requests.
+    const LAST_INSERT_ID_QUERY: &'static str = "SELECT LAST_INSERT_ID() AS id";
```
ethowitz, line 58: Just drawing attention to my comment here to make sure my logic checks out 😇

src/tokenserver/db/models.rs
```diff
@@ -61,152 +70,294 @@ impl TokenserverDb {
...
-                   users.keys_changed_at, users.created_at, nodes.node
-              FROM users
-              JOIN nodes
-                ON nodes.id = users.nodeid
-             WHERE users.email = ?
+    /// Get the most current user record for the given email and service ID. This function also
+    /// marks any old user records as replaced, in case of data races that may have occurred
+    /// during row creation.
+    fn get_user_sync(&self, params: params::GetUser) -> DbResult<results::GetUser> {
+        const QUERY: &str = r#"
```
ethowitz, line 77: I updated this query to more closely match the query used on the legacy Tokenserver. The analogous database method on the legacy Tokenserver is [here](https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/assignment/sqlnode/sql.py#L285)

src/tokenserver/db/models.rs
```diff
@@ -61,152 +70,294 @@ impl TokenserverDb {
...
-            .bind::<Integer, _>(node.backoff)
-            .execute(&self.inner.conn)?;
+        Ok(user)
+    }
 
-        let query = r#"
-            SELECT id
-              FROM nodes
+    /// Mark users matching the given email and service ID as replaced.
+    fn replace_users_sync(&self, params: params::ReplaceUsers) -> DbResult<results::ReplaceUsers> {
```
ethowitz, line 136: Analogous legacy Tokenserver DB method is [here](https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/assignment/sqlnode/sql.py#L516)

src/tokenserver/db/models.rs
```diff
@@ -61,152 +70,294 @@ impl TokenserverDb {
...
             .map_err(Into::into)
     }
 
-    #[cfg(test)]
-    fn post_service_sync(&self, service: params::PostService) -> DbResult<results::PostService> {
-        let query = r#"
-            INSERT INTO services (service, pattern)
-            VALUES (?, ?)
+    /// Mark the user with the given uid and service ID as being replaced.
+    fn replace_user_sync(&self, params: params::ReplaceUser) -> DbResult<results::ReplaceUser> {
```
ethowitz, line 158: Analogous legacy Tokenserver DB method is [here](https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/assignment/sqlnode/sql.py#L526)

src/tokenserver/db/models.rs
```diff
@@ -61,152 +70,294 @@ impl TokenserverDb {
...
+            .bind::<Integer, _>(params.service_id)
+            .bind::<Bigint, _>(params.uid)
+            .execute(&self.inner.conn)
+            .map(|_| ())
+            .map_err(Into::into)
+    }
+
+    /// Update the user with the given email and service ID with the given `generation` and
+    /// `keys_changed_at`.
+    fn put_user_sync(&self, params: params::PutUser) -> DbResult<results::PutUser> {
```
ethowitz, line 177: Analogous legacy Tokenserver DB method is [here](https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/assignment/sqlnode/sql.py#L397). Note that my approach is different here: I plan to move any `client_state`-checking logic to the handler. That seems more like "business logic" to me, and I'd prefer to keep the database methods slim.

src/tokenserver/db/models.rs
```diff
@@ -61,152 +70,294 @@ impl TokenserverDb {
...
+            .bind::<Text, _>(&params.email)
+            .bind::<Bigint, _>(params.generation)
+            .bind::<Nullable<Bigint>, _>(params.keys_changed_at)
+            .execute(&self.inner.conn)
+            .map(|_| ())
             .map_err(Into::into)
     }
 
-    #[cfg(test)]
+    /// Create a new user.
```
ethowitz, line 206: There's not really an analogous database method on the legacy Tokenserver. User "creation" is referred to as ["allocation"](https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/assignment/sqlnode/sql.py#L368), and I'm not sure how I'm going to handle that yet. I may move some of the allocation logic to the handler

ethowitz: The query I used is based off of the one [here](https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/assignment/sqlnode/sql.py#L58)

src/tokenserver/db/models.rs
jrconlin: I'll note that `LAST_INSERT_ID()` is problematic, but then so is damn near any other mechanism.

src/tokenserver/db/models.rs
jrconlin: well, [here](https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/assignment/sqlnode/sql.py#L44)

src/tokenserver/db/models.rs
jrconlin: FWIW, some distributed databases charge per MB output, so moving filtering logic into the query can make some sense. For the data we're talking about here, it's not cost effective. (There are also some other storage systems that charge on data collected before filtering is done, and then charge on data output as well, but that's not the case here.)

src/tokenserver/handlers.rs
jrconlin: I was thinking that we were doing something around shifting the time up by 1000 in sync, but that's for the TTL.

jrconlin (approved):

src/tokenserver/db/models.rs
ethowitz: Problematic enough not to use it?

src/tokenserver/db/models.rs
ethowitz: I was linking to the actual Python methods that served as inspiration for these Rust methods, but yeah, I should probably have included links to the queries as well

ethowitz:
Had to rebase

jrconlin (approved):

PR #1139, opened 2021-08-09
removed send from async_trait for DbPool

### LINKED ISSUE #1138: Remove Send from async_trait for DbPool so that it's required for DbPool implementations
Follow up issue based on comment:

> Since that bug was filed, there was some refactoring that went on with DbPool. The `Send` trait is still required because `DbPool` is exchanged between threads by `HttpServer`. I don't believe that it's necessary to remove the `Send` trait from `Box<DbPool>`.

https://github.com/mozilla-services/syncstorage-rs/pull/1134#issuecomment-893879805

This bug doesn't say to remove the `Send` bound, but to require it for `DbPool` implementations (by removing the `?Send`, which allows non-`Send` implementations). Therefore, this issue should be reopened since it's not fully fixed (or split it in two).

_Originally posted by @AzureMarker in https://github.com/mozilla-services/syncstorage-rs/issues/697#issuecomment-894590847_

tiftran:
## Description

removed send from async_trait for DbPool so that send is required

## Testing

How should reviewers test?

## Issue(s)

Closes #1138.

### DIFF
--- a/src/db/mock.rs
+++ b/src/db/mock.rs
@@ -14,7 +14,7 @@ impl MockDbPool {
     }
 }
 
-#[async_trait(?Send)]
+#[async_trait]
 impl DbPool for MockDbPool {
     async fn get<'a>(&'a self) -> ApiResult<Box<dyn Db<'a>>> {
         Ok(Box::new(MockDb::new()) as Box<dyn Db<'a>>)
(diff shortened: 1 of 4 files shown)


tiftran:
So here's my second attempt...

jrconlin (approved):

PR #1140, opened 2021-08-10
chore: update actix-http version due to RUSTSEC-2021-0081

tiftran:
## Description
RUSTSEC-2021-0081 appeared due to current version of actix-http being 2.2.0. 

## Testing
RUSTSEC-2021-0081 should not be appearing in `cargo audit`

## Issue(s)

Closes [link](link).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -53,9 +53,9 @@ dependencies = [
 
 [[package]]
 name = "actix-http"
-version = "2.2.0"
+version = "2.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "452299e87817ae5673910e53c243484ca38be3828db819b6011736fc6982e874"
+checksum = "5cb8958da437716f3f31b0e76f8daf36554128517d7df37ceba7df00f09622ee"
 dependencies = [
  "actix-codec",
  "actix-connect",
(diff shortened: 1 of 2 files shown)


tiftran:
Not sure if this is the right way/process to go about this, but I noticed this in #1139

jrconlin:
Normally we usually just do a `cargo update` and then commit the `Cargo.lock`.

Rust versioning is a bit odd, but the crux of it is Use the Most Significant Digit of the version. So 
`actix-http = "2"`
says to use any version of `actix-http` so long as it's major version is `2` and
`actix-http = "2.2.1"` 
says to only use the 2.2.1 version of `actix-http`. 

Which actual version you use is defined in the `Cargo.lock` file, so a `cargo update` goes and applies the rules you've specified in `Cargo.toml`.

jrconlin (approved):

AzureMarker:
> `actix-http = "2.2.1"`
> says to only use the 2.2.1 version of actix-http.

@jrconlin This actually says "use any version greater than 2.2.1 but less than 3.0.0". To force a specific version, use `actix-http = "=2.2.1"`.
More info: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html

PR #1145, opened 2021-08-26
feat: Tokenserver: Add validations and user updating for generation, keys_changed_at, and client_state

### LINKED ISSUE #866: Implement generation and keys_changed_at in the rust tokenserver route

ethowitz:
## Description

Adds a series of validations involving the claims provided in a Tokenserver request.

## Testing

* Enable test mode by setting `tokenserver.test_mode_enabled = true` in `local.toml`
* Send test requests using [httpie](https://httpie.io/) like so:
```
http GET localhost:8000/1.0/sync/1.5 'Authorization:Bearer <your token here>` `X-KeyID:<your X-KeyID here>`
```

When test mode is activated, the only part of the JWT that matters is the payload, so a token of the format `.<base64-encoded payload>.` (note the placement of the `.`s) should work fine. The payload should have the following claims:
* `client_id`: The client ID doesn't matter for testing this PR
* `scope`: The scope doesn't matter for testing this PR
* `fxa-generation`: This value is mapped to `generation` in the code
* `fxa-profileChangedAt`: This value doesn't matter for testing this PR

The `X-KeyID` header is of the format `<keys changed at>-<base64-encoded client state>`.

The following cases should be tested:
* The generation provided in the request is lower than that currently stored on the user
* The keys changed at provided in the request is lower than that currently stored on the user
* An old client state is used
* Keys changed it is being updated without an update to generation
* Client state is being updated without a change in keys changed at
* Client state is being updated without a change in generation

## Issue(s)

Closes #866

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,7 +38,7 @@ commands:
           command: |
             cargo fmt -- --check
             # https://github.com/bodil/sized-chunks/issues/11
-            cargo audit --ignore RUSTSEC-2020-0041 --ignore RUSTSEC-2021-0078 --ignore RUSTSEC-2021-0079
+            cargo audit --ignore RUSTSEC-2020-0041 --ignore RUSTSEC-2021-0078 --ignore RUSTSEC-2021-0079 --ignore RUSTSEC-2020-0159 --ignore RUSTSEC-2020-0071
   python-check:
     steps:
       - run:


ethowitz:
I want to wait to merge this until [this pull request](https://github.com/mozilla-services/syncstorage-rs/pull/1152) is merged

src/tokenserver/db/models.rs
```diff
@@ -94,6 +90,7 @@ impl TokenserverDb {
         }
 
         raw_users.sort_by_key(|raw_user| (raw_user.generation, raw_user.created_at));
+        raw_users.reverse();
```
ethowitz, line 93: This was a bug (we were in ascending order instead of descending order)

src/tokenserver/error.rs
```diff
@@ -78,13 +80,13 @@ impl TokenserverError {
         }
     }
 
-    pub fn unsupported(description: &'static str) -> Self {
+    pub fn unsupported(description: &'static str, name: String) -> Self {
```
ethowitz, line 83: These changes are unrelated to the original issue (#866) but address a failing integration test. In short, when a client requests an invalid application or version, we need to include the errant application/version in the error message

tools/integration_tests/tokenserver/test_authorization.py
```diff
@@ -517,3 +517,21 @@ def test_x_client_state_must_have_same_client_state_as_key_id(self):
         self.assertEqual(res.json, expected_error_response)
         headers['X-Client-State'] = '616161'
         res = self.app.get('/1.0/sync/1.5', headers=headers)
+
+    def test_x_key_id_header_required(self):
```
ethowitz, line 521: The new Tokenserver will not accept any requests that do not have an `X-KeyID` header, since it only supports OAuth

ethowitz:
This is ready for review now

PR #1151, opened 2021-09-16
feat: Tokenserver: Add support for client-specified token duration

### LINKED ISSUE #1050: Tokenserver: Add support for client-specified token duration
The old Tokenserver allows clients to specify a shorter token duration via a query parameter. Let's replicate this functionality in the Rust version.

ethowitz:
## Description

Setting the `duration` query parameter adjusts the expiry time of the token returned by Tokenserver. If the query parameter cannot be parsed or if the query parameter is greater than the default token duration set on the server, the default token duration is used instead.

## Testing

This can be tested by sending requests to Tokenserver using HTTPie like so:
```bash
http GET localhost:8000/1.0/sync/1.5 'Authorization:Bearer <insert JWT here>' 'X-KeyID:<insert kid here>' duration==100
```
You can inspect the 'duration' field of the JSON returned by the endpoint to ensure its value matches your expectations.

## Issue(s)

Closes #1050

### DIFF
--- a/src/tokenserver/extractors.rs
+++ b/src/tokenserver/extractors.rs
@@ -104,7 +104,7 @@ impl FromRequest for TokenserverRequest {
 
                 // An error in the "duration" query parameter should never cause a request to fail.
                 // Instead, we should simply resort to using the default token duration.
-                params.duration.clone().and_then(|duration_string| {
+                params.duration.as_ref().and_then(|duration_string| {
                     match duration_string.parse::<u64>() {
                         // The specified token duration should never be greater than the default
                         // token duration set on the server.
--- a/src/tokenserver/extractors.rs
+++ b/src/tokenserver/extractors.rs
@@ -99,7 +99,20 @@ impl FromRequest for TokenserverRequest {
 
                 db.get_user(params::GetUser { email, service_id }).await?
             };
-            let TokenDuration(duration) = TokenDuration::extract(&req).await?;
+            let duration = {
+                let params = Query::<QueryParams>::extract(&req).await?;
+
+                // An error in the "duration" query parameter should never cause a request to fail.
+                // Instead, we should simply resort to using the default token duration.
+                params.duration.clone().and_then(|duration_string| {
+                    match duration_string.parse::<u64>() {
+                        // The specified token duration should never be greater than the default
+                        // token duration set on the server.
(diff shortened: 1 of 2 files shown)


src/tokenserver/extractors.rs
```diff
@@ -102,13 +111,46 @@ impl FromRequest for TokenserverRequest {
...
             };
 
             Ok(tokenserver_request)
         })
     }
 }
 
+#[derive(Deserialize)]
+struct QueryParams {
+    pub duration: Option<String>,
```
pjenvey, line 124: serde can make this conversion for us

```suggestion
    pub duration: Option<u64>,
```

src/tokenserver/extractors.rs
```diff
@@ -102,13 +111,46 @@ impl FromRequest for TokenserverRequest {
...
+}
+
+struct TokenDuration(Option<u64>);
+
+impl FromRequest for TokenDuration {
```
pjenvey, line 129: nit: I'd argue `TokenDuration` doesn't warrant a struct and its own extractor if its to only be extracted once in `TokenserverRequest` but keeping it isn't a big deal.

pjenvey (changes requested):

src/tokenserver/extractors.rs
ethowitz: Won't the request fail if we let serde attempt the conversion to `u64` and the client sends a string? In that case, we just want to ignore the query param, but we don't want the request to fail altogether

src/tokenserver/extractors.rs
```diff
@@ -91,6 +99,20 @@ impl FromRequest for TokenserverRequest {
 
                 db.get_user(params::GetUser { email, service_id }).await?
             };
+            let duration = {
+                let params = Query::<QueryParams>::extract(&req).await?;
+
+                // An error in the "duration" query parameter should never cause a request to fail.
+                // Instead, we should simply resort to using the default token duration.
+                params.duration.clone().and_then(|duration_string| {
```
pjenvey, line 107: nit:

```suggestion
                params.duration.as_ref().and_then(|duration_string| {
```

pjenvey (approved):
looks good but I added another small nit

src/tokenserver/extractors.rs
ethowitz: Thanks for the catch! Updated

pjenvey (approved):

PR #1171, opened 2021-11-18
fix: resolve intermittent Tokenserver test failure

### LINKED ISSUE #1170: Fix Tokenserver test failure
There is a [test failure](https://app.circleci.com/pipelines/github/mozilla-services/syncstorage-rs/3603/workflows/0221d7de-48de-4abe-89bb-2d4fe6d2bb8a/jobs/13200?invite=true#step-112-40) on master relating to a test added in 0ac30958de5cfca0f3d44dfb479b615cae7ede27. This test does not fail locally and did not fail on the feature branch.

ethowitz:
## Description

Add `time.sleep`s to fix an intermittent timestamp issue in a unit test

## Testing

The failing test should pass on CI

## Issue(s)

Closes #1170

### DIFF
--- a/tools/tokenserver/test_database.py
+++ b/tools/tokenserver/test_database.py
@@ -183,7 +183,6 @@ class TestDatabase(unittest.TestCase):
         time.sleep(0.1)
         # That should be a total of 7 old records.
         old_records = list(self.database.get_old_user_records(0))
-        print([record[-1] for record in old_records])
         self.assertEqual(len(old_records), 7)
         # Sleep briefly to ensure that time has passed since the last
         # operation
--- a/tools/tokenserver/test_database.py
+++ b/tools/tokenserver/test_database.py
@@ -162,60 +162,54 @@ class TestDatabase(unittest.TestCase):
         # Do a sleep halfway through so we can test use of grace period.
         email1 = 'test1@mozilla.com'
         user1 = self.database.allocate_user(email1)
+        # We have to sleep between every user create/update operation: if two
+        # users are created with the same timestamp, it can lead to a
+        # situation where two active user records exist for a single email.
+        time.sleep(0.1)
         self.database.update_user(user1, client_state='a')
+        time.sleep(0.1)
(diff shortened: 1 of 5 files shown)


ethowitz:
This is still not working...converting to a draft

ethowitz:
Whew. The problem was that users were sometimes being created with the same timestamp, which caused a problem [here](https://github.com/mozilla-services/syncstorage-rs/blob/fix/1170-fix-tokenserver-test-failure/tools/tokenserver/database.py#L394): since the two most recently created users share the same timestamp, _neither_ of those users were being replaced, resulting in two "active" user records for a given email address.

jrconlin:
oh, huh.
I hate to say it, but I think I remember a similar issue cropping up with Sync. Basically, things were working far faster in the current environment and what used to take a few nanos to complete was effectively being done in parallel. Guessing you hit that too.

tools/tokenserver/test_database.py
```diff
@@ -162,20 +162,32 @@ def test_cleanup_of_old_records(self):
         # Do a sleep halfway through so we can test use of grace period.
         email1 = 'test1@mozilla.com'
         user1 = self.database.allocate_user(email1)
+        # We have to sleep between every user create/update operation: if two
```
jrconlin, line 165: I do wonder if this might eventually bite us (e.g. if the client is also in rust and is on a particularly close node, they may respond faster than one ms.) We might want to add an issue to check for active user records before writing a second and returning a 409 CONFLICT if one is detected.

jrconlin (approved):

tools/tokenserver/test_database.py
ethowitz: I'm not sure that the risk of this happening on the new Tokenserver is any greater than the risk of it happening on the old Tokenserver 🤔 Requests on the new Tokenserver will still take well over 1 ms to complete, so I _think_ the only way this could happen would be if a client decided to bombard us with requests, and I think the old Tokenserver is just as vulnerable to that

PR #1176, opened 2021-11-22
feat: add missing Tokenserver response fields

### LINKED ISSUE #1173: Add hashalg and node_type fields to TokenserverResponse
The `TokenserverResponse` type is missing two fields (`hashalg` and `node_type`), causing an integration test to fail. These fields should be added so the response matches that of the legacy Tokenserver.

**Acceptance Criteria**
* Add a hashalg field that always has value `"sha256"`
* Add a node_type field that has value `"mysql"` or `"spanner"`, depending on a tokenserver.node_type setting

ethowitz:
## Description

Adds `hashalg` and `node_type` to the Tokenserver response.

## Testing

Sending a request to Tokenserver will yield a response with the two added fields. Alternatively, the end-to-end integration tests test that these fields are present.

## Issue(s)

Closes #1173

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -176,6 +176,7 @@ impl Settings {
         s.set_default("tokenserver.fxa_email_domain", "test.com")?;
         s.set_default("tokenserver.fxa_metrics_hash_secret", "secret")?;
         s.set_default("tokenserver.test_mode_enabled", false)?;
+        s.set_default("tokenserver.node_type", "spanner")?;
 
         // Merge the config file if supplied
         if let Some(config_filename) = filename {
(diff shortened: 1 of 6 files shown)


src/settings.rs
```diff
@@ -176,6 +176,7 @@ impl Settings {
         s.set_default("tokenserver.fxa_email_domain", "test.com")?;
         s.set_default("tokenserver.fxa_metrics_hash_secret", "secret")?;
         s.set_default("tokenserver.test_mode_enabled", false)?;
+        s.set_default("tokenserver.node_type", "spanner")?;
```
ethowitz, line 179: It's sufficient to use a setting to set the node type, since a given instance of Tokenserver will only ever be responsible for Spanner nodes **or** MySQL nodes, but never both

jrconlin (approved):

jrconlin (approved):

PR #1177, opened 2021-11-26
changes to allow cors allowed origin and max age to be configurable

### LINKED ISSUE #930: Tune the actix-cors permissions for the newer library
Per https://github.com/mozilla-services/syncstorage-rs/issues/928, we did a quick patch that relies on using nginx CORS wrappers for fine-tuned security. 

We should add better support around origin and other aspects of the CORS specification, due to increased actix-cors permission guidance. This may include adding additional options.

tiftran:
## Description

changes to allow cors allowed origin and max age to be configurable

## Testing

How should reviewers test?

## Issue(s)

Closes #930

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -3,6 +3,7 @@ use std::{cmp::min, env};
 
 use actix_cors::Cors;
 use config::{Config, ConfigError, Environment, File};
+use http::method::Method;
 use serde::{de::Deserializer, Deserialize, Serialize};
 use url::Url;
 
@@ -10,6 +11,10 @@ use crate::db::spanner::models::MAX_SPANNER_LOAD_SIZE;
 use crate::error::ApiError;
 use crate::tokenserver::settings::Settings as TokenserverSettings;
 use crate::web::auth::hkdf_expand_32;
+use crate::web::{
+    X_LAST_MODIFIED, X_VERIFY_CODE, X_WEAVE_BYTES, X_WEAVE_NEXT_OFFSET, X_WEAVE_RECORDS,
+    X_WEAVE_TIMESTAMP, X_WEAVE_TOTAL_BYTES, X_WEAVE_TOTAL_RECORDS,
+};
 
 static DEFAULT_PORT: u16 = 8000;
 
@@ -93,7 +98,6 @@ pub struct Settings {
     pub cors_max_age: Option<usize>,
     pub cors_allowed_methods: Option<Vec<String>>,
     pub cors_allowed_headers: Option<Vec<String>>,
-
 }
 
 impl Default for Settings {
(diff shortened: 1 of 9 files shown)


src/settings.rs
```diff
@@ -272,6 +279,25 @@ impl Settings {
...
         format!("http://{}:{} ({}) {}", self.host, self.port, db, quota)
     }
+
+    pub fn build_cors(&self) -> Cors {
+        // Followed by the "official middleware" so they run first.
+        // actix is getting increasingly tighter about CORS headers. Our server is
+        // not a huge risk but does deliver XHR JSON content.
+        // For now, let's be permissive and use NGINX (the wrapping server)
+        // for finer grained specification.
+        let mut cors = Cors::permissive();
```
tiftran, line 289: it wasn't too happy with me when I used `Cors::default()`...  so I used permissive and below overrode allowed origin, if it was available?

tools/integration_tests/run.py
```diff
@@ -37,17 +37,20 @@ def terminate_process(process):
...
         # TODO we should change this to watch for a log message on startup
         # to know when to continue instead of sleeping for a fixed amount
         time.sleep(20)
 
         return the_server_subprocess
 
-    os.environ.setdefault('SYNC_MASTER_SECRET', 'secret0')
+    os.environ.setdefault("SYNC_MASTER_SECRET", "secret0")
+    os.environ.setdefault("SYNC_CORS_MAX_AGE", "555")
+    os.environ.setdefault("SYNC_CORS_ALLOWED_ORIGIN", "localhost")
```
tiftran, line 52: is this kinda jank to make integration tests feasible?

jrconlin: Yep. Since we're creating a test server, doing this sort of thing is fine. We're just setting initialization parameters.

src/settings.rs
jrconlin: IIRC, `Cors::default()` is completely restrictive. You're supposed to add what you're using. `Cors::permissive()` is the opposite. For good security hygene we should probably be as explicit.

so:
```
cors = cors
	.allow_methods(vec![ "DELETE", "GET", "POST", "PUT"])
	.allow_headers(vec![
		"Authorization",
		"Content-Type",
		"UserAgent",
		crate::web::X_LAST_MODIFIED,
		crate::web::X_WEAVE_TIMESTAMP,
		crate::web::X_WEAVE_NEXT_OFFSET,
		crate::web::X_WEAVE_RECORDS,
		// might want to add these to `crate::web` so we're consistent.
		"X-Weave-Bytes",
		"X-Weave-Total-Records",
		"X-Weave-Total-Bytes",
		"X-Verify-Code",
		// I think this is only used for tests.
		"TEST_IDLES"
	]);
```

Definitely want to verify those. I just did a quick scan of `server/mod.rs` for the methods and a file scan for `headers().get(` to find the Header values. Obviously, before we push we'll want QA to run a client test to make sure things are working right. You could always fire up a local profile that points to a locally running server if you want to make their lives  bit easier.

jrconlin (approved):

PR #1181, opened 2021-12-02
chore: update 12/2/21

ethowitz:
## Description

Updates for the new rust version

## Testing

No changes to functionality

## Issue(s)

N/A

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM rust:1.54-buster as builder
+FROM rust:1.57-buster as builder
 WORKDIR /app
 ADD . /app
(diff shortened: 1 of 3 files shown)


jrconlin:
Gonna suggest holding off on this for ~ 24 hours.

The docker images are usually updated the day after a given release. That way you can update the Dockerfile versions as well.

pjenvey (commented):
We might as well upgrade the rust docker version in our Dockerfile as well here

jrconlin:
Sigh.

Guessing that there was a goof in the auto deploy for this package. Happens now and then. We can retry this Monday or so.

pjenvey (approved):

PR #1182, opened 2021-12-02
route get / requests to sync docs

### LINKED ISSUE #1097: Route `GET /` requests to Sync documentation
per transparency bug, we should 302 users who send HTTP requests to our server root to the Sync documentation.

tiftran:
## Description

route GET "/" to docs

## Testing

How should reviewers test?

## Issue(s)

Closes #1097 .

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -19,8 +19,6 @@ use crate::web::{handlers, middleware};
 
 pub const BSO_ID_REGEX: &str = r"[ -~]{1,64}";
 pub const COLLECTION_ID_REGEX: &str = r"[a-zA-Z0-9._-]{1,32}";
-pub const SYNC_DOCS_URL: &str =
-    "https://mozilla-services.readthedocs.io/en/latest/storage/apis-1.5.html";
 const MYSQL_UID_REGEX: &str = r"[0-9]{1,10}";
 const SYNC_VERSION_PATH: &str = "1.5";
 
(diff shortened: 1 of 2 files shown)


jrconlin (dismissed):
LGTM. (Just need to fix up the two checks. Drop the unused field and change the 
```
impl Default ... 
```
to 
```
#[derive(Default, ...)]
```

tiftran:
@jrconlin just wanna check if i have the "right" doc

jrconlin:
Great question.

The idea behind this is if someone sees this host URL, they can go to the root and get something useful. I think, in that case, pointing them to "https://www.mozilla.org/firefox/sync/" might be better than the API docs. 

I'm also wondering if it might be a good idea to pull the SYNC_DOCS_URL from a `settings` variable, that way, if we ever want to change things, we don't need to patch the code, just tweak the setting value. 

You can pass the `settings` as a variable to the macro.

tiftran:
> I'm also wondering if it might be a good idea to pull the SYNC_DOCS_URL from a `settings` variable, that way, if we ever want to change things, we don't need to patch the code, just tweak the setting value.

After spending time with @jrconlin  at considering the above option, it seems that it may require more worth than its worth. We decided to leave it as is for now, and take another look at it in the future.

jrconlin:
As a bit of follow-up, this is a case where we eventually need to construct a static string from a value being passed into a `move` block, so there's a lot going on. While it would be nice to change the value without needing a re-compile, the site we point to shouldn't change very often, so doing a redeploy would just be more annoying than fatal. I've filed [a potential follow up issue](https://github.com/mozilla-services/syncstorage-rs/issues/1183) with some additional info if others are interested in trying it out.

jrconlin (approved):

PR #1184, opened 2021-12-07
test: add Tokenserver load tests

### LINKED ISSUE #1107: Tokenserver: Add load tests
The legacy Tokenserver has a suite of load tests [here](https://github.com/mozilla-services/tokenserver/tree/master/loadtest), which could be adapted to work with the new Tokenserver. These load tests currently stub out the calls to FxA for OAuth token verification; in order to do that in Rust, we'd need to support starting the server in a "test mode" that stubs out the OAuth verifier. While this would work, it might mischaracterize the performance of Tokenserver, since the web requests to FxA will likely contribute substantially to the overall performance of Tokenserver.

ethowitz:
## Description

Adds a suite of locust load tests for Tokenserver.

## Testing

The load test suite currently doesn't run well against the new Tokenserver (I think there is an issue without how I configured the database; I'll look into this separately). Running it against the old Tokenserver should work without issue (you'll need to configure the old Tokenserver to verify OAuth tokens through the mock service at https://mock-oauth-stage.dev.lcip.org).

## Issue(s)

Closes #1107

### DIFF
--- a/src/tokenserver/db/pool.rs
+++ b/src/tokenserver/db/pool.rs
@@ -78,8 +78,8 @@ impl From<actix_web::error::BlockingError<DbError>> for DbError {
 #[async_trait]
 impl DbPool for TokenserverPool {
     async fn get(&self) -> Result<Box<dyn Db>, DbError> {
-        let pool = self.inner.clone();
-        let conn = block(move || pool.get().map_err(DbError::from)).await?;
+        let pool = self.clone();
+        let conn = block(move || pool.inner.get().map_err(DbError::from)).await?;
 
         Ok(Box::new(TokenserverDb::new(conn)) as Box<dyn Db>)
     }
--- a/src/tokenserver/extractors.rs
+++ b/src/tokenserver/extractors.rs
@@ -348,7 +348,7 @@ impl FromRequest for KeyId {
             }
 
             let x_key_id = headers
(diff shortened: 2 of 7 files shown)


src/tokenserver/db/pool.rs
```diff
@@ -78,8 +78,8 @@ impl From<actix_web::error::BlockingError<DbError>> for DbError {
 #[async_trait]
 impl DbPool for TokenserverPool {
     async fn get(&self) -> Result<Box<dyn Db>, DbError> {
-        let pool = self.inner.clone();
-        let conn = block(move || pool.get().map_err(DbError::from)).await?;
+        let pool = self.clone();
+        let conn = block(move || pool.inner.get().map_err(DbError::from)).await?;
```
ethowitz, line 82: This was a bug -- we should be cloning `TokenserverPool` not the r2d2 `Pool`

src/tokenserver/extractors.rs
```diff
@@ -348,7 +348,7 @@ impl FromRequest for KeyId {
             }
 
             let x_key_id = headers
-                .get("X-KeyId")
+                .get("X-KeyID")
```
ethowitz, line 351: I don't think this matters, but good for consistency

jrconlin (approved):

PR #1190, opened 2021-12-20
fix: move I/O calls to blocking threadpool

### LINKED ISSUE #1188: Move I/O calls to the blocking threadpool
Currently, the [`verify_token`](https://github.com/mozilla-services/syncstorage-rs/blob/master/src/tokenserver/extractors.rs#L313), [`get_or_create_user`](https://github.com/mozilla-services/syncstorage-rs/blob/master/src/tokenserver/db/models.rs#L472), and [`allocate_user`](https://github.com/mozilla-services/syncstorage-rs/blob/master/src/tokenserver/db/models.rs#L423) function calls all run on the worker threads Actix initializes to handle requests. This means that, when these calls are waiting for I/O, they are blocking other work from happening. This wouldn't be as much of an issue if Actix spawned many worker threads, but normally, there are only as many workers as there are CPUs on the host. The best solution to this problem is to move these long-running I/O tasks to the blocking threadpool via [`block`](https://docs.rs/actix-web/1.0.3/actix_web/web/fn.block.html), where there are many more threads available to handle concurrent I/O tasks.

ethowitz:
## Description

Moves blocking I/O method calls to the blocking threadpool.

## Testing

Running the load test suite against Tokenserver with 100 users should result in no errors.

## Issue(s)

Closes #1188

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ docker_stop_spanner:
 	docker-compose -f docker-compose.spanner.yaml down
 
 run:
-	RUST_LOG=debug RUST_BACKTRACE=full cargo run -- --config config/local.toml --features tokenserver_test_mode
+	RUST_LOG=debug RUST_BACKTRACE=full cargo run --features tokenserver_test_mode -- --config config/local.toml
 
 run_spanner:
 	GOOGLE_APPLICATION_CREDENTIALS=$(PATH_TO_SYNC_SPANNER_KEYS) GRPC_DEFAULT_SSL_ROOTS_FILE_PATH=$(PATH_TO_GRPC_CERT) make run
--- a/src/tokenserver/db/models.rs
+++ b/src/tokenserver/db/models.rs
@@ -411,16 +411,17 @@ impl TokenserverDb {
             let start = SystemTime::now();
             start.duration_since(UNIX_EPOCH).unwrap().as_millis() as i64
         };
-        let uid = self.post_user_sync(params::PostUser {
-            service_id: params.service_id,
-            email: params.email.clone(),
-            generation: params.generation,
-            client_state: params.client_state.clone(),
-            created_at,
-            node_id: node.id,
-            keys_changed_at: params.keys_changed_at,
-        })?
-        .id;
+        let uid = self
+            .post_user_sync(params::PostUser {
+                service_id: params.service_id,
+                email: params.email.clone(),
+                generation: params.generation,
+                client_state: params.client_state.clone(),
+                created_at,
+                node_id: node.id,
+                keys_changed_at: params.keys_changed_at,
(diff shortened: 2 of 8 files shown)


Makefile
```diff
@@ -36,7 +36,7 @@ docker_stop_spanner:
 	docker-compose -f docker-compose.spanner.yaml down
 
 run:
-	RUST_LOG=debug RUST_BACKTRACE=full cargo run -- --config config/local.toml --features tokenserver_test_mode
+	RUST_LOG=debug RUST_BACKTRACE=full cargo run --features tokenserver_test_mode -- --config config/local.toml
```
ethowitz, line 39: This was a bug: https://doc.rust-lang.org/cargo/commands/cargo-run.html#description

src/tokenserver/db/models.rs
```diff
@@ -264,6 +265,171 @@ impl TokenserverDb {
...
+            .map_err(Into::into)
+    }
+
+    /// Gets the user with the given email and service ID, or if one doesn't exist, allocates a new
+    /// user.
+    fn get_or_create_user_sync(
+        &self,
+        params: params::GetOrCreateUser,
+    ) -> DbResult<results::GetOrCreateUser> {
+        let mut raw_users = self.get_users_sync(params::GetUsers {
```
ethowitz, line 293: I couldn't figure out how to have an async `get_or_create_users` method that runs on the blocking threadpool, since defining that method on the trait would result in the the method body running on an actix worker thread (though the database calls would still run on the blocking threadpool). This seemed preferable given those two options, but this method as it's defined will block a thread for quite a long time, since there are no `await`s. Definitely open to suggestions here

jrconlin: Hashing a few things out, so my apologies in advance.
There are two blocking operations in this thread, both of which potentially block the Actix event loop. `get_users_sync`, and `allocate_user_sync`, both of which use the `dbpool` which runs on it's own system thread. `allocate_user_sync` does three operations, while `get_users_sync` just does one GET. If we wanted to free up the actix event loop during those db calls, we could probably wrap those in futures. 

We might want to just put some profile metrics on `allocate_user_sync` to see what sort of delays it introduces before we optimize things.

jrconlin (approved):

src/tokenserver/db/models.rs
pjenvey: It could potentially be async, since it no longer has any direct calls to diesel now. From an async block it'd call the async versions of the individual db calls it makes -- it looks like it has 3 of them -- each would up separately/individually called via the thread pool.

I'm not terribly concerned about the way it is currently but that's always an option.

pjenvey (approved):

PR #1199, opened 2022-01-03
refactor: remove static service IDs

### LINKED ISSUE #1144: Remove statically-defined service records
The new Tokenserver includes a migration that creates two service records for `sync-1.1` and `sync-1.5` with static database IDs. This is incompatible with the services table that exists on the old Tokenserver. We'll need to make sure the new Tokenserver is compatible with the Tokenserver database as it currently exists, so let's remove this migration and make the requisite changes to the code to allow for service records with arbitrary IDs.

ethowitz:
## Description

Removes the two statically-defined services from the database via a new migration. Makes updates to the application code and integration tests to reflect the change in schema.

## Testing

Passing unit and integration tests should sufficiently show that the change was successful.

## Issue(s)

Closes #1144

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,7 +38,7 @@ commands:
           command: |
             cargo fmt -- --check
             # https://github.com/bodil/sized-chunks/issues/11
-            cargo audit --ignore RUSTSEC-2020-0041 --ignore RUSTSEC-2021-0078 --ignore RUSTSEC-2021-0079 --ignore RUSTSEC-2020-0159 --ignore RUSTSEC-2020-0071 --ignore RUSTSEC-2021-0124
+            cargo audit --ignore RUSTSEC-2020-0041 --ignore RUSTSEC-2021-0078 --ignore RUSTSEC-2021-0079 --ignore RUSTSEC-2020-0159 --ignore RUSTSEC-2020-0071 --ignore RUSTSEC-2021-0124 --ignore RUSTSEC-2021-0131
   python-check:
     steps:
       - run:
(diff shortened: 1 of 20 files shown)


.circleci/config.yml
```diff
@@ -38,7 +38,7 @@ commands:
           command: |
             cargo fmt -- --check
             # https://github.com/bodil/sized-chunks/issues/11
-            cargo audit --ignore RUSTSEC-2020-0041 --ignore RUSTSEC-2021-0078 --ignore RUSTSEC-2021-0079 --ignore RUSTSEC-2020-0159 --ignore RUSTSEC-2020-0071 --ignore RUSTSEC-2021-0124
+            cargo audit --ignore RUSTSEC-2020-0041 --ignore RUSTSEC-2021-0078 --ignore RUSTSEC-2021-0079 --ignore RUSTSEC-2020-0159 --ignore RUSTSEC-2020-0071 --ignore RUSTSEC-2021-0124 --ignore RUSTSEC-2021-0131
```
jrconlin, line 41: Let's put these in a `.cargo/audit.toml` file.

jrconlin (changes requested):
Looks good. We probably want to note that this is a breaking change, since we're technically no longer offering a service that we didn't really offer before.

.circleci/config.yml
ethowitz: Good call!

ethowitz:
@jrconlin Where do you think we should note this? In the code?

jrconlin:
I think adding it to the commit message and making sure that it's rolled up into the Changelog should be good.
Basically, we're logging the point in time we made the change, should someone need to do discovery later.

jrconlin (approved):

PR #1200, opened 2022-01-05
feat: add Tokenserver metrics

### LINKED ISSUE #1108: Tokenserver: Add metrics
Let's add metrics equivalent to those emitted by the legacy Tokenserver. We should also look into adding additional metrics if it would give us enhanced visibility in the health of the service.

ethowitz:
## Description

Emits one log line and several useful metrics per Tokenserver request.

## Testing

I tested this by running the integration tests with `RUST_LOG=debug` and ensuring that the output included the expected log lines. I wasn't sure how to test the metrics.

## Issue(s)

Closes #1108


jrconlin (dismissed):

src/tokenserver/db/models.rs
```diff
@@ -211,6 +214,9 @@ impl TokenserverDb {
         "#;
         const DEFAULT_CAPACITY_RELEASE_RATE: f32 = 0.1;
 
+        let mut metrics = self.metrics.clone();
+        metrics.start_timer("tokenserver.storage.get_best_node", None);
```
pjenvey, line 218: You may have intended this but I just wanted to note that Cadence hardcodes a prefix that we set to the value of `Settings::statsd_label` so this emits "syncstorage.tokenserver.storage.get_best_node".

We could overwrite tokenserver's `statsd_label` to use "tokenserver" instead or look to the future with "syncstorage.tokenserver" or maybe it makes more sense to keep it as you have it 🤷‍♂️

pjenvey (approved):

src/tokenserver/db/models.rs
ethowitz: Ooo I didn't notice this. I think it makes sense to overwrite `statsd_label` to `"syncstorage.tokenserver"` for Tokenserver. I'll make the change

src/server/mod.rs
```diff
@@ -241,7 +245,11 @@ impl Server {
         let tokenserver_state = if settings.tokenserver.enabled {
             Some(tokenserver::ServerState::from_settings(
                 &settings.tokenserver,
-                metrics.clone(),
+                metrics::metrics_from_opts(
```
ethowitz, line 248: Tokenserver will have its own separate metrics client -- is that okay? It seems like it shouldn't be a problem to me, but I wanted to double-check

src/server/metrics.rs
```diff
@@ -167,18 +175,21 @@ pub fn metrics_from_req(req: &HttpRequest) -> Result<Box<StatsdClient>, Error> {
         .clone())
 }
 
-/// Create a cadence StatsdClient from the given options
-pub fn metrics_from_opts(opts: &Settings) -> Result<StatsdClient, ApiError> {
-    let builder = if let Some(statsd_host) = opts.statsd_host.as_ref() {
+pub fn metrics_from_opts(
+    label: String,
+    host: Option<String>,
```
pjenvey, line 180: not terribly important but this could take references instead. `Option::as_deref` can easily give you an `Option<&str>` from the host field `Option<String>`

```suggestion
    label: &str
    host: Option<&str>,
```

src/server/mod.rs
pjenvey: Yea, it's not a big deal for now.

src/server/metrics.rs
ethowitz: Nice catch!

pjenvey (approved):

Issue #766, opened 2020-08-06
0.5.5 SSL issues


pjenvey:
The `features=[ "openssl"]` causes SSL issues, seen on stage logs:

```
2020-08-06T19:42:59.394543836Z Corruption detected. E 
2020-08-06T19:42:59.394603261Z error:14187180:SSL routines:ssl_do_config:bad value E 
2020-08-06T19:42:59.394609172Z     Decryption error: TSI_DATA_CORRUPTED E 
```

One might assume the `target_vendor="ubuntu"` wouldn't affect the `debian:buster-slim` docker env we use, but there's no other reason this would begin happening on 0.5.5.

jrconlin:
Prior to this version, the app would fail during init because of conflicts in openssl and the grpcio included boringssl. 

It's targeted to `ubuntu` because I was unaware of it impacting anything else, but there appears to be a different issue at play.

Does it make sense to just set `feature="openssl"` universally?

jrconlin:
Huh, are we terminating TLS on our machines or on an ELB equivalent? I'm guessing we're doing it on each server and we run a round-robin load balancer?

pjenvey:
> Does it make sense to just set `feature="openssl"` universally?

I doubt that would affect the docker build since I'm guessing the ubuntu target is being triggered. Did we confirm it's being enabled on there from `cargo tree -e features`?

> Huh, are we terminating TLS on our machines or on an ELB equivalent? I'm guessing we're doing it on each server and we run a round-robin load balancer?

We don't terminate TLS, the LB does. The GRPC connections to Spanner are over TLS though.

pjenvey:
JR confirmed the grpcio openssl feature is not being picked up on the Docker build, so it's not the cause.

(also `rustc --print=cfg` shows what the current env's `vendor` is).

All we know is this started happening in 0.5.5 on stage. It first occurred an hour after it was deployed. There's very little difference between 0.5.4 and 0.5.5

https://github.com/mozilla-services/syncstorage-rs/compare/0.5.4...0.5.5

0.5.4 was on stage for a couple days before 0.5.5 (there wasn't a lot of traffic, but neither was there on 0.5.5). I can't easily tell if the libssl-dev install on the docker changed in between the 2 releases (this step was skipped because it hit the docker layer caching) but it shouldn't matter because grpcio's using boringssl underneath anyway.

jrconlin:
[...]

PR #1216, opened 2022-02-04
feat: add BrowserID support for Tokenserver

### LINKED ISSUE #1215: Add BrowserID support to Tokenserver
We've determined that there are enough Sync users still using BrowserID that we should continue support for it. The legacy Tokenserver verifies BrowserID assertions by making a request to FxA – the new Tokenserver should do the same.

**Acceptance Criteria**
* Add a BrowserIdVerifier that implements VerifyToken and makes a request to FxA to verify BrowserID assertions. Reference this class from the legacy Tokenserver: https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/verifiers.py#L112
* In the extractors, check whether the Authorization header includes "Bearer" or "BrowserID" and parse the token accordingly
* Add unit tests and integration tests to ensure that the functionality exactly duplicates that of the legacy Tokenserver

ethowitz:
## Description
* Refactor the `support.rs` file into a new `tokenserver::auth` module
* Add a verifier responsible for verifying BrowserID assertions by making a request to FxA
* Simplify integration testing by removing Tokenserver's "test mode" and instead adding a server to mock FxA responses (this allows us to test the OAuth and BrowserID verifiers more thoroughly since we're not just stubbing them out)
* Add request timeout settings for both the OAuth and BrowserID verifiers
* Adjust a number of the client state values in the unit and integration tests to be more straightforward (the tests now make it more obvious that the `X-Client-State` header contains the hex-encoded client state bytes and the `X-KeyID` header contains the base64-encoded client state bytes
* Add a number of missing doc comments and regular comments to make the logic clearer

## Testing

The new verifier has very comprehensive unit and integration test coverage.

## Issue(s)

Closes #1215


ethowitz:
Not ready for review yet -- I'm going to leave a bunch of my own comments on this PR to make it easier to understand. When that's done, I'll convert it from a draft to a real PR

Makefile
```diff
@@ -36,10 +36,10 @@ docker_stop_spanner:
 	docker-compose -f docker-compose.spanner.yaml down
 
 run:
-	RUST_LOG=debug RUST_BACKTRACE=full cargo run --features tokenserver_test_mode -- --config config/local.toml
+	RUST_LOG=debug RUST_BACKTRACE=full cargo run --features grpcio/openssl -- --config config/local.toml
```
ethowitz, line 39: The build was failing on my machine because of a conflict between openssl and boringssl. This resolves the issue (and I see that there's a comment relating to this in Cargo.toml). Is it worth leaving this here to prevent issues in the future?

docker-compose.mysql.yaml
```diff
@@ -28,6 +28,13 @@ services:
             MYSQL_USER: test
             MYSQL_PASSWORD: test
 
+    mock-fxa-server:
```
ethowitz, line 31: Tokenserver's "test mode" stubbed out the main token verifiers for fake ones, prevent us from being able to test the verifiers themselves with integration tests. I removed the test mode in favor of adding a local server that mocks responses from FxA. This way, we don't need to stub out the verifiers in the Rust code.

src/tokenserver/README.md
```diff
@@ -24,30 +24,12 @@ This functionality was previously provided by a [Python service](https://github.
...
-| `tokenserver.database_url` | `"mysql://root@127.0.0.1/tokenserver_rs"` | database DSN |
-| `tokenserver.database_pool_max_size` | `None` | the maximum number of connections in the database pool |
-| `tokenserver.database_pool_min_idle` | `None` | the minimum number of idle database connections to maintain at all times |
-| `tokenserver.database_pool_connection_timeout` | `Some(30)` | the timeout (in seconds) when waiting for an available connection |
-| `tokenserver.fxa_metrics_hash_secret` | `"secret"` | the secret used to hash users' FxA UIDs |
-| `tokenserver.fxa_email_domain` | `"api.accounts.firefox.com"` | the email domain used to contruct the FxA email address from the user's FxA UID |
-| `tokenserver.fxa_oauth_server_url` | `None` | the URL of the FxA OAuth server to be used to verify user's OAuth tokens |
-| `tokenserver.test_mode_enabled` | `false` | whether to enable Tokenserver's [test mode](#test-mode) |
+You can find example settings for Tokenserver in [config/local.example.toml](../../config/local.example.toml). The available settings are described in doc comments [here](../../src/tokenserver/settings.rs).
```
ethowitz, line 27: I figured it made sense not to have to maintain documentation in two places

src/tokenserver/auth/browserid.rs
```diff
@@ -0,0 +1,469 @@
...
+            })?;
+
+        // If FxA responds with an HTTP status other than 200, report a 503 to the client
+        if response.status() != StatusCode::OK {
+            return Err(TokenserverError::resource_unavailable());
+        }
+
+        // If FxA responds with an invalid response body, report a 503 to the client
+        let response_body = response
+            .json::<VerifyResponse>()
```
ethowitz, line 78: Due to serde/Rust's type system, there is some deviation between how the old and new Tokenserver handles a faulty FxA response. For certain situations (e.g. a `fxa-generation` claim that isn't an integer), the old Tokenserver returns a 401, but for others (e.g. FxA responds with malformed JSON), it returns a 503. I would have had to write a pretty lengthy custom `Deserializer` to handle these cases the same way, and I figured it would be better just to handle all of them as 503 errors anyway, since they _would_ be errors internal to Sync.

src/tokenserver/auth/browserid.rs
```diff
@@ -0,0 +1,469 @@
...
+    audience: String,
+    #[serde(rename(serialize = "trustedIssuers"))]
+    trusted_issuers: [String; 1],
+}
+
+/// The response returned by the FxA BrowserID verifier for a token verification request.
+#[derive(Deserialize, Serialize)]
+#[serde(tag = "status", rename_all = "lowercase")]
+enum VerifyResponse {
+    Okay {
```
ethowitz, line 136: The old Tokenserver looks specifically for a ["status" field set to "okay"](https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/verifiers.py#L161), but the structure of success and error responses differ, making it possible for serde to distinguish between them without having to parse the "status" field at all.

ethowitz, line 136: The old Tokenserver manually checks to see if the `status` field is set to "okay", but I am relying on serde's support for [internally-tagged enums](https://serde.rs/enum-representations.html#internally-tagged) to express this check in the type system. A status other than "okay" or "failure" would result in a 503 error (which differs from the old Tokenserver, which would return a 401), but the [structure of error vs success responses](https://github.com/mozilla/fxa/tree/main/packages/browserid-verifier#error-response) is well-documented, and it seemed reasonable to me to implement it this way. I think it would be considered an internal error if FxA sent us a different status.

src/tokenserver/auth/browserid.rs
```diff
@@ -0,0 +1,469 @@
...
+            // If the fxa-tokenVerified claim is true or missing, return true.
+            Some(Some(true)) | None => true,
+            // If the fxa-tokenVerified claim is false or null, return false.
+            Some(Some(false)) | Some(None) => false,
+        }
+    }
+}
+
+// Approach inspired by: https://github.com/serde-rs/serde/issues/984#issuecomment-314143738
+fn strict_deserialize<'de, T, D>(deserializer: D) -> Result<Option<T>, D::Error>
```
ethowitz, line 207: I use a custom deserializer here to differentiate between missing fields and null fields, since these cases are handled differently.

src/tokenserver/auth/oauth.rs
```diff
@@ -0,0 +1,93 @@
...
+                        .map_err(|_| TokenserverError::invalid_credentials("Unauthorized"))
+                }
+                None => Err(TokenserverError::invalid_credentials("Unauthorized")),
+            }
+        });
+
+        // The PyFxA OAuth client does not offer a way to set a request timeout, so we set one here
+        // by timing out the future if the verification process blocks this thread for longer
+        // than the specified number of seconds.
+        time::timeout(Duration::new(self.timeout, 0), fut)
```
ethowitz, line 88: The old Tokenserver [sets a `timeout` attribute on the OAuth client object](https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/verifiers.py#L198), but I don't see such an attribute referenced in the code: https://github.com/mozilla/PyFxA/blob/main/fxa/oauth.py#L24

src/tokenserver/auth/browserid.rs
```diff
@@ -0,0 +1,469 @@
...
+    }
+}
+
+#[async_trait]
+impl VerifyToken for RemoteVerifier {
+    type Output = VerifyOutput;
+
+    /// Verifies a BrowserID assertion. Returns `VerifyOutput` for valid assertions and a
+    /// `TokenserverError` for invalid assertions.
+    async fn verify(&self, assertion: String) -> Result<VerifyOutput, TokenserverError> {
```
ethowitz, line 51: This is comparable to [this method](https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/verifiers.py#L137) on the old Tokenserver. Note that the Python function throws a number of different exceptions, which are turned into errors reported to the client [here](https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/views.py#L180)

src/tokenserver/db/models.rs
```diff
@@ -1080,7 +1080,7 @@ mod tests {
             service_id,
             email: email1.to_owned(),
             generation: 1,
-            client_state: "616161".to_owned(),
```
ethowitz, line 1083: The unit and integration tests previously set the client state equal to "aaa" (which is not valid hex), even though the client state is stored in the database as hex. I updated the client state values throughout the unit and integration tests to improve the consistency and make things easier to understand.

src/tokenserver/extractors.rs
```diff
@@ -72,62 +69,73 @@ impl TokenserverRequest {
     /// of the FxA server may not have been sending all the expected fields, and
     /// that some clients do not report the `generation` timestamp.
     fn validate(&self) -> Result<(), TokenserverError> {
+        /// Returns true only if both arguments are Some and opta > optb.
+        fn gt(opta: Option<i64>, optb: Option<i64>) -> bool {
```
ethowitz, line 73: Open to suggestions here...I wanted to just compare the options themselves using comparison operators, but the semantics of `Option`'s impl for `PartialOrd` didn't work for my use case: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=2073cf8cbebae1806f073b04c3110dd2

src/tokenserver/handlers.rs
```diff
@@ -63,7 +66,7 @@ pub async fn get_tokenserver_result(
         id: token,
         key: derived_secret,
         uid: updates.uid,
-        api_endpoint: format!("{:}/1.5/{:}", req.user.node, req.user.uid),
+        api_endpoint: format!("{:}/1.5/{:}", req.user.node, updates.uid),
```
ethowitz, line 69: This was a bug :(

ethowitz: But it was caught by the integration tests :)

tools/integration_tests/run.py
```diff
@@ -50,16 +50,22 @@ def start_server():
...
-    os.environ['SYNC_TOKENSERVER__TEST_MODE_ENABLED'] = 'true'
+    mock_server_port = os.environ.get('MOCK_FXA_SERVER_PORT', 6000)
+    os.environ["SYNC_TOKENSERVER__FXA_BROWSERID_SERVER_URL"] = "http://localhost:%s/v2" % mock_server_port
+    os.environ["SYNC_TOKENSERVER__FXA_OAUTH_SERVER_URL"] = "http://localhost:%s" % mock_server_port
     the_server_subprocess = start_server()
     try:
         res = 0
         res |= run_live_functional_tests(TestStorage, sys.argv)
-        res |= run_local_tests()
+        os.environ.setdefault("TOKENSERVER_AUTH_METHOD", "oauth")
```
ethowitz, line 60: We run all the local tests with both BrowserID and OAuth requests to be extra safe here

tools/integration_tests/tokenserver/mock_fxa_server.py
```diff
@@ -0,0 +1,72 @@
+from wsgiref.simple_server import make_server as _make_server
+from pyramid.config import Configurator
+from pyramid.response import Response
+from pyramid.view import view_config
+import json
+import os
+
+
+@view_config(route_name='mock_oauth_verify', renderer='json')
+def _mock_oauth_verify(request):
```
ethowitz, line 10: These mock verify endpoints work the same way as [this mock server](https://github.com/mozilla-services/tokenserver/blob/master/loadtest/mock-oauth-cfn.yml) from the old Tokenserver's load tests. Instead of sending an opaque JWT token, the client sending the request to Tokenserver encodes the response it desires from FxA in the `Authorization` header (e.g. `Authorization: { "body": { .. }, "status": 200 }

Note that this server is stubbing out FxA, so the flow is:
User --> Tokenserver --> FxA

Tokenserver is oblivious to the fact that the client is sending it bogus tokens since it just passes it along to the verification server.

tools/integration_tests/tokenserver/test_support.py
```diff
@@ -60,20 +71,65 @@ def tearDown(self):
 
         self.database.close()
 
-    def _forge_oauth_token(self, generation=None, sub='test', scope='scope'):
+    def _build_oauth_headers(self, generation=None, user='test',
```
ethowitz, line 74: `_build_oauth_headers()` and `_build_browserid_headers()` are responsible for building the bogus auth headers I referenced [here](https://github.com/mozilla-services/syncstorage-rs/pull/1216/files#r799825107)

src/tokenserver/auth/browserid.rs
```diff
@@ -0,0 +1,479 @@
...
+
+/// The verifier used to verify BrowserID assertions.
+#[derive(Clone)]
+pub struct RemoteVerifier {
+    audience: String,
+    issuer: String,
+    fxa_verifier_url: String,
+    // Note that we do not need to use an Arc here, since Reqwest's blocking client uses an Arc
+    // internally
+    request_client: ReqwestClient,
```
ethowitz, line 32: If `request::blocking::Client` didn't use an Arc internally, we'd need to use one here to maintain keep-alive connections between threads

src/tokenserver/auth/oauth.rs
```diff
@@ -0,0 +1,113 @@
...
+    pub fxa_uid: String,
+    pub generation: Option<i64>,
+}
+
+/// The verifier used to verify OAuth tokens.
+#[derive(Clone)]
+pub struct RemoteVerifier {
+    // Note that we do not need to use an Arc here, since Py is already a reference-counted
+    // pointer
+    inner: Py<PyAny>,
```
ethowitz, line 30: If `Py` didn't use an Arc internally, we'd need to use one here to maintain keep-alive connections between threads

ethowitz: https://docs.rs/pyo3/0.15.1/pyo3/struct.Py.html#impl-Clone

src/tokenserver/auth/browserid.rs
```diff
@@ -0,0 +1,486 @@
...
+
+/// The verifier used to verify BrowserID assertions.
+#[derive(Clone)]
+pub struct RemoteVerifier {
+    audience: String,
+    issuer: String,
+    fxa_verifier_url: String,
+    // We need to use an Arc here because reqwest's blocking client doesn't use one internally,
+    // and we want to share keep-alive connections across threads
+    request_client: Arc<ReqwestClient>,
```
ethowitz, line 32: The async request client _does_ use an Arc internally, but I specifically chose to use the blocking one. We don't want these BrowserID requests to FxA happening on Actix's worker threads; instead, I use `web::block` to invoke the requests on threads in the blocking threadpool. If I had used the async client, I would have had to wrap each request invocation with `web::block(futures::executor::block_on(|| { .. }))` instead of simply using `web::block`.

jrconlin:
will continue review tomorrow, but need to reboot and don't want to lose these.

src/tokenserver/auth/browserid.rs
ethowitz: These fields aren't mandatory, and I think older clients may not send them. The legacy Tokenserver is built to handle situations where the fields aren't sent: https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/views.py#L355-L361

tools/integration_tests/tokenserver/test_support.py
ethowitz: That's way better :) My Python skills are rusty

Makefile
pjenvey: Do you have output of this build failure?

I think this ends up disabling boringssl entirely in grpcio. We avoided doing this in the past but I can't track down why (possibly it wasn't even an option when we originally built syncstorage). There's a couple concerns about doing this though:

- we should test it extensively on stage before it rolls out further
- I think it should be the default so all the unit/integration tests are running against this option

src/tokenserver/auth/browserid.rs
pjenvey: > We don't want these BrowserID requests to FxA happening on Actix's worker threads;

Why not? The `verify` method is async so it seems we could use the async `reqwest` client -- `.await`ing it instead of the `block`/`block_on` combo.

src/tokenserver/auth/browserid.rs
```diff
@@ -0,0 +1,486 @@
...
+                        // If we are unable to reach the FxA server, report a 503 to the client
+                        TokenserverError::resource_unavailable()
+                    } else {
+                        // If any other error occurs during the request, report a 401 to the client
+                        TokenserverError::invalid_credentials("Unauthorized")
+                    }
+                })?;
+
+            // If FxA responds with an HTTP status other than 200, report a 503 to the client
+            if response.status() != StatusCode::OK {
```
pjenvey, line 83: nit: You might be able to kill this block entirely by using `error_for_status` along with including `e.is_status()` along w/ the `if e.is_connect` check.

src/tokenserver/auth/browserid.rs
```diff
@@ -0,0 +1,486 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use mockito::{self, Mock};
+
+    #[actix_rt::test]
+    async fn test_browserid_verifier_success() {
+        let body = r#"{
```
pjenvey, line 237: No request for changes here but I'll just note I like using serde's `json!` macro for these things (you would need to `to_string()` its result if you really need the text version). The downside however, as with any macro, is potentially added compile times.

pjenvey (commented):
still working my way through this one, sorry

pjenvey (changes requested):

src/tokenserver/auth/browserid.rs
ethowitz: Yup, that's it. I'll definitely add some more comments to make it more obvious whats going on 👍🏻

tools/integration_tests/tokenserver/mock_fxa_server.py
ethowitz: That exception is from a test that sends it improperly-formatted JSON -- an exception is thrown by `json.loads` and a 5XX error is returned. Do you think we should catch that exception and return a response explicitly?

src/tokenserver/auth/browserid.rs
ethowitz: Ah I think I had myself confused. Since the requests are async, it's okay if they run on the worker threads since they wouldn't be blocking them. Thanks for the catch

src/tokenserver/auth/browserid.rs
```diff
@@ -0,0 +1,486 @@
...
+
+    fn try_from(settings: &Settings) -> Result<Self, Error> {
+        Ok(Self {
+            audience: settings.fxa_browserid_audience.clone(),
+            issuer: settings.fxa_browserid_issuer.clone(),
+            request_client: Arc::new(
+                ReqwestClient::builder()
+                    .timeout(Duration::new(settings.fxa_browserid_request_timeout, 0))
+                    .build()
+                    .map_err(|_| Error::from(()))?,
```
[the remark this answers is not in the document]
ethowitz: Actually after looking a bit more closely, this is just the code that creates the request client on startup -- we're already returning (panicking with) a descriptive error [here](https://github.com/mozilla-services/syncstorage-rs/blob/feat/tokenserver-add-browserid-support/src/tokenserver/mod.rs#L50) if we run into problems. Do you still think we'd benefit from returning something here?

src/tokenserver/auth/browserid.rs
```diff
@@ -0,0 +1,486 @@
...
+#[derive(Deserialize, Serialize)]
+struct IdpClaims {
+    #[serde(rename = "fxa-deviceId")]
+    pub device_id: Option<String>,
+    #[serde(
+        default,
+        rename = "fxa-generation",
+        deserialize_with = "strict_deserialize"
+    )]
+    generation: Option<Option<i64>>,
```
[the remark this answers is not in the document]
ethowitz: Related to [this thread](https://github.com/mozilla-services/syncstorage-rs/pull/1216#discussion_r799818023)

The second `Option` indicates whether the field was present or not in the JSON returned by FxA. I chose to use an `Option` here instead of a `Result` because it's not actually an error if one of these fields is missing in the response

docker-compose.e2e.spanner.yaml
```diff
@@ -17,22 +17,25 @@ services:
...
+          SYNC_TOKENSERVER__FXA_BROWSERID_AUDIENCE: "https://token.stage.mozaws.net/"
+          SYNC_TOKENSERVER__FXA_BROWSERID_ISSUER: "api-accounts.stage.mozaws.net"
           SYNC_TOKENSERVER__FXA_EMAIL_DOMAIN: api-accounts.stage.mozaws.net
           SYNC_TOKENSERVER__FXA_METRICS_HASH_SECRET: secret0
-          SYNC_TOKENSERVER__FXA_OAUTH_SERVER_URL: https://oauth.stage.mozaws.net
           TOKENSERVER_HOST: http://localhost:8000
         entrypoint: >
           /bin/sh -c "
-            sleep 28; pip3 install -r /app/tools/integration_tests/requirements.txt && python3 /app/tools/integration_tests/run.py 'http://localhost:8000#secret0'
+            python3 /app/tools/integration_tests/run.py 'http://localhost:8000#secret0'
```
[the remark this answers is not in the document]
ethowitz: Ah yeah, my removing that was not intentional 😅

src/tokenserver/extractors.rs
ethowitz: I tried my hand at writing my own (very simple) `opt_cmp!` macro to accomplish something similar, but IMO it's a bit more readable. It lets you do something like `opt_cmp!(opta < optb)`, where `opta` and `optb` are both `Option`s. Let me know what you think!

src/tokenserver/auth/browserid.rs
```diff
@@ -65,9 +65,8 @@ impl VerifyToken for RemoteVerifier {
             })
             .send()
             .await
-            .and_then(Response::error_for_status)
             .map_err(|e| {
-                if e.is_connect() || e.is_status() {
+                if e.is_connect() {
```
ethowitz, line 69: I had to revert this back because it wasn't handling 2XX errors other than 200 correctly (the old Tokenserver returns a 503 to the client if FxA returns anything other than a 200, not just anything other than a 2XX)

ethowitz:
@pjenvey @jrconlin Rebased! Ready for another pass whenever you get a chance

Makefile
pjenvey: Finally tracked down the issue I was thinking of: [syncstorage-rs#766](https://github.com/mozilla-services/syncstorage-rs/issues/766) (CC: @jrconlin).

Basically: we enabled openssl narrowed to solely on Ubuntu. Due to a cargo issue I believe it became enabled on more platforms such as Debian used in our Docker, resulting in "Corruption detected" errors on stage.

This [grpc.io issue](https://github.com/grpc/grpc/issues/23144#issuecomment-761025661) suggests it could still be an issue. We could try enabling it again and see if the issue persists. Like I mentioned earlier, we would probably want it on by default in Cargo.toml so all tests run against it.

Another potential alternative we could try is using `rusttls` for reqwest instead.

src/tokenserver/extractors.rs
```diff
@@ -314,39 +330,104 @@ impl FromRequest for TokenData {
...
-
-            web::block(move || oauth_verifier.verify_token(auth.token()))
-                .await
-                .map_err(TokenserverError::from)
-                .map_err(Into::into)
+            metrics.start_timer("tokenserver.token_verification", None);
+
+            match token {
+                Token::BrowserIdAssertion(assertion) => {
+                    let verify_output = state.browserid_verifier.clone().verify(assertion).await?;
```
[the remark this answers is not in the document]
ethowitz: Ah no, that was from a previous iteration when I was passing the verifier into another thread. Thanks for the catch!

ethowitz:
@pjenvey Are we reasonably confident that rusttls is safe? If so, it's probably easiest just to use that

pjenvey:
> @pjenvey Are we reasonably confident that rusttls is safe? If so, it's probably easiest just to use that

In the past foxsec has recommended sticking with more proven options like openssl/necko. I don't recall if they explicitly veto'd rustls usage though, we could ping them about its usage here. Plus it's had [an audit and a significant effort](https://www.memorysafety.org/initiative/rustls/) put into it recently.

(if this is even an option, I noted in slack you had issues trying rustls previously w/ reqwest)

PR #1218, opened 2022-02-09
bug: fix Tokenserver metrics

### LINKED ISSUE #1214: Fix Tokenserver metrics
When `tokenserver.disable_syncstorage` is set to true, there is no primary application state; the only application state that exists is Tokenserver's, which doesn't store the Metrics object. In the Tokenserver extractors, we're using the `Metrics: FromRequest` impl to get access to the metrics, which looks at the syncstorage application state. To resolve this, we'll either want to manually reach into the Tokenserver application state or create a wrapper type around the metrics so we can impl `FromRequest` specifically for Tokenserver's metrics object.

ethowitz:
## Description

Adds a `TokenserverMetrics` wrapper type to allow for a Tokenserver-specific `FromRequest` impl for `Metrics` (the existing impl doesn't work because it looks for the `Metrics` object in the syncstorage `ServerState` instead of the Tokenserver `ServerState`

## Testing

I tested this by running Tokenserver, sending a request, and ensuring that the "metric error: No App State" error message was no longer being logged.

## Issue(s)

Closes #1214

### DIFF
--- a/src/server/metrics.rs
+++ b/src/server/metrics.rs
@@ -23,9 +23,9 @@ pub struct MetricTimer {
 
 #[derive(Debug, Clone)]
 pub struct Metrics {
-    client: Option<StatsdClient>,
-    tags: Option<Tags>,
-    timer: Option<MetricTimer>,
+    pub client: Option<StatsdClient>,
+    pub tags: Option<Tags>,
+    pub timer: Option<MetricTimer>,
 }
 
 impl Drop for Metrics {
--- a/src/server/metrics.rs
+++ b/src/server/metrics.rs
@@ -63,21 +63,27 @@ impl FromRequest for Metrics {
     type Future = Ready<Result<Self, Self::Error>>;
 
     fn from_request(req: &HttpRequest, _: &mut Payload) -> Self::Future {
-        let exts = req.extensions();
-        let def_tags = Tags::from(req.head());
-        let tags = exts.get::<Tags>().unwrap_or(&def_tags);
+        future::ok(metrics_from_request(
+            req,
+            req.app_data::<Data<ServerState>>()
+                .map(|state| state.metrics.clone()),
+        ))
+    }
+}
 
-        future::ok(Metrics {
-            client: match req.app_data::<Data<ServerState>>() {
-                Some(v) => Some(*v.metrics.clone()),
-                None => {
-                    warn!("⚠️ metric error: No App State");
-                    None
-                }
-            },
-            tags: Some(tags.clone()),
-            timer: None,
-        })
+pub fn metrics_from_request(req: &HttpRequest, client: Option<Box<StatsdClient>>) -> Metrics {
+    let exts = req.extensions();
+    let def_tags = Tags::from(req.head());
+    let tags = exts.get::<Tags>().unwrap_or(&def_tags);
+
+    if client.is_none() {
+        warn!("⚠️ metric error: No App State");
+    }
+
+    Metrics {
+        client: client.as_deref().cloned(),
+        tags: Some(tags.clone()),
+        timer: None,
     }
 }
 
--- a/src/tokenserver/extractors.rs
+++ b/src/tokenserver/extractors.rs
@@ -12,7 +12,7 @@ use actix_web::{
     Error, FromRequest, HttpRequest,
 };
 use actix_web_httpauth::extractors::bearer::BearerAuth;
-use futures::future::LocalBoxFuture;
+use futures::future::{self, LocalBoxFuture, Ready};
 use hmac::{Hmac, Mac, NewMac};
 use lazy_static::lazy_static;
 use regex::Regex;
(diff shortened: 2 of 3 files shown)


jrconlin:
Ah, ok, this took me longer than it should have to understand.

If I understand correctly:
The problem is that Actix `ServerState` is "clever" in that it stores things by type. Since we want to have two, distinct metrics systems (one for `syncstorage`, another for `tokenserver`) we can't just store two different instances of `Metrics` to ServerState, because they'd wind up overwriting each other, thus the need to wrap Metrics with `TokenserverMetrics()`. 

It might(?) be a better idea to perhaps create a `Metrics::from_request_sync(&HttpRequest) -> Result<Self, Self::Error>` that does the actual `Metrics` build that `Metrics::from_request(...)` calls, since I can easily see this being a pattern for other apps.

ethowitz:
I believe we can store two different instances of `Metrics` without issue (the Tokenserver `Metrics` instance is just a clone of the syncstorage one), but the problem concerns how we extract the `Metrics` object from the application state. Since Tokenserver and syncstorage share the same `Metric`s type, there could previously only be one `FromRequest` impl shared by Tokenserver and syncstorage for `Metrics`, and that impl specifically extracted the `Metrics` object from syncstorage's `ServerState` and not Tokenserver's. Adding a wrapper type here allowed me to write a second `FromRequest` impl to extract the Tokenserver `Metrics` object. I experimented with adding `ServerState` as a type parameter to differentiate between the Tokenserver and syncstorage `Metrics` objects, but it involved using [`PhantomData`](https://doc.rust-lang.org/std/marker/struct.PhantomData.html), and it seemed like more trouble than it was worth.

I'm not sure I follow your suggestion about adding `Metrics::from_request_sync(&HttpRequest) -> Result<Self, Self::Error>` -- wouldn't we run into the same problem?

jrconlin:
Yeah, I agree about the two instances of Metrics.

My comment about the adding a `from_request_sync(&HttpRequest)` method was more about reducing some of the code duplication between the two. Both require just the `HttpRequest` in order to create themselves. You could wrap both inside of a `from_request(...)` call.

e.g. 
```rust
impl FromRequest for TokenserverMetrics {
  ...
  fn from_request(req: &HttpRequest, _payload: &mut Payload) -> Self::Future {
    match Metrics::from_request_sync(req){
      Ok(v) => future::ok(TokenserverMetrics(v)),
      Err(e) => future::err(e)
    }
  }
}
```

and basically do the same for `FromRequest for Metrics` and move all the actual creation stuff to `from_request_sync(...)`

ethowitz:
Let me know if this is what you had in mind! I couldn't quite create a `from_request_sync` function that just took a request, since the function has to know how to extract the proper app data from the request. I also couldn't just include `ServerState` as a type parameter to resolve this, since the function would need to access the `metrics` attribute on the server state, which it can't do on data of a generic type `T`. I would have had to introduce a trait implemented by both the syncstorage and Tokenserver `ServerState`s, and that seemed a bit heavyweight for what we're trying to accomplish here.

jrconlin:
Huh. 

So, this is what I had come up with just goofing around:
```rust
impl FromRequest for Metrics {
    type Config = ();
    type Error = Error;
    type Future = Ready<Result<Self, Self::Error>>;

    fn from_request(req: &HttpRequest, _: &mut Payload) -> Self::Future {
        match Self::from_request_sync(req) {
            Ok(v) => future::ok(v),
            Err(e) => future::err(e)
        }
    }
}

impl Metrics {
    pub fn from_request_sync(req: &HttpRequest) -> Result<Self, Error> {
        let exts = req.extensions();
        let def_tags = Tags::from(req.head());
        let tags = exts.get::<Tags>().unwrap_or(&def_tags);

        Ok(Metrics {
            client: match req.app_data::<Data<ServerState>>() {
                Some(v) => Some(*v.metrics.clone()),
                None => {
                    warn!("⚠️ metric error: No App State");
                    None
                }
            },
            tags: Some(tags.clone()),
            timer: None,
        })
    }
}
```
which basically extracts the state from the request. It may not be super efficient if you've already got the client, though so 🤷🏻‍♀️

ethowitz:
Yeah so unfortunately this won't work because there are two different `ServerState`s: the one from syncstorage-rs and the one from tokenserver. If we pass the `ServerState` as a type parameter, we get something like this:
```rust
pub fn from_request_sync<T>(req: &HttpRequest) -> Result<Self, Error> {
        let exts = req.extensions();
        let def_tags = Tags::from(req.head());
        let tags = exts.get::<Tags>().unwrap_or(&def_tags);

        Ok(Metrics {
            client: match req.app_data::<Data<T>>() {
                Some(v) => Some(*v.metrics.clone()), // We can't call `v.metrics` here, since v is of type `T`, and we don't know anything about T
                None => {
                    warn!("⚠️ metric error: No App State");
                    None
                }
            },
            tags: Some(tags.clone()),
            timer: None,
        })
    }
```
But it doesn't compile, since we can't actually access the `metrics` attribute on an arbitrary type `T`. The only way to make this work would be to add a trait implemented on both `ServerState` types that includes a `metrics()` method. Then, we could do something like:
```rust
trait OurNewTrait {
    fn metrics(&self) -> Metrics;
}

pub fn from_request_sync(req: &HttpRequest) -> Result<Self, Error>
where T: OurNewTrait
{
        let exts = req.extensions();
        let def_tags = Tags::from(req.head());
        let tags = exts.get::<Tags>().unwrap_or(&def_tags);

        Ok(Metrics {
            client: match req.app_data::<Data<T>>() {
                Some(v) => Some(*v.metrics().clone()),
                None => {
                    warn!("⚠️ metric error: No App State");
                    None
                }
            },
            tags: Some(tags.clone()),
            timer: None,
        })
    }
```
But I think this approach ends up being a bit heavyweight for our use case. What do you think?

jrconlin:
Yeah, no, you're right, that's too heavy. Let's stick with what you had originally and pass the client down as a variable as well.

pjenvey (approved):

PR #1219, opened 2022-02-10
test: Add BrowserId support to Tokenserver load tests

### LINKED ISSUE #1213: Add load test on Tokenserver for BrowserID-based requests
With the addition of BrowserID support, we should update the Tokenserver load tests to include test cases for requests using BrowserID. It might make sense to adjust the proportion of BrowserID to OAuth requests to match that currently found in production.

ethowitz:
## Description

Adds support for BrowserID for the Tokenserver load tests.

## Testing

I ran each of the load tests in isolation against the new Tokenserver to ensure that the correct code path was followed and the appropriate response was generated. I also ran the whole suite of load tests against a local copy of Tokenserver up to ~260 requests per second, and everything seemed to be working as expected.

## Issue(s)

Closes #1213


jrconlin (approved):

tools/tokenserver/loadtests/locustfile.py
```diff
@@ -19,50 +21,79 @@ def __init__(self, *args, **kwargs):
...
         # When a user's password changes, the generation number increases.
         self.generation_counter += 1
         token = self._make_oauth_token(self.email)
 
-        self._do_token_exchange(token)
+        self._do_token_exchange_via_oauth(token)
+
+    @task(100)
+    def test_browserid_success(self):
+        assertion = self._make_browserid_assertion(self.email)
```
pjenvey, line 73: I'm having trouble confirming this in the old Python code but aren't browserid assertions encoded somehow (vs just being a json dumps here)?

Or is this currently intended to solely work with the mock verification service (and would need tweaking to work with say a live stage env)?

ethowitz:
I've updated these tests to use FxA's built-in mock verifier for BrowserID assertions. Just like the load tests from the Python Tokenserver, these will still rely on a mock service to verify OAuth tokens

pjenvey (dismissed):

pjenvey (approved):

PR #1221, opened 2022-02-11
refactor: use the builder pattern for `TokenserverError`s

### LINKED ISSUE #1220: Use builder pattern for `TokenserverError`s
Constructing a `TokenserverError` is arduous in situations where an error needs to be tailored to a very specific situation. In particular, where callers can't make use of the utility methods defined on `TokenserverError`, constructing an error requires knowledge of the `ErrorLocation` type and requires access to the internal fields of the struct. This could be simplified by using the builder pattern, which will make it much easier for callers to tailor errors to their liking in an ergonomic way. It will also make it much easier to create errors that are very _similar_ to the errors returned by the utility methods with small alterations.

**Acceptance Criteria**
* Make `ErrorLocation` private to the `tokenserver::error` module
* Make the fields of `TokenserverError` private
* Add a `TokenserverErrorBuilder` type that allows callers to construct `TokenserverErrors`

ethowitz:
## Description

* Add a `TokenserverErrorBuilder` type to improve the construction of `TokenserverError`s
* Make `ErrorLocation` private to the `tokenserver::error` module
* Make the fields of `TokenserverError` private

## Testing

N/A

## Issue(s)

Closes #1220

### DIFF
--- a/src/tokenserver/auth/browserid.rs
+++ b/src/tokenserver/auth/browserid.rs
@@ -77,7 +77,7 @@ impl VerifyToken for RemoteVerifier {
             })?;
 
         if response.status() != StatusCode::OK {
-            return Err(TokenserverError::resource_unavailable());
+            return Err(ErrorBuilder::resource_unavailable().build());
         }
 
         // If FxA responds with an invalid response body, report a 503 to the client
--- a/src/tokenserver/auth/browserid.rs
+++ b/src/tokenserver/auth/browserid.rs
@@ -4,7 +4,7 @@ use serde::{de::Deserializer, Deserialize, Serialize};
 
 use super::VerifyToken;
 use crate::tokenserver::{
-    error::{ErrorLocation, TokenserverError},
+    error::{TokenserverError, TokenserverErrorBuilder as ErrorBuilder},
     settings::Settings,
 };
 
@@ -69,10 +69,10 @@ impl VerifyToken for RemoteVerifier {
                 if e.is_connect() {
                     // If we are unable to reach the FxA server or if FxA responds with an HTTP
                     // status other than 200, report a 503 to the client
-                    TokenserverError::resource_unavailable()
+                    ErrorBuilder::resource_unavailable().build()
                 } else {
                     // If any other error occurs during the request, report a 401 to the client
-                    TokenserverError::invalid_credentials("Unauthorized")
+                    ErrorBuilder::invalid_credentials("Unauthorized").build()
                 }
             })?;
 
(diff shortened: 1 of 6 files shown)


ethowitz:
This is low priority, but I think it improves the readability of error handling on Tokenserver

jrconlin:
I'm not really sure that this helps vs. using something like `Default`. The content of the error doesn't really need to be private. Nothing really needs to be protected in there since the Error is just going to be passed along and eventually rendered. I can see using a builder for a protected resource like a network handle (where you don't want anything altering behavior radically), or some form of getter/setter where you need to have potential side effects or validation routines to deal with. 

If you want to experiment with builder routines, that's fine 👍🏻. I'm just not sure that they're really worth the effort here.

ethowitz:
Yeah, as I was implementing this, I noticed that it wasn't really going to result in less code necessary to construct an error. There were two primary benefits I saw, though:
* Requiring callers to know of and use `ErrorLocation` makes error construction less ergonomic. Making the contents of  `TokenserverError` private and using `TokenserverErrorBuilder::in_body` et al. means clients don't need to `use` the `ErrorLocation` type (or even know of its existence)
* The majority of error cases fall into one of a relatively small number of repeatedly-used categories of errors, but there are a few cases where just one or two of the error fields are different. If it weren't for these cases, we could make the `TokenserverError` fields private and solely use the predefined constructors (`invalid_generation()`, `invalid_keys_changed_at()`, etc.) to build errors (or we could even just use an enum to represent the different error cases). But given that it's just those few cases that prevent us from doing that, providing the ability to build off of the most commonly-used cases seemed valuable

But ultimately, I'm happy to forego this change if you don't think it's worth it (or if it makes things actively worse 😅). I just wanted to see what it looked like, and I was interested to know what you and Phil thought. Maybe it's still possible to use enums...I'll give it some more thought. It just feels complicated to build `TokenserverError`s the way they're structured now

jrconlin:
Well, I don't think it makes things actively worse. I suppose I'd feel differently if we used a similar pattern for other errors (like `ApiError`), but we don't. Really, the biggest change (aside from remembering to call `.build()` for this specific class of error) is [here](https://github.com/mozilla-services/syncstorage-rs/pull/1221/files#diff-adbbfb1506014ad6d9fa53bff7d4007ca46017a1a3fdeb2d69a6585bb39e7b25R93-R97) where I'm guessing we should probably create an `TokenserverError::invalid_timestamp()` like we do [other errors](https://github.com/mozilla-services/syncstorage-rs/blob/master/src/tokenserver/error.rs#L34). I guess I can see having creator functions for all of these, which set the default values for things like `location` and the `status` string. I guess I just don't see needing to create an `invalid_credentials().in_body()` at some point in the future, or you make these proper builders and you do something like:

```rust
   ErrorBuilder::new("Unauthorized").status("invalid_credentials").in_body().build()
```
for all the errors, but then I'm not sure we're getting anything drastically better than calling the Default pattern.

ethowitz:
Yeah, I see your point. I do think `TokenserverError`s are a bit different than `ApiErrors` in that there isn't a small, enumerable set of reusable errors (the legacy Tokenserver has a number of errors that look similar to one another but aren't exactly the same), but I think I agree that the builder pattern doesn't necessarily add much given that we can do something like:
```rust
TokenserverError {
    status: "different status than the one usually used for an invalid generation error",
    ..TokenserverError::invalid_generation()
}
```
Which means the primary benefit of using the builder pattern would be making `ErrorLocation` private, which on its own, probably isn't a compelling enough reason for this change. I don't feel particularly strongly about this change, so I'll go ahead and close this PR. I appreciate the discussion and your insight 🙂

PR #1222, opened 2022-02-16
feat: Add a spanner prune tool to remove old, unaccessed data

jrconlin:
This adds a python `prune.py` script, which will search for:

 * *"abandoned" records*, these are records that have not been updated in
`SYNC_ABANDON_AGE` (366 days)
 * *no longer accessible `fxa_kid` records*, these are older `fxa_kid`
records for a given `fxa_uid`. Only the most recent `fxa_kid` should be
kept.

Closes: CONSVC-1552


jrconlin:
Aged out.

PR #1224, opened 2022-02-23
feat: add context to `TokenserverError`s

### LINKED ISSUE #1223: Add context to `TokenserverError`s
Many Tokenserver errors don't give detailed information to the client about exactly what occurred. Unfortunately, this means that the error in Sentry also won't contain a lot of detailed information. Let's add additional context to the `TokenserverError`s to make sure we have specific error information in Sentry. This additional information should not be sent to the client.

ethowitz:
## Description

* Add a `context` field to `TokenserverError` to be included in the its `Display` impl but **not** its `Serialize` impl (so the new field doesn't get reported to clients)
* Add context for all of the errors being reported by Tokenserver

## Testing

There should be no change in the errors received by the client -- the new field should only be visible when `Display`ing a `TokenserverError` (e.g. in Sentry). The integration tests verify this.

## Issue(s)

Closes #1223


jrconlin (dismissed):

src/tokenserver/auth/browserid.rs
```diff
@@ -69,22 +69,44 @@ impl VerifyToken for RemoteVerifier {
...
+                        ),
+                        ..TokenserverError::invalid_credentials("Unauthorized")
+                    }
                 }
             })?;
 
         if response.status() != StatusCode::OK {
-            return Err(TokenserverError::resource_unavailable());
+            return Err(TokenserverError {
+                context: "FxA returned a status code other than 200".to_owned(),
```
jrconlin, line 93: Do we want to include the status code?

src/tokenserver/auth/browserid.rs
```diff
@@ -300,7 +330,10 @@ mod tests {
             let error = verifier.verify(assertion.to_owned()).await.unwrap_err();
             mock.assert();
 
-            let expected_error = TokenserverError::resource_unavailable();
+            let expected_error = TokenserverError {
+                context: "FxA returned a status code other than 200".to_owned(),
```
jrconlin, line 334: also, maybe include status code?

jrconlin (dismissed):

pjenvey (approved):

PR #1225, opened 2022-02-23
shell.nix: fix for recent grpc

Mic92:
## Description

see title

## Testing

If you use nix, you can run `nix-shell` and `cargo build`

## Issue(s)

-

PR #1261, opened 2022-03-21
fix: Add recent grpc to nix build

jrconlin:
This is a resubmit of https://github.com/mozilla-services/syncstorage-rs/pull/1225 (offered by @Mic92). The original patch
could not be accepted due to lack of commit signatures.


clouserw:
It's been 4 years, I think we can call this expired at this point.  If we want to do the feature we'd have to start from the beginning to understand the patchset.

PR #1274, opened 2022-04-01
feat: emit 4XX errors as metrics instead of Sentry events

ethowitz:
## Description

Instead of emitting 4XX errors as Sentry events, emit them as metrics. The label is `http_{}`, where `{}` is the HTTP status. Each metric also includes the Weave error code as a tag.

## Testing

I wasn't quite sure how to test this, so I am open to suggestions 🤔


jrconlin (dismissed):

src/web/middleware/mod.rs
```diff
@@ -58,3 +62,42 @@ impl SyncServerRequest for HttpRequest {
         HawkIdentifier::extrude(self, method.as_str(), self.uri(), ci, secrets)
     }
 }
+
+pub fn emit_metric_for_4xx_error(
```
pjenvey, line 66: The sentry middleware emits metrics similarly -- doing so for any error returning `Some` from `metric_label()` (maybe we should call it the `error_reporting` middleware instead).

Many of the `is_client_error` variants already have metric labels so they're already being reported. So I'd argue this isn't needed.

FYI we have a [dashboard panel started](https://earthangel-b40313e5.influxcloud.net/d/5kIPrfbWj/sync-rs-prod?orgId=1&refresh=1m&viewPanel=77) for showing these. It was only showing up a couple of the hawk variants but I've just updated it to show everything underneath "syncstorage.request." via a regex.

pjenvey (changes requested):
> I wasn't quite sure how to test this, so I am open to suggestions 🤔

Testing the metrics output isn't super critical but it's probably worth adding a very basic test. You can copy merino or contile's usage of the cadence `SpyMetricsSink` e.g. [here](https://github.com/mozilla-services/contile/blob/f9649c5/src/server/img_storage.rs#L626) also [here](https://github.com/mozilla-services/contile/blob/25a551b/src/web/test.rs#L695)

src/web/middleware/mod.rs
ethowitz: Gotcha, I'll remove this then 👍🏻

pjenvey (approved):

PR #1281, opened 2022-04-14
refactor: add common crates

### LINKED ISSUE #1275: Create *-common and syncstorage crates
Add the following crates (dependencies are listed as sub-bullets):

* syncstorage-db-common: DbError, Db, DbPool, DbErrorKind
  * syncstorage-common
* tokenserver-common: TokenserverError
* syncstorage-common: Macros and other utilities
* syncstorage: The main crate for the project
  * syncstorage-db-common
  * tokenserver-common
  * syncstorage-common

We will also need to crate a workspace akin to [Merino’s](https://github.com/mozilla-services/merino/blob/main/Cargo.toml). The most difficult part of this work will be removing global dependencies on ApiError and other globally-available data structures: Rust does not allow circular dependencies among crates, so we will need to refactor the code such that a given child on the dependency tree does not rely on any code from its parents. We should also take care to avoid pulling in major dependencies unnecessarily (e.g. only the syncstorage crate should rely on Actix Web).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1743)

ethowitz:
## Description

* Add the following crates:
  - Top-level `syncstorage` crate
  - `syncstorage-db-common`
  - `syncstorage-common`
  - `tokenserver-common`
* Remove the following dependencies from the database code:
    - `ApiError`
    - Actix Web
    - `HawkIdentifier`
* Make `DbError.kind` and `ApiError.kind` private to simplify API

Things I was unsure about:
* I had to remove a lot of the `#[cfg(test)]` annotations, since anything pinned to the `test` feature isn't made public to other crates. I don't think this is a big deal but just wanted to mention it

## Testing

No change in functionality

## Issue(s)

Closes #1275


ethowitz:
Not yet ready for review. I want to remove `syncstorage/src/db`'s dependency on actix if I can

ethowitz:
I made the opinionated decision to remove syncstorage-rs's dependency on Actix's `web::block` in favor of tokio's `task::spawn_blocking`. This prevents us from having to pull Actix in as a dependency for libraries that don't otherwise need it. I'm definitely open to reverting the change if the reviewer thinks it isn't worth it, but I'm trying to separate out the concerns of each of the new crates to reduce the cognitive complexity of how the pieces fit together.

pjenvey:
> * I had to remove a lot of the `#[cfg(test)]` annotations, since anything pinned to the `test` feature isn't made public to other crates. I don't think this is a big deal but just wanted to mention it

FYI at one point the unit tests were in a separate crate (IIRC so they ran on nightly to utilize nightly only async/await 😄 ) and instead of wrapping things in `cfg(test)` we had a [db_test](https://github.com/mozilla-services/syncstorage-rs/blob/0.1.0/db-tests/Cargo.toml#L13) feature they could opt into. I don't think it's worth doing this again though, the unit tests really belong with the syncstorage crate IMO.

> I made the opinionated decision to remove syncstorage-rs's dependency on Actix's `web::block` in favor of tokio's `task::spawn_blocking`.

It looks like actix_web 4.0's `block` was changed from `actix_threadpool` to just be tokio's `spawn_blocking` underneath so we were going to end up with it anyway. So switching over is fine but with one caveat, that `ACTIX_THREADPOOL` env var we adjust in settings will no longer be relevant, so we should remove it.

I can't recall tokio `spawn_blocking`'s typical thread pool size, we should double check what it will end up being for us. If it seems too small we can always just load test first and only worry about it if performance looks degraded.

ethowitz:
Looks like the default upper limit on tokio's blocking threads is 512: https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.max_blocking_threads

I think that should be plenty given that we currently set ACTIX_THREADPOOL to be `max(num_cpus * 5, db_pool_max_size)`

syncstorage/src/db/transaction.rs
```diff
@@ -134,7 +135,8 @@ impl DbTransactionPool {
                         self.collection.clone(),
                         self.bso_opt.clone(),
                     )
-                    .await?;
+                    .await
+                    .map_err(ApiError::from)?;
```
pjenvey, line 139: Looks like this'll continue working:
```suggestion
                    .await?;
```

syncstorage/src/tokenserver/auth/oauth.rs
```diff
@@ -121,6 +137,7 @@ impl VerifyToken for RemoteVerifier {
                 context: "OAuth verification timeout".to_owned(),
                 ..TokenserverError::resource_unavailable()
             })?
+            .map_err(|_| TokenserverError::resource_unavailable())?
             .map_err(Into::into)
```
pjenvey, line 141:
```suggestion
            .map_err(|_| TokenserverError::resource_unavailable())?
```

syncstorage/src/web/extractors.rs
```diff
@@ -112,6 +117,17 @@ impl BatchBsoBody {
     }
 }
 
+impl From<BatchBsoBody> for PostCollectionBso {
```
pjenvey, line 120: Leave this for now but I'll note we had a TODO comment somewhere that dissapeared to kill one of these types. I logged https://github.com/mozilla-services/syncstorage-rs/issues/1296 to revisit it later.

syncstorage/src/web/handlers.rs
```diff
@@ -340,28 +344,19 @@ pub async fn post_collection_batch(
...
-                        }
-                    };
-                    failed.extend(
-                        bso_ids
-                            .clone()
-                            .into_iter()
-                            .map(|id| (id, "db error".to_owned())),
-                    )
-                }
+                Err(e) if e.is_conflict() || e.is_quota() => return Err(e.into()),
```
pjenvey, line 351: 🎉

pjenvey (changes requested):
Just a couple small things.

Thanks for carefully doing this, I really appreciate the git history being preserved.

ethowitz:
Thanks for the review @pjenvey -- this was a long one!

ethowitz:
One thing I just thought of: I changed the database methods to return a `DbError` instead of an `ApiError`, and I think that makes the backtrace that gets reported as part of the final `ApiError` less granular, since the conversion to an `ApiError` happens in the handler, and it's the conversion from `DbError` to `ApiError` that triggers the backtrace collection. Should we add a backtrace to `DbError` to fix this?

syncstorage-db-common/src/lib.rs
```diff
@@ -91,6 +80,23 @@ pub struct PoolState {
     pub idle_connections: u32,
 }
 
+impl From<diesel::r2d2::State> for PoolState {
```
jrconlin, line 83: 🎉

syncstorage-db-common/src/util.rs
```diff
@@ -52,19 +53,14 @@ impl SyncTimestamp {
...
     /// Create a `SyncTimestamp` from an i64
-    ///
-    /// Only called from the db module
-    pub(super) fn from_i64(val: i64) -> Result<Self, DbError> {
+    pub fn from_i64(val: i64) -> Result<Self, DbError> {
         if val < 0 {
-            Err(DbErrorKind::Integrity(
-                "Invalid modified i64 (< 0)".to_owned(),
-            ))?;
+            return Err(DbErrorKind::Integrity("Invalid modified i64 (< 0)".to_owned()).into());
```
jrconlin, line 58: ugh, yeah, the previous version was clever. Thanks!

jrconlin (approved):
Looks good!

I'm a little concerned about some of the test functions potentially "leaking" into non-test, since those tend to act without safety checks, but I'm fine with flagging them softly for later potential cleanup. Since they seem to be fairly isolated, perhaps we create a "test" trait we apply? 🤷🏻‍♂️

syncstorage/src/error.rs
```diff
@@ -322,34 +260,17 @@ where
...
-    ($from:ty, $to:ty, $to_kind:expr) => {
-        impl From<$from> for $to {
-            fn from(inner: $from) -> $to {
-                $to_kind(inner).into()
-            }
+impl From<DbError> for ApiError {
+    fn from(db_error: DbError) -> Self {
+        Self {
+            status: db_error.status,
+            backtrace: db_error.backtrace.clone(),
```
pjenvey, line 269: I won't hold up this PR any longer over this but we could avoid this clone -- maybe adding a `backtrace2` method that defers to the inner `DbError` + making the `ApiError::backtrace` field a private `Option<Backtrace>`.

`thiserror` seems to provide a nice way of automatically implementing this for the `std::Error::backtrace` method but that method's still unstable, unfortunately making many backtrace things like this a bit clunky.

pjenvey (approved):

syncstorage/src/error.rs
ethowitz: I think we could also use [mem::take](https://doc.rust-lang.org/stable/std/mem/fn.take.html) here, but I'm not sure if that's actually much better than clone (i.e. would memory be allocated for the newly-created default value?)

syncstorage/src/db/spanner/models.rs
```diff
@@ -2128,13 +2104,11 @@ impl<'a> Db<'a> for SpannerDb {
         }
     }
 
-    #[cfg(test)]
     fn create_collection(&self, name: String) -> DbFuture<'_, i32> {
```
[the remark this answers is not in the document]
ethowitz: Yeah I think it might be wise to do something with them. Putting them into a separate trait would at least make them slightly harder to use by mistake 🤔 I'll log an issue for this

syncstorage/src/error.rs
ethowitz: The answer appears to be yes, as `Backtrace::default()` just calls `Backtrace::new()`, which just creates a backtrace :(

PR #1286, opened 2022-04-21
feat: don't run Tokenserver migrations on startup

ethowitz:
## Description

Since we're connecting to the pre-existing Tokenserver database, we should not be running the migrations on server startup.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,3 @@ run_spanner:
 
 test:
 	SYNC_DATABASE_URL=$(SYNC_DATABASE_URL) SYNC_TOKENSERVER__DATABASE_URL=$(SYNC_TOKENSERVER__DATABASE_URL) RUST_TEST_THREADS=1 cargo test
-
-tokenserver_migrate:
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -44,10 +44,13 @@ python:
 	venv/bin/python -m pip install -r requirements.txt
 
 run: python
-	PATH=./venv/bin:$(PATH) RUST_LOG=debug RUST_BACKTRACE=full cargo run -- --config config/local.toml
+	PATH="./venv/bin:$(PATH)" RUST_LOG=debug RUST_BACKTRACE=full cargo run -- --config config/local.toml
```
ethowitz, line 47: This fixes an unrelated bug that causes `make run` to fail if an item in your path has any whitespace characters in it.

pjenvey (changes requested):
This needs the e2e tests in CI to run them

ethowitz:
Removed the `tokenserver_migrate` make task since the migrations will run automatically for any non-release build

ethowitz:
The e2e tests run against a release build...I will add a line to run the migrations manually in the docker-compose

pjenvey (approved):

jrconlin (approved):

PR #1287, opened 2022-04-26
feat: adjust ACTIX_THREADPOOL based on Tokenserver max pool size

ethowitz:
## Description

We should be setting `ACTIX_THREADPOOL` according to the max pool sizes of both the Syncstorage and Tokenserver databases. Because Actix Web 4 removes the dependency on actix-threadpool in favor of tokio's `task::spawn_blocking` method, we won't need to do this post-upgrade. Tokio, by default, configures a high number of threads in its blocking threadpool (512), which is higher than the `num_cpus * 5` default set by actix-threadpool.

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -264,20 +264,34 @@ impl Settings {
                 }
 
                 if !s.uses_spanner() {
-                    if let Some(database_pool_max_size) = s.database_pool_max_size {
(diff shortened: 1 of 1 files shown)


ethowitz:
Oh yeah, good call 👍🏻

PR #1294, opened 2022-04-28
bug: fix metrics and BrowserID error context

ethowitz:
## Description

Three small changes:
* We were extracting the Tokenserver metrics reporter incorrectly, which I noticed when we weren't receiving one of the metrics correctly
* The Tokenserver statsd label already has `tokenserver.` in it, so remove that prefix from the places in the code where we emit metrics
* Add context to unrecognized BrowserID verification errors that have a reason message

### DIFF
--- a/src/tokenserver/handlers.rs
+++ b/src/tokenserver/handlers.rs
@@ -41,7 +41,7 @@ pub async fn get_tokenserver_result(
     let (token, derived_secret) = {
         let token_plaintext = get_token_plaintext(&req, &updates)?;
 
-        metrics.start_timer("tokenserver.token_creation", None);
+        metrics.start_timer("token_creation", None);
         // Get the token and secret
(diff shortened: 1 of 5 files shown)


src/tokenserver/handlers.rs
```diff
@@ -35,7 +34,7 @@ pub struct TokenserverResult {
 pub async fn get_tokenserver_result(
     req: TokenserverRequest,
     db: Box<dyn Db>,
-    mut metrics: Metrics,
+    TokenserverMetrics(mut metrics): TokenserverMetrics,
```
ethowitz, line 37: Recently learned you could destructure in function arguments 🦀

jrconlin: TIL. neat!

jrconlin (approved):

PR #1295, opened 2022-04-28
refactor/remove unnecessary syncserverrequest impl

taddes:
## Description

This task entailed removing unnecessary impl blocks for syncserverrequest, as they were not used. The implementation of the TryFrom trait for HawkIdentifier followed by moving it into `extractors.rs` sufficiently tidied up the code in `mod.rs` scoping the functionality to where the structure is defined and various traits implemented for `HawkIdentifier`

In the Makefile, added a quoting to avoid certain paths with space characters breaking the build. @ethowitz was able to clear this issue up and had dealt with it in the past.

## Testing

Ensuring compilation should be sufficient. Ran into some formatting deltas between CircleCI and clippy in initial tests.

## Issue(s)

Closes [link](link).
https://mozilla-hub.atlassian.net/browse/CONSVC-1693

### DIFF
--- a/syncstorage/src/db/transaction.rs
+++ b/syncstorage/src/db/transaction.rs
@@ -241,13 +241,10 @@ impl FromRequest for DbTransactionPool {
                 }
             };
             let method = req.method().clone();
-            let user_id = match HawkIdentifier::try_from(&req) {
-                Ok(v) => v,
-                Err(e) => {
-                    warn!("⚠️ Bad Hawk Id: {:?}", e; "user_agent"=> useragent);
-                    return Err(e);
-                }
-            };
+            let user_id = HawkIdentifier::try_from(&req).map_err(|e| {
+                warn!("⚠️ Bad Hawk Id: {:?}", e; "user_agent"=> useragent);
+                e
+            })?;
             let bso = BsoParam::extrude(req.head(), &mut req.extensions_mut()).ok();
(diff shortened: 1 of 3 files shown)


src/web/extractors.rs
```diff
@@ -1091,6 +1092,26 @@ impl HawkIdentifier {
...
 
+impl TryFrom<&HttpRequest> for HawkIdentifier {
+    type Error = Error;
+
+    fn try_from(req: &HttpRequest) -> Result<HawkIdentifier, Error> {
+        if DOCKER_FLOW_ENDPOINTS.contains(&req.uri().path().to_lowercase().as_str()) {
+            return Ok(HawkIdentifier::cmd_dummy());
+        }
+        let method = req.method().clone();
+        // NOTE: `connection_info()` gets a mutable reference lock on `extensions()`, so
```
jrconlin, line 1103: Looking at what goes on here, the sequence is all blocking code. Is the concern that we don't want the risk of something modifying the `extensions`?

ethowitz (changes requested):
Looks great, @taddes! Just a couple of suggestions 👨🏻‍💻

ethowitz:
I think this needs another rebase after the merging of #1281. Happy to assist if it would be helpful :)

ethowitz (approved):
Looks great @taddes! :shipit:

PR #1306, opened 2022-05-09
refactor: add settings crates

### LINKED ISSUE #1276: Add settings crates
Add the following crates (they should be independent from each other):

* syncstorage-settings: code related to syncstorage settings
* tokenserver-settings: code related to Tokenserver settings

Currently, the Tokenserver settings are nested within the Syncstorage settings. Ideally, we should remove this dependency relationship so that Syncstorage and Tokenserver can be run independently from one-another.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1744)

ethowitz:
## Description

I revamped the naming in this PR to make things clearer. At first, I considered simply adding a "syncstorage-settings" crate with a "tokenserver-settings" crate as a dependency, but I don't think that scheme makes sense right now. Our eventual goal is to roll Tokenserver into Syncstorage so they run as one service, but for right now, Tokenserver and Syncstorage exist separately. If we made the top-level settings crate the Syncstorage settings crate, a bunch of the settings would either need to be `Option`s or they would need to be set but unused in the Tokenserver deployment. I think it's cleaner to have separate, composable settings crates for Tokenserver and Syncstorage, with settings common to the two in a separate crate.

As I implemented this, the name "syncstorage" quickly became overloaded. It was unclear whether "syncstorage" was referring to both Tokenserver and Syncstorage (i.e. the whole server) or just the actual Syncstorage part. Following the naming convention [here](https://github.com/mozilla-services/syncserver), I think it makes sense to rename syncstorage-rs to "syncserver-rs". I've updated the names of the crates to reflect this, but we may want to consider renaming this repository at some point too.

If we do roll Tokenserver into Syncstorage some day, it might make sense to revert the naming to "syncstorage-rs," but for now, I think it would be too confusing not to rename things. We need the flexibility to run Syncstorage and Tokenserver separately (for our production deployment) and together (for self-hosted users), and I think it would be much cleaner and more obvious to think of Tokenserver and Syncstorage as smaller, composable parts of a larger system.

## ❗ Breaking Changes ❗ 
Because of the naming changes in this PR, the names of environment variables and settings in `local.toml` will need to change. Specifically, there are some settings that were previously global in scope that are now scoped specifically to Syncstorage. These settings will need to be prefixed with `SYNC_SYNCSTORAGE__` for env vars and `syncstorage.` for settings in `local.toml`. The Syncstorage-specific settings can be found in `syncstorage-settings/src/lib.rs`.

## Testing

There should be no change in functionality. Successful compilation and runs of the unit and integration tests should be sufficient

## Issue(s)

Closes #1276


jrconlin (dismissed):

ethowitz:
FWIW, I think we can avoid having to change the `SYNC_` prefix to `SYNC__` if we use [`Environment::prefix_separator()`](https://docs.rs/config/latest/config/struct.Environment.html#method.prefix_separator) ([relevant PR](https://github.com/mehcode/config-rs/pull/292)), so this may not depend on upgrading the config crate

ethowitz:
I plan to make substantial updates to the README in a future ticket to reflect the new crate structure and naming scheme.

tokenserver-settings/src/lib.rs
```diff
@@ -74,12 +75,12 @@ pub struct Jwk {
...
     fn default() -> Settings {
         Settings {
-            database_url: "mysql://root@127.0.0.1/tokenserver_rs".to_owned(),
+            database_url: "mysql://root@127.0.0.1/tokenserver".to_owned(),
             database_pool_max_size: None,
             database_pool_min_idle: None,
             database_pool_connection_timeout: Some(30),
             enabled: false,
-            fxa_email_domain: "api.accounts.firefox.com".to_owned(),
+            fxa_email_domain: "api-accounts.stage.mozaws.net".to_owned(),
```
jrconlin, line 83: Is this temporary, or do we always want to point to the stage server as the default?

syncserver-settings/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+                _ => {
+                    error!("Configuration error: Other: {:?}", &e);
+                    e
+                }
+            }
+        })
+    }
+
+    pub fn test_settings() -> Self {
+        let mut settings =
```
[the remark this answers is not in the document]
ethowitz: Good call! I'll add a unit test that covers that case

tokenserver-settings/src/lib.rs
ethowitz: I think it would make sense always to point to stage as the default. That way, in dev environments, people don't mistakenly hit prod if the necessary settings are unset. (The rest of the settings in this file already pointed to stage -- I just updated this one to make it consistent with the others.) Open to other suggestions though!

.circleci/config.yml
```diff
@@ -48,7 +48,7 @@ commands:
       - run:
           name: Rust Clippy
           command: |
-            cargo clippy --all --all-targets --all-features -- -D warnings
+            cargo clippy --workspace --all-targets --all-features -- -D warnings
```
ethowitz, line 51: [`--all` is a deprecated alias for `--workspace'](https://doc.rust-lang.org/cargo/commands/cargo-test.html#package-selection)

jrconlin (approved):

jrconlin (commented):
I see the README.md still refers to building the vendored `mozilla-rust-sdk`. I believe we switched to running the crate version of this?

Otherwise looks good.

ethowitz:
@jrconlin Oh, did I accidentally add a change that references the vendored `mozilla-rust-sdk`? It looks like the current version of the README references it, and I agree that we should update that, though maybe in a separate PR?

ethowitz:
(had to rebase)

jrconlin (approved):

pjenvey (dismissed):

jrconlin (approved):

Issue #1309, opened 2022-05-18
Send dbg! batch logs to stdout


data-sync-user:
We are currently using the `dbg!` macro to emit a log line when we append during a batch operation [here](https://github.com/mozilla-services/syncstorage-rs/blob/master/syncstorage/src/web/handlers.rs#L368). This macro sends log lines to stderr, but given that this isn’t an error, we should be emitting these logs to stdout. Let’s use slog’s trace! or debug! macros instead.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1788)

data-sync-user:
➤ Philip Jenvey commented:

I’ll note that we currently direct MozLog style logs (the default) to stdout, however when disabled (w/ SYNC_HUMAN_LOGS=true) we log to stderr. They should probably go to the same place.

It’s not uncommon for logging systems to output to stderr by default (e.g. slog’s TermDecorator and Python's basicConfig() default there). The old UNIX standard recommends stderr for "error messages and diagnostics" which some log statements fall under. Additionally stderr isn’t usually buffered whereas stdout is, which can be helpful for seeing the logging statements immediately.

We definitely do want to avoid dbg! as everything should be under the fully configurable logging system. dbg! is usually just a trace! that should only be enabled for debugging purposes.

PR #1314, opened 2022-05-24
refactor: replaced dbg! with trace macro

taddes:
## Description

Simple change to replace `dbg!` marco with the slog-rs `trace!` macro, since the `dbg!` was resulting in stderr output, not to stdout.

## Testing
Compiles and matches patterns in rest of `handlers.rs` file.
Had to make alterations within the `.circleci` yaml file. @pjenvey was essential in identifying issue where cargo build pulls in most recent rust version which is often at odds with the specified version in the Docker build, which is explicitly hard-coded. 
Commented out commands to build rust and update which would override specified version in CI/CD. In the future, these lines can be commented back in and a `|` inserted at start of command directive.
Also, it was necessary to add the `no_output_timeout` directive to `cargo build` to allow for longer compilation times.

How should reviewers test?
Ensure GCP logs show  output of trace!("Batch: Appending to {}", &new_batch.id); replacing previous `dbg!` [here](https://github.com/mozilla-services/syncstorage-rs/blob/master/syncstorage/src/web/handlers.rs#L368)

## Issue(s)

Closes [link](https://github.com/mozilla-services/syncstorage-rs/issues/1309).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@
 #
 version: 2.1
 commands:
-  setup-rust:
+  display-rust:
     steps:
       - run:
           name: Display Rust Version
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ commands:
     steps:
       - run:
           name: Setup Rust
-          command: |
+          command: 
             # rustup install stable
             # rustup default stable
             # rustup update
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,9 +11,9 @@ commands:
       - run:
           name: Setup Rust
           command: |
-            rustup install stable
-            rustup default stable
-            rustup update
+            # rustup install stable
+            # rustup default stable
+            # rustup update
             rustc --version
   setup-rust-check:
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,7 +57,7 @@ commands:
       - run:
           name: cargo build
           command: cargo build
-          timeout: 20m
+          no_output_timeout: 20m
   setup-gcp-grpc:
     steps:
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,7 @@ commands:
   setup-rust:
     steps:
       - run:
-          name: Setup Rust
+          name: Display Rust Version
           command: 
             rustc --version
   setup-rust-check:
@@ -54,7 +54,6 @@ commands:
       - run:
           name: cargo build
           command: cargo build
-          no_output_timeout: 20m
   setup-gcp-grpc:
     steps:
       - run:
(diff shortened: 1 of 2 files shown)


taddes:
@pjenvey hoping to discuss this and get some additional context on `slog-rs` and if there is a better pattern. Also need to ensure I can verify the output from slog in GCP.

.circleci/config.yml
```diff
@@ -10,10 +10,10 @@ commands:
     steps:
       - run:
           name: Setup Rust
-          command: |
-            rustup install stable
```
ethowitz, line 14: Were these commands commented out just because there was a temporary mismatch between the Rust version specified in the Dockerfile and the Rust version being pulled in here? If so, we can probably the Dockerfile and uncomment these lines now, right? Looks like there's a 1.61 buster image now: https://hub.docker.com/_/rust

ethowitz: Alternatively, do we need these commands at all if we always want to use the Rust version given to us by the Docker image?

jrconlin: FWIW, Rust Docker images usually are available the day after a rust release. It's been quite some time since there's been a breaking change between versions of rust, or a rust update causing something like `fmt` or `clippy` to fail. I'd be curious why those were commented out.

.circleci/config.yml
```diff
@@ -57,6 +57,7 @@ commands:
       - run:
           name: cargo build
           command: cargo build
+          no_output_timeout: 20m
```
jrconlin, line 60: Isn't cargo build fairly chatty? 

20m of no output seems a bit long. Why are we using this value?

ethowitz: It might be worth investigating ways to cut down on the syncstorage build time as part of the sync modernization work because it is getting quite lengthy

ethowitz: FWIW, it appears the source of this may actually be a regression in Rust 1.61.0 itself. I've opened a PR (https://github.com/mozilla-services/syncstorage-rs/pull/1326) to address this by pinning Rust to 1.60.0. As @pjenvey noted [here](https://mozilla.slack.com/archives/CK6MA5133/p1654120513569029?thread_ts=1654114975.555839&cid=CK6MA5133), we should probably start pinning to a particular Rust version anyway (instead of just grabbing the latest version) to ensure that the version in the Dockerfile and the version in the CircleCI file always match.

taddes: Agreed @ethowitz! Since that PR is now merged, how would you like to proceed with closing this out given the changes in the CircleCI config? Thanks!

ethowitz: You can go ahead and rebase with the master branch to pull the changes in. I think we can remove this line (the one that extends the timeout to 20 minutes). From what I can tell, I think we should also remove the `rustup` commands you commented out: we're using circleci images that have the correct version of rust pre-installed, so I think those commands prob aren't necessary

taddes: Ok, I made the changes to the file, only leaving the output of what Rust version is currently built, though is there any merit to leaving the 20 min timeout? I'm letting it run now to see how long it may take to build again given these changes.

ethowitz (approved):
This looks good to me!

PR #1326, opened 2022-06-02
chore: pin Rust 1.60.0

ethowitz:
## Description

Rust 1.61.0 uses substantially more memory than 1.60.0 in our CI builds, preventing them from completing (and preventing us from merging anything new). This PR pins Rust 1.60.0 until the issue is resolved upstream.

Upstream issue: https://github.com/rust-lang/rust/issues/97549

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,8 +11,8 @@ commands:
       - run:
           name: Setup Rust
           command: |
-            rustup install stable
-            rustup default stable
+            rustup install 1.60.0
+            rustup default 1.60.0
             rustup update
             rustc --version
   setup-rust-check:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -163,7 +163,7 @@ commands:
 jobs:
   checks:
     docker:
-      - image: circleci/rust:latest
+      - image: cimg/rust:1.60.0
         auth:
           username: $DOCKER_USER
           password: $DOCKER_PASS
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -163,7 +163,7 @@ commands:
 jobs:
   checks:
     docker:
-      - image: circleci/rust:latest
+      - image: cimg/rust:1.60.0
```
ethowitz, line 166: Using `cimg` here upgrades us to using CircleCI's new build images (`circleci` uses the legacy build images)

pjenvey (approved):

jrconlin:
Heh, I was experimenting to see if there was a way we could convert the rust version to a Variable, because we pin to it for a bunch of different stuff (circleci, Dockerfile, etc.) We *might* be able to use an environment variable for it, but everything seemed complicated and prone to failure.

taddes:
@jrconlin  Where in such a large collection of crates or in such a project project would env vars typically sit? At the root, or say in the build directory? Is it a concern that said variable could be missed and result in regressions that are hard to find or fix? It does sound like a good idea though

jrconlin:
Yeah, that's the biggest problem. Technically, the circleci config file _should_ be the one that's at the top level since it fires off all the other containers, but it's not super easy to deal with. It _might_ be possible to store them as environment variables, but then you'd have to go into the circleci admin panel to update the values, but that's a pain and prone to forgetting. There might be other methods as well, but they're equally complex and fragile.

PR #1327, opened 2022-06-02
bug: convert `DbError`s to `TokenserverError`s

### LINKED ISSUE #1316: Tokenserver: Convert `DbError`s to `TokenserverError`s before returning to the client
The logs posted in #1313 indicate that the Firefox Tokenserver client is failing to parse the error being returned by Tokenserver ([this is the part of the client causing the error](https://searchfox.org/mozilla-central/source/services/common/tokenserverclient.js#304)) in some situations. I believe this is because the error being returned is a `DbError` and not a `TokenserverError`. We should be converting `DbError`s returned by Tokenserver to `TokenserverError`s before returning so the errors are in the right format. We can enforce this with the type system by changing [this return type](https://github.com/mozilla-services/syncstorage-rs/blob/master/syncstorage/src/tokenserver/handlers.rs#L39) to `Result<HttpResponse, TokenserverError>`, which will require any response from the Tokenserver handler to return a `TokenserverError`.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1796)

ethowitz:
## Description

The Tokenserver handler should always return a `TokenserverError` to ensure that errors are of the format expected by Tokenserver clients. In order to accomplish this, I needed to implement `From<DbError>` for `TokenserverError`. I treat any unhandled `DbError` as an internal error, since [SQLAlchemy errors on the Python Tokenserver](https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/assignment/sqlnode/sql.py#L264) raise `BackendError`s, which return 503s to the client.

## Testing

I've added an integration test to ensure that a `DbError` response has the correct format

Closes #1316


jrconlin (approved):
I like this approach. I wonder how much I can steal for the autopush redo.

ethowitz:
Just updated [this comment](https://github.com/mozilla-services/syncstorage-rs/pull/1327/files#diff-3a406538eae7c752e865db72534ae9ac0bbbd50c2abf71fae444a2cbb69233a0R249) to reflect the fact that we're returning 503 and not 500 for unhandled database errors

tokenserver-common/src/error.rs
```diff
@@ -52,31 +73,31 @@ impl TokenserverError {
         }
     }
 
-    pub fn invalid_key_id(description: &'static str) -> Self {
+    pub fn invalid_key_id(description: String) -> Self {
```
pjenvey, line 76: Alternatively you could take `description: &str` for these to avoid `to_owned()` at the call sites. You'll need to pass `&foo` when you have owned `String`s and it'll add an extra clone, but that's not the common case and IMO the small clone's worth it for convenience (but feel free to disagree and leave as is).

pjenvey (approved):

tokenserver-common/src/error.rs
pjenvey: (and when you *need* to avoid that extra clone you can take an Into<Cow<&str>> but it's rarely worth that effort)

ethowitz: I think I had done it this way to make the clones obvious: in my mind, passing by reference suggests that the function doesn't need ownership, and it may suggest to the caller that the function is less expensive than it really is. But now I'm not so sure 🤔

ethowitz: Perhaps having the function take `Into<String>` is a good compromise? It allows flexibility, but it signals to the caller that the function does in fact need ownership

pjenvey: I'm ok with how it is now 😄  but good point, `Into<String>` is simpler and works similarly and isn't too much trouble to specify as a generic type.

Issue #545, opened 2020-03-25
Convert actix-web middleware to async await


fzzzy:
In our Actix middleware `Service` implementations, we invoke `and_then()` on the `Future` returned by [`Service::call`](https://docs.rs/actix-service/1.0.6/actix_service/trait.Service.html#tymethod.call). Now that Rust supports async/await, this is no longer idiomatic. Instead, we should wrap the `call()` invocation in a `Box::pin(async move { ... })` block as we do in the extractors. This would allow us to use async/await.

pjenvey:
#714 (`wrapfn`) might help in getting rid of the async (at least the transform bits), or at least may help in their refactoring.

ethowitz:
[JIRA](https://mozilla-hub.atlassian.net/browse/CONSVC-1684)

ethowitz:
I could be wrong about this, but since the middleware traits aren't defined with `#[async_trait]`, I don't think we can apply `#[async_trait]` to the middleware implementation, so I'm not sure if this is possible

Issue #544, opened 2020-03-25
Convert actix-web extractors to async await


fzzzy:
In the `FromRequest` implementations for `BsoBodies` and `BsoBody`, we invoke `and_then()` on the `Future` returned by another `from_request()` invocation. Now that Rust supports async/await, this is no longer idiomatic. Instead, we should wrap the `from_request()` invocation in a `Box::pin(async move { ... })` block as we do in the other extractors. This would allow us to use async/await.

Emmanuel-Melon:
@fzzzy, I'd like to work on this issue.

Emmanuel-Melon:
@fzzzy Rust doesn't support Async in traits so we can either utilize
- [Async Trait](https://github.com/dtolnay/async-trait)
- A custom workaround (Still investigating)

fzzzy:
Thanks. I looked at b15cf0d and I don't think we want to use the dyn Future approach. Did you try using Async Trait? That is what we are planning on moving to in the whole codebase eventually.

In the meantime, you can find other places in the codebase where we work around this. We do it by using a non-async function in the impl, but then using an async closure inside that to be able to use async syntax. However, I think this can also be replaced with Async Trait.

Emmanuel-Melon:
> Thanks. I looked at [b15cf0d](https://github.com/mozilla-services/syncstorage-rs/commit/b15cf0df7911118c9a054d0ba7ce191b2791fa75) and I don't think we want to use the dyn Future approach. Did you try using Async Trait? That is what we are planning on moving to in the whole codebase eventually.
> 
> In the meantime, you can find other places in the codebase where we work around this. We do it by using a non-async function in the impl, but then using an async closure inside that to be able to use async syntax. However, I think this can also be replaced with Async Trait.

Yes, I've suggested using async trait in the previous comment and I wanted a confirmation on whether I should introduce new dependencies.

So, can I go back to writing these functions using async trait?

fzzzy:
Yes, please do. Hopefully it works easily. If not, discuss it in the channel.

ethowitz:
I could be wrong about this, but since the `FromRequest` trait isn't defined with `#[async_trait]`, I don't think we can apply `#[async_trait]` to a `FromRequest` implementation, so I'm not sure if this is possible

pjenvey:
AFAIK you're right: it's not possible to use `async-trait` for `FromRequest` impls

ethowitz:
[...]

PR #1338, opened 2022-06-14
refactor: convert actix web middleware to async await

taddes:
## Description

Modify no longer idiomatic uses of `and_then()` in instances of the `call()` function that return a future. While some other methods were worked though, an example in the Actix middleware docs were the prototype for this solution, as it is almost identical. See [https://actix.rs/docs/middleware/](url)

```
    fn call(&self, req: ServiceRequest) -> Self::Future {
        println!("Hi from start. You requested: {}", req.path());

        let fut = self.service.call(req);

        Box::pin(async move {
            let res = fut.await?;

            println!("Hi from response");
            Ok(res)
        })
    }
```

## Testing
Compilation successful, correct future returned. 

## Issue(s)

Closes 545.  [https://github.com/mozilla-services/syncstorage-rs/issues/545](link).


ethowitz:
Hmm I think the issue linked in the description should be #545 and not #544

ethowitz (approved):
Looks great, @taddes!

taddes:
Thank you!

PR #1347, opened 2022-06-30
feat: Return 412 when handling a ':' in offset.

### LINKED ISSUE #1345: Return 412 if encountering a ':' in the `offset`
[https://sentry.io/organizations/mozilla/issues/3286812377/?project=6255069](https://sentry.io/organizations/mozilla/issues/3286812377/?project=6255069&query=is%3Aunresolved)

We’re seeing a number of requests from iOS that contain ':' in the offset query parameter. It is not known what is originating these requests, but it was proposed that returning a 412 may break whatever loop these clients are in.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1868)

jrconlin:
## Description

Return a 412 if encountering a ':' in the offset. This possibly resolves an issue with iOS clients.

(also includes updates for rust 1.46)

## Testing

Unit test included.

## Issue(s)

Closes #1345


pjenvey (dismissed):
This looks good, just those 2 comment changes could be killed. This does seem to be triggered solely by iOS. There's another issue in the same sentry event triggered by desktop Linux that I'll log a separate issue for.

syncstorage/src/web/extractors.rs
```diff
@@ -1187,13 +1185,35 @@ impl From<Offset> for params::Offset {
...
         // issue559: Disable ':' support for now: simply parse as i64 as
         // previously (it was u64 previously but i64's close enough)
+        //
+        // We were seeing a number of errors where iOS clients would return
+        // offsets that contained ':' (e.g. '123456790:123') This is a valid
+        // form that the original python server used, but is not implemented
+        // with this version. Returning a 412 was suggested in order to
+        // break any loop that might exist on the client.
+        if s.contains(':') {
+            return Err(ValidationErrorKind::FromDetails(
```
ethowitz, line 1199: Hmm, is the error returned here being overwritten by the [`map_err` call on line 1288](https://github.com/mozilla-services/syncstorage-rs/blob/54321126d81710d2150f2b7d05b2a1ccb0338e89/syncstorage/src/web/extractors.rs#L1288)? If so, I think that would overwrite the 412 error code with a 400 (I made some changes to the `test_invalid_offset` unit test, and I think I was able to confirm this).

I think we may need to pass through the name and label from the original offset error to the one on line 1288?

ethowitz (changes requested):

pjenvey (changes requested):
It looks like the issue Ethan pointed out (I just unresolved it) still stands.

I think the special `PRECONDITION_FAILED` `ValidationError` is getting eaten in potentially 2 places. The first one's due to the fact that a serde deserializer constructs the `Offset`, so we're not easily able to propagate the error out of there via its `Deserializer Error` (I have to think about how to get around this again, it's been a while..).

Once that's solved, there's another map_err call in `BsoQueryParams::from_request` that would also eat it.

Let's also add some kind of integration test for this case, here's an easy one for the e2e test suite:

```python
    def test_offset_with_colon(self):
        res = self.app.get(
            self.root + "/storage/xxx_col2?limit=10000&offset=123456:123",
            status=412
        )
```

PR #1355, opened 2022-07-06
Refactor/convert actix extractors to async await

taddes:
## Description

In the `FromRequest` implementations for `BsoBodies` and `BsoBody`, we had invoked `and_then()` on the Future returned by another `from_request()` invocation. Now that Rust supports async/await, this is no longer idiomatic. Instead, we wrap the `from_request()` invocation in a Box::pin(async move { ... }) block as we do in the other extractors. This would allows us to use async/await.

Some novel issues were posed in that it was problematic to define the futures within the async blocks. However, the solution was to define the future and then pass it to the `async move` block contained in the `Box::pin`, at which point the Future would be consumed by `.await`. While not immediately obvious, this manages the issue smoothly by passing the future in and avoiding an additional mutable reference that would confuse the `'static` lifetime requirement. There is precedent for this in the Actix docs and examples.

## Testing

Compilation. Any commentary on the implementation is welcome, especially for readability and the new code being idiomatic.

## Issue(s)

Closes 544


syncstorage/src/web/extractors.rs
```diff
@@ -419,49 +419,50 @@ impl FromRequest for BsoBody {
...
+                > max_payload_size
+            {
+                let err: ApiError = ValidationErrorKind::FromDetails(
+                    "payload too large".to_owned(),
+                    RequestErrorLocation::Body,
+                    Some("bso".to_owned()),
+                    label!("request.validate.payload_too_large"),
+                )
+                .into();
+                return Err(err.into());
```
pjenvey, line 456: Like above I think we can simplify these 2 while you're here (but they still need `.into()` because we don't use `?` here:
```suggestion
                return Err(ValidationErrorKind::FromDetails(
                    "payload too large".to_owned(),
                    RequestErrorLocation::Body,
                    Some("bso".to_owned()),
                    label!("request.validate.payload_too_large"),
                ).into());
```

pjenvey (changes requested):
This looks good, I think you can go even further and move the rest of this method into the async block -- it will need the pattern used in e.g. `BsoRequest::from_request`:

```
        let req = req.clone();
        let mut payload = payload.take();
```

The `req.clone()` lets request move into the async closure since it's borrowed. `payload.take()` is similar, but it grabs the request body payload, replacing the one passed in with an empty payload. (I think actix-web's API here wants us to strictly read the request body once then be done with it, since it's basically an async Stream)

syncstorage/src/web/extractors.rs
```diff
@@ -419,49 +419,50 @@ impl FromRequest for BsoBody {
...
-                    .unwrap_or_default()
-                    > max_payload_size
-                {
+        // We declare fut here to avoid moving a mutable ref of payload into the async block.
+        // See examples in actix documentation for this style of implementation.
+        let fut = <Json<BsoBody>>::from_request(req, payload);
+
+        Box::pin(async move {
+            let bso = fut
+                .map_err(|e| {
```
pjenvey, line 428: This is `map_err` on the Future itself -- with it now being `.await`'d it will be more idiomatic (and simpler to reason about) to call `map_err` on the `Result` returned from the `await` itself. i.e. just change the order of calls here.

Also we can simplify this error conversion because we're now using `?` here which does the `.into()` for us. We also have a `impl From<ValidationErrorKind> for Error` so we don't need the extra `.into()` `ApiError` step either.

```
                let bso = fut
                .await
                .map_err(|e| {
                    warn!("⚠️ Could not parse BSO Body: {:?}", e);
                    ValidationErrorKind::FromDetails(
                        e.to_string(),
                        RequestErrorLocation::Body,
                        Some("bso".to_owned()),
                        label!("request.validate.bad_bso_body"),
                    )
                })?;

```

syncstorage/src/web/extractors.rs
taddes: This is an excellent point and solid refactor, thank you for catching that. Will resolve.

syncstorage/src/web/extractors.rs
```diff
@@ -446,20 +446,25 @@ impl FromRequest for BsoBody {
...
-                    ValidationErrorKind::FromValidationErrors(e, RequestErrorLocation::Body, None)
-                        .into();
-                return Err(err.into());
+                return Err(ValidationErrorKind::FromValidationErrors(
+                    e,
+                    RequestErrorLocation::Body,
+                    None,
+                )
+                .into());
+                // let err: ApiError =
```
jrconlin, line 464: drop the commented code?

taddes: Yup @jrconlin , that's a WIP from a recent commit as I'm working on refactoring, had that on my list to remove. Getting my head around the two types of implementations and comparing syntax. Thx!

taddes (commented):
Implemented changes as requested

syncstorage/src/web/extractors.rs
```diff
@@ -374,94 +374,95 @@ impl FromRequest for BsoBody {
...
+        // req.clone() allows move into async block since it is borrowed
+        // payload.take() grabs request body payload, replacing the one passed in
+        // with an empty payload so we strictly read the request body payload once
+        // and dispense with it
+        let req = req.clone();
+        let mut payload = payload.take();
+
+        Box::pin(async move {
+            // Only try and parse the body if its a valid content-type
+            let ctype = match ContentType::parse(&req) {
```
pjenvey, line 386: This could further refactor into `map_err(..)?` and similarly below w/ the `req.app_data` (which could use `ok_or_else` since it's an `Option`)

pjenvey (approved):

PR #1357, opened 2022-07-12
bug: rework Tokenserver load tests for local OAuth verification

ethowitz:
## Description
Now that we're caching FxA's public OAuth JWK on the Rust Tokenserver, OAuth token verification is happening local to Tokenserver: when a JWK is passed in, the PyFxA library verifies tokens locally using PyJWT instead of making requests to FxA's OAuth server. The load tests were previously written under the assumption that tokens were always verified remotely, so this PR adapts them to "forge" valid tokens.

The idea is that we can generate a set of RSA keys, use the public key to be the primary JWK used by Tokenserver for verification, and use the private key to generate OAuth tokens with valid signatures in the load tests.

## Testing
I tested this by following the steps in the load testing README and ensuring that each of the test cases passed successfully.


tools/tokenserver/loadtests/mock-fxa-server/main.py
```diff
@@ -1,20 +0,0 @@
-import functions_framework
```
ethowitz, line 1: We don't need this file anymore now that we don't use a mock OAuth service for load testing

tools/tokenserver/loadtests/locustfile.py
```diff
@@ -60,15 +76,18 @@ def test_oauth_success(self):
 
     @task(100)
     def test_invalid_oauth(self):
-        token = self._make_oauth_token(status=400)
+        token = self._make_oauth_token(
+            self.email,
+            key=INVALID_OAUTH_PRIVATE_KEY
```
ethowitz, line 81: To create an invalid token, we sign the JWT with a private key that doesn't correspond with the public key set on Tokenserver. In this case, we just generate a new private key with every run of the load tests.

ethowitz:
Once someone else agrees that this approach is reasonable, I will open a PR to update the necessary env vars on Tokenserver dev

tools/tokenserver/loadtests/README.md
```diff
@@ -4,37 +4,68 @@ This directory contains everything needed to run the suite of load tests for Tok
...
-```sh
-functions-framework --target mock_fxa_server --debug
-```
-Note that you'll need to install `functions-framework` via `pip3 install functions-framework`. The load tests will use FxA stage to verify BrowserID assertions.
-3. Configure Tokenserver to verify OAuth tokens through the mock FxA service and BrowserID assertions through FxA stage. This is done by setting the following environment variables:
-```
-SYNC_TOKENSERVER__FXA_BROWSERID_AUDIENCE=https://token.stage.mozaws.net
-SYNC_TOKENSERVER__FXA_BROWSERID_ISSUER=api-accounts.stage.mozaws.net
-SYNC_TOKENSERVER__BROWSERID_VERIFIER_URL=https://verifier.stage.mozaws.net/v2
+    ```sh
```
jrconlin, line 7: Super minor nit, but I wonder if we should run the `.md` files through `prettier` or something. There are a lot of formatting warnings about no empty lines around headers and code blocks. Totally fine if we don't.

tools/tokenserver/loadtests/README.md
```diff
@@ -4,37 +4,68 @@ This directory contains everything needed to run the suite of load tests for Tok
...
+    jwk.dumps(public_key, kty="RSA")
+    ```
+    This will give you a key of the form
+    ```json
+    {
+        "n": ...,
+        "e": ...,
+        "kty": "RSA"
+    }
+    ```
```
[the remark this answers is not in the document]
ethowitz: Good call!

tools/tokenserver/loadtests/README.md
ethowitz: Are you thinking we should run `prettier` in CI?

jrconlin: I don't know if we modify the file enough to warrant that, nor do I feel like it's a blocker. We might want to run the files through at least once so that they're consistent.

pjenvey (approved):
Looks good.

As I'm revisiting JWK caching I had a question -- when you ported over the ["hardcoded JWK" feature](https://github.com/mozilla-services/syncstorage-rs/pull/1307) from the python version, did we confirm that tokenserver stage/prod are using it?

Reason I ask is I noticed [PyFxA also does at least some caching of verification requests](https://github.com/mozilla/PyFxA/blob/main/fxa/oauth.py#L268) (bbangert mentioned similar clients typically do some caching). Might this caching be sufficient vs the "hardcoding"?

ethowitz:
@pjenvey Yeah, the values are set for both stage and prod. If desktop and mobile clients cache their own tokens, I'm _guessing_ most (or at least many) of the requests Tokenserver receives are for never-before-seen OAuth tokens from FxA, which IMO would make caching the JWK worth it. If you want, I can look into the expiry time for tokens granted by FxA and then query the Tokenserver logs to see how many repeated requests we get from a given UID within the expiry timeframe. That should tell us (very roughly) the number of hits we get in the PyFxA cache (although each container will have its own cache, and there's no guarantee that a given user will be sent to the same container twice)

PR #1360, opened 2022-07-13
perf: remove unnecessary database calls for Tokenserver

ethowitz:
## Description

A few more changes I wanted to make before we load test:
* Only call `Db::put_user` if there are changes that need to be made to the user record
* Remove an unnecessary call to `Db::get_node_id` when the Spanner node ID is cached
* Cache `spanner_node_id` and `service_id` on `TokenserverPool` and `TokenserverDb` instead of on `ServerState` (this isn't really a performance improvement; just a refactor I thought made sense given the changes here)
* Add metrics for some of the most common database methods

## Testing

These changes should be transparent. The caching of the Spanner node ID already has test coverage, and user updates have integration test coverage.

### DIFF
--- a/syncstorage/src/tokenserver/mod.rs
+++ b/syncstorage/src/tokenserver/mod.rs
@@ -34,9 +34,7 @@ pub struct ServerState {
     pub browserid_verifier: Box<dyn VerifyToken<Output = browserid::VerifyOutput>>,
     pub node_capacity_release_rate: Option<f32>,
     pub node_type: NodeType,
-    pub service_id: Option<i32>,
     pub metrics: Box<StatsdClient>,
-    pub spanner_node_id: Option<i32>,
 }
 
 impl ServerState {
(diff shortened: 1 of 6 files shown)


syncstorage/src/tokenserver/mod.rs
```diff
@@ -52,8 +50,8 @@ impl ServerState {
         let use_test_transactions = false;
 
         TokenserverPool::new(settings, &Metrics::from(&metrics), use_test_transactions)
-            .map(|db_pool| {
-                let service_id = db_pool
+            .map(|mut db_pool| {
+                db_pool.service_id = db_pool
```
pjenvey, line 54: I suppose this could always potentially fail leaving `None` but I doubt it's even worth caring about. Maybe let's add a comment?

pjenvey (commented):
Have you seen any local load test improvements with this PR?

pjenvey (dismissed):

syncstorage/src/tokenserver/mod.rs
ethowitz: Just making sure I understand, you mean a comment to explain the unlikelihood of the query failing?

pjenvey: Yea, exactly

pjenvey: What I meant was it could also happen due to say a temporary network hiccup

ethowitz: Ah, got it

pjenvey (approved):

PR #1370, opened 2022-08-03
feat: use Actix to spawn blocking threads

ethowitz:
## Description

In a previous commit, I refactored syncstorage-rs to use tokio's `task::spawn_blocking` to add blocking tasks to tokio's blocking threadpool. Since the tokio runtime is started up by Actix, we don't have easy control over the number of threads in the blocking threadpool. We should stick with using Actix's [`web::block`](https://docs.rs/actix-web/3.3.3/actix_web/web/fn.block.html) to spawn blocking tasks, since we can control the size of the blocking threadpool using `ACTIX_THREADPOOL`.

This PR switches back to using `web::block`. It also:
* Adds the ability for the server to correctly set `ACTIX_THREADPOOL` based on the sizes of the Syncstorage and Tokenserver database connection pools
* Adds an `additional_blocking_threads_for_fxa_requests` setting to add more blocking threads when the OAuth JWK is not cached on the server. The server will throw an error at startup if this setting is unset when the JWK is not cached to ensure that there are always enough blocking threads in the threadpool to allow for the max number of database connections and requests to FxA
* Refactors `Settings` to set the defaults for `database_pool_max_size` and `tokenserver.database_pool_max_size` in a single place

## Testing

N/A

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ target
 .#*
 service-account.json
 .sentryclirc
+.envrc
 
 config/local.toml
 tools/tokenserver/loadtests/*.pem
--- a/syncstorage/src/db/mod.rs
+++ b/syncstorage/src/db/mod.rs
@@ -9,13 +9,13 @@ pub mod transaction;
 
 use std::time::Duration;
 
+use actix_web::{error::BlockingError, web};
 use cadence::{Gauged, StatsdClient};
-use futures::TryFutureExt;
 use syncstorage_db_common::{
     error::{DbError, DbErrorKind},
     results, DbPool, GetPoolState, PoolState,
 };
-use tokio::{self, task, time};
+use tokio::{self, time};
 use url::Url;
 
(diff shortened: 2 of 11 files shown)


ethowitz:
Actually, I can't tell if there's a real difference between using `task::spawn_blocking` and `web::block` (i.e. are Actix's and Tokio's blocking threadpools the same?). Either way, I think it's preferable to use `web::block` (since that's the function exposed by Actix) until we upgrade to Actix 4.

pjenvey (approved):

ethowitz:
@pjenvey I added the ability to manually set a value for `ACTIX_THREADPOOL`, which will be helpful for Tokenserver load testing

pjenvey (dismissed):

ethowitz:
Sorry for the noise, had to make a bug fix 🙂

pjenvey (approved):

PR #1374, opened 2022-08-15
refactor: convert middleware to `wrap_fn` paradigm

### LINKED ISSUE #714: Implement middleware with `wrap_fn`
[`wrap_fn`] is a lightweight way to create Actix middleware. We might want to use it to reduce boilerplate in the middlewares that aren't converted into extractors.

```rust
App::new()
    .wrap_fn(|request, service| {
        // Add a `text/plain` content type header to every response
        let fut = service.call(request);
        async {
            let mut response = fut.await?;
            response.headers_mut().insert(
               CONTENT_TYPE, HeaderValue::from_static("text/plain"),
            );
            Ok(response)
        }
    })
```

[`wrap_fn`]: https://docs.rs/actix-web/2.0.0/actix_web/struct.App.html#method.wrap_fn

This should be completed before #1249, since the middleware APIs changed in that version (there’s no point in having our middleware conform to the new APIs only to switch to using wrap_fn in the end.

taddes:
## Description

Implemented middleware with `wrap_fn()` to remove significant boilerplate. Moved bodies of `Service::call()` methods into the function body of `wrap_fn` and registered it in `mod.rs`, which passes in the service and request. Each middleware contains its own logic scoped to the individual module.

Also made some minor refactors due to clippy warnings for `Eq` and the [manual_range_contains warning ](https://rust-lang.github.io/rust-clippy/master/index.html#manual_range_contains)which I had not yet seen. Hopefully it is relatively clear given the range check.

## Testing

Compilation, clarity. 

## Issue(s)

Closes #714

### DIFF
--- a/syncserver/src/main.rs
+++ b/syncserver/src/main.rs
@@ -32,7 +32,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
     let settings = Settings::with_env_and_config_file(args.flag_config.as_deref())?;
     init_logging(!settings.human_logs).expect("Logging failed to initialize");
     debug!("Starting up...");
-    // Set DSN environment variable to enable Sentry.
+    // Set SENTRY_DSN environment variable to enable Sentry.
     // Avoid its default reqwest transport for now due to issues w/
     // likely grpcio's boringssl
     let curl_transport_factory = |options: &sentry::ClientOptions| {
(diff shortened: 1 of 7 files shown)


ethowitz (changes requested):
Looks good so far! Just a few small suggestions. This branch will need a rebase, since I made some changes to the Sentry middleware last week; I'll take a closer look at the Sentry middleware changes after the rebase is pushed up

ethowitz:
In the next syncstorage and tokenserver release, we'll want to verify on stage that these middlewares are working as expected before we promote to production

jrconlin (approved):
much nicer without the boilerplate stuff.

ethowitz (approved):

PR #1375, opened 2022-08-17
feat: add `__error__` endpoint to Tokenserver

### LINKED ISSUE #1364: Add an /__error__ endpoint to Rust tokenserver
`Server::tokenserver_only_with_settings` doesn't setup an `/__error__` endpoint like syncstorage does. It should add one as these are used to verify Sentry integration by generating an Error/Sentry event.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1921)

ethowitz:
## Description

* Add the `/__error__` Dockerflow endpoint to Tokenserver
* Report `TokenserverError`s to Sentry

## Testing

The endpoint itself can be tested by running the server and sending `GET /__error__` to ensure that an error is correctly generated. I wasn't sure how to test integration with Sentry.

## Issue(s)

Closes #1364

### DIFF
--- a/syncstorage/src/tokenserver/handlers.rs
+++ b/syncstorage/src/tokenserver/handlers.rs
@@ -220,3 +220,13 @@ pub async fn heartbeat(db: Box<dyn Db>) -> Result<HttpResponse, Error> {
         }
     }
 }
+
+/// Generates an error to test the Sentry integration
+pub async fn test_error() -> Result<HttpResponse, TokenserverError> {
+    error!("Test Error");
+    Err(TokenserverError {
+        context: "Test error for Sentry".to_owned(),
+        description: "Test error for Sentry".to_owned(),
+        ..TokenserverError::internal_error()
+    })
+}
--- a/syncstorage/src/web/middleware/sentry.rs
+++ b/syncstorage/src/web/middleware/sentry.rs
@@ -144,17 +144,23 @@ where
                                 metrics.incr(&label);
                             }
                         }
-                        if !apie.is_reportable() {
+
+                        if apie.is_reportable() {
+                            report(&tags, event_from_error(apie));
+                        } else {
                             trace!("Sentry: Not reporting error: {:?}", apie);
-                            return Ok(sresp);
                         }
-                        report(&tags, event_from_error(apie));
                     } else if let Some(tokenserver_error) = e.as_error::<TokenserverError>() {
-                        if tokenserver_error.http_status.is_server_error() {
+                        if let Some(metrics) = metrics {
+                            if let Some(label) = tokenserver_error.metric_label() {
+                                metrics.incr(&label);
+                            }
+                        }
+
+                        if tokenserver_error.is_reportable() {
                             report(&tags, event_from_error(tokenserver_error));
                         } else {
                             trace!("Sentry: Not reporting error: {:?}", tokenserver_error);
-                            return Ok(sresp);
                         }
                     }
                 }
--- a/syncstorage/src/web/middleware/sentry.rs
+++ b/syncstorage/src/web/middleware/sentry.rs
@@ -160,7 +160,7 @@ where
         }
     }
 
-    if err.is_reportable() {
+    if err.is_sentry_event() {
         report(tags, event_from_error(err));
     } else {
         trace!("Sentry: Not reporting error: {:?}", err);
@@ -171,7 +171,7 @@ where
 ///
 /// `sentry::event_from_error` can't access `std::Error` backtraces as its
 /// `backtrace()` method is currently Rust nightly only. This function works
-/// against `HandlerError` instead to access its backtrace.
+/// against `ReportableError` instead to access its backtrace.
 pub fn event_from_error<E>(err: &E) -> Event<'static>
 where
     E: ReportableError + StdError + 'static,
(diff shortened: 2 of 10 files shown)


syncstorage/src/web/middleware/sentry.rs
```diff
@@ -161,12 +169,15 @@ where
...
+pub fn event_from_error<E>(err: &E) -> Event<'static>
+where
+    E: ErrorBacktrace + StdError + 'static,
+{
     let mut exceptions = vec![exception_from_error_with_backtrace(err)];
 
     let mut source = err.source();
     while let Some(err) = source {
-        let exception = if let Some(err) = err.downcast_ref() {
+        let exception = if let Some(err) = err.downcast_ref::<E>() {
```
ethowitz, line 180: If I understand correctly, this is drilling down to the original source of the error by repeatedly calling `Error::source`. The type parameter was previously omitted, but before this change, I _believe_ that `err.downcast_ref()` was being inferred to be `err.downcast_ref::<ApiError>()`.

I just want to make sure that explicitly passing `E` as the type param here doesn't change the functionality at all

syncstorage/src/tokenserver/handlers.rs
```diff
@@ -220,3 +220,13 @@ pub async fn heartbeat(db: Box<dyn Db>) -> Result<HttpResponse, Error> {
         }
     }
 }
+
+/// Generates an error to test the Sentry integration
+pub async fn test_error() -> Result<HttpResponse, TokenserverError> {
```
ethowitz, line 225: From what I could tell, [these two lines](https://github.com/mozilla-services/syncstorage-rs/blob/master/syncstorage/src/web/handlers.rs#L704-L705) weren't necessary to generate a test error, but I could definitely be wrong here

syncstorage/src/web/middleware/sentry.rs
```diff
@@ -148,6 +149,13 @@ where
                             return Ok(sresp);
                         }
                         report(&tags, event_from_error(apie));
+                    } else if let Some(tokenserver_error) = e.as_error::<TokenserverError>() {
```
jrconlin, line 152: Do we want to do the same sort of logic (optionally checking to see if this is a "metric" error) for Tokenserver Errors?
(I guess we generate less non-actionable errors in TokenServer)
Also wondering if we should have the same basic pattern for both error types (have `is_reportable()`)

ethowitz: Actually, I do think it'd be super useful to report BrowserID vs OAuth 4XX errors as metrics. I've added that, following the same `is_reportable()` type of pattern we are using for `ApiError`

jrconlin: cool. We're duplicating the lines 148-151. I wonder if we can break those out so that we're not repeating things?

ethowitz: Done!

jrconlin (approved):
I'm sure there's someone out there that would say "Use a `match` for the conditional!"

I'm not that someone. 

This is readable and works, ship it.

tokenserver-common/src/error.rs
```diff
@@ -262,3 +273,24 @@ impl From<TokenserverError> for HttpResponse {
         ResponseError::error_response(&inner)
     }
 }
+
+impl ReportableError for TokenserverError {
+    fn error_backtrace(&self) -> String {
+        format!("{:#?}", self.backtrace)
+    }
+
+    fn is_reportable(&self) -> bool {
```
pjenvey, line 282: Can you rename this method while you're here? We've been meaning to change it everywhere (already done in contile) to make it clearer:

```suggestion
    fn is_sentry_event(&self) -> bool {
```

I think `ReportableError` still works as the trait name (`is_reportable` wasn't clear as to where it would be reported to, sentry or metrics, but the trait handles both)

syncstorage/src/web/middleware/sentry.rs
```diff
@@ -156,17 +150,37 @@ where
...
+    } else {
+        trace!("Sentry: Not reporting error: {:?}", err);
+    }
+}
+
 /// Custom `sentry::event_from_error` for `ApiError`
 ///
 /// `sentry::event_from_error` can't access `std::Error` backtraces as its
 /// `backtrace()` method is currently Rust nightly only. This function works
 /// against `HandlerError` instead to access its backtrace.
```
pjenvey, line 174:
```suggestion
/// against `ReportableError` instead to access its backtrace.
```

syncstorage/src/web/middleware/sentry.rs
```diff
@@ -186,11 +200,12 @@ pub fn event_from_error(err: &ApiError) -> Event<'static> {
 /// Custom `exception_from_error` support function for `ApiError`
```
pjenvey, line 200:
```suggestion
/// Custom `exception_from_error` support function for `ReportableError`
```

syncstorage/src/web/middleware/sentry.rs
pjenvey: That's right, so now it's `Reportable + etc` so it should work the same AFAICT.

pjenvey (changes requested):
a couple more things

syncstorage/src/web/middleware/sentry.rs
```diff
@@ -186,11 +200,12 @@ pub fn event_from_error(err: &ApiError) -> Event<'static> {
...
 ///
 /// Based moreso on sentry_failure's `exception_from_single_fail`.
-fn exception_from_error_with_backtrace(err: &ApiError) -> sentry::protocol::Exception {
+fn exception_from_error_with_backtrace<E>(err: &E) -> sentry::protocol::Exception
+where
+    E: ReportableError + StdError + ?Sized,
```
pjenvey, line 205: Looks like this `?Sized` isn't needed
```suggestion
    E: ReportableError + StdError,
```

tokenserver-common/src/error.rs
ethowitz: Can do 👍🏻

pjenvey (approved):

PR #1376, opened 2022-08-17
feat: update Tokenserver load tests to test user allocation

ethowitz:
## Description
Locust works by creating an instance of the `TokenserverTestUser` class for each user you specify when you run the load tests. For example, if you wanted to generate 4000 requests/second, you could instantiate 4000 users, where each user sends ~1 request/second. The problem with this approach is that there is a significant difference in the cost of handling a request from a previously-seen UID and a newly-seen UID (namely, Tokenserver needs to allocate a new user when a UID has not been seen before).

Previously, I was using the object ID of each instance of the `TokenserverTestUser` as the FxA UID, but this meant that, once we reached the peak number of Locust "users," no new users were being allocated. This doesn't accurately represent production load, since real load will include new user allocation. To address this, I've added a test case that periodically rotates the `fxa_uid` for each instance of the `TokenserverTestUser` class.

## Testing
This can be tested by commenting out all the load tests except for the `change_user_id` test case, and ensuring that the `fxa_uid` is rotated with each run of the test case by adding `print(self.fxa_uid, self.email)` to the end of the `change_user_id` function.

### DIFF
--- a/tools/tokenserver/loadtests/locustfile.py
+++ b/tools/tokenserver/loadtests/locustfile.py
@@ -1,5 +1,6 @@
 from base64 import urlsafe_b64encode as b64encode
 import binascii
+import hashlib
 import jwt
 import os
 import time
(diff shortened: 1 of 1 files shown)


jrconlin (approved):

ethowitz:
I realized I forgot to cast the original object ID to a string (🤦🏻). I also made use of the [constant_throughput](https://docs.locust.io/en/stable/api.html#locust.wait_time.constant_throughput) function, which I think is a better choice for our use case, since it makes it easier to generate a consistent amount of traffic

jrconlin (commented):
darn you weak typing. ;)

ethowitz:
I looked at the number of users we're actually allocating daily, and I've determined that the amount of traffic this generates is negligible compared to the number of other requests we get

PR #1397, opened 2022-09-09
bug: fix Tokenserver generation and keys_changed_at handling

ethowitz:
## Description

There were a few missing cases relating to how we handle generation and keys_changed_at:
* When Tokenserver gets an updated keys_changed_at in a request but that request has no generation and the generation on the user record is less than the new keys_changed_at, it should set the generation equal to the updated keys_changed_at
  * Rust Tokenserver was setting the generation to be the keys_changed_at in the request even when the request didn't contain an updated keys_changed_at
* Once Tokenserver sees a client state and keys_changed_at for a user, it requires both of them in every request going forward (this is only relevant for BrowserID requests, since all OAuth requests must contain both keys_changed_at and client state in the form of an `X-KeyID` header
  * Rust Tokenserver was allowing clients to revert to omitting these fields even after it had already seen them. Prior to adding support for BrowserID, this case didn't need to be handled since every request included both keys_changed_at and client state (and the server errored out if they weren't included). When we introduced BrowserID support, we allowed BrowserID clients to omit `X-Client-State` and `keys_changed_at` but didn't make sure we were enforcing that requests contain them if we had already seen these values for a given user.
* This is minor and may not be something we need to handle, but: The database schema allows users.keys_changed_at to be null. When the Python Tokenserver loads the value into memory, it converts null values to 0, which means that null and 0 values from the database are both treated as 0 when running the consistency checks. This has the implication that clients can `technically` have a keys_changed_at of 0 in the database and then revert to not sending a keys_changed_at. In reality, I'm sure why any user would have a keys_changed_at of 0 in the database, but for completeness's sake, I added support to handle this the same way in the Rust Tokenserver

## Testing

I added integration tests to handle the missing cases.

### DIFF
--- a/syncstorage/src/settings.rs
+++ b/syncstorage/src/settings.rs
@@ -289,6 +289,7 @@ impl Settings {
         s.set_default("tokenserver.node_type", "spanner")?;
         s.set_default("tokenserver.statsd_label", "syncstorage.tokenserver")?;
         s.set_default("tokenserver.run_migrations", cfg!(test))?;
+        s.set_default("tokenserver.token_duration", 3600)?;
 
         // Set Cors defaults
         s.set_default(
--- a/syncstorage/src/tokenserver/db/models.rs
+++ b/syncstorage/src/tokenserver/db/models.rs
@@ -164,7 +164,7 @@ impl TokenserverDb {
              WHERE service = ?
                AND email = ?
                AND generation <= ?
-               AND COALESCE(keys_changed_at, 0) <= COALESCE(?, 0)
+               AND COALESCE(keys_changed_at, 0) <= COALESCE(?, keys_changed_at, 0)
                AND replaced_at IS NULL
         "#;
 
--- a/syncstorage/src/tokenserver/extractors.rs
+++ b/syncstorage/src/tokenserver/extractors.rs
@@ -440,6 +440,15 @@ impl FromRequest for AuthData {
 
             let TokenserverMetrics(mut metrics) = TokenserverMetrics::extract(&req).await?;
 
+            // The Python Tokenserver treats zero values and null values both as being
+            // null, so for consistency, we need to convert a `Some(0)` value to `None`
+            fn convert_zero_to_none(generation_or_keys_changed_at: Option<i64>) -> Option<i64> {
+                match generation_or_keys_changed_at {
+                    Some(0) => None,
+                    _ => generation_or_keys_changed_at,
+                }
+            }
+
             match token {
                 Token::BrowserIdAssertion(assertion) => {
                     let mut tags = Tags::default();
@@ -463,8 +472,8 @@ impl FromRequest for AuthData {
                         device_id: verify_output.device_id,
                         email: verify_output.email.clone(),
                         fxa_uid: fxa_uid.to_owned(),
-                        generation: verify_output.generation,
-                        keys_changed_at: verify_output.keys_changed_at,
+                        generation: convert_zero_to_none(verify_output.generation),
+                        keys_changed_at: convert_zero_to_none(verify_output.keys_changed_at),
                     })
                 }
                 Token::OAuthToken(token) => {
@@ -485,8 +494,8 @@ impl FromRequest for AuthData {
                         email,
                         device_id: None,
                         fxa_uid,
-                        generation: verify_output.generation,
-                        keys_changed_at: Some(key_id.keys_changed_at),
+                        generation: convert_zero_to_none(verify_output.generation),
+                        keys_changed_at: convert_zero_to_none(Some(key_id.keys_changed_at)),
                     })
                 }
             }
--- a/syncstorage/src/tokenserver/extractors.rs
+++ b/syncstorage/src/tokenserver/extractors.rs
@@ -31,7 +31,6 @@ lazy_static! {
     static ref CLIENT_STATE_REGEX: Regex = Regex::new("^[a-zA-Z0-9._-]{1,32}$").unwrap();
 }
 
-const DEFAULT_TOKEN_DURATION: u64 = 5 * 60;
 const SYNC_SERVICE_NAME: &str = "sync-1.5";
 
 /// Information from the request needed to process a Tokenserver request.
@@ -108,6 +107,14 @@ impl TokenserverRequest {
             return Err(TokenserverError::invalid_client_state(error_message));
         }
 
+        // If the client previously reported a client state, every subsequent request must include
+        // one. Note that this is only relevant for BrowserID requests, since OAuth requests must
+        // always include a client state.
+        if !self.user.client_state.is_empty() && self.auth_data.client_state.is_empty() {
+            let error_message = "Unacceptable client-state value empty string".to_owned();
+            return Err(TokenserverError::invalid_client_state(error_message));
+        }
+
         // If the client state on the request differs from the most recently-used client state, it must
         // be accompanied by a valid change in generation (if the client reports a generation).
         if self.auth_data.client_state != self.user.client_state
@@ -143,6 +150,26 @@ impl TokenserverRequest {
         if opt_cmp!(user_keys_changed_at > auth_keys_changed_at) {
             return Err(TokenserverError {
                 context: "New keys_changed_at less than previously-seen keys_changed_at".to_owned(),
+                ..TokenserverError::invalid_keys_changed_at()
+            });
+        }
+
+        // If there's no keys_changed_at on the request, there must be no value stored on the user
+        // record. Note that this is only relevant for BrowserID requests, since OAuth requests
+        // must always include a keys_changed_at header. The Python Tokenserver converts a NULL
+        // keys_changed_at on the user record to 0 in memory, which means that NULL
+        // keys_changed_ats are treated equivalently to 0 keys_changed_ats. This would allow users
+        // with a 0 keys_changed_at on their user record to hold off on sending a keys_changed_at
+        // in requests even though the value in the database is non-NULL. To be thorough, we
+        // handle this case here.
+        if auth_keys_changed_at.is_none()
+            && matches!(user_keys_changed_at, Some(inner) if inner != 0)
+        {
+            let context =
+                "No keys_changed_at sent for a user for whom we've already seen a keys_changed_at"
+                    .to_owned();
+            return Err(TokenserverError {
+                context,
                 ..TokenserverError::invalid_keys_changed_at()
             });
         }
@@ -252,7 +279,7 @@ impl FromRequest for TokenserverRequest {
                     match duration_string.parse::<u64>() {
                         // The specified token duration should never be greater than the default
                         // token duration set on the server.
-                        Ok(duration) if duration <= DEFAULT_TOKEN_DURATION => Some(duration),
+                        Ok(duration) if duration <= state.token_duration => Some(duration),
                         _ => None,
                     }
                 })
@@ -265,7 +292,7 @@ impl FromRequest for TokenserverRequest {
                 hashed_fxa_uid,
                 hashed_device_id,
                 service_id,
-                duration: duration.unwrap_or(DEFAULT_TOKEN_DURATION),
+                duration: duration.unwrap_or(state.token_duration),
                 node_type: state.node_type,
             };
 
@@ -683,6 +710,8 @@ mod tests {
         static ref SERVER_LIMITS: Arc<ServerLimits> = Arc::new(ServerLimits::default());
     }
 
+    const TOKEN_DURATION: u64 = 3600;
+
     #[actix_rt::test]
     async fn test_valid_tokenserver_request() {
         let fxa_uid = "test123";
@@ -1064,7 +1093,7 @@ mod tests {
             hashed_fxa_uid: "abcdef".to_owned(),
             hashed_device_id: "abcdef".to_owned(),
             service_id: 1,
-            duration: DEFAULT_TOKEN_DURATION,
+            duration: TOKEN_DURATION,
             node_type: NodeType::default(),
         };
 
@@ -1107,7 +1136,7 @@ mod tests {
             hashed_fxa_uid: "abcdef".to_owned(),
             hashed_device_id: "abcdef".to_owned(),
             service_id: 1,
-            duration: DEFAULT_TOKEN_DURATION,
+            duration: TOKEN_DURATION,
             node_type: NodeType::default(),
         };
 
@@ -1149,7 +1178,7 @@ mod tests {
             hashed_fxa_uid: "abcdef".to_owned(),
             hashed_device_id: "abcdef".to_owned(),
             service_id: 1,
-            duration: DEFAULT_TOKEN_DURATION,
+            duration: TOKEN_DURATION,
             node_type: NodeType::default(),
         };
 
@@ -1192,7 +1221,7 @@ mod tests {
             hashed_fxa_uid: "abcdef".to_owned(),
             hashed_device_id: "abcdef".to_owned(),
             service_id: 1,
-            duration: DEFAULT_TOKEN_DURATION,
+            duration: TOKEN_DURATION,
             node_type: NodeType::default(),
         };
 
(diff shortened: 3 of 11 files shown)


syncstorage/src/settings.rs
```diff
@@ -289,6 +289,7 @@ impl Settings {
         s.set_default("tokenserver.node_type", "spanner")?;
         s.set_default("tokenserver.statsd_label", "syncstorage.tokenserver")?;
         s.set_default("tokenserver.run_migrations", cfg!(test))?;
+        s.set_default("tokenserver.token_duration", 3600)?;
```
ethowitz, line 292: The Python Tokenserver has a default duration of 300 seconds, but it has the duration set to 3600 in the settings: https://github.com/mozilla-services/cloudops-deployment/blob/master/projects/tokenserver/puppet/modules/tokenserver/templates/settings.ini.erb#L55

syncstorage/src/tokenserver/db/models.rs
```diff
@@ -164,7 +164,7 @@ impl TokenserverDb {
              WHERE service = ?
                AND email = ?
                AND generation <= ?
-               AND COALESCE(keys_changed_at, 0) <= COALESCE(?, 0)
+               AND COALESCE(keys_changed_at, 0) <= COALESCE(?, keys_changed_at, 0)
```
ethowitz, line 167: I had left this bit out the query, but the Python Tokenserver has it: https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/assignment/sqlnode/sql.py#L83

I'm not entirely sure what its purpose is, but it seems like these queries should match

syncstorage/src/tokenserver/handlers.rs
```diff
@@ -108,28 +112,60 @@ fn get_token_plaintext(
...
+            // larger than the generation stored on the user record, set the user's generation to be
+            // the keys_changed_at on the request.
+            Some(request_keys_changed_at) if request_keys_changed_at > req.user.generation => {
+                request_keys_changed_at
+            }
+            // If there's not a generation on the request and the keys_changed_at on the request is
+            // less than or equal to the generation stored on the user record, just use the user's
+            // current generation.
+            _ => req.user.generation,
+        },
```
ethowitz, line 168: This whole section was very difficult to understand, so I reworked it all and added tons more comments. I marked unreachable paths with `unreachable!` to ensure that they are, in fact, unreachable paths

ethowitz: The relevant portion of the Python Tokenserver code is here if you want to compare the two: https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/views.py#L326

syncstorage/src/tokenserver/extractors.rs
```diff
@@ -98,6 +97,14 @@ impl TokenserverRequest {
             });
         }
 
+        // If the client previously reported a client state, every subsequent request must include
+        // one. Note that this is only relevant for BrowserID requests, since OAuth requests must
+        // always include a client state.
+        if !self.user.client_state.is_empty() && self.auth_data.client_state.is_empty() {
```
ethowitz, line 103: This check needs to happen before the check for an already-used client state, since `""` could also be a previously used client state, which would cause `""` to invoke a "stale value" error instead of an "empty string" error. (The Python Tokenserver does this too: https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/views.py#L419)

A small detail with likely no material impact, but best to keep the Rust Tokenserver as close as possible to the Python Tokenserver

syncstorage/src/tokenserver/handlers.rs
```diff
@@ -84,7 +83,12 @@ fn get_token_plaintext(
             })?;
         let client_state_b64 = base64::encode_config(&client_state, base64::URL_SAFE_NO_PAD);
 
-        format!("{:013}-{:}", updates.keys_changed_at, client_state_b64)
+        format!(
+            "{:013}-{:}",
+            // We fall back to using the user's generation here, which matches FxA's behavior
```
ethowitz, line 88: https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/views.py#L477

syncstorage/src/tokenserver/handlers.rs
```diff
@@ -108,28 +112,72 @@ fn get_token_plaintext(
...
+    };
+
+    let generation = match req.auth_data.generation {
+        // If there's a generation in the request and it's greater than or equal to that stored on
+        // the user record, update to the value in the request.
+        Some(request_generation) if request_generation >= req.user.generation => request_generation,
+        // If there's a generation in the request and it's smaller than that stored on the user
+        // record, we've already returned an error.
+        Some(_request_generation) => unreachable!(),
+        None => match (req.auth_data.keys_changed_at, req.user.keys_changed_at) {
```
ethowitz, line 157: I thought very hard about the best way to organize this section to make it as readable as possible, but I am very open to suggestions if there's a clearer way!

syncstorage/src/tokenserver/handlers.rs
```diff
@@ -108,28 +112,72 @@ fn get_token_plaintext(
...
+            // keys_changed_at.
+            (Some(request_keys_changed_at), None)
+                if request_keys_changed_at > req.user.generation =>
+            {
+                request_keys_changed_at
+            }
+            // If the request has a keys_changed_at but the above conditions don't hold OR if the
+            // request doesn't have a keys_changed_at, just keep the same generation.
+            (_, _) => req.user.generation,
+        },
```
ethowitz, line 180: I decided to rework this whole section to make it much more obvious. Using `cmp::max` in tandem with `unwrap(0)` made this logic extremely difficult to follow, and I specifically wanted to highlight each possible case explicitly. I also wanted to mark unreachable code paths with `unreachable!` for extra clarity (and safety!)

The relevant portion of the Python Tokenserver is here: https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/views.py#L326

tools/integration_tests/tokenserver/test_support.py
```diff
@@ -17,7 +17,6 @@
 
 
 class TestCase:
-    AUTH_METHOD = os.environ.get('TOKENSERVER_AUTH_METHOD', 'oauth')
```
ethowitz, line 20: This line of code is run a single time when this class definition is read by the interpreter and **not** every time the `TestCase` symbol is referenced (which I think was my original assumption). We need the Tokenserver auth method to be read from the env var every time we run the tests in this class, which is exactly what `setUpClass` is for. I've moved this and the definition of the `_build_auth_headers` function there

ethowitz:
Alright, take two! This is good to go now (I hope)

syncstorage/src/tokenserver/db/models.rs
jrconlin: Ah, ok.
So, check that the value of `keys_changed_at` or `0` is greater than the passed value, `keys_changed_at`, or `0` . 

Yeah, I can see where that would be good. 

If, for whatever reason, the value you're passing is `null` you could with up with `keys_changed_at <= 0` rather than `keys_changed_at <= keys_changed_at`

syncstorage/src/tokenserver/handlers.rs
jrconlin: I like the idea of the heavily commented `match` case, mostly because it forces all the cases to be addressed.

syncstorage/src/tokenserver/settings.rs
```diff
@@ -102,6 +104,7 @@ impl Default for Settings {
             run_migrations: cfg!(test),
             spanner_node_id: None,
             additional_blocking_threads_for_fxa_requests: None,
+            token_duration: 3600,
```
jrconlin, line 107: In some other repos, we put constants like this into their own file or block, so that they're easier to find. 
Might be a follow up issue and a #goodfirstbug .

jrconlin (dismissed):

pjenvey:
From your slack message: "This caused the rust tokenserver to update the generation from zero to be equal to the user’s updated keys_changed_at for any user who made a request that didn’t include generation but did include a keys changed at."

Just wanted to clarify, it sounds like the request causing the extra write included a `keys_changed_at` -- was the `keys_changed_at` actually "updated" though (request.keys_changed_at > user.keys_changed_at in the db) or just included in the request?

syncstorage/src/tokenserver/handlers.rs
```diff
@@ -108,28 +112,72 @@ fn get_token_plaintext(
...
+            // request but not currently on the user record, set the user's generation to the new
+            // keys_changed_at.
+            (Some(request_keys_changed_at), None)
+                if request_keys_changed_at > req.user.generation =>
+            {
+                request_keys_changed_at
+            }
+            // If the request has a keys_changed_at but the above conditions don't hold OR if the
+            // request doesn't have a keys_changed_at, just keep the same generation.
+            (_, _) => req.user.generation,
```
pjenvey, line 179: Wouldn't this be the equivalent?

```suggestion
        None => match keys_changed_at {
            // No generation on the request but an updated keys_changed_at greater than the user's
            // current generation: set the user's generation to the new keys_changed_at.
            Some(keys_changed_at) if keys_changed_at > req.user.generation => keys_changed_at,
            // Otherwise just keep the same generation.
            _ => req.user.generation,
```

Granted I get you're after the verbosity here but it it's basically the same as the above check (minus the 0 case which wouldn't trigger the `> generation` anyway)

jrconlin: FWIW, I'm ok with the verbosity, since it's outlining and documenting potential states and results (a bit like having a matrix). 
I suppose those could be captured in just the documentation, though.

syncstorage/src/tokenserver/db/models.rs
ethowitz: Ah! I see

ethowitz:
>Just wanted to clarify, it sounds like the request causing the extra write included a keys_changed_at -- was the keys_changed_at actually "updated" though (request.keys_changed_at > user.keys_changed_at in the db) or just included in the request?

The request causing the extra write did include a keys_changed_at, but that keys_changed_at was the same as the one we already had, so the keys_changed_at wasn't updated. It was the generation that was erroneously updated to be the keys_changed_at in this situation. If the request had included a new keys_changed_at (i.e. a keys_changed_at greater than the one we've already seen for the user) but no generation, we _would_ want to update the generation to be the keys_changed_at

syncstorage/src/tokenserver/handlers.rs
ethowitz: Yeah I considered simplifying down to that, but I think it hides a number of distinct cases. I thought it would be preferable to be as explicit as possible so each edge case is clear

pjenvey (approved):
This fix looks good.

A couple more questions though:

If this comment is accurate:

    #   * `keys_changed_at` is a monotonic timestamp, and increases every time
    #     the user's keys change. This is a type of auth-related change, so
    #     `keys_changed_at` <= `generation` at all times.

How exactly would a `keys_changed_at` with the same value as in the db have triggered an update to generation? If I'm reading this right it would have hit this previous code path (when `req.auth_data.generation` is `None`): `cmp::max(req.auth_data.keys_changed_at, Some(req.user.generation))` -- which should have always returned `req.user.generation` if we're guaranteed `keys_changed_at` was always `<=` to it (which wouldn't have triggered a `generation` update AFAICT).

I'm wondering what we think the `generation` values were updated to and if there was any impact to those clients.

ethowitz:
The requests that triggered the update didn't include generations, and the records in the database had generations == 0 (generation is optional -- clients don't have to report it) for the records in question. So `cmp::max(req.auth_data.keys_changed_at, Some(req.user.generation))` would have evaluated to `req.auth_data.keys_changed_at` in every case for those requests.

In terms of impact, I think there are a few interesting scenarios, but none cause any trouble. I've confirmed with the FxA team that we should never be getting a `generation` that's less than `keys_changed_at`, since the `generation` is always updated in addition to `keys_changed_at` when a user's keys are updated. For the requests that caused the superfluous database writes, even though the users didn't report a `generation` to us, we know that the `generation` on FxA **must be greater than or equal to `keys_changed_at`**, so the erroneously-set value in Tokenserver's database actually represents the smallest possible value the `generation` could be in FxA's database, since we set the `generation` to the current value of `keys_changed_at`.

This presents us with a few cases:

1. A user presents a generation that's equal to the one we erroneously set
  * If the `keys_changed_at` is being updated in the request, the user would be 401ed as expected, since a keys_changed_at update must be accompanied by a generation update
  * If the `keys_changed_at` is not being updated, this would just set the `generation` to this value in the database. Given that it's already set, there would be no change and the user would get a token as expected
2. A user presents a `generation` that's less than the one we erroneously set
  * In this instance, the user would be 401ed. Python Tokenserver would _technically_ allow a user with a `keys_changed_at` set but no `generation` set to begin reporting a `generation` that's less than the current `keys_changed_at` (provided that the `keys_changed_at` is not being updated in the request). **However**, we should never run into this situation, since the FxA team confirmed that the `generation` should never be less than `keys_changed_at`. A bad actor couldn't even manipulate these values, given that they are included as claims in the JWT signed by FxA's private key. The only way we'd run into this is if there were a bug on FxA
3. A user presents a `generation` that's greater than the one we erroneously set
  * If the `keys_changed_at` is being updated in the request
    * If it's < the `keys_changed_at` in the request, they'd be 401ed as expected, since `generation` must be >= `keys_changed_at`
    * If it's >= the `keys_changed_at` on the request, the new value would be persisted to the database
  * If the `keys_changed_at` is not being updated in the request, we'd just set the `generation` to be the new value
4. A user presents a `keys_changed_at` that's larger than the one we have stored, which means the new `keys_changed_at` is > the `generation` in the db and > the `keys_changed_at` in the db
  * If the user presents a `generation` in that request, it would need to be higher than the `keys_changed_at` in the request anyway, so the new value would be set
  * If the user doesn't present a `generation`, the `keys_changed_at` in the database would be updated to the new value, and since the `generation` in the database is < the new `keys_changed_at`, [it would be updated to the new `keys_changed_at`](https://github.com/mozilla-services/syncstorage-rs/blob/bug/fix-tokenserver-generation-keys-changed-at/syncstorage/src/tokenserver/handlers.rs#L164)
5. A user presents a new client state
  * If the `generation` reported in the request is <= the one we erroneously set, this would pose a problem. But a client state update necessitates a `keys_changed_at` update, so if the `generation` reported is <= `generation` we set, that would mean it’s also less than the `keys_changed_at` on the request, which is invalid. (Also, as stated above, we should never be getting a `generation` that's less than the one we set, since FxA doesn't allow for that behavior)
  * If there's no `generation` reported, it would be updated to the new `keys_changed_at` value, since a client state update necessitates a `keys_changed_at` update, and [we set `generation` to `keys_changed_at` when we get a new `keys_changed_at` that's greater than the `generation` on the user record](https://github.com/mozilla-services/syncstorage-rs/blob/bug/fix-tokenserver-generation-keys-changed-at/syncstorage/src/tokenserver/handlers.rs#L164)

There are also certain situations on FxA where a user may have a `generation` of 0. The Python Tokenserver treats 0 and null equivalently, so I've updated the Rust Tokenserver to do the same and added integration test coverage. I've also ensured that the new integration tests succeed against the Python Tokenserver. The most interesting case above is probably 2, since we could potentially be locking the affected users out of their accounts if we were to receive a `generation` less than the one we set, but because FxA would never provide us with such a value, I don't think it's worth taking any action.

pjenvey:
This makes sense, thank you for digging into this w/ the FxA team. I think we're safe to roll this out after quick verification on stage.

I think #2 can occur naturally when a client happens to be "behind" an auth change (say due to lagging at the exact time another client triggers `generation` update), however that should be quite rare. This suggests we could metric this occurrence (in general, not just solely because of this bug) in case FxA ever violates the contract. If that sounds like a good idea we can add that in a separate issue and at a later time/release.

pjenvey (approved):

PR #1406, opened 2022-09-25
perf: always verify OAuth tokens on blocking thread

ethowitz:
## Description

When an OAuth token being verified is not a JWT or the key being used to verify the signature is not a JWK, PyFxA falls back to verifying the token by making a request to FxA: https://github.com/mozilla/PyFxA/blob/main/fxa/oauth.py#L303

This means that, even if the JWK is cached, it's possible that a request will be dispatched, which would block the Actix worker from processing any other requests until that request returns. To account for this, we should always be verifying tokens with PyFxA in a thread on the blocking threadpool.

In practice, OAuth tokens being sent by Firefox clients should always be well-formed, but if a bad actor were to spam `GET /1.0/sync/1.5` with garbage in the `Authorization` header, it could block the worker threads and significantly affect performance for other users.

## Testing

This change is pretty trivial: the integration tests should ensure that OAuth verification is continuing to work properly.

### DIFF
--- a/syncstorage/src/tokenserver/auth/oauth.rs
+++ b/syncstorage/src/tokenserver/auth/oauth.rs
@@ -30,7 +30,6 @@ pub struct Verifier {
     // pointer
     inner: Py<PyAny>,
     timeout: u64,
-    jwk_is_cached: bool,
 }
 
 impl Verifier {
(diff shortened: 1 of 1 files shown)


ethowitz:
We may want to increase the size of Actix's blocking threadpool, though I'll probably wait to see what the load tests indicate first

pjenvey (approved):

PR #1407, opened 2022-09-30
refactor: add database crates

### LINKED ISSUE #1277: Add database crates
Add the following crates (dependencies are noted as sub-bullets):

* syncstorage-db: Code that neatly combines and exposes the functionality of syncstorage-mysql and synsctorage-spanner for consumption by syncstorage-web
  * syncstorage-mysql: MySQL impls for the database traits in syncstorage-db-common
  * syncstorage-spanner: Spanner impls for the database traits in syncstorage-db-common
* tokenserver-db: Tokenserver database-related code
* syncstorage-mysql
* syncstorage-spanner

syncstorage-db and tokenserver-db should provide an easy-to-use high-level API exposing database functionality. This functionality will eventually be consumed by syncstorage-web and tokenserver-web.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DISCO-1940)

ethowitz:
## Overview

This PR breaks up syncstorage's and tokenserver's database functionality into a number of new crates:

* syncserver-db-common: Database-related code that is shared between syncstorage and tokenserver
  * Does not depend on any other database crates
* syncstorage-db-common: Database-related code that is shared between the MySQL and Spanner backends of syncstorage. Note that this crate is not used in any way by Tokenserver
  * Depends on syncserver-db-common
* syncstorage-db: Acts as an interface between the two database backends and the `syncserver` crate
  * Depends on:
    * syncstorage-mysql
    * syncstorage-spanner
    * syncserver-db-common
    * syncstorage-db-common 
* syncstorage-mysql: The code that powers syncstorage's MySQL database backend
  * Depends on:
    * syncstorage-db-common
    * syncserver-db-common
* syncstoarge-spanner: The code the powers syncstorage's Spanner database backend
  * Depends on:
    * syncstorage-db-common
    * syncserver-db-common
* tokenserver-db: The code that powers tokenserver's MySQL database backend
  * Depends on:
    * syncserver-db-common

I'd recommend generating the docs using `cargo doc --no-deps --workspace` to view the public interfaces for each crate in the workspace. I tried to keep the public interfaces for each crate as simple as possible.

## Crate Relationships

It might be useful to visualize how the new crates relate to one another (I generated this using [`cargo depgraph`](https://crates.io/crates/cargo-depgraph/)):
graph

The idea is that `syncserver` should not need to know any details about the underlying database backend. `syncstorage-db` acts as an interface to the database that reveals as little as possible about what's happening under the hood. `syncstorage-db` has two features (`mysql` and `spanner`) that enable the MySQL and Spanner backends, respectively.

## `cargo` features

I added two `cargo` features to the top-level `syncserver` crate: `mysql` and `spanner`. These features represent the two syncstorage database backends. This was an important addition because the Spanner backend pulls in `protobuf` and `grpcio`, which both contribute very substantially to the time it takes to build syncstorage. Giving developers the ability to omit the code related to Spanner when developing locally substantially improves compile times. By my measurements, compiling with the `spanner` feature takes about 4.5 minutes, whereas compiling with the `mysql` feature takes only about 1.5 minutes.

Adding these `cargo` features also reduces the size of the compiled binary, since we are omitting substantial amounts of code we never use.

## Error Handling

The toughest part about this work was figuring out how to handle errors. Originally, we had a single `DbError` which encapsulated knowledge of both the Spanner and MySQL backends. Decoupling the Spanner and MySQL backend meant that each needed to have its own error type that included only knowledge of its own backend. To resolve this, I defined a number of different error types in different places and nested them using `enum`s.

* syncserver-db-common:
  * `MysqlError`: An error representing low-level, MySQL-specific errors (e.g. connection errors, migration errors, errors that arise during queries, etc.). Basically, this wraps a few `diesel` errors. This type is shared by the syncstorage MySQL backend and the Tokenserver MySQL backend
* syncstorage-db-common:
  * `SyncstorageDbError`: An error that represents **syncstorage-specific** errors that arise during database queries. It represents things like `CollectionNotFound`, `BsoNotFound`, quota errors, batch errors, etc. This type doesn't wrap any lower-level errors, and the errors are not specific to MySQL or Spanner.
* syncstorage-mysql:
  * `DbError`: This is a top-level database error for the MySQL backend. It encapsulates both `MysqlError` and `SyncstorageDbError`, representing any error that may occur when processing a MySQL query for syncstorage.
* syncstorage-spanner:
  * `DbError`: This is a top-level database error for the Spanner backend. It encapsulates `SyncstorageDbError` and lower-level Spanner errors (there was no need for a `SpannerError` type akin to the `MysqlError` type, since the Tokenserver backend doesn't use Spanner). It represents any error that may occur when processing a Spanner query for syncstorage.
* tokenserver-mysql:
  * `DbError`: This is a top-level database error for tokenserver. It encapsulates `MysqlError` and any other internal errors that may occur when processing a MySQL query for tokenserver.

I also made the `*ErrorKind` types private. This was a more opinionated decision on my part, but IMO, it simplifies the public APIs of the error types, making them easier to understand and use by clients. I include a number of helper methods on the base error types to allow clients to access information hidden in the `*ErrorKind` types.

## Use of `debug_assertions`

Unfortunately, it's not possible to pass through items defined conditionally under the `test` feature to other crates (i.e. if something is defined with `#[cfg(test)]` in a crate, it won't be rendered visible to parent crates). There are numerous testing utilities that we define under the `test` feature, and I thought it would be wrong to include them in release builds. To remedy this, I define testing utilities with the `debug_assertions` feature instead, which only compiles the testing utilities in debug builds. This feels a little inelegant, but I thought it would be preferable to including testing utilities in release builds or defining our own new `testing` feature, which would further complicate compiling syncserver.

## CI

This PR introduces new cargo features (one for the Spanner backend and one for the MySQL backend), which are mutually exclusive.  This makes it impossible to build a single Docker image against which we can run the MySQL and Spanner e2e tests. Instead, we need to build **two** separate Docker images: one compiled with the MySQL database backend and one compiled with the Spanner database backend. Our Docker image builds are quite lengthy, so I introduced [`cargo chef`](https://github.com/LukeMathWalker/cargo-chef) to facilitate the caching of build dependencies between CI runs. This dramatically reduces build times, since only our own crates need to be recompiled with code changes -- the dependencies are redownloaded and recompiled only if our Cargo.tomls change.

Additionally, I configured the MySQL and Spanner images/e2e tests to run in parallel, further improving build times. We might want to consider using a larger CircleCI resource class to allow `cargo` to build more crates in parallel and to improve single-threaded CPU performance.

## Testing

Given that no functionality should be changing, ensuring compilation and that the unit and integration tests pass should be sufficient. We should definitely run through the end-to-end QA tests on stage before we deploy this to production.

## Issue(s)

Closes #1277


syncserver-db-common/src/lib.rs
```diff
@@ -97,212 +38,33 @@ impl From<deadpool::Status> for PoolState {
...
-    pub fxa_uid: String,
-    pub fxa_kid: String,
+pub async fn run_on_blocking_threadpool<F, T, E, M>(f: F, e: M) -> Result<T, E>
+where
+    F: FnOnce() -> Result<T, E> + Send + 'static,
+    T: Send + 'static,
+    E: fmt::Debug + Send + 'static,
+    M: FnOnce(String) -> E,
+{
+    web::block(f)
```
ethowitz, line 48: Unfortunately, that we have to use Actix's blocking threadpool instead of tokio's forces actix web to be a dependency of this crate

syncserver/src/tokenserver/handlers.rs
```diff
@@ -121,7 +121,7 @@ struct UserUpdates {
 
 async fn update_user(
     req: &TokenserverRequest,
-    db: Box<dyn Db>,
+    db: Box<dyn DbTrait>,
```
ethowitz, line 124: It would be nice to get rid of these trait objects now that we don't need to heterogeneously handle MySQL and Spanner database pools, but we still need them to handle `MockDb` and `MockDbPool` in the unit tests, since some unit tests use the mock data structures and some use the real database data structures

syncserver/src/web/extractors.rs
```diff
@@ -1173,7 +1170,10 @@ impl FromRequest for HawkIdentifier {
                 "tokenserver_origin",
                 &hawk_id.tokenserver_origin.to_string(),
             );
-            tags.commit(&mut exts);
```
ethowitz, line 1176: I removed this method from the `Tags` type because it depended on Actix Web's request extensions. Now that the `Tags` lives in `syncserver-common` (which has no other dependency on Actix), I decided to remove it. It's only called in two places anyway

syncstorage-db/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+pub type Db = syncstorage_spanner::SpannerDb;
+
+pub use syncserver_db_common::{GetPoolState, PoolState};
+pub use syncstorage_db_common::error::DbErrorIntrospect;
+
+pub use syncstorage_db_common::{
+    params, results,
+    util::{to_rfc3339, SyncTimestamp},
+    DbPoolTrait, DbTrait, Sorting, UserIdentifier,
+};
```
ethowitz, line 37: I re-export a bunch of stuff here so the syncserver crate only needs to deal with a simple, flat interface

syncstorage-db-common/src/lib.rs
```diff
@@ -0,0 +1,304 @@
...
+
+/// The ttl to use for rows that are never supposed to expire (in seconds)
+pub const DEFAULT_BSO_TTL: u32 = 2_100_000_000;
+
+/// Non-standard collections will be allocated IDs beginning with this value
+pub const FIRST_CUSTOM_COLLECTION_ID: i32 = 101;
+
+#[async_trait]
+pub trait DbPoolTrait: Sync + Send + Debug + GetPoolState {
+    type Error;
```
ethowitz, line 52: We need to introduce an associated type here because implementers of this trait have their own `DbError` types

syncstorage-mysql/src/error.rs
```diff
@@ -0,0 +1,138 @@
...
+
+    pub fn quota() -> Self {
+        DbErrorKind::Common(SyncstorageDbError::quota()).into()
+    }
+}
+
+#[derive(Debug, Error)]
+enum DbErrorKind {
+    #[error("{}", _0)]
+    Common(SyncstorageDbError),
```
ethowitz, line 49: I struggled a bit with the naming here. `Common` is supposed to be a Syncstorage database error that is not specific to the database backend

syncstorage-spanner/src/error.rs
```diff
@@ -0,0 +1,145 @@
...
+    Expired,
+
+    #[error("A database error occurred: {}", _0)]
+    Grpc(#[from] grpcio::Error),
+
+    #[error("Database integrity error: {}", _0)]
+    Integrity(String),
+
+    #[error("Spanner data load too large: {}", _0)]
+    TooLarge(String),
```
ethowitz, line 72: These could have been pulled into a `SpannerDbError` type akin to the `MysqlDbError` type, but the reason I made the `MysqlDbError` type was because it's also used by Tokenserver

ethowitz:
This can't be merged until #1276 is merged

syncserver/src/tokenserver/handlers.rs
jrconlin: Going to say that we probably shouldn't. I know YAGNI is a thing, but if Autopush is any guide, having a level of database abstraction can be a benefit in the future. Even if it's just encapsulating one "type" right now, there's a reasonable chance it may wrap future ones later. (e.g. we have to move to a new provider, or we need to move to a different data store, etc.)

syncstorage-spanner/src/batch.rs
```diff
@@ -98,7 +98,7 @@ pub async fn append_async(db: &SpannerDb, params: params::AppendToBatch) -> Resu
     if !exists {
         // NOTE: db tests expects this but it doesn't seem necessary w/ the
         // handler validating the batch before appends
-        Err(DbErrorKind::BatchNotFound)?
+        return Err(DbError::batch_not_found());
```
jrconlin, line 101: I know that adding the `?` does the same thing, but I agree that this is more readable, particularly when fast reading the code.

jrconlin (changes requested):
Looks good, aside from the Makefile pointing to your directory.

Makefile
```diff
@@ -40,14 +44,28 @@ python:
 	python3 -m venv venv
 	venv/bin/python -m pip install -r requirements.txt
 
-run: python
-	PATH="./venv/bin:$(PATH)" RUST_LOG=debug RUST_BACKTRACE=full cargo run -- --config config/local.toml
+run_mysql: python
+	PATH="./venv/bin:$(PATH)" \
+		# See https://github.com/PyO3/pyo3/issues/1741 for discussion re: why we need to set the
+		# below env var
+		PYTHONPATH=/Users/edonowitz/Dev/syncstorage-rs/venv/lib/python3.9/site-packages \
```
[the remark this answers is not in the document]
ethowitz: Oops!

pjenvey (commented):
Looks good. Wow on the `cargo chef` changes, thanks for adding those!

Of course on a huge PR like this I really only have one thing to bikeshed, and it's mostly just food for thought: the `Trait` suffix for `Db/PoolTrait` feels odd -- I don't think I've ever seen it used like that in Rust -- and a bit superfluous when there's a `dyn` beside it (which is most of the time when referencing traits). One alternative could be to keep the trait names as is and rename `syncstorage_db::Db` and `DbPool` to `DbImpl/DbPoolImpl` (I saw [a complex trait setup in servo](https://github.com/servo/servo/blob/master/components/layout_thread/lib.rs#L2025) that does this) -- and maybe the `Impl` kinda helps mark that the type's swapped via the feature flag and backed by a trait?

syncstorage-mysql/src/error.rs
```diff
@@ -0,0 +1,144 @@
...
+    #[error("{}", _0)]
+    Mysql(MysqlError),
+}
+
+impl From<DbErrorKind> for DbError {
+    fn from(kind: DbErrorKind) -> Self {
+        match &kind {
+            DbErrorKind::Common(dbe) => Self {
+                status: dbe.status,
+                backtrace: Box::new(dbe.backtrace.clone()),
```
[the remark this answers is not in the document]
ethowitz: I think the `backtrace` method approach sounds good 👍🏻 I created a ticket [here](https://mozilla-hub.atlassian.net/browse/SYNC-3548)

ethowitz:
@pjenvey I like your idea about removing the `Trait` suffix and adding an `Impl` suffix to the types that implement the traits. I'll make the change!

syncserver/Cargo.toml
```diff
@@ -13,46 +13,23 @@ default-run = "syncserver"
 [dependencies]
 actix-http = "2"
 actix-web = "3"
-actix-rt = "1"          # Pin to 1.0, due to dependencies on Tokio
+actix-rt = "1"                                                       # Pin to 1.0, due to dependencies on Tokio
 actix-cors = "0.5"
-actix-service = "1.0.6"
 async-trait = "0.1.40"
```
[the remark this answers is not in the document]
ethowitz: Good call! I created a separate ticket to do this: https://mozilla-hub.atlassian.net/browse/SYNC-3549

syncserver-common/Cargo.toml
```diff
@@ -4,5 +4,16 @@ version = "0.13.0"
 edition = "2021"
```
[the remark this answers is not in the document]
ethowitz: Similar to above, I created a ticket to track this: https://mozilla-hub.atlassian.net/browse/SYNC-3550

syncstorage-db/src/tests/db.rs
```diff
@@ -16,9 +19,9 @@ lazy_static! {
 }
 
 #[tokio::test]
-async fn bso_successfully_updates_single_values() -> Result<()> {
+async fn bso_successfully_updates_single_values() -> Result<(), DbError> {
```
jrconlin, line 22: Bit surprised you didn't create a type for this, just to save keystrokes, but it's fine.

jrconlin (approved):

syncstorage-db/src/tests/db.rs
ethowitz: There were a few different `Result`/`DbResult`/`DbFuture` aliases, and I thought it would make things easier to follow if we were more explicit with our return types but 🤷🏻 perhaps that's just me

ethowitz:
CI was stuck so I did a `git commit --amend --no-edit` and force pushed...I still see that GitHub appears to be waiting for the status of the `e2e-tests` job, which no longer exists in the CircleCI file...

jrconlin:
I removed the requirement for the missing e2e-test from the master branch protection settings. 

I am curious why they're missing.

ethowitz:
@jrconlin The original `e2e-tests` job was removed and replaced with `mysql-e2e-tests` and `spanner-e2e-tests` -- should I update the branch protection to reflect that?

jrconlin:
Ah, yeah. Sorry, forgot about that.

ethowitz:
@jrconlin Ok I updated the branch protection settings, and that seems to have fixed the issue. I think this needs another approval before I can merge

jrconlin (approved):

Issue #1279, opened 2022-04-07
Add web crates


data-sync-user:
Add the following crates:

* syncstorage-web: Neatly wraps all of the database-related Syncstorage code and exposes it via an HTTP API provided by Actix Web
* tokenserver-web: Neatly wraps all of the database-related Tokenserver code and exposes it via an HTTP API provided by Actix Web

These two crates should **not** depend on each other; rather, they should provide data structures that can be composed to create a final server. That is, it should be possible to run Syncstorage and Tokenserver separately **or** together. The Syncstorage and Tokenserver [Apps](https://docs.rs/actix-web/3.3.3/actix_web/struct.App.html) are currently very tightly-coupled – breaking this dependency may be difficult.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1747)

PR #1413, opened 2022-10-03
refactor: add tokenserver-auth crate

### LINKED ISSUE #1278: Add tokenserver-auth crate
Add a tokenserver-auth crate that contains all of the code responsible for verifying OAuth tokens/BrowserId assertions and crafting tokens to return to Sync clients. This should be the only crate that depends on pyo3. This crate will be consumed exclusively by tokenserver-web.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1746)

ethowitz:
## Description

Adds a new `tokenserver-auth` crate whose responsibility is to expose functionality related to:

1. Verifying OAuth tokens and BrowserID assertions from FxA
2. Creating tokens and the request-signing secrets derived from those tokens

Its only consumer is the `syncserver` crate (though this will change to be the `tokenserver-web` crate with #1279).

## Testing

There is no change in functionality -- successful compilation and passing unit/integration tests should be enough to verify that everything is working as it should.

## Issue(s)

Closes #1278


syncserver-common/Cargo.toml
```diff
@@ -4,6 +4,7 @@ version = "0.12.3"
 edition = "2021"
 
 [dependencies]
+actix-web = "3"
```
ethowitz, line 7: It is unfortunate that actix web needs to be a dependency here. since only reason we pull it in is to use the `web::block` function to spawn tasks on blocking threads, which doesn't have anything to do with the syncstorage/tokenserver web API. Actix 4 exposes settings related to tokio's blocking threadpool, so once we upgrade to Actix 4, we'll be able to replace this dependency with tokio.

jrconlin (dismissed):

pjenvey (dismissed):

jrconlin (approved):

pjenvey (approved):

PR #1417, opened 2022-10-06
feat: emit metrics about the number of blocking tasks

### LINKED ISSUE #1416: Report blocking threadpool statistics
Given that Tokenserver makes extensive use of Actix’s blocking threadpool, it would be good for us to report metrics for the number of tasks spawned on the blocking threadpool. There’s no way for us to get these statistics directly from the Actix server, so we’ll need to keep count ourselves.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1998)

ethowitz:
## Description

Increment the counter when the blocking task is spawned, and decrement the counter when the future returns. This would be visualized on a per-pod basis in grafana. My (possibly incorrect) assumption here is that each pod would have its own counter, and when a pod starts up, its counter is 0.
## Issue(s)

Closes #1416

### DIFF
--- a/syncstorage/src/db/mod.rs
+++ b/syncstorage/src/db/mod.rs
@@ -69,13 +69,19 @@ pub fn spawn_pool_periodic_reporter<T: GetPoolState + Send + 'static>(
     Ok(())
 }
 
-pub async fn run_on_blocking_threadpool<F, T>(f: F) -> Result<T, DbError>
+pub async fn run_on_blocking_threadpool<F, T>(metrics: Metrics, f: F) -> Result<T, DbError>
 where
     F: FnOnce() -> Result<T, DbError> + Send + 'static,
     T: Send + 'static,
 {
-    web::block(f).await.map_err(|e| match e {
+    metrics.incr("blocking_tasks");
+
+    let result = web::block(f).await.map_err(|e| match e {
         BlockingError::Error(e) => e,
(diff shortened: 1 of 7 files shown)


syncstorage/src/db/mod.rs
```diff
@@ -69,13 +69,24 @@ pub fn spawn_pool_periodic_reporter<T: GetPoolState + Send + 'static>(
...
+/// WARNING: Calling `web::block` anywhere else will result in inaccurate threadpool metrics being
+/// reported. If you want to spawn a task on Actix's blocking threadpool, you **must** use this
+/// function.
+pub async fn run_on_blocking_threadpool<F, T>(metrics: Metrics, f: F) -> Result<T, DbError>
 where
     F: FnOnce() -> Result<T, DbError> + Send + 'static,
     T: Send + 'static,
 {
-    web::block(f).await.map_err(|e| match e {
+    metrics.incr("blocking_tasks");
```
ethowitz, line 82: I don't think there's a way to get this statistics from the Actix server/tokio runtime directly, so this was my best attempt to keeping track of these tasks. The main downside to this (as reported above) is that if someone calls `web::block` directly somewhere else in the code base, the task spawned there would not be reported in the metrics

ethowitz:
I think my assumption above is incorrect -- I'll do some more research. Sorry for the noise!

PR #1418, opened 2022-10-06
feat: report blocking threadpool statistics

### LINKED ISSUE #1416: Report blocking threadpool statistics

ethowitz:
## Description

Increment the counter when a task is added to the blocking threadpool; decrement the counter when the future returns. The counters are tagged with the hostname of each pod.

## Issue(s)

Closes #1416


syncstorage/src/db/mod.rs
```diff
@@ -69,13 +77,24 @@ pub fn spawn_pool_periodic_reporter<T: GetPoolState + Send + 'static>(
     Ok(())
 }
 
-pub async fn run_on_blocking_threadpool<F, T>(f: F) -> Result<T, DbError>
+/// Runs a function as a task on Actix's blocking threadpool.
+///
+/// WARNING: Calling `web::block` anywhere else will result in inaccurate threadpool metrics being
+/// reported. If you want to spawn a task on Actix's blocking threadpool, you **must** use this
+/// function.
+pub async fn run_on_blocking_threadpool<F, T>(metrics: Metrics, f: F) -> Result<T, DbError>
```
ethowitz, line 85: I don't love that we need to pass the `Metrics` into the function, but I couldn't think of a clean way to avoid it

syncstorage/src/db/mod.rs
```diff
@@ -69,13 +77,24 @@ pub fn spawn_pool_periodic_reporter<T: GetPoolState + Send + 'static>(
...
 where
     F: FnOnce() -> Result<T, DbError> + Send + 'static,
     T: Send + 'static,
 {
-    web::block(f).await.map_err(|e| match e {
+    metrics.incr_with_tag("blocking_tasks", "hostname", &HOSTNAME);
```
ethowitz, line 90: I _think_ we do this elsewhere to get separate counters for each tag

pjenvey, line 90: This should probably be a `gauge` instead of a `counter`.

With an actual `gauge` we'd have to maintain the count on our side, however. An easy way would to do so would be to `Ordering::Relaxed` inc/dec say an `AtomicI64` somewhere (as much as I discourage globals, it could even be a module level static for this use case), and emit its value in our existing periodic reporter (which emits every 10 seconds, we could make that more often if necessary).

However I'll admit it's not totally clear to me how such different types ultimately result in influxdb/grafana (especially because the influxdb wire protocol treats everything as a number, it's just a matter of who's doing the tallying). Maybe the counter isn't much of a difference here. A `gauge` is at least be a bit more efficient for telegraf/influxdb to handle though.

ethowitz: @pjenvey I hadn't heard of `gauge` before, but I agree that I like it here more than a counter 👍🏻 Do you think we should report changes to the gauge with every new blocking task, or should we spawn something that periodically reports the threadpool stats?

I'm thinking I could create a `BlockingThreadpool` struct that we keep in `ServerState` that has its own copy of `Metrics` and stores up-to-date counts of the number of blocking tasks. That would also get rid of the need for any globals

pjenvey: @ethowitz I think the existing `db::spawn_pool_periodic_reporter` could emit it (but then maybe it needs to be renamed). It's currently hardcoded to report every 10 seconds. We could make that configurable if it needs to emit more often.

jrconlin (approved):

ethowitz:
Ok, I created a new `BlockingThreadpool` type that tracks the number of active blocking tasks with an `AtomicU64`. I think that we could technically use `Ordering::RelAcq` for the writes to the task count, since x86 is strongly-ordered anyway (the [nomicon states that `Ordering::Relaxed` doesn't provide a benefit on strongly-ordered platforms](https://doc.rust-lang.org/nomicon/atomics.html#relaxed)), and x86 platforms are strongly-ordered. Ultimately, I doubt it matters much, though.

syncserver/src/server/mod.rs
```diff
@@ -371,3 +396,92 @@ pub fn build_cors(settings: &Settings) -> Cors {
...
+    /// Runs a function as a task on the blocking threadpool.
+    ///
+    /// WARNING: Spawning a blocking task through means other than calling this method will
+    /// result in inaccurate threadpool metrics being reported. If you want to spawn a task on
+    /// the blocking threadpool, you **must** use this function.
+    pub async fn spawn<F, T, E>(&self, f: F) -> Result<T, E>
+    where
+        F: FnOnce() -> Result<T, E> + Send + 'static,
+        T: Send + 'static,
+        E: fmt::Debug + Send + InternalError + 'static,
```
ethowitz, line 468: Requiring `E: InternalError` prevents callers from having to pass a function manually to construct `E`'s version of an internal error

syncserver/src/server/mod.rs
```diff
@@ -371,3 +396,92 @@ pub fn build_cors(settings: &Settings) -> Cors {
...
+            time::delay_for(interval).await;
+        }
+    });
+
+    Ok(())
+}
+
+/// A threadpool on which callers can spawn non-CPU-bound tasks that block their thread (this is
+/// mostly useful for running I/O tasks). `BlockingThreadpool` intentionally does not implement
+/// `Clone`: `Arc`s are not used internally, so a `BlockingThreadpool` should be instantiated once
```
ethowitz, line 451: An alternative would have been to encapsulate an `Arc<AtomicU64>` and allow clients to `clone` `BlockingThreadpool`s, but I thought it would be more obvious to have a single instance of `BlockingThreadpool` represent the threadpool in its entirety

jrconlin:
I like this approach. I like that the `BlockingThreadPool` is essentially a singleton. We should probably note the dependency on x86 ordering for the self-hosting crowd. (Maybe a comment indicating a performance concern?). I also like the `InternalError` trait.

jrconlin (dismissed):

jrconlin:
@ethowitz !! Last PR is showing as unverified! that means we may not be able to land this. Might want to recommit, with signature, and do a force push so that it can land.

ethowitz:
Nooooo I don't know why this happened

jrconlin (approved):

ethowitz:
@jrconlin Merging in master introduced a compiler error -- should be all good now 👍🏻

jrconlin (approved):

pjenvey (approved):

PR #1430, opened 2022-10-26
fix syncstorage DB cfg key in default config

s644:
## Description

This patch fixes the path for syncstorage database key. In my setup (Debian, self compiled, NO docker) I wasn't able to connect to my database with another user/password than root/nopwd via config file. I figured out that the key's parent was missing.

## Testing

Run `make run` with and without default values (e.g. some user db user than root)


clouserw:
Thanks for the patch.  This is already in place at https://github.com/mozilla-services/syncstorage-rs/blob/master/config/local.example.toml#L47

PR #1431, opened 2022-10-30
chore: make docker files more podman friendly

mb:
Make docker files more podman friendly by specifying that images come from docker.io. It would be possible to specify the dependency as `docker.io/mysql` instead of `docker.io/library/mysql`. The doc seems to prefer the longer version, so I used it in this PR as well.

## Description

Specify registry from which podman should pull the images. From the doc:

> **Note**: Podman searches in different registries. Therefore it is recommend to use the full image name (`docker.io/library/httpd` instead of `httpd`) to ensure, that you are using the correct image. [[1]](https://podman.io/getting-started/#searching-pulling--listing-images)

On Arch Linux it doesn't seem to find the image without the `docker.io`-prefix, hence I created this PR.

## Testing - How should reviewers test?
Test by verifying that the images still build and start with docker/docker-compose. And maybe by running them with podman/podman-compose

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM rust:1.64-buster as builder
+FROM docker.io/library/rust:1.64-buster as builder
 WORKDIR /app
 ADD . /app
 ENV PATH=$PATH:/root/.cargo/bin
@@ -21,7 +21,7 @@ RUN \
     cargo install --path ./syncserver --locked --root /app && \
     cargo install --path ./syncserver --locked --root /app --bin purge_ttl
 
-FROM debian:buster-slim
+FROM docker.io/library/debian:buster-slim
 WORKDIR /app
 COPY --from=builder /app/requirements.txt /app
 COPY --from=builder /app/mysql_pubkey.asc /app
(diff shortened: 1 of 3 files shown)


jrconlin:
Thanks! Unfortunately, we can't land this as is because the commits are not verified. (See our [Contributing](https://github.com/mozilla-services/syncstorage-rs/blob/master/CONTRIBUTING.md) doc). 
I did add https://github.com/mozilla-services/syncstorage-rs/issues/1433 this could be included in a future fix.

mb:
Thanks for linking the contributing doc (sorry, I didn't read it beforehand). I hope the PR is now in line with the guidelines :). First time signing a git commit :tada:

pjenvey:
I'm not opposed to changing to fully qualified names here but I think this is more of a podman, OS, or docker-compose issue, see:

https://github.com/containers/podman/issues/8896#issuecomment-1359043284
https://github.com/containers/podman/issues/13234

I can see why podman would recommend fully qualified names with its multiple docker registry support: so that there's less of a chance for conflicting or incorrect images.

However for compatibility sake, any docker-like/OCI system should be treating "short names" as the equivalent of a docker.io entry, otherwise the large majority of the existing docker ecosystem is just not going to work.

pjenvey:
This was applied in #1479 & #1492

PR #1435, opened 2022-11-01
chore: switch from mariadb libmysqlclient to mysql's

### LINKED ISSUE #1434: Switch from mariadb libmysqlclient to mysql's
mariadb’s libmysqlclient doesn’t support TLS by default, whereas the mysql impl does.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-2079)

pjenvey:
as it defaults to TLS connections. also diesel's ssl_mode relies on the mysql specific MYSQL_OPT_SSL_MODE option

Closes #1434


mysql_pubkey.asc
```diff
@@ -0,0 +1,49 @@
+-----BEGIN PGP PUBLIC KEY BLOCK-----
```
ethowitz, line 1: Was this file included intentionally?

pjenvey: Yes, as it must be imported via apt-key to pull from the mysql apt repo. The key could be pulled from the MIT key server every time instead as described in their [docs](https://dev.mysql.com/doc/mysql-apt-repo-quick-guide/en/#repo-qg-apt-repo-manual-setup), however I've committed it because pgp.mit.edu was unavailable at some point yesterday while I was testing this (and very slow to pull from when it was available).

We can get rid of this in the future likely by switching off debian entirely to something like ubuntu that provides the mysql libmysqlclient on their apt servers.

ethowitz:
Should we consider pulling in the AWS RDS TLS certs? https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html

pjenvey:
> Should we consider pulling in the AWS RDS TLS certs? https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html

These aren't necessary for enabling TLS. However they could be needed for (potentially) enabling hostname cert verification, which isn't on by default. This would be nice to have (preventing man in the middle attacks) but will require some further effort to switch on.

ethowitz:
Yeah, I wasn't sure whether we wanted to enable cert verification now or at some point in the future, but I'm good with waiting :+1:

pjenvey:
https://github.com/mozilla-services/syncstorage-rs/issues/1438 for cert verification

PR #1437, opened 2022-11-01
Fix high cardinality metrics tags

### LINKED ISSUE #1436: Remove user agent information from metric tags
We are currently including a lot of high cardinality user agent tags with the metrics we emit from syncstorage and tokenserver. InfluxDB doesn’t support tags with high cardinality, so we’ll need to remove these. We should also explore improving the Metrics and Tags interfaces to make it harder to include high cardinality accidentally and easier to include tags/extras through the request lifecycle.

[Jira](https://mozilla-hub.atlassian.net/browse/SYNC-3463)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-2081)

ethowitz:
## Description

The primary purpose of this change is to reduce the cardinality of the tags we're including with InfluxDB metrics by removing user agent data. I also wanted to make it more ergonomic to use (and more difficult to misuse) tags in general, so I made a few other improvements to our tagging code:
* Previously, there were several ways to construct a `Metrics` object, including ways that didn't involve deriving the object from the `HttpRequest`. This is bad because a `Metrics` that isn't derived from `HttpRequest` won't include any tags that were persisted to the request extensions. I removed the trait implementations and functions that enabled this; `FromRequest` is now the only way to construct a `Metrics` (without building a new client altogether using a host and port). This means that, if a tag is added to the request extensions, it will necessarily be included in all subsequent metric emissions in the request lifecycle
* The `Tags` type used to represent the tags we were including in Sentry errors **as well as** the tags we were passing in to `Metrics` methods (even though `Tags` included high-cardinality data, which metrics shouldn't support). It seemed like `Tags`'s scope was a little overloaded and its interface made it easy to accidentally include high-cardinality data with metrics data. I decided to simplify things by removing `Tags` altogether and adding a `Taggable` trait, which is implemented for all `T: HttpMessage`. It includes `add_tag` and `add_extra` methods, so that adding a tag or extra to the request extensions is simplified (i.e. it just involves calling `req.add_tag(key, value)`). I also added doc comments to highlight the important difference in cardinality between tags and extras
* All tags in the request extensions will be included in both Sentry errors and metrics; extras will only be included in Sentry errors

## Issue(s)

Closes #1436

### DIFF
--- a/syncserver/src/server/metrics.rs
+++ b/syncserver/src/server/metrics.rs
@@ -1,44 +1,45 @@
+use std::collections::HashMap;
 use std::net::UdpSocket;
 use std::time::Instant;
 
-use actix_web::{
-    dev::Payload, error::ErrorInternalServerError, web::Data, Error, FromRequest, HttpRequest,
-};
+use actix_web::{dev::Payload, web::Data, FromRequest, HttpRequest};
 use cadence::{
     BufferedUdpMetricSink, Counted, Metric, NopMetricSink, QueuingMetricSink, StatsdClient, Timed,
 };
 use futures::future;
 use futures::future::Ready;
+use slog::{Key, Record, KV};
 
 use crate::error::ApiError;
 use crate::server::ServerState;
-use crate::web::tags::Tags;
+use crate::tokenserver;
+use crate::web::tags::Taggable;
 
 #[derive(Debug, Clone)]
 pub struct MetricTimer {
     pub label: String,
     pub start: Instant,
-    pub tags: Tags,
+    pub tags: HashMap<String, String>,
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Default, Clone)]
 pub struct Metrics {
     pub client: Option<StatsdClient>,
-    pub tags: Option<Tags>,
+    pub tags: HashMap<String, String>,
     pub timer: Option<MetricTimer>,
 }
 
 impl Drop for Metrics {
     fn drop(&mut self) {
-        let tags = self.tags.clone().unwrap_or_default();
+        let tags = self.tags.clone();
         if let Some(client) = self.client.as_ref() {
             if let Some(timer) = self.timer.as_ref() {
                 let lapse = (Instant::now() - timer.start).as_millis() as u64;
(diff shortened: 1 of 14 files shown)


syncserver/src/server/metrics.rs
```diff
@@ -59,39 +60,38 @@ impl Drop for Metrics {
...
-    }
+        let client = {
+            let syncstorage_metrics = req
+                .app_data::<Data<ServerState>>()
+                .map(|state| state.metrics.clone());
+            let tokenserver_metrics = req
+                .app_data::<Data<tokenserver::ServerState>>()
+                .map(|state| state.metrics.clone());
+
+            syncstorage_metrics.or(tokenserver_metrics)
```
ethowitz, line 75: This lets the extractor work for syncstorage _or_ tokenserver

syncserver/src/tokenserver/extractors.rs
```diff
@@ -455,9 +451,12 @@ impl FromRequest for AuthData {
...
                 Token::BrowserIdAssertion(assertion) => {
-                    let mut tags = Tags::default();
-                    tags.tags
-                        .insert("token_type".to_owned(), "BrowserID".to_owned());
+                    // Add a tag to the request extensions
+                    req.add_tag("token_type".to_owned(), "BrowserID".to_owned());
+
+                    // Start a timer with the same tag
+                    let mut tags = HashMap::default();
+                    tags.insert("token_type".to_owned(), "BrowserID".to_owned());
```
jrconlin, line 459: duplicating the tag feels a bit awkward, but not quite sure how to resolve it. maybe add an `extend()` in a future PR?

jrconlin (dismissed):

jrconlin:
I really like turning Tags into a trait. It tidies a lot of things up. Thanks!

ethowitz:
Ugh, I have to figure out why some of these commits aren't verified...

ethowitz:
@jrconlin Okay sorry, just squashed my commits and made sure the commit on this branch is signed properly, but I'll need one more ✅

jrconlin (approved):

PR #1439, opened 2022-11-01
chore: tag 0.12.5

pjenvey:


syncstorage-common/Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[package]
```
ethowitz, line 1: Do we want this file here? The setting crates PR removed the syncstorage-common crate

pjenvey: Thanks, good catch! I thought this was deleted, merge fail.

ethowitz:
When this PR is merged, will be it be squashed into one commit? If so, that squashed commit would include the MySQL client library update -- is that intended?

pjenvey:
> When this PR is merged, will be it be squashed into one commit? If so, that squashed commit would include the MySQL client library update -- is that intended?

I temporarily enable merge commits for a branch like this (I really don't want them off 100% I just don't want it to by the default for merging PRs)

ethowitz (approved):

PR #1441, opened 2022-11-18
chore: upgrade to Rust 1.65

ethowitz:


tokenserver-common/src/error.rs
```diff
@@ -19,7 +19,7 @@ pub struct TokenserverError {
     /// For internal use only. Used to report any additional context behind an error to
     /// distinguish between similar errors in Sentry.
     pub context: String,
-    pub backtrace: Backtrace,
+    pub backtrace: Box<Backtrace>,
```
ethowitz, line 22: The [`result_large_err`](https://rust-lang.github.io/rust-clippy/master/#result_large_err) clippy lint was saying that `TokenserverError` was too large. `Box`ing the backtrace field resolved the issue, but I don't quite understand why. This is the definition of `Backtrace`:
```rust
pub struct Backtrace {
    // Frames here are listed from top-to-bottom of the stack
    frames: Vec<BacktraceFrame>,
    // The index we believe is the actual start of the backtrace, omitting
    // frames like `Backtrace::new` and `backtrace::trace`.
    actual_start_index: usize,
}
```
`Vec`s are stored on the heap and `usize`s are trivially small, so 🤷🏻 it doesn't seem that big to me

pjenvey (approved):

tokenserver-common/src/error.rs
ethowitz: Ah, I should have read through the rest of the lint's documentation

PR #1447, opened 2022-12-16
bug: fix CORS issue

ethowitz:
## Description

I must have made a mistake during a rebase -- https://github.com/mozilla-services/syncstorage-rs/commit/221705b7ea74c6dddffd1c5289c53b3ad2cc7522 wasn't properly applied to the settings crate reorg branch, and we weren't allowing all origins. This PR should resolve it

### DIFF
--- a/syncserver-settings/src/lib.rs
+++ b/syncserver-settings/src/lib.rs
@@ -176,7 +176,7 @@ impl Default for Settings {
             statsd_host: Some("localhost".to_owned()),
             statsd_port: 8125,
             human_logs: false,
-            cors_allowed_origin: None,
+            cors_allowed_origin: Some("*".to_owned()),
             cors_allowed_methods: Some(
                 ["DELETE", "GET", "POST", "PUT"]
                     .into_iter()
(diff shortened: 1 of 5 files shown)


syncserver-settings/src/lib.rs
```diff
@@ -176,7 +176,7 @@ impl Default for Settings {
             statsd_host: Some("localhost".to_owned()),
             statsd_port: 8125,
             human_logs: false,
-            cors_allowed_origin: None,
```
jrconlin, line 179: Ah, yeah, I think actix changed this default a while ago. `None` now means None, not all.

Sorry, should have caught that.

jrconlin (dismissed):

jrconlin (approved):

pjenvey (commented):
This has the old incorrect spelling of 'UserAgent' as a default.

Let's switch:

```
                    "Authorization",
                    "Content-Type",
                    "UserAgent",
```
to the constants from actix-web (`use actix_web::http::header::{AUTHORIZATION, CONTENT_TYPE, USER_AGENT};`)
-> 

```
                AUTHORIZATION,
                CONTENT_TYPE,
                USER_AGENT,
```

ethowitz:
@pjenvey Do you think it's worth including Actix as a dependency for this crate just for the header definitions? Either way, I'll fix that misspelling

pjenvey:
Ah -- no, it doesn't seem worth it just for those constants.

ethowitz:
I also updated the docker resource class here. We wanted to do it anyway, and it'll help get this fix out a little faster :)

pjenvey (approved):

PR #1449, opened 2022-12-21
Fix database_url key in config template

brevilo:
## Description

* Regression in #1306
* `database_url` got moved into (and is now expected in) a new `syncstorage` parent

To quote the original PR description:

    Because of the naming changes in this PR, the names of environment variables and settings in local.toml will need to change. 

Yet that exact change wasn't done as part of the PR. This is particularly important as the README uses the example config as a template when setting up a new server.

## Testing
1. Use the original config as a template and your configured `database_url` won't be used at all
2. Merge the patch and test again
3. Works


brevilo:
🤷‍♂️

PR #1451, opened 2023-01-06
chore: updates for Rust 1.66

ethowitz:


syncserver/src/db/mysql/models.rs
```diff
@@ -489,7 +489,7 @@ impl MysqlDb {
                 bso::expiry,
             ))
             .filter(bso::user_id.eq(user_id))
-            .filter(bso::collection_id.eq(collection_id as i32)) // XXX:
+            .filter(bso::collection_id.eq(collection_id)) // XXX:
```
pjenvey, line 492: Can you kill these while you're here, they were about the casting.

```suggestion
            .filter(bso::collection_id.eq(collection_id))
```

syncserver/src/db/mysql/models.rs
```diff
@@ -572,7 +572,7 @@ impl MysqlDb {
         let mut query = bso::table
             .select(bso::id)
             .filter(bso::user_id.eq(user_id))
-            .filter(bso::collection_id.eq(collection_id as i32)) // XXX:
+            .filter(bso::collection_id.eq(collection_id)) // XXX:
```
pjenvey, line 575:
```suggestion
            .filter(bso::collection_id.eq(collection_id))
```

pjenvey (changes requested):

pjenvey (approved):

PR #1452, opened 2023-01-13
DRAFT: refactor: add web crates

ethowitz:
A draft PR for the addition of web-related crates to syncstorage. It is in decent shape (and is _mostly_ rebased with the main branch), but it doesn't completely compile at this point. I think a good place to start would be to deal with the compile errors. Then, the main unfinished portion is the dockerflow endpoint registration code.


syncserver/src/lib.rs
```diff
@@ -3,12 +3,243 @@
...
+            .wrap(middleware::rejectua::RejectUA::<syncstorage_web::ServerState>::default())
+            // Followed by the "official middleware" so they run first.
+            // actix is getting increasingly tighter about CORS headers. Our server is
+            // not a huge risk but does deliver XHR JSON content.
+            // For now, let's be permissive and use NGINX (the wrapping server)
+            // for finer grained specification.
+            .wrap(build_cors($settings))
+            .wrap(syncstorage_web::middleware::EmitTokenserverOrigin::default())
+            .configure(
+                syncstorage_web::get_configurator(
```
ethowitz, line 54: The idea here is that `syncstorage-web` and `tokenserver-web` will each have their own `get_configurator` function, which is responsible for registering the routes specific to each of the two services. The `syncserver` crate doesn't need to know anything about the particular implementations in order to make use of them; `syncserver`'s role is to decide whether to include Syncstorage, Tokenserver, or both, depending on what cargo features are set

syncserver/src/lib.rs
```diff
@@ -3,12 +3,243 @@
...
+
+#[cfg(all(feature = "syncstorage", not(feature = "tokenserver")))]
+#[macro_export]
+macro_rules! build_app {
+    ($settings: expr) => {
+        App::new()
+            .data(Arc::new($settings.master_secret.clone()))
+            // Middleware is applied LIFO
+            // These will wrap all outbound responses with matching status codes.
+            .wrap(ErrorHandlers::new().handler(
```
ethowitz, line 35: These middleware registrations should be replaced with the new `wrap_fn` paradigm (we already do this on the main branch, I just didn't get a chance to update it here)

syncserver/src/lib.rs
```diff
@@ -3,12 +3,243 @@
...
+}
+
+#[cfg(all(feature = "syncstorage", feature = "tokenserver"))]
+#[macro_export]
+macro_rules! build_app {
+    ($settings: expr, $blocking_threadpool: expr) => {
+        App::new()
+            .data(Arc::new($settings.master_secret.clone()))
+            .service(
+                web::scope("/token")
```
ethowitz, line 104: I include the same prefixes that exist for syncserver here https://github.com/mozilla-services/syncserver/blob/c116417d653432db9470098236fb117adb349c36/syncserver/__init__.py#L158-L159

syncserver/src/lib.rs
```diff
@@ -3,12 +3,243 @@
...
+/// This function configures the following endpoints:
+/// * GET /__error__
+/// * GET /__version__
+/// * If syncstorage and tokenserver are both enabled, `GET __heartbeat__`, since we need to
+///   combine the `GET __heartbeat__` endpoints configured by syncstorage and tokenserver
+///   in their `get_configurator` functions
+///
+/// Any other dockerflow endpoints are configured individually by syncstorage and tokenserver.
+#[cfg(not(all(feature = "syncstorage", feature = "tokenserver")))]
+fn configure_dockerflow(cfg: &mut ServiceConfig) {
```
ethowitz, line 154: The dockerflow endpoints that are general enough to be defined here are defined here

syncserver/src/middleware/sentry.rs
```diff
@@ -76,9 +72,9 @@ pub fn report_error(
             Some(e) => {
                 let metrics = MetricsWrapper::extract(sresp.request()).await.unwrap().0;
 
-                if let Some(apie) = e.as_error::<ApiError>() {
+                if let Some(apie) = e.as_error::<syncstorage_web::ApiError>() {
```
ethowitz, line 75: We may want to revisit this approach? At the very least, we should double check that this is working as expected on stage before we deploy to prod

syncstorage-web/src/metrics.rs
```diff
@@ -0,0 +1,30 @@
+use actix_web::{dev::Payload, web::Data, FromRequest, HttpRequest};
+use futures::future;
+use futures::future::Ready;
+use syncserver_common::{Metrics, Taggable};
+
+use super::ServerState;
+
+pub struct MetricsWrapper(pub Metrics);
```
ethowitz, line 8: This serves the same function as `TokenserverMetrics` (it's a newtype that lets us define `FromRequest` on `Metrics`). We should probably be consistent and:
1. Rename it to `SyncstorageMetrics`; and
2. Move it to the extractors file

tokenserver-web/src/error.rs
```diff
@@ -0,0 +1,116 @@
...
+    ser::{SerializeMap, Serializer},
+    Serialize,
+};
+use syncserver_common::ReportableError;
+use thiserror::Error;
+use tokenserver_common::{ErrorLocation, TokenserverError};
+use tokenserver_db::DbError;
+
+#[derive(Debug, Error)]
+pub struct ApiError(TokenserverError);
```
ethowitz, line 14: I define a top-level `ApiError` that mirrors Syncstorage's `ApiError` for consistency. I chose to do this rather than just rename `TokenserverError` to `ApiError` because `TokenserverError` has to live in `tokenserver_common`, which doesn't include Actix, and we need to define `ResponseError` for `ApiError`

tools/integration_tests/run.py
```diff
@@ -13,39 +14,69 @@
 DEBUG_BUILD = "target/debug/syncserver"
 RELEASE_BUILD = "/app/bin/syncserver"
 
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description="Run the syncstorage and tokenserver integration tests"
+    )
+    parser.add_argument(
```
ethowitz, line 21: I began the process of updating this file to allow us to run different test suites using command line flags. We'll need to do this in order to selectively run test suites in our CI pipelines for Syncstorage and Tokenserver

jrconlin:
I recommend closing this branch since it's over a year out of date and is causing too many conflicts. 
It may be worth keeping the goal of this PR in mind when working on syncstorage, and consider moving the web components to their own crate, but that should be left to the engineer handling the tickets.

ethowitz:
I am inclined to agree :)

Issue #1051, opened 2021-04-26
Tokenserver: Add node assignment logic


ethowitz:
We'll need to translate the node assignment logic from the old Tokenserver. Most of that logic lives in [this file](https://github.com/mozilla-services/tokenserver/blob/0fbf534dbe305f83f8cb43f8090c9dd4fde78805/tokenserver/assignment/sqlnode/sql.py). As part of this work, we'll need to handle changes to the `cilent_state` field in the request.

jrconlin:
[Jira ticket](https://jira.mozilla.com/browse/CONSVC-160)

jackyzy823:
Something for self-hosting:
1.  For self-hosting , old syncserver implement  use [staticnode assignment](https://github.com/mozilla-services/syncserver/blob/master/syncserver/staticnode.py) . which results different database structure of `users` . So either staticnode assignment or a db upgrade script needs to be implemented for self-hosting.

2. for self-hosting, old syncserver implement mounts tokenserver and server-syncstorage under two routes: `/token/` and `/storage/` , for `syncstorage-rs` should we implement this for self-hosting to avoid user change `identity.sync.tokenserver.uri` in `about:config`

ethowitz:
Hey @jackyzy823, the work to make syncstorage-rs self-hostable will be done as part of a separate project. But to your points:
1. For simplicity's sake, I'm thinking it might make sense to use the same node assignment logic and database schema for self-hosting. Rather than add separate static node-assignment logic, we could just have a single node in the database to which users would always be assigned (in which case you’re right that we’d need to provide a script to migrate the database to the new format)
2. When we complete the work to make syncstorage-rs self-hostable, I expect we'll take an approach similar to the one we used for the old syncstorage. I agree that it would make sense to keep the configuration as consistent as possible so users don't have to do too much work to upgrade.

PR #1453, opened 2023-01-14
Fix the URL people should set to use their custom tokenserver

Ekleog:
I tried running with the recommended `/token/1.0/sync/1.5`. It led to logs of:
```
INFO {"ua.os.family":"Linux","uri.path":"/token/1.0/sync/1.5","ua.name":"Firefox","ua.os.ver":"UNKNOWN","ua.browser.ver":"108.0","ua":"108.0","ua.browser.family":"Firefox","uri.method":"GET"}
```
which returned 404 and failed to sync.

Once fixing my firefox parameter, I now get the following log line:
```
INFO {"ua.browser.family":"Firefox","uid":"[censored]","metrics_uid":"[censored]","ua":"108.0","ua.os.family":"Linux","ua.name":"Firefox","uri.method":"GET","ua.os.ver":"UNKNOWN","first_seen_at":"[censored]","uri.path":"/1.0/sync/1.5","ua.browser.ver":"108.0"}
```
and sync actually works.

I guess the `/token` part of the URL originally came from a reverse-proxy that automatically needs it? Anyway, I can confirm that things work for me without it, and don't with it.

(For full disclosure, I do have a nginx rproxy in front of my syncserver; but it's configured as passthrough, and given the log lines copied above report the expected `uri.path`, I don't think my configuration of it is wrong)

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,7 +200,7 @@ This will walk you through the steps to connect this project to your local copy 
     sync-1.5 = "http://localhost:8000/1.5/1"```
 
 
-3. In Firefox, go to `about:config`. Change `identity.sync.tokenserver.uri` to `http://localhost:5000/token/1.0/sync/1.5`.
+3. In Firefox, go to `about:config`. Change `identity.sync.tokenserver.uri` to `http://localhost:5000/1.0/sync/1.5`
 4. Restart Firefox. Now, try syncing. You should see new BSOs in your local MySQL instance.
 
 ## Logging


jackyzy823:
FYI.

This configuration of url prefix came from the old sync server . The tokenserver's path is mounted under  /token 

I have asked about this question. https://github.com/mozilla-services/syncstorage-rs/issues/1051#issuecomment-924885375.

jrconlin (approved):

jrconlin:
Sorry about the delay. Thanks for this!

pjenvey:
This was applied in #1492, thank you!

Issue #1461, opened 2023-02-14
Convert Cargo.toml files to use `workspace` dependency trees


jrconlin:
Cargo now allows sub-crates to use [workspaces](https://doc.rust-lang.org/cargo/reference/workspaces.html). 
At the least, we should copy the `[package]` values down from a central Cargo.toml file.

e.g. in `syncserver-common` specify

```toml
[package]
name  = "syncserver-common"
version.workspace=true
edition.workspace=true
authors.workspace=true
```

This would ensure that all crates are updated to the latest version when the central `syncstorage-rs/Cargo.toml` file is.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DISCO-2245)

Issue #1362, opened 2022-07-15
Update 'latest' tag on Docker Hub


besendorf:
The 'latest' tag on docker hub has last been updated 2 years ago. Meanwhile there are other tag with newer version. Please update the latest tag so people can use `image: mozilla/syncstorage-rs:latest` in their docker-compose files.

latest: https://hub.docker.com/layers/syncstorage-rs/mozilla/syncstorage-rs/latest/images/sha256-e5501334ec1f6a0f135a46c1c96d1fe4b7f67cd4a76d8acbff524d57a80b13e4?context=explore
0.12.0: https://hub.docker.com/layers/syncstorage-rs/mozilla/syncstorage-rs/0.12.0/images/sha256-8063995f0e3e06c1390ae3a2dee5690528706bfaddbc0e34c0d37b6a17704a6a?context=explore

ictabc:
Any news about this, should be an easy fix?

pjenvey:
#488 appears to be the cause of this change

PR #1462, opened 2023-02-14
Feat/1461 update

jrconlin:
## Description

Update package description and some dependencies to use `workspace`
inheritance.

* Also updates circleci to produce dockerhub image.

## Testing

No functional changes

## Issue(s)

Closes https://github.com/mozilla-services/syncstorage-rs/issues/1461
Issue https://github.com/mozilla-services/syncstorage-rs/issues/1362


syncserver-common/src/metrics.rs
```diff
@@ -1,5 +1,6 @@
 use std::collections::HashMap;
 use std::net::UdpSocket;
+use std::sync::Arc;
```
mhammond, line 3: did functionality changes sneak in here accidentally?

mhammond (commented):
Eeek, sorry I left this hanging, but I missed it entirely! While the workspace changes looks fine, there's also functionality changes here which I can't really review.

syncserver-common/src/metrics.rs
jrconlin: A recent change in StatsdClient changed the way that it is supposed to be called. Previously, you would clone the client, but now, it's a singleton that needs to be reference counted.

pjenvey:
Per https://github.com/mozilla-services/syncstorage-rs/issues/1362#issuecomment-1430543424 the "latest" tag change made to the cirleci config was [originally made by SRE](https://github.com/mozilla-services/syncstorage-rs/pull/488), looping them in as reviewers.

.circleci/config.yml
```diff
@@ -363,7 +363,7 @@ jobs:
           command: |
             export UTILS_DOCKERHUB_REPO=mozilla/sync-spanner-py-utils
             if [ "${CIRCLE_BRANCH}" == "master" ]; then
-              DOCKER_TAG="${CIRCLE_SHA1}"
+              DOCKER_TAG="latest"
```
cvalaas, line 366: could,should we make the tags "latest,${CIRCLE_SHA1}" ?

cvalaas: (I don't know if that's a valid syntax...)

cvalaas: It's not, no. Okay - well, looks like the SHA tag is only used to trigger dev builds, which I don't think is a thing we're using or even looking at, so I'd say we could switch back to latest.

cvalaas (approved):

cvalaas (approved):

PR #1464, opened 2023-03-03
fix: handle nullable (None) keys_changed_at values

pjenvey:
## Description

a regression from the Python 3 conversion (max's changes)

(against release/0.13)

## Issue(s)

SVCSE-3633


jrconlin (approved):
pending flake8

PR #1488, opened 2023-10-10
fix: switch more test flags to cfg(debug_assertions)

pjenvey:
switch mutation usage to an explicit flag

fix the db tests' uid handling

## Issue(s)

Closes SYNC-3955


pjenvey:
We still do not run these in CI but it should be possible, I'm just punting on it for now: https://mozilla-hub.atlassian.net/browse/SYNC-3957

jrconlin (approved):

PR #1491, opened 2023-10-18
feat: add dynamic routing headers to all Spanner ops

### LINKED ISSUE #1490: Add dynamic routing headers to the Spanner client
Per a recent [incident](https://docs.google.com/document/d/1HKim9Ho2aFVK6OM3FKv12JeUPVoiaHDzUzd2K5v8N_E/edit#heading=h.p64b1qfksx4w) we’ve determined our client does not support “dynamic routing”.

This feature appears to consist of an additional gRPC metadata (header) added to certain requests. This header is defined in [AIP-4222 Routing headers](https://google.aip.dev/client-libraries/4222).

Spanner’s current [protobuf definition](https://github.com/googleapis/googleapis/blob/1cfcea4/google/spanner/v1/spanner.proto) includes some usages of the implicit `google.api.http` header. Spanner API calls with such header definitions should result in added `x-goog-request-params` gRPC metadata headers added to their requests.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/SYNC-3922)

pjenvey:
## Description

and support the seemingly undocumented new leader aware routing header recently added to other spanner clients (via a setting, disabled by default)

## Issue(s)

SYNC-3922
Closes #1490

### DIFF
--- a/syncstorage-spanner/src/manager/session.rs
+++ b/syncstorage-spanner/src/manager/session.rs
@@ -72,7 +72,7 @@ pub struct SpannerSessionSettings {
 }
 
 impl SpannerSessionSettings {
-    pub fn from(settings: &Settings) -> Result<Self, DbError> {
+    pub fn from_settings(settings: &Settings) -> Result<Self, DbError> {
         let database = settings
             .spanner_database_name()
             .ok_or_else(|| {
--- a/syncstorage-spanner/src/manager/session.rs
+++ b/syncstorage-spanner/src/manager/session.rs
@@ -4,10 +4,11 @@ use google_cloud_rust_raw::spanner::v1::{
     spanner::{CreateSessionRequest, GetSessionRequest, Session},
     spanner_grpc::SpannerClient,
 };
-use grpcio::{CallOption, ChannelBuilder, ChannelCredentials, Environment, MetadataBuilder};
+use grpcio::{CallOption, ChannelBuilder, ChannelCredentials, Environment};
 use syncserver_common::{BlockingThreadpool, Metrics};
+use syncstorage_settings::Settings;
 
-use crate::error::DbError;
+use crate::{error::DbError, metadata::MetadataBuilder};
 
 const SPANNER_ADDRESS: &str = "spanner.googleapis.com:443";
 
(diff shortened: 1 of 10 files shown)


syncstorage-spanner/src/metadata.rs
```diff
@@ -0,0 +1,129 @@
...
+    pub fn route_to_leader(mut self, route_to_leader: bool) -> Self {
+        self.route_to_leader = route_to_leader;
+        self
+    }
+
+    /// Build the [grpcio::Metadata]
+    pub fn build(self) -> Result<grpcio::Metadata, grpcio::Error> {
+        let mut meta = grpcio::MetadataBuilder::new();
+        meta.add_str(PREFIX_KEY, self.prefix)?;
+        meta.add_str(METRICS_KEY, "gcp-grpc-rs")?;
```
jrconlin, line 64: Let's make this a defined constant (so that we can reuse it in the test below.)

syncstorage-spanner/src/manager/session.rs
```diff
@@ -21,25 +22,103 @@ pub struct SpannerSession {
...
+    pub max_idle: Option<u32>,
+
+    /// For tests: disables transactions from committing
+    pub(crate) use_test_transactions: bool,
+    /// Spanner emulator hostname when set to Spanner emulator mode
+    pub emulator_host: Option<String>,
+}
+
+impl SpannerSessionSettings {
+    pub fn from(settings: &Settings) -> Result<Self, DbError> {
```
jrconlin, line 75: Could this be a `TryFrom`?

jrconlin (changes requested):

syncstorage-spanner/src/metadata.rs
pjenvey: I'd punt on that for now. Ideally we'd have some kind of more fleshed out "user-agent" like e.g. the python has:

https://github.com/googleapis/python-api-core/blob/main/google/api_core/gapic_v1/client_info.py
https://github.com/googleapis/python-api-core/blob/main/google/api_core/client_info.py

Those are in the more generic library (more along the lines of google-cloud-rust). I'm not sure if individual client libs such as python-spanner further customize these headers.

syncstorage-spanner/src/manager/session.rs
pjenvey: Last time we discussed this I came to the conclusion that I'd rather leave the `Try`/`TryFrom` for simpler type conversions that feel a bit more implicit (like `let foo: i64 = 1i32.into();`) and be more explicit for things that are more along the lines of construction or deserialization (like `serde_json::from_str`);

Renamed it `from` -> `from_settings` like we have elsewhere to distinguish it.

jrconlin (approved):

syncstorage-spanner/src/manager/session.rs
pjenvey: or `str::parse` is maybe a better example

jrconlin (approved):

PR #1506, opened 2023-12-07
bug: Use google specified UA for x-goog-api-client

jrconlin:
## Description

Use suggested UA for GRPCIO header. This uses the syncserver crate version. We *may* want to use the version of `google-cloud-rust` instead?

Closes SYNC-4047

### DIFF
--- a/syncstorage-spanner/src/metadata.rs
+++ b/syncstorage-spanner/src/metadata.rs
@@ -1,7 +1,3 @@
-use std::sync::OnceLock;
-
-use google_cloud_rust_raw::VERSION;
-
 /// gRPC metadata Resource prefix header
 ///
 /// Generic across Google APIs. This "improves routing by the backend" as
@@ -31,8 +27,9 @@ const ROUTING_KEY: &str = "x-goog-request-params";
 /// transactions to the leader region." as described by other Spanner clients
 const LEADER_AWARE_KEY: &str = "x-goog-spanner-route-to-leader";
 
-/// Our user agent value for [METRICS_KEY]
-static USER_AGENT: OnceLock<String> = OnceLock::new();
+/// The USER_AGENT string is a static value specified by Google.
+/// It's meaning is not to be known to the uninitiated.
+const USER_AGENT: &str = "gl-external/1.0 gccl/1.0";
 
 /// Builds the [grpcio::Metadata] for all db operations
 #[derive(Default)]
@@ -67,10 +64,9 @@ impl<'a> MetadataBuilder<'a> {
     /// Build the [grpcio::Metadata]
     pub fn build(self) -> Result<grpcio::Metadata, grpcio::Error> {
         let mut meta = grpcio::MetadataBuilder::new();
-        let ua = USER_AGENT.get_or_init(|| format!("gl-external/{VERSION} gccl/{VERSION}"));
 
         meta.add_str(PREFIX_KEY, self.prefix)?;
-        meta.add_str(METRICS_KEY, ua)?;
+        meta.add_str(METRICS_KEY, USER_AGENT)?;
         if self.route_to_leader {
             meta.add_str(LEADER_AWARE_KEY, "true")?;
         }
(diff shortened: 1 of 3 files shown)


syncstorage-spanner/src/metadata.rs
```diff
@@ -28,7 +28,7 @@ const ROUTING_KEY: &str = "x-goog-request-params";
 const LEADER_AWARE_KEY: &str = "x-goog-spanner-route-to-leader";
 
 /// Our user agent value for [METRICS_KEY]
-const USER_AGENT: &str = "gcp-grpc-rs";
+const USER_AGENT: &str = "gl-external/{version} gccl/{version}";
```
pjenvey, line 31: not super critical but we might as well only do this once: we could use `lazy_static` but the new `OnceLock` [e.g.](https://github.com/mozilla-services/autopush-rs/blob/master/autopush-common/src/logging.rs#L13) is probably preferable to it at this point.

Though in practice `OnceLock`'s API is also a bit funny for this (because you either need to move this `format!` looking constant string into a runtime code block or maybe put the entire `OnceLock` into the runtime code block instead, keeping the string template here as const module level).

So maybe concat! is the simplest:

```suggestion
const USER_AGENT: &str = concat!(
    "gl-external/",
    env!("CARGO_PKG_VERSION"),
    " gccl/",
    env!("CARGO_PKG_VERSION")
);
```

jrconlin: Sounds good. I also built https://github.com/mozilla-services/google-cloud-rust/pull/38 which exposes the VERSION of the actual rust library. I may fix up this one to use it instead.

jrconlin: Probably want to hold off merging this until we get an authoritative answer from Google about some of the version number questions.

jrconlin: This is no longer valid (or needed)

syncstorage-spanner/src/metadata.rs
```diff
@@ -27,8 +27,9 @@ const ROUTING_KEY: &str = "x-goog-request-params";
 /// transactions to the leader region." as described by other Spanner clients
 const LEADER_AWARE_KEY: &str = "x-goog-spanner-route-to-leader";
 
-/// Our user agent value for [METRICS_KEY]
-const USER_AGENT: &str = "gcp-grpc-rs";
+/// The USER_AGENT string is a static value specified by Google.
+/// It's meaning is not to be known to the uninitiated.
```
pjenvey, line 31:
```suggestion
/// Its meaning is not to be known to the uninitiated.
```

syncstorage-spanner/src/metadata.rs
```diff
@@ -93,11 +95,12 @@ mod tests {
         MetadataBuilder, LEADER_AWARE_KEY, METRICS_KEY, PREFIX_KEY, ROUTING_KEY, USER_AGENT,
     };
 
-    pub const DB: &str = "/projects/sync/instances/test/databases/sync1";
-    pub const SESSION: &str = "/projects/sync/instances/test/databases/sync1/sessions/f00B4r_quuX";
+    // Resource paths should not start with a "/"
+    pub const DB: &str = "projects/sync/instances/test/databases/sync1";
+    pub const SESSION: &str = "projects/sync/instances/test/databases/sync1/sessions/f00B4r_quuX";
```
pjenvey, line 100: FYI I double checked that we're in fact sending such values without a leading slash, this was just a mistake in the test's data.

pjenvey (changes requested):

syncstorage-spanner/src/metadata.rs
jrconlin: Yep, just wanted to make sure that the test values reflected the correct value, as well as the guidance was captured in the code.

pjenvey (approved):

syncstorage-spanner/src/metadata.rs
pjenvey: :100:

PR #1513, opened 2024-01-17
feat: Removes pyo3 and derives tokens directly in Rust

tarikeshaq:
## Description
We were using pyo3 in two places to call into Python code:
1. When verifying an FxA oauth access token, calling into [PyFxA](https://github.com/mozilla/PyFxA)'s [verify_token](https://github.com/mozilla/PyFxA/blob/6c3f803b3c27c665f417b0c5bd3ca79add8e2027/fxa/oauth.py#L257) function - to verify an oauth token, which would first try to validate the token directly using a JWK stored in our environment variables, or otherwise it would trigger a `/v1/verify/` call on FxA's oauth server, asking FxA to verify for us
2. When generating a new Token to be used against storage server, we called into [tokenlib](https://github.com/mozilla-services/tokenlib)'s [`make_token`](https://github.com/mozilla-services/tokenlib/blob/91ec9e2c922e55306eddba1394590a88f3b10602/tokenlib/__init__.py#L96) and [`get_derived_secret`](https://github.com/mozilla-services/tokenlib/blob/91ec9e2c922e55306eddba1394590a88f3b10602/tokenlib/__init__.py#L158)

This PR replaces both usages of pyo3 with Rust code implemented directly in the tokenserver.

## Testing
I've tested a local firefox connected to FxA's stage, and my local environment with tokenserver and syncstorage and verified that:
- Tokenserver can verify tokens properly if given the JWK's in its env variables
- Tokenserver can send the network request properly to FxA when it cannot verify the token locally (think FxA rotated their JWKs)
- Tokenserver can issue tokens, and those tokens can be used correctly for sync against syncstorage

My local firefox syncs successfully with the changes in this PR, however, **Out of abundance of caution we should run end-to-end tests and deploy this first on a canary if we plan to take it**

## TODOs
- [x] Add unit tests for both `verify` and `get_token_and_derived_secret`
   - [x] verify
   - [x] token generation
- [x] Document the hidden policies here so they're not hidden, eg: how the tokenserver tokens are derived, and how it validates the FxA access token, this was documented in the code for now as doc strings on a public API

## Issue(s)

Closes [SYNC-3528](https://mozilla-hub.atlassian.net/browse/SYNC-3528).

[SYNC-3528]: https://mozilla-hub.atlassian.net/browse/SYNC-3528?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Dockerfile
```diff
@@ -38,7 +38,6 @@ RUN \
     # https://dev.mysql.com/doc/mysql-apt-repo-quick-guide/en/#repo-qg-apt-repo-manual-setup
     apt-get -q update && \
     apt-get -q install -y --no-install-recommends libmysqlclient-dev cmake golang-go python3-dev python3-pip python3-setuptools python3-wheel && \
-    pip3 install -r requirements.txt && \
```
tarikeshaq, line 41: I'm not sure about the changes to the dockerfile here, so someone more familiar with the whole setup should take a look!

tokenserver-auth/src/crypto.rs
```diff
@@ -0,0 +1,37 @@
...
+            .map_err(|_| TokenserverError::internal_error())?;
+        Ok(okm.to_vec())
+    }
+
+    fn hmac_sign(&self, key: &[u8], payload: &[u8]) -> Result<Vec<u8>, Self::Error> {
+        let mut mac: Hmac<Sha256> =
+            Hmac::new_from_slice(key).map_err(|_| TokenserverError::internal_error())?;
+        mac.update(payload);
+        Ok(mac.finalize().into_bytes().to_vec())
+    }
```
tarikeshaq, line 36: This will require a careful review since they're trying to do crypto!

tokenserver-auth/src/lib.rs
```diff
@@ -75,36 +57,37 @@ impl Tokenlib {
...
+        token_bytes.extend_from_slice(&signature);
+        let token = base64::engine::general_purpose::URL_SAFE.encode(token_bytes);
+        // Now that we finialized the token, lets generate our per token secret
+        let mut info = Vec::with_capacity(HKDF_INFO_DERIVE.len() + token.as_bytes().len());
+        info.extend_from_slice(HKDF_INFO_DERIVE);
+        info.extend_from_slice(token.as_bytes());
+
+        let per_token_secret = crypto_lib.hkdf(shared_secret, Some(salt.as_bytes()), &info)?;
+        let per_token_secret = base64::engine::general_purpose::URL_SAFE.encode(per_token_secret);
+        Ok((token, per_token_secret))
```
tarikeshaq, line 90: Ditto a careful review of this, this is a fairly mechanical translation from tokenlib, but is still calling into crypto bits

tokenserver-auth/src/oauth.rs
```diff
@@ -27,69 +39,97 @@ pub struct VerifyOutput {
...
+        let url = Url::parse(&self.server_url).map_err(|_| TokenserverError::internal_error())?;
+        let url = url
+            .join("v1/jwks")
+            .map_err(|_| TokenserverError::internal_error())?;
+        client
+            .get(url)
+            .send()
+            .await
+            .map_err(|_| TokenserverError::internal_error())?
+            .json::<Vec<Jwk>>()
```
tarikeshaq, line 130: note to self, this is not right, the response is a json in the shape:
```json
{
  "keys": []
}
```

Cargo.toml
```diff
@@ -38,7 +38,9 @@ docopt = "1.1"
 env_logger = "0.10"
 futures = { version = "0.3", features = ["compat"] }
 hex = "0.4"
+hmac = "0.12"
 http = "0.2"
+jsonwebtoken = "9.2"
```
tarikeshaq, line 43: Note I'm introducing a dependency here, jsonwebtoken is widely used, with over 14 Million downloads (https://crates.io/crates/jsonwebtoken).

However, jsonwebtoken does pull in [`ring`](https://crates.io/crates/ring/) directly. That might be okay, because rustls is already in our dependency tree and it depends on `ring` for crypto but wanted to make sure this is called out explicitly. There are other options to validate jwks natively in Rust if we go another direction although might be a tad bit more painful to implement

Cargo.lock
```diff
@@ -1789,6 +1785,21 @@ dependencies = [
...
+[[package]]
+name = "jsonwebtoken"
+version = "9.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5c7ea04a7c5c055c175f189b6dc6ba036fd62306b58c66c9f6389036c503a3f4"
+dependencies = [
+ "base64 0.21.5",
+ "js-sys",
+ "pem",
+ "ring 0.17.7",
```
tarikeshaq, line 1797: note to self that jsonwebtoken's requirment for ring and serde is not the same as what's already in here, might need some version bumping somewhere

tools/integration_tests/tokenserver/test_e2e.py
```diff
@@ -215,14 +215,14 @@ def test_valid_oauth_request(self):
...
+        payload_dict = json.loads(payload_str)
         # The `id` payload should include a field indicating the origin of the
         # token
         self.assertEqual(payload_dict['tokenserver_origin'], 'rust')
         signing_secret = self.TOKEN_SIGNING_SECRET
-        expected_token = tokenlib.make_token(payload_dict,
-                                             secret=signing_secret)
-        expected_signature = urlsafe_b64decode(expected_token)[-32:]
+        tm = tokenlib.TokenManager(secret=signing_secret)
+        expected_signature = tm._get_signature(payload_str.encode('utf8'))
```
tarikeshaq, line 225: This was fun to track down, the `tokenlib.make_token` serializes its input to json using [`json.dumps`](https://github.com/mozilla-services/tokenlib/blob/91ec9e2c922e55306eddba1394590a88f3b10602/tokenlib/__init__.py#L108).

Turns out `json.dumps` keeps some spaces in the serialized json, so if `serde_json` would have spit out `{"foo":"bar"}`, `json.dumps` dumps `{"foo": "bar"}` note the extra space!

So the extra spaces end up making it so that my implementation produces different signatures!

tarikeshaq:
Alrighty this should be in a good stage for a first-round review 

@jrconlin @pjenvey I know there is a lot going in the autopush world, so no rush getting to this - it is **not** urgent and I wouldn't want to interrupt your flow over there but I'll keep my eye on the patch for when ya'll get the time

tokenserver-auth/src/lib.rs
tarikeshaq: Also note to self to add some tests with test vectors that verify that a token generated using this function, can have its signature verified by any party that also has the signing secret

tokenserver-auth/src/oauth.rs
```diff
@@ -100,69 +213,28 @@ impl VerifyToken for Verifier {
     /// Verifies an OAuth token. Returns `VerifyOutput` for valid tokens and a `TokenserverError`
     /// for invalid tokens.
     async fn verify(&self, token: String) -> Result<VerifyOutput, TokenserverError> {
```
tarikeshaq, line 215: Ditto for myself to test with test vectors here that generate a key pair, generate an access token JWT using the private key, then sets the public key as the `Verifier`'s `jwks` - the tests should cover all of:
- Invalid signatures from all the jwks is an invalid credentials error
- No keys, or improperly formatted token falls back to triggering requests to FxA
   - We trust FxA's word on it if it returns valid claims
   - We also return an invalid credential error if any error is returned from the FxA request
- Expired tokens are an invalid credential error right away
- If we don't have any keys in our environment, we call FxA's `v1/jwks` to get keys
- Test the OK case of verifying the token against the right JWKs
- Make sure that even if the tokens are OK, if the claims don't include the sync scope we also return an incorrect credentials error

tokenserver-auth/src/lib.rs
```diff
@@ -1,17 +1,21 @@
...
-use pyo3::{
-    prelude::{IntoPy, PyErr, PyModule, PyObject, Python},
-    types::IntoPyDict,
-};
+use rand::RngCore;
 use serde::{Deserialize, Serialize};
 use tokenserver_common::TokenserverError;
 
+const HKDF_SIGNING_INFO: &[u8] = b"services.mozilla.com/tokenlib/v1/signing";
+const HKDF_INFO_DERIVE: &[u8] = b"services.mozilla.com/tokenlib/v1/derive/";
```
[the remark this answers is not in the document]
jrconlin: Can you add that comment to the code before the `const` declarations? That would be useful for folk in the future.

tokenserver-auth/src/lib.rs
jrconlin: I'm not sure how feasible it might be, but one of the things we did with Autopush was take a value generated using the old python code and see if we can match that value using rust. (e.g. you could generate a signature using a set of known values using the existing python and validate the rust by producing a matching signature, using the same values as inputs.)

tokenserver-auth/src/crypto.rs
```diff
@@ -0,0 +1,135 @@
...
+            Hmac::new_from_slice(key).map_err(|_| TokenserverError::internal_error())?;
+        mac.update(payload);
+        mac.verify_slice(signature)
+            .map_err(|_| TokenserverError::internal_error())?;
+        Ok(())
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum OAuthVerifyError {
```
jrconlin, line 53: Should we add a `ReportableError` trait to this so that we can get metrics? 

Looks like this gets eaten by `VerifyToken::verify()`, but might be good to track a few errors to catch any anomalies. (e.g. `InvalidKey` or `DecodingError`)

tarikeshaq: That's a great idea! One thing I wanted to do is track why tokenserver is seemingly hitting FxA often, and tracking those errors will get us exactly there - I'll give adding that trait a shot

tokenserver-auth/src/lib.rs
tarikeshaq: That's a great suggestion, I think that's what the end-to-end tests were trying to do in https://github.com/mozilla-services/syncstorage-rs/pull/1513#discussion_r1456756218

The signatures didn't match at first, but I debugged it was because the python library's json serializer keeps extra spaces that `serde_json` doesn't 😬 Calling the python library's signature function does create matching signatures and the end-to-end tests pass

Cargo.lock
tarikeshaq: Closing the loop here, looks like when we bumped our version on actix, we pulled ring 0.17 - and the only dependant of ring 0.16, which keeps it in our dependency tree is `hawk`. Looking at `hawk`'s main, [they already have the right version of `ring` on their main branch](https://github.com/taskcluster/rust-hawk/blob/41f435cbaf31994f1f655dcefa7434079d221fa4/Cargo.toml#L29) so they just need to cut a release to get rid of the duplicate dependency

tarikeshaq: FWIW I asked the taskcluster folks if they could cut a new release for `hawk` so we can eliminate that dependency

tarikeshaq: https://github.com/mozilla-services/syncstorage-rs/pull/1516 fixes this!

jrconlin (commented):
Still going through the tests for oauth.rs

ethowitz:
Just stopping by to say this is awesome and I'm glad it's happening 🎉

tarikeshaq:
This is why I love working in the open - Great to see you @ethowitz!!

tokenserver-auth/src/oauth.rs
```diff
@@ -24,145 +50,485 @@ pub struct VerifyOutput {
...
+        let url = self
+            .server_url
+            .join("v1/verify")
+            .map_err(|_| TokenserverError::internal_error())?;
+        Ok(client
+            .post(url)
+            .json(&VerifyRequest { token })
+            .send()
+            .await
+            .map_err(|_| TokenserverError::invalid_credentials("Unauthorized".to_string()))
```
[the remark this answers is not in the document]
tarikeshaq: Yeah, that's a great call out - both here and the other errors we were eating up! I added context to the tokenserver errors, which if I understand how this is setup correctly, will make it way to sentry or logs depending on the type of the error

jrconlin:
Part of me would love to put the python replacement stuff behind a feature flag, maybe break the functions out of  `tokenserver-auth/lib.rs` and into either `lib-py.rs` or `lib-rs.rs` (same with `oauth.rs`) and bring them in via a flag? That would let us land this code but keep it out of production use while it gets reviewed. 

Fortunately a _LOT_ of this PR is general cleanup.

tokenserver-auth/src/crypto.rs
```diff
@@ -0,0 +1,152 @@
...
+    fn hmac_verify(&self, key: &[u8], payload: &[u8], signature: &[u8]) -> Result<(), Self::Error>;
+}
+
+/// An implementation for the needed cryptographic using
+///    the hmac crate for hmac and hkdf crate for hkdf
+pub struct CryptoImpl {}
+
+impl Crypto for CryptoImpl {
+    type Error = TokenserverError;
+    fn hkdf(&self, secret: &str, salt: Option<&[u8]>, info: &[u8]) -> Result<Vec<u8>, Self::Error> {
```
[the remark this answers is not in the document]
tarikeshaq: Yeah the difference is the hawk function we're looking at is generating the Hmac key, which happens (in `hmac_sign` for us, and unlike hawk we hard-code the algorithm to be sha256, although we could expand the API if the use case comes!)

We end up using a salt here in the HKDF for the per-token secret, using the salt that is randomly generated and inserted into the token. That all said, a lot of this is out of my comfort zone and I tried to keep it true to what the python was doing and not re-invent anything

tarikeshaq:
Thank you @jrconlin!! I like the suggestion of using rust features for this so it's safer (also easier to rollback and canary, etc etc)

I went ahead and moved the python implementation behind a rust feature, and removing that feature will have the Rust implementation

tarikeshaq:
(the ci failures are unrelated and should be fixed in #1518 )

jrconlin (dismissed):
Oh, awesome! You were able to break the python bit into it's own feature.

jrconlin (approved):

PR #1514, opened 2024-01-18
Upgrading to Actix 4.x

### LINKED ISSUE #1249: Upgrade to Actix 4
Now that Actix 4 has been released, we should upgrade syncstorage-rs to the latest version. [Here](https://docs.google.com/document/d/18_ZxdDhUzjUUn9vHUVg5E8pF46vq7s_-Y-8V5thsYqA/edit#) is a list of the changes that are required for the upgrade. As part of this change, we should make use of Actix’s [Scopes](https://docs.rs/actix-web/3.3.3/actix_web/struct.Scope.html) to have `syncstorage-web` and `tokenserver-web` set their own middleware in `*::get_configurator`. `syncserver` should only be responsible for middleware that is general to both syncstorage and tokenserver. This change would allow us to collapse the three `build_app!` variants into a single macro.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1699)

### LINKED ISSUE #1258: Revisit ignored `cargo audit` warnings
We've ignored a number of the warnings emitted by `cargo audit` due to an inability to upgrade the necessary dependencies. After the upgrade to Actix 4, we'll have updated many of our dependencies and should revisit the ignored warnings to see if they are still necessary.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1706)

bendk:
## Description

Updating dependencies:
  - actix-web 4.x
  - actix-http 3.x
  - actix-rt 2.x
  - actix-cors 0.6.x
  - tokio 1.x
  - reqwest 0.11.x
  - deadpool 0.7.x

The unit tests are passing for me, but I'm not sure if this is ready to merge yet.  One reason I'm making a PR is that I think it will run the integration tests, which I gave up on getting to work on my machine.

I'm not sure how to land these changes.  I think some of these changes can be factored out and landed without upgrading any dependencies, I'll try to open a PR for that.  Other than that, I'm not sure if this can be split apart more.  I don't think it makes sense to upgrade any one of the dependencies without upgrading the others since that will result it duplicate crates getting compiled in (for example, the tests fail if you don't upgrade `reqwest` because the tokio runtimes get mixed up).

## Issue(s)

Closes #1249
Closes #1258

### DIFF
--- a/syncserver/src/server/mod.rs
+++ b/syncserver/src/server/mod.rs
@@ -1,13 +1,13 @@
 //! Main application server
 
-use std::{env, sync::Arc, time::Duration};
+use std::{convert::Infallible, env, sync::Arc, time::Duration};
 
 use actix_cors::Cors;
 use actix_web::{
     dev::{self, Payload},
     http::StatusCode,
     http::{header::LOCATION, Method},
-    middleware::errhandlers::ErrorHandlers,
+    middleware::ErrorHandlers,
     web::{self, Data},
     App, FromRequest, HttpRequest, HttpResponse, HttpServer,
 };
@@ -75,14 +75,14 @@ macro_rules! build_app {
     ($syncstorage_state: expr, $tokenserver_state: expr, $secrets: expr, $limits: expr, $cors: expr) => {
         App::new()
             .configure(|cfg| {
-                cfg.data($syncstorage_state);
+                cfg.app_data(Data::new($syncstorage_state));
 
                 if let Some(tokenserver_state) = $tokenserver_state {
                     let state = tokenserver_state.clone();
-                    cfg.data(state);
+                    cfg.app_data(Data::new(state));
                 }
             })
-            .data($secrets)
+            .app_data(Data::new($secrets))
             // Middleware is applied LIFO
             // These will wrap all outbound responses with matching status codes.
             .wrap(ErrorHandlers::new().handler(StatusCode::NOT_FOUND, ApiError::render_404))
@@ -165,7 +165,7 @@ macro_rules! build_app {
                                        */
             )))
             .service(
-                web::resource("/__version__").route(web::get().to(|_: HttpRequest| {
+                web::resource("/__version__").route(web::get().to(|_: HttpRequest| async {
                     // return the contents of the version.json file created by circleci
                     // and stored in the docker root
                     HttpResponse::Ok()
@@ -174,11 +174,13 @@ macro_rules! build_app {
                 })),
             )
             .service(web::resource("/__error__").route(web::get().to(handlers::test_error)))
-            .service(web::resource("/").route(web::get().to(|_: HttpRequest| {
-                HttpResponse::Found()
-                    .header(LOCATION, SYNC_DOCS_URL)
-                    .finish()
-            })))
+            .service(
+                web::resource("/").route(web::get().to(|_: HttpRequest| async {
+                    HttpResponse::Found()
+                        .insert_header((LOCATION, SYNC_DOCS_URL))
+                        .finish()
+                })),
+            )
     };
 }
 
@@ -186,8 +188,8 @@ macro_rules! build_app {
 macro_rules! build_app_without_syncstorage {
     ($state: expr, $secrets: expr, $cors: expr) => {
         App::new()
-            .data($state)
-            .data($secrets)
+            .app_data(Data::new($state))
+            .app_data(Data::new($secrets))
             // Middleware is applied LIFO
             // These will wrap all outbound responses with matching status codes.
             .wrap(ErrorHandlers::new().handler(StatusCode::NOT_FOUND, ApiError::render_404))
@@ -212,16 +214,16 @@ macro_rules! build_app_without_syncstorage {
                 web::resource("/__heartbeat__")
                     .route(web::get().to(tokenserver::handlers::heartbeat)),
             )
-            .service(
-                web::resource("/__lbheartbeat__").route(web::get().to(|_: HttpRequest| {
+            .service(web::resource("/__lbheartbeat__").route(web::get().to(
+                |_: HttpRequest| async {
                     // used by the load balancers, just return OK.
                     HttpResponse::Ok()
                         .content_type("application/json")
                         .body("{}")
-                })),
-            )
+                },
+            )))
             .service(
-                web::resource("/__version__").route(web::get().to(|_: HttpRequest| {
+                web::resource("/__version__").route(web::get().to(|_: HttpRequest| async {
                     // return the contents of the version.json file created by circleci
                     // and stored in the docker root
                     HttpResponse::Ok()
@@ -232,11 +234,13 @@ macro_rules! build_app_without_syncstorage {
             .service(
                 web::resource("/__error__").route(web::get().to(tokenserver::handlers::test_error)),
             )
-            .service(web::resource("/").route(web::get().to(|_: HttpRequest| {
-                HttpResponse::Found()
-                    .header(LOCATION, SYNC_DOCS_URL)
-                    .finish()
-            })))
+            .service(
+                web::resource("/").route(web::get().to(|_: HttpRequest| async {
+                    HttpResponse::Found()
+                        .insert_header((LOCATION, SYNC_DOCS_URL))
+                        .finish()
+                })),
+            )
     };
 }
 
@@ -310,7 +314,7 @@ impl Server {
         });
 
         if let Some(keep_alive) = actix_keep_alive {
-            server = server.keep_alive(keep_alive as usize);
+            server = server.keep_alive(Duration::from_secs(keep_alive as u64));
         }
 
         let server = server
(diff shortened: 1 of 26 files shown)


syncserver/src/web/extractors.rs
```diff
@@ -134,7 +134,7 @@ impl From<BatchBsoBody> for PostCollectionBso {
 // This will pull the first accepted content type listed, or the highest rated non-accepted type.
 fn get_accepted(req: &HttpRequest, accepted: &[&str], default: &'static str) -> String {
     let mut candidates = Accept::parse(req).unwrap_or_else(|_| {
-        Accept(vec![qitem(
+        Accept(vec![QualityItem::max(
```
bendk, line 137: I got this from Ethan's doc, but please double check it because I have no clue what's actually going on here.

syncserver/src/server/mod.rs
```diff
@@ -310,7 +314,7 @@ impl Server {
         });
 
         if let Some(keep_alive) = actix_keep_alive {
-            server = server.keep_alive(keep_alive as usize);
+            server = server.keep_alive(Duration::from_secs(keep_alive as u64));
```
bendk, line 317: I'm pretty sure `keep_alive` is in seconds, but I'd love a double check on this one as well.

syncserver/src/web/transaction.rs
```diff
@@ -197,11 +195,10 @@ impl DbTransactionPool {
 }
 
 impl FromRequest for DbTransactionPool {
-    type Error = Error;
+    type Error = actix_web::Error;
```
bendk, line 198: This was another change that felt kind of magical to me and I'd love some double-checking on.  This used to return `actix_http::Error`, which I think may have been a type alias for the `actix_web` version, but I'm not sure.  The reason the change is necessary, is `actix_http::Error` no longer implements `From<T: ResponseError>` after the upgrade.

syncserver/src/server/test.rs
```diff
@@ -149,7 +149,7 @@ fn create_request(
             let hn = HeaderName::from_lowercase(ln.as_bytes())
                 .expect("Could not get hn in create_request");
             let hv = HeaderValue::from_str(v.as_str()).expect("Could not get hv in create_request");
-            req = req.header(hn, hv);
+            req = req.insert_header((hn, hv));
```
bendk, line 152: The docs claim that `append_header` is equivalent to the old `header` method, but that wasn't true for `TestRequest`.  Here, we need to use `insert_header` to override the header set by `set_json` in order to make the `invalid_content_type` test pass.

Maybe we should keep the calls to `header` and update them in another pass, I'm not sure though.

bendk:
> I think some of these changes can be factored out and landed without upgrading any dependencies, I'll try to open a PR for that.

After testing it out some more, I don't think there's much than can be done before the upgrades.

syncserver/src/server/mod.rs
jrconlin: Yes, actix-http 2.2.2 [used Seconds](https://docs.rs/actix-http/2.2.2/actix_http/enum.KeepAlive.html#variant.Timeout)

syncserver/src/server/test.rs
jrconlin: Yeah, the original [`header`](https://docs.rs/actix-web/3.3.3/src/actix_web/test.rs.html#461-469) just set the header in the request directly, and looking at the more recent code [`append_header`](https://docs.rs/actix-web/4.4.1/actix_web/test/struct.TestRequest.html#method.append_header) adds headers of the Key to the request, and [`insert_header`](https://docs.rs/actix-web/4.4.1/actix_web/test/struct.TestRequest.html#method.insert_header) replaces them.

I think we want to use `insert_header` for all of these since we want to create or replace any default header, not just add a value to it.

syncserver/src/server/test.rs
```diff
@@ -131,15 +135,15 @@ fn create_request(
     let settings = get_test_settings();
     let mut req = test::TestRequest::with_uri(path)
         .method(method.clone())
-        .header(
+        .append_header((
```
jrconlin, line 138:
```suggestion
        .insert_header((
```

syncserver/src/server/test.rs
```diff
@@ -131,15 +135,15 @@ fn create_request(
...
             "Authorization",
             create_hawk_header(method.as_str(), settings.port, path),
-        )
-        .header("Accept", "application/json")
-        .header(
+        ))
+        .append_header(("Accept", "application/json"))
```
jrconlin, line 142:
```suggestion
        .insert_header(("Accept", "application/json"))
```

syncserver/src/server/test.rs
```diff
@@ -131,15 +135,15 @@ fn create_request(
...
+        .append_header((
```
jrconlin, line 143:
```suggestion
        .insert_header((
```

syncserver/src/tokenserver/extractors.rs
```diff
@@ -748,9 +740,9 @@ mod tests {
         let req = TestRequest::default()
             .data(state)
             .data(Arc::clone(&SECRETS))
-            .header("authorization", "Bearer fake_token")
-            .header("accept", "application/json,text/plain:q=0.5")
-            .header("x-keyid", "0000000001234-qqo")
+            .append_header(("authorization", "Bearer fake_token"))
```
jrconlin, line 743: I think we're ok with using `append_header` in this case, since there's probably no existing header to add the value to, but we might want to swap this with `insert_header` to override any existing ones.

syncserver/src/web/extractors.rs
jrconlin: Oh, wow. Yeah, this is a fairly convoluted way to deal with a very convoluted field specification. Basically, this is constructing the `Accept` header's `;q=` value, set to the max `1.0` for the Mime qualifier for whatever is specifed as `default`. (e.g. for `txt` it should produce something like `Accept: plain/txt:q=1.0`)

syncserver/src/web/transaction.rs
jrconlin: Yeah, this is pretty common for Actix, so I'm not surprised they moved `Error` again. Looks like the current version re-exports `actix_http::error::Error` which appears to be more about [HTTP processing errors](https://docs.rs/actix-http/3.5.1/src/actix_http/error.rs.html#81), rather than [`actix-web::error::Error`](https://docs.rs/actix-web/latest/actix_web/error/struct.Error.html), which appears to be very much about responses.

jrconlin (changes requested):
That's awesome that you were able to bump up tokio with this. That clears a lot of the CVEs in audit.toml as well. (We might want to do a separate commit that does a `cargo update` and then checks to see what we can purge out of  `.cargo/audit.toml`)

I think the only thing we might want to do is swap out most of the `.append_header()` calls for `.insert_header()`. My read of the docs and the source seems to imply that `.append_header()` adds the provided value to the list of values for that `key` where `.insert_header()` overwrites whatever was there before. Tests are passing because most of the time there's no prior header, but we probably should be paranoid about that. 
Thanks!

jrconlin (commented):
That's awesome that you were able to bump up tokio with this. That clears a lot of the CVEs in audit.toml as well. (We might want to do a separate commit that does a `cargo update` and then checks to see what we can purge out of  `.cargo/audit.toml`)

I think the only thing we might want to do is swap out most of the `.append_header()` calls for `.insert_header()`. My read of the docs and the source seems to imply that `.append_header()` adds the provided value to the list of values for that `key` where `.insert_header()` overwrites whatever was there before. Tests are passing because most of the time there's no prior header, but we probably should be paranoid about that. 
Thanks!

syncserver/src/server/test.rs
bendk: Yeah, that makes sense to me too.  I'll update the code to use `insert_header`

bendk:
Updated the code to use `insert_header`, ran `cargo update` and removed the audit ignores.  I'm starting to feel good about this one, although I don't have a good way to judge the risk.  What should be the next step?

jrconlin (approved):

jrconlin:
This looks like a fairly light lift. For heavier things, we'd want to run a load test to see if there were any large impacts to performance, but I don't think that's a concern here since the upgrades aren't impacting much of the code flow or design. (Mostly it's function name changes and a bit of ownership changes.)

If you're interested, the somewhat crufty load test script we have is [here](https://github.com/mozilla-services/syncstorage-loadtest), I'll reach out to Kat to see if there's a more recent script.

syncserver/src/server/mod.rs
pjenvey: The `keep_alive` method takes a `Duration` so the question's about if `settings.actix_keep_alive` is in seconds -- it is.

We usually document that in the Setting field's docstring but this one isn't documented -- let's clarify that it's seconds [here](https://github.com/mozilla-services/syncstorage-rs/blob/cc8d584/syncserver-settings/src/lib.rs)

pjenvey (changes requested):

syncserver/src/server/mod.rs
bendk: Good call, I just updated that.

jrconlin (approved):

pjenvey (approved):

jrconlin:
*sigh* I really wish that github would flag unsigned commits earlier rather than as the very last check.

PR #1518, opened 2024-02-08
Upgrades rustc in CI and Docker to 1.75

tarikeshaq:
## Description
Looks like our Rust version was still 1.72, which is OK, but it seems like cargo-audit now depends on clap 4.5, which itself required 1.74 (see https://app.circleci.com/pipelines/github/mozilla-services/syncstorage-rs/4909/workflows/427c8cef-e69e-46f3-947f-aa9dc61b3353/jobs/18546)

**I'm keeping it as a draft till we see the tests pass**

Also, It might not be a good idea to upgrade all the way to 1.76, especially since 1.76 just got released today, so I'm open to making this 1.75 (or 1.74)

I also fixed the clippy issues that came with the upgrade

### DIFF
--- a/syncstorage-spanner/src/models.rs
+++ b/syncstorage-spanner/src/models.rs
@@ -1842,10 +1842,7 @@ impl SpannerDb {
             sqlparam_types.insert("payload".to_owned(), payload.spanner_type());
             sqlparams.insert("payload".to_string(), payload.into_spanner_value());
             let now_millis = timestamp.as_i64();
-            let ttl = bso.ttl.map_or(i64::from(DEFAULT_BSO_TTL), |ttl| {
-                ttl.try_into()
-                    .expect("Could not get ttl in put_bso_async_test")
-            }) * 1000;
+            let ttl = bso.ttl.map_or(i64::from(DEFAULT_BSO_TTL), |ttl| ttl.into()) * 1000;
             let expirystring = to_rfc3339(now_millis + ttl)?;
             debug!(
                 "!!!!! [test] INSERT expirystring:{:?}, timestamp:{:?}, ttl:{:?}",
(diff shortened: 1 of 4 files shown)


tarikeshaq:
Well, it looks like circle ci image for 1.76 isn't even there yet, so using 1.75 seems like a good call eitherways

syncstorage-spanner/src/models.rs
```diff
@@ -1842,10 +1842,7 @@ impl SpannerDb {
             sqlparam_types.insert("payload".to_owned(), payload.spanner_type());
             sqlparams.insert("payload".to_string(), payload.into_spanner_value());
             let now_millis = timestamp.as_i64();
-            let ttl = bso.ttl.map_or(i64::from(DEFAULT_BSO_TTL), |ttl| {
-                ttl.try_into()
-                    .expect("Could not get ttl in put_bso_async_test")
-            }) * 1000;
+            let ttl = bso.ttl.map_or(i64::from(DEFAULT_BSO_TTL), |ttl| ttl.into()) * 1000;
```
jrconlin, line 1845: Why are we dropping the `expect`?

pjenvey: Because `try_into` -> `into`, I guess the `try_into` wasn't necessary (maybe the new clippy caught that?)

jrconlin (approved):

syncstorage-spanner/src/models.rs
tarikeshaq: Yup! Looks like there is a [new clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#/unnecessary_fallible_conversions) that catches uses of `try_into` and `try_from` when `From<T>` and `Into<T>` are implemented

PR #1519, opened 2024-02-09
chore: Cleans up TLS dependencies

tarikeshaq:
## Description
We had a few places where we were importing native-tls (which pulls openssl in linux boxes) unintentionally. First was `reqwest`, where we enable the `rustls-tls` feature in a couple of places, but [**it was a no-op** in one](https://docs.rs/reqwest/latest/reqwest/tls/index.html#default-tls) but since we did call `use_rustls_tls` in our [tokenserver browserid verifier](https://github.com/mozilla-services/syncstorage-rs/blob/8ff1b6385eb6ff206750221f4c73259aa73b79d6/tokenserver-auth/src/browserid.rs#L41) that used rustls properly. The second place was sentry, where even though we were using the `curl` feature, the `transport` feature is enabled by default, which pulls in request again, with the `native-tls` feature

Regardless, having **both** rustls and native-tls in our dependency tree could be a future foot gun, so this PR removes native-tls (alternatively, we can go the route of removing rustls but we seem to have consciously chose that, where native-tls snuck in with `reqwest` and `sentry`)

What this PR does not touch, is `boring-ssl` that is pulled in using grpc-io.

One possible improvement we can make to this, that fits well with https://mozilla-hub.atlassian.net/browse/SYNC-4127 is to setup rust features, where we can support multiple tls backends... but I'm struggling to see the value from that (i.e why do we need to support multiple backends in a binary?)


syncserver/Cargo.toml
```diff
@@ -42,10 +41,6 @@ hostname = "0.3.1"
 hawk = "5.0"
 hmac = "0.12"
 mime = "0.3"
-reqwest = { workspace = true, features = [
```
tarikeshaq, line 45: Looks like both reqwest and `env_logger` weren't being used by this crate, so I removed them while cleaning this up

tokenserver-auth/Cargo.toml
```diff
@@ -15,7 +15,7 @@ serde_json.workspace = true
 async-trait = "0.1.40"
 dyn-clone = "1.0.4"
 pyo3 = { version = "0.20", features = ["auto-initialize"] }
-reqwest = { workspace = true, features = ["json", "rustls-tls"] }
```
tarikeshaq, line 18: Since rust features are additive, and we specify the rustls feature in the workspace, there is no need to specify it here

jrconlin (approved):

PR #1520, opened 2024-02-11
feat: Syncstorage inital sqlite support

### LINKED ISSUE #498: Support sqlite/postgresql backend

Eragonfr:
## Description

Re-implementation of #935 on the newer codebase.  
Thanks for @mqus for the first implementation.

## Issue(s)

Partially fixes #498 

## Todo

- [x] Code Cleanup
- [x] Commit history cleanup (I'm not really working in a clean way)
- [x] Add Sqlite clippy checks to CircleCI
- [x] docker container (Thanks @bcspragu for the help)
- [x] e2e tests image (Thanks @bcspragu for the help)
- [ ] pass the e2e tests

Current state is that it runs on my computer, but it's probably not clean at all, there's still a lot of duplicate code, and some SQL queries are still in MySQL-flavored SQL but could be used by the SQLite backend.


Eragonfr:
To pass the checks I have the choice between upgrading diesel or ignoring the issue with libsqlite-sys (which tbh is probably a non-issue because we should not be using SQLite printf anyway).

Eragonfr:
I choose to ignore the libsqlite3-sys CVE because it's not impacting this project. That should let me run the other tests in the CI.

.cargo/audit.toml
```diff
@@ -1,2 +1,2 @@
 [advisories]
-ignore = []
+ignore = ["RUSTSEC-2022-0090"]
```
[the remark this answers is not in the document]
Eragonfr: You're right. Is it enough with what I changed in the latest commit ?

syncserver-db-common/src/test.rs
```diff
@@ -1,6 +1,7 @@
 use diesel::{
     mysql::MysqlConnection,
     r2d2::{CustomizeConnection, Error as PoolError},
+    sqlite::SqliteConnection,
```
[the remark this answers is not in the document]
Eragonfr: I think that I can easily add a feature flag, at least for SQLite and MySQL.
I don't want to touch the code related to spanner because I can't test it.

Makefile
```diff
@@ -13,13 +13,17 @@ PATH_TO_GRPC_CERT = ../server-syncstorage/local/lib/python2.7/site-packages/grpc
 SRC_ROOT = $(shell pwd)
 PYTHON_SITE_PACKGES = $(shell $(SRC_ROOT)/venv/bin/python -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())")
 
+clippy_sqlite:
+	# Matches what's run in circleci
+	cargo clippy --workspace --all-targets --no-default-features --features=syncstorage-db/sqlite,tokenserver-db/sqlite -- -D warnings
+
```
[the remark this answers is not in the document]
Eragonfr: I did not saw that PR.

You want to remove tokenserver completely. Is that right ?

I don't understand what it means for my PR.
Should I change how I handle tokenserver ?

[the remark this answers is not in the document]
jrconlin: I don't believe that it impacts much of your PR, since you're only focused on the Storage part. There may be a few opportunities for folk running their own "Stand Alone" version to have just one database running instead of two, and depending on how we decide to do this for Spanner, there may be some additional functions we introduce to the database layer. We'll try to isolate and highlight those so that we don't cause more work.

syncserver-db-common/src/test.rs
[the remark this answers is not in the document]
jrconlin: You may be able to feature flag a few things (like this `use`:

```rust
#[cfg(feature = "sqlite")]
use diesel::sqlite::SqliteConnection;
```
(and then flag the parts that include `SqliteConnection`)

For now, don't worry about it. As you get closer to the final version, we can work together to try and isolate things so that folk only include what they need. (This would include all the Spanner stuff, too.)

.cargo/audit.toml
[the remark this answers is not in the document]
jrconlin: Yes! That's perfect. Thank you.

Eragonfr:
> I'm really jazzed about this PR, thanks for all the work on it! I noticed this in the PR desc:
> 
> > * [ ]   docker container **(Help needed)**
> 
> If this is a question of building a Docker container from this codebase that bundles all the SQLite deps and nothing more, I'm happy to take on that work.

Something like that yes. There might be some docker-compose configuration files to make too.
I gave multiple tries to Docker but I never seems to get a "standard" docker environment. I always get something with some broken networking, nuked firewall or something else. At this point I've given up on that technology. (docker should release VMs that I can start and get a "standard" env with networking and stuff)

That would leave me to clean up my code

bcspragu:
Sounds good, I'll give it a shot!

bcspragu:
Okay, I think I've got something approximately working in https://github.com/Eragonfr/syncstorage-rs/compare/syncstorage-sqlite...bcspragu:syncstorage-rs:syncstorage-sqlite

E.g.

```bash
docker build -t mozsync-sqlite --build-arg DATABASE_BACKEND=sqlite .

# The --userns line is a rootless Podman thing, can be ignored mostly
docker run --rm -it \
    --userns=keep-id:uid=10001,gid=10001 \
    -v $PWD/temp-data-do-not-submit:/data \
    -e "SYNC_HOST=0.0.0.0" \
    -e "SYNC_MASTER_SECRET=secret0" \
    -e "SYNC_SYNCSTORAGE__DATABASE_URL=sqlite:///data/syncdb.sqlite" \
    -e "SYNC_TOKENSERVER__DATABASE_URL=sqlite:///data/tokenserverdb.sqlite" \
    -e "SYNC_TOKENSERVER__RUN_MIGRATIONS=true" \
    mozsync-sqlite
```

Seems to start up successfully, but I'm not brave enough to connect it to my real Firefox sync yet

Eragonfr:
Thanks @bcspragu !

bcspragu:
For sure! Also, looking at the current `ci/circleci: build-and-test` failure, that's a result of missing `--features` flags here: https://github.com/mozilla-services/syncstorage-rs/blob/4a503f8c36fe070e11df43a8ce0b3c71358e983c/.circleci/config.yml#L67

Not sure what the team would prefer here, but there's two options:

1. If you're just looking for a sanity check, test against MySQL (e.g `--features=syncstorage-db/mysql,tokenserver-db/mysql`)
2. Run separate builds for each DB backend
3. [Actually, this is probably what we want] Fix [the default flags](https://github.com/mozilla-services/syncstorage-rs/blob/4a503f8c36fe070e11df43a8ce0b3c71358e983c/syncserver/Cargo.toml#L65) to use MySQL correctly

---

Unrelated, one thing I forgot to mention is that regardless of which features are enabled, MySQL client libs are required on the host (e.g. there are places where it isn't gated behind the feature flag, it gets linked in regardless). I started adding the relevant `#cfg(feature(...))` in places where I noticed it, but again, not sure what

Eragonfr:
I thought of dropping the default flags because the user should choose their storage backend instead of relying on a default one.


---

We want to have the proper `#cfg(feature(...))` when it's needed without breaking the spanner backend. (It uses MySQL for the token server database)

Eragonfr:
The ci fails on sqlite e2e tests. I think it is because the /data/ folder is missing, which means that the database can't be created and that makes the ci to fail.
I need some help to figure out how to either create this folder or to not need it in the container.

The checks are unrelated and are probably failing on the master branch too.

Eragonfr:
:woman_facepalming: I'm dumb the e2e tests can't work because they require a connection to the database from the python script and syncserver at the same time. Which is simply not possible with sqlite.  
There was no need for my last height commits.  
But the e2e tests are not usable with sqlite.

JulianWgs:
Hi @Eragonfr , thanks for your hard work! I am trying to get this to run locally, but I can't execute [this query](https://github.com/Eragonfr/syncstorage-rs/blob/1c08e8602dd0ce726a9165b90ae881368e948eca/README.md?plain=1#L245-L254) against the token database, because the database file is empty. Am I doing something wrong? Could you try running the server with a missing token database and check if the server correctly sets it up? Thanks again! I am really looking forward to this!

Eragonfr:
@JulianWgs I also have the issue. I think that I broke the auto migration when doing some refactor.
The tokenserver part is giving me some trouble to get it to use SQLite queries, it doesn't pass its unit tests and there is quite a lot of its queries that doesn't run.

Eragonfr:
I really don't like how I do my rebase. It's seems to be quite messy.

jrconlin:
> I really don't like how I do my rebase. It's seems to be quite messy.

I don't think anyone ever does like how they do their rebases.

Would you like me to give it a quick review to see if there's anything I might be able to help with?

Cargo.toml
```diff
@@ -23,6 +27,7 @@ authors = [
   "Ben Bangert <ben@groovie.org>",
   "Phil Jenvey <pjenvey@underboss.org>",
   "Mozilla Services Engineering <services-engineering+code@mozilla.com>",
+  "Eragon <eragon+github@eragon.re>",
```
jrconlin, line 30: :heart: (Seriously, yes, you should add your name here. You're doing a lot.)

tokenserver-db/src/lib.rs
```diff
@@ -1,10 +1,14 @@
 extern crate diesel;
-#[macro_use]
 extern crate diesel_migrations;
 #[macro_use]
 extern crate slog_scope;
 
-mod error;
+use diesel::r2d2::{ConnectionManager, PooledConnection};
```
jrconlin, line 6: I wonder if we should add a feature check here, kind of like we do for spanner. One of the features should be selected (either "mysql" or "sqlite") because otherwise things are not going to work well at all.

tokenserver-db-mysql/src/models.rs
```diff
@@ -0,0 +1,108 @@
+pub const LAST_INSERT_ID_QUERY: &'static str = "SELECT LAST_INSERT_ID() AS id";
```
jrconlin, line 1: nit: can we put a new line between these so they're a bit easier to scan?

syncserver-db-common/src/error.rs
```diff
@@ -8,14 +8,14 @@ use thiserror::Error;
 /// Error specific to any MySQL database backend. These errors are not related to the syncstorage
```
jrconlin, line 8: Hmm. Technically, we should only pull this file in if we have some SQL database we're using, so the `lib.rs` should only include it if appropriate feature flags are set. That could add a lot of flag logic though...

tokenserver-db-sqlite/src/models.rs
```diff
@@ -0,0 +1,111 @@
+pub const LAST_INSERT_ID_QUERY: &'static str = "SELECT LAST_INSERT_ROWID() AS id";
```
jrconlin, line 1: Same as with the mysql file, it'd be nice to add a bit of space here so they're easier to scan.

jrconlin (commented):
I definitely wouldn't beat yourself up about the refactoring. You're introducing a new data storage system, and I like that you did a bunch of things well (like extracting the queries to their own files, putting the `Pool` behind a type for easy abstraction, etc.)

I didn't go into the guts of the sqlite stuff yet, but otherwise just a few minor nits.

THANK YOU for doing all this.

Eragonfr:
> Would you like me to give it a quick review to see if there's anything I might be able to help with?

If you have any advices about how I can fix the tests for sqlite.
Because we can't have both the server and the integration test script connected to the database.
And I really don't know how to fix that.

jrconlin:
> If you have any advices about how I can fix the tests for sqlite.

Ah, because only one application can have open access to the file at one time. Hmm... 
I suppose one possible option might be to have a "test-mode" where the server and integration tests do "blocking writes". (The db file is only opened for a specific read/write operation, and then closed after the transaction is completed.) It's not great, but it would allow for some hand off of the file between the two. How one would do that while using the connection pool would be a bit of a trick, though.

I'm guessing setting the `database_pool_min_idle` to `0`, or maybe modifying the `lock_for_*()` functions under `syncstorage-*` to maybe open and close the files at the point where the locks are requested, and then doing something similar with optionally stub calls in the integration tests. 

I'm not sure, since it's been a while since I had to dig into that code, but that would be my first thought. I'll try to see if I can come up with any other ideas when I get a chance to dig in more.

syncserver-db-common/src/error.rs
```diff
@@ -8,14 +8,14 @@ use thiserror::Error;
...
 /// or tokenserver application logic; rather, they are lower-level errors arising from diesel.
 #[derive(Debug)]
-pub struct MysqlError {
-    kind: MysqlErrorKind,
+pub struct SqlError {
```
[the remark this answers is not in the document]
Eragonfr: Maybe I can split that into a separate PR that can get merged before this one.
Because I really think that the renaming should be done before adding the new backend.
If I do the renaming later I will mess up things and I won't know if something is MySQL specific or generic.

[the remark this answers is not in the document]
jrconlin: Yeah, I do think the rename makes a lot of sense, particularly if we're going to have more than one SQL backend. 
I'm fine with boosting the rename PR since it'll also reduce the size and complexity of this one. 

(Granted, I've also been debating making the default build "spanner", since that's what we use for production, but that's beside the point.)

syncserver-db-common/src/error.rs
Eragonfr: I think that fb10b2c is good

tokenserver-db/src/lib.rs
Eragonfr: Do you know of a way to force the user to choose a feature ? Something like a compile time panic if no feature is enabled.

syncserver-db-common/src/error.rs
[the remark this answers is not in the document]
Eragonfr: Here's the new PR #1619 
It's based of the master branch. (I can revert the second commit if needed but that was triggering clippy)

tokenserver-db/src/lib.rs
Bowarc: build.rs
```rust
If ![cfg!(feature = "feat1"), cfg!(feature = "feat2")].iter().any(){
    panic("You need to enable at least one feature")
}
```

Im on my phone atm, so idk if this compiles, but it seems reasonable

jrconlin: Ah! Sorry, I thought you saw the trick we do in [syncstorage-db](https://github.com/mozilla-services/syncstorage-rs/blob/master/syncstorage-db/src/lib.rs#L37C1-L37C57) 

```
#[cfg(not(any(feature = "mysql", feature = "spanner")))]
```
basically, just add `sqlite` into that mix. There's also a line above that which does a check to make sure you don't specify more than one storage engine.

Eragonfr: Turns out I already added this to [syncstorage-db](https://github.com/eragonfr/syncstorage-rs/blob/syncstorage-sqlite/syncstorage-db/src/lib.rs#L46-L47)
```rust
#[cfg(not(any(feature = "mysql", feature = "spanner", feature = "sqlite")))]
compile_error!("exactly one of the \"mysql\", \"spanner\" and \"sqlite\" features must be enabled");
```

Do you want the same thing for tokenserver-db ?

Cargo.toml
Eragonfr: Thank you

tokenserver-db/src/lib.rs
jrconlin: It's not a bad idea.

The long(-ish) term idea is that we eventually roll tokenserver into syncstorage so that there's one less weird thing hanging out on it's own. Tokenserver will definitely stay MySQL based for the forseeable future, but I don't know if you want to do the same work to convert it to Sqlite. (Much of tokenserver deals with juggling users between various "nodes" and "clusters" which can be abandoned, even for many stand-alone systems. You could also leave that as an exercise for some future person, and just make notes about the things that would need changing.)

Eragonfr: I added the feature checks in [d8c556b](https://github.com/mozilla-services/syncstorage-rs/pull/1520/commits/d8c556bb49e0c379d9e05b8637ba5af1cf8f3e5c)

I already almost have tokenserver fully working on SQLite. There's only 8 tests that still fails. But it should be pretty easy to fix.

Eragonfr:
> I'm guessing setting the `database_pool_min_idle` to `0`, or maybe modifying the `lock_for_*()` functions under `syncstorage-*` to maybe open and close the files at the point where the locks are requested, and then doing something similar with optionally stub calls in the integration tests.

Setting the `database_pool_min_idle` to `0` might be working to close the file after each transaction from syncstorage/tokenserver.

But we need to find a way of doing the same for the python scripts

Eragonfr:
I'm trying to clean up the imports, crates versions, etc
Some refactor that should not change the way the code works. But it should be cleaner to read and maintain.

Eragonfr:
If anyone has an idea on how to get the e2e tests to run, I'll take it.
What is needed is for syncstorage to close the sqlite database, the test script can then open the database.  
The test script then closes the database and query syncstorage for the next test.
Syncstorage can then re-open the database for the transaction.
Rince and repeat.

But I can't figure out how to force the Rust SQLite driver (diesel with r2d2) and the Python driver (sqlalchemy) to close the database after each transaction.

Eragonfr:
I tried, it only made the MySQL e2e test to fail [c8f21bd](https://github.com/mozilla-services/syncstorage-rs/pull/1520/commits/c8f21bdc3c650468fa5b2a0752225a991b1bea1d)

Eragonfr:
At this point [ffd1bd9](https://github.com/mozilla-services/syncstorage-rs/pull/1520/commits/ffd1bd96bbece460edb9f55d9edb4ace1634554f) it's certain that it's python that holds the lock on the database but I really don't know how I can force it to close the file

kyz:
> But I can't figure out how to force the Rust SQLite driver (diesel with r2d2) and the Python driver (sqlalchemy) to close the database after each transaction.

I'm just a passer-by offering my opinion, and I haven't tested this, but hopefully it helps.

SQLite supports a [busy handler](https://www.sqlite.org/pragma.html#pragma_busy_timeout) and [WAL mode](https://www.sqlite.org/wal.html) to allow multiple writers and readers, across multiple processes, to use the same database without blocking each other. It's still only allowing one write transaction at a time, but processes automatically coordinate with each other through a shared memory file as to who has the lock.

Here is a good article on StackOverflow on [enabling WAL with Diesel / r2d2](https://stackoverflow.com/questions/57123453/how-to-use-diesel-with-sqlite-connections-and-avoid-database-is-locked-type-of), including some sample code for inspiration. 

I imagine you would only want to enable this for tests where you want this concurrent access.

I haven't used SQLAlchemy, but the Python tests should use the same SQLite pragmas as Rust to access the database file, e.g. `conn.execute("PRAGMA journal_mode = WAL; PRAGMA synchronous = NORMAL; PRAGMA busy_timeout = 10000;")`

Eragonfr:
WAL mode seems like a good thing to have even in production. As it significantly speeds up SQLite.

But even when it's used I still have a connection that holds the lock and makes the others fail their transactions.

Eragonfr:
I GOT THE TESTS TO RUN!!!
I modified them for SQLite, but I should be able to add only two conditions and everything else will run exactly the same on MySQL and SQLite

Eragonfr:
Finger crossed… this might fail because of python and test lib version mismatch

Eragonfr:
Oops… the `checks` part will fail until we change something in it.
```
error: cannot install package `cargo-audit 0.21.2`, it requires rustc 1.81.0 or newer, while the currently active rustc version is 1.78.0
`cargo-audit 0.21.1` supports rustc 1.74.0
```

Eragonfr:
[db5b1dc](https://github.com/mozilla-services/syncstorage-rs/pull/1520/commits/db5b1dcd288eee04c317901cacc06ca4d6ae1926) Will fail I haven't managed to get the ORM system to work exactly as needed to pass the tests

almereyda:
There are three errors, one that shows in all environments, one in the checks and one in the SQLite one:

1. All environments fail with importing a type from SQLAlchemy:

```py
  File "/app/tools/integration_tests/run.py", line 11, in <module>
    from tokenserver.run import run_end_to_end_tests, run_local_tests
  File "/app/tools/integration_tests/tokenserver/run.py", line 6, in <module>
    from tokenserver.test_authorization import TestAuthorization
  File "/app/tools/integration_tests/tokenserver/test_authorization.py", line 5, in <module>
    from tokenserver.test_support import TestCase
  File "/app/tools/integration_tests/tokenserver/test_support.py", line 28, in <module>
    from tokenserver.tables import Users, Nodes, Services
  File "/app/tools/integration_tests/tokenserver/tables.py", line 1, in <module>
    from sqlalchemy import Integer, String, Null, BigInteger, Index
ImportError: cannot import name 'Null' from 'sqlalchemy' (/usr/local/lib/python3.9/dist-packages/sqlalchemy/__init__.py)
```

As far as I can see it the used SQLAlchemy version does not have a type called `Null`. But there is one called `NullType`.

- https://github.com/mozilla-services/syncstorage-rs/blob/324a6d8144169af34daf14a9aa76c35741293d49/tools/tokenserver/requirements.txt#L5
- https://docs.sqlalchemy.org/en/14/core/type_api.html#sqlalchemy.types.NullType

Maybe that's already it?

2. The checks issue is related to a version and dependency jump from a dependency:

```
error: cannot install package `cargo-audit 0.21.2`, it requires rustc 1.81.0 or newer, while the currently active rustc version is 1.78.0
`cargo-audit 0.21.1` supports rustc 1.74.0
```

Does the patch upgrade from `cargo-audit` justify moving to `rustc` >= 1.81.0? I suppose this error would also apply on other branches. Maybe this can be fixed on `mozilla-services:master` and be rebased here?

3. https://app.circleci.com/pipelines/github/mozilla-services/syncstorage-rs/5579/workflows/ee50f1bb-8d13-4fb3-b1e1-c7bed79be30a/jobs/22461?invite=true#step-106-410_334

```
called `Result::unwrap()` on an `Err` value: ApiError { kind: Db(DbError { kind: Sql(SqlError { kind: DieselConnection(BadConnection("Unable to open the database file")), status: 500, backtrace:    0: <syncserver_db_common::error::SqlError as core::convert::From<syncserver_db_common::error::SqlErrorKind>>::from
```

Might this be to a read-only filesystem on the CI runner? To the contrary, the databases in the other tests boot up just fine.

Eragonfr:
> There are three errors, one that shows in all e2e tests, one in the checks and one in the SQLite one:
> 
>     1. All environments fail with importing a type from SQLAlchemy:
> 
> As far as I can see it the used SQLAlchemy version does not have a type called `Null`. But there is one called `NullType`.

Interesting, this one is related to version mismatch between my dev env and our test env I think.
I'm running python 3.13 with latest sqlalchemy.

>     * https://github.com/mozilla-services/syncstorage-rs/blob/324a6d8144169af34daf14a9aa76c35741293d49/tools/tokenserver/requirements.txt#L5
> 
>     * https://docs.sqlalchemy.org/en/14/core/type_api.html#sqlalchemy.types.NullType
> 
> 
> Maybe that's already it?
> 
>     2. The checks issue is related to a version and dependency jump from a dependency:

I already knew that
See https://github.com/mozilla-services/syncstorage-rs/pull/1520#issuecomment-2708415736
Already fixed in master on march 10.

> Might this be to a read-only filesystem on the CI runner? To the contrary, the databases in the other tests boot up just fine.

docker FS is not read only
This is just me that fucked up the SQLite DB path and it points to a folder that doesn't exists
There's a good reason why the db in the other tests is able to start, one test uses MySQL only and the other uses MySQL+spanner. None of them uses SQLite except the sqlite one.

Eragonfr:
The move to nextest and my exams stopped me from keeping up with the main branch.
I'll try to get it working with the new test setup

Eragonfr:
I haven't keep my branch up to date with the new modifications to tokenserver-db.  
I don't have time to rebase and fix it.
The main issue with my code for a year is that it can't run e2e tests because of the nature of sqlite (and maybe my architectural choices)

I won't try to fix it anymore, if someone want to try to take it over, please do. But I expect that the database architecture will change quite a bit because of the current work being done on tokenserver

shyeyian:
> I haven't keep my branch up to date with the new modifications to tokenserver-db. I don't have time to rebase and fix it. The main issue with my code for a year is that it can't run e2e tests because of the nature of sqlite (and maybe my architectural choices)
> 
> I won't try to fix it anymore, if someone want to try to take it over, please do. But I expect that the database architecture will change quite a bit because of the current work being done on tokenserver

Hi @Eragonfr, thank you for all your work on this PR :)

I continued your work in a new PR. A lot of my implementation is inherited from your code and developed further to fit the current codebase. (Your notes about SQLite migrations, URLs, transactions are super powerful!!)

I am still quite new to open source (BTW, I'm a Rustacean...), so I hope I have continued your work in a reasonable way. I credited you, @mqus, and @bcspragu in the new PR.

Thank you again :)

Eragonfr:
I'm quite surprised that my ramblings in this bad, non-native, english could help you @shyeyian .
But I'm glad that I could help in some ways.

I'm glad that someone took to the task of finishing it. Good luck with the e2e tests, I never got them to work.
I did my best in this PR but with time I got tired of it and stopped trying to self-host syncstorage.
I'll close it as yours is more up to date.
And I hope to see your work merged and available for all in the future.

PR #1522, opened 2024-02-26
docs: Remove reference to legacy vendored library

jrconlin:
## Description

Remove documentation instruction that points to old vendor library


README.md
```diff
@@ -156,34 +160,40 @@ DDL_STATEMENTS=$(
...
 To run an application server that points to the local Spanner emulator:
+
 ```sh
 SYNC_SYNCSTORAGE__SPANNER_EMULATOR_HOST=localhost:9010 make run_spanner
 ```
 
 ### Running via Docker
-This requires access to the mozilla-rust-sdk which is now available at `/vendor/mozilla-rust-sdk`.
+
+This requires access to [Google Cloud Rust (raw)](https://github.com/mozilla-services/google-cloud-rust-raw/) crate. Please note that due to interdependencies, you will need to ensure that `grpcio` and `protobuf` match the version used by `google-cloud-rust-raw`.
```
taddes, line 184: Hrm, I get a 404 not found when following this link. Is the `raw` crate private? I see the path to `google-cloud-rust` works fine here: https://github.com/mozilla-services/google-cloud-rust. However the listing for `google-cloud-rust-raw` is active in `lib.rs`: https://lib.rs/crates/google-cloud-rust-raw

taddes (changes requested):
Thanks @jrconlin , just an issue with the url to `google-cloud-rust-raw` (private?) that I mentioned in the comment. Don't have enough context to know which of those two repos (raw or not raw) it should resolve to.

README.md
jrconlin: No, I pasted in the wrong link. It should be https://crates.io/crates/google-cloud-rust-raw/ (sigh)

taddes (approved):
LGTM @jrconlin 👍 . Looks like re-run needed for mysql-e2e-tests , couldn't re-init from failed on my end, perhaps I need permissions access to do so for sync. May be intermittent fail, know that happens in CI.

PR #1524, opened 2024-02-28
Feat/override

jrconlin:
## Description

The thought is that in `update_user`, we don't call `db.get_node_id()` to override the user's `node_id` unless the `client_state` doesn't match between the auth and the user records. We provide the spanner node, but we only pass up the resolved `node` string. This PR passes the node_id back up, checks that it matches the override if it's present, and resets the user record if it doesn't match.


jrconlin:
closing in favor of a more modern branch

PR #1531, opened 2024-04-03
feat: Remove support for BrowserID

jrconlin:
## Description

This PR removes BrowserID support from Syncstorage. 
In addition it includes several additional elements I needed to perform local testing and validation.
These included:
  * Updating some library dependencies (to fix audit concerns).
  * Updating the mysql docker image to use supported CircleCI images
  * Locking spanner image to avoid a parse error with the current version
  * Replacing several hard coded test secret values with ones that can optionally read from an environment variable

## Testing

Testing is included

## Issue(s)

Closes: SYNC-3684


pjenvey (dismissed):

docker-compose.spanner.yaml
```diff
@@ -1,62 +1,69 @@
...
-            - sync-db-setup
-        environment:
-            SYNC_HOST: 0.0.0.0
-            SYNC_MASTER_SECRET: secret0
-            SYNC_SYNCSTORAGE__DATABASE_URL: spanner://projects/test-project/instances/test-instance/databases/test-database
-            SYNC_SYNCSTORAGE__SPANNER_EMULATOR_HOST: sync-db:9010
-            SYNC_TOKENSERVER__DATABASE_URL: mysql://test:test@tokenserver-db:3306/tokenserver
-            SYNC_TOKENSERVER__RUN_MIGRATIONS: "true"
+  sync-db:
+    # Getting sporadic errors in spanner.
```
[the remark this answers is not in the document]
jrconlin: Yep. They are. Apparently the most recent Spanner emulator produces the error I noted. That's why it's locked down to a specific version that does work.

tools/integration_tests/run.py
```diff
@@ -52,21 +52,15 @@ def start_server():
...
     mock_fxa_server_url = os.environ["MOCK_FXA_SERVER_URL"]
     url = "%s/v2" % mock_fxa_server_url
-    os.environ["SYNC_TOKENSERVER__FXA_BROWSERID_SERVER_URL"] = url
     os.environ["SYNC_TOKENSERVER__FXA_OAUTH_SERVER_URL"] = mock_fxa_server_url
     the_server_subprocess = start_server()
     try:
         res = 0
         res |= run_live_functional_tests(TestStorage, sys.argv)
         os.environ["TOKENSERVER_AUTH_METHOD"] = "oauth"
-        res |= run_local_tests(include_browserid_specific_tests=False)
```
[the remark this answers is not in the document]
jrconlin: yeah, sorry, over-aggressive pruning there.

syncserver/src/tokenserver/extractors.rs
```diff
@@ -98,14 +98,6 @@ impl TokenserverRequest {
             });
         }
 
-        // If the client previously reported a client state, every subsequent request must include
-        // one. Note that this is only relevant for BrowserID requests, since OAuth requests must
-        // always include a client state.
-        if !self.user.client_state.is_empty() && self.auth_data.client_state.is_empty() {
```
pjenvey, line 104: Double checking these docs stating these sections are "only relevant for BrowserID", I'm not entirely sure about this one -- where do we enforce a non empty `client_state` for oauth? It's a bit more complicated by the fact that we check different places for it (via `X-KeyID` and `X-Client-State`) headers.

I probably need to stare at this code a little more closely but I'm also wondering if a test for this case might be helpful (if it doesn't exist already) -- could be a separate issue.

syncserver/src/tokenserver/extractors.rs
```diff
@@ -159,26 +151,6 @@ impl TokenserverRequest {
...
 
-        // If there's no keys_changed_at on the request, there must be no value stored on the user
-        // record. Note that this is only relevant for BrowserID requests, since OAuth requests
-        // must always include a keys_changed_at header. The Python Tokenserver converts a NULL
-        // keys_changed_at on the user record to 0 in memory, which means that NULL
-        // keys_changed_ats are treated equivalently to 0 keys_changed_ats. This would allow users
-        // with a 0 keys_changed_at on their user record to hold off on sending a keys_changed_at
-        // in requests even though the value in the database is non-NULL. To be thorough, we
-        // handle this case here.
-        if auth_keys_changed_at.is_none()
```
pjenvey, line 170: ..whereas this case seems to definitely be true, `keys_changed_at` is enforced for oauth via the `KeyId` struct, ensuring one exists and is valid and is converted to i64.

syncserver/src/tokenserver/extractors.rs
```diff
@@ -209,12 +181,8 @@ impl FromRequest for TokenserverRequest {
             log_items_mutator.insert("metrics_uid".to_owned(), hashed_fxa_uid.clone());
 
             // To preserve anonymity, compute a hash of the FxA device ID to be used for reporting
-            // metrics. Only requests using BrowserID will have a device ID, so use "none" as
-            // a placeholder for OAuth requests.
-            let hashed_device_id = {
-                let device_id = auth_data.device_id.as_deref().unwrap_or("none");
```
pjenvey, line 215: I think we should kill the `AuthData::device_id` field entirely now as well (now always `None` and no longer used?)

pjenvey (changes requested):

syncserver/src/tokenserver/extractors.rs
jrconlin: Yeah, I didn't want to touch that because it could get complicated, fast. The problem is that `device_id` gets converted to `hashed_device_id`, which has a lot of dependencies, most notably, as part of the HAWK header request. I'll remove the `device_id`, but I'm also going to file a ticket to see about removing `hashed_device_id`, because I think we may need to check the HAWK header bits to make sure that there's no weirdness that could happen if that field goes away.

syncserver/src/tokenserver/extractors.rs
jrconlin: meh, leaving the code and removing the comment seems reasonable.

syncserver/src/tokenserver/extractors.rs
jrconlin: Ah, right! I got thrown by the comment. I'll add this back and remove the browserID stuff.

pjenvey: Sorry, what I meant was I agreed with removing this section, I was just being verbose about how I audited the extractor code and it matched what the comment said: that oauth will always supply a `keys_changed_at` value. So let's nuke it!

jrconlin: Hrm. So, let me counter.
As I understand this step currently, it's a counter for a record being potentially out of step for unknown reasons.
The modified code checks to see if there's an `auth_keys_changed_at` and then checks to see if the user record may have a `0` due to a previously incorrectly written record.  I think that may still be a valid thing to test for.

pjenvey: Just to clarify: the code checks if there's *not* a `keys_changed_at` in the request, which the original comment suggests never happens on oauth, but that's not quite correct. Per the new, more accurate comment added: a `0` value is treated as `None` even on oauth. 👍

pjenvey (approved):
Even a seemingly simple killing code task in tokenserver is tricky 😓 (like everything else in tokenserver :rage1:), thanks @jrconlin

PR #1543, opened 2024-04-23
feat: special case purging of users previously migrated to Spanner

pjenvey:
don't issue deletes to their current data on syncstorage if their old "migration records" point to it

Closes SYNC-4225

### DIFF
--- a/tools/tokenserver/purge_old_records.py
+++ b/tools/tokenserver/purge_old_records.py
@@ -108,12 +108,14 @@ def purge_old_records(
                         database.delete_user_record(row.uid)
                     counter += 1
                 elif force:
+                    delete_sd = not points_to_active(database, row, override_node)
                     logger.info(
                         "Forcing tokenserver record delete: "
-                        f"{row.uid} on {row.node}"
+                        f"{row.uid} on {row.node} "
+                        f"(deleting service data: {delete_sd})"
                     )
                     if not dryrun:
-                        try:
+                        if delete_sd:
                             # Attempt to delete the user information from
                             # the existing data set. This may fail, either
                             # because the HawkAuth is referring to an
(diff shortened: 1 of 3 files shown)


jrconlin:
Looks like a `cargo update` might help with the audit.

tools/tokenserver/purge_old_records.py
```diff
@@ -108,34 +108,38 @@ def purge_old_records(
...
+                                    override_node=override_node
+                                )
+                            except requests.HTTPError:
+                                logger.warn(
+                                    "Delete failed for user "
+                                    f"{row.uid} [{row.node}]"
+                                )
+                                if override_node:
+                                    # Assume override_node should be reachable
+                                    raise
```
pjenvey, line 142: Note @jrconlin I just newly added this (and I forgot to commit it separately)

tools/tokenserver/purge_old_records.py
```diff
@@ -183,6 +188,23 @@ def delete_service_data(user, secret, timeout=60, dryrun=False):
...
+    node) would delete their active data, which should be avoided
+    """
+    if override_node and replaced_at_row.node != override_node:
+        # NOTE: Users who never connected after being migrated could be
+        # assigned a spanner node record by get_user (TODO: rename get_user ->
+        # get_or_assign_user)
+        user = database.get_user(replaced_at_row.email)
+        if (user["generation"] == replaced_at_row.generation and
+            user["client_state"] == replaced_at_row.client_state):
+            return True
```
jrconlin, line 205:
```suggestion
        return (user["generation"] == replaced_at_row.generation and
            user["client_state"] == replaced_at_row.client_state)
```

jrconlin:
Yeah, I kinda wonder if we should standardize to black for python stuff, since autopush wants to run both. It'd save us a bit of hassle to just have the various IDEs run black on save and be done with that sort of thing.

jrconlin (approved):

taddes (approved):
LGTM, just a few small suggestions if you want to consider.

taddes:
How do you feel if I were to draft a quick PR to give us those same Makefile commands to run linting and style enforcement? At the least, flake8. That way it's consistent as @jrconlin  mentioned.

jrconlin:
We're going to land this PR as is, mostly because of time constraints, but @taddes has a few good suggestions for clean up. Let's move those to a different PR (feel free to create one @taddes) and we can land that later.

PR #1545, opened 2024-04-25
fix: take keys_changed_at into account w/ migrated records' special case

pjenvey:
and make the comparison more clear/explicit as the goal is to check if the `fxa_kid`s match

Issue SYNC-4225


jrconlin (approved):

taddes (approved):
LGTM, great addition and context.

PR #1547, opened 2024-04-26
feat: Allow uid range for purge function (SYNC-4246)

jrconlin:
## Description

In an attempt to parallelize this script after a very long delay, specify a range so that multiple scripts can try to process different ranges of the database.

## Issue(s)

Issue  SYNC-4246.

### DIFF
--- a/tools/tokenserver/database.py
+++ b/tools/tokenserver/database.py
@@ -101,6 +101,26 @@ offset
     :offset
 """)
 
+_GET_OLD_USER_RECORDS_FOR_SERVICE_RANGE = """\
+select
+    uid, email, generation, keys_changed_at, client_state,
+    nodes.node, nodes.downed, created_at, replaced_at
+from
+    users left outer join nodes on users.nodeid = nodes.id
+where
+    users.service = :service
+and
+    ::RANGE::
+and
+    replaced_at is not null and replaced_at < :timestamp
+order by
+    replaced_at desc, uid desc
+limit
+    :limit
+offset
+    :offset
+"""
+
 
 _GET_ALL_USER_RECORDS_FOR_SERVICE = sqltext("""\
 select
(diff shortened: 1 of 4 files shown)


tools/tokenserver/purge_old_records.py
```diff
@@ -45,6 +45,7 @@ def purge_old_records(
     dryrun=False,
     force=False,
     override_node=None,
+    uid_range=None,
```
pjenvey, line 48: `uid_range` is never used in this method (it's not passing `range` to `get_old_user_records`)

tools/tokenserver/database.py
```diff
@@ -101,6 +101,26 @@
...
+_GET_OLD_USER_RECORDS_FOR_SERVICE_RANGE = """\
+select
+    uid, email, generation, keys_changed_at, client_state,
+    nodes.node, nodes.downed, created_at, replaced_at
+from
+    users left outer join nodes on users.nodeid = nodes.id
+where
+    users.service = :service
+and
+    ::RANGE::
```
pjenvey, line 113: nit: I'd prefer adding this to `_GET_OLD_USER_RECORDS_FOR_SERVICE` and only adding the entire `and ` clause when needed but not going to block on this

pjenvey (changes requested):

pjenvey:
Also we can just skip the e2e tests for now until that gets resolved, here's a branch doing that:

https://github.com/mozilla-services/syncstorage-rs/compare/test/disable-tokenserver-e2e

tools/tokenserver/database.py
jrconlin: I thought about doing that, but since I'm basically doing a string replacement inside of the sql query, (and since this is technically only required for the bulk cleanup we're doing) I figured that it would be cleaner to break these apart.

tools/tokenserver/purge_old_records.py
jrconlin: Argh. Missed that when doing the clean-up. Thanks.

tools/tokenserver/purge_old_records.py
```diff
@@ -69,6 +70,7 @@ def purge_old_records(
                 "grace_period": grace_period,
                 "limit": max_per_loop,
                 "offset": offset,
+                "range": uid_range,
             }
             rows = list(database.get_old_user_records(**kwds))
             if not rows:
```
pjenvey, line 76: one more thing: let's also add the range to the "Fetched rows.." log message a few lines below

pjenvey (changes requested):

pjenvey (approved):

PR #1550, opened 2024-04-30
Feat/range2

jrconlin:
## Description

Correct the val names to match the options. In addition, address a few other things:

* always use the last arg as the "secret". 
* break out the LOGGER (logging still seems a bit broken locally, this fixes it in some instances)
* `range` is a keyword, so avoid using it where and when possible.

Issue: SYNC-4246

### DIFF
--- a/tools/tokenserver/purge_old_records.py
+++ b/tools/tokenserver/purge_old_records.py
@@ -70,6 +70,7 @@ def purge_old_records(
                 "grace_period": grace_period,
                 "limit": max_per_loop,
                 "offset": offset,
+                "range": uid_range,
             }
             rows = list(database.get_old_user_records(**kwds))
             if not rows:
--- a/tools/tokenserver/purge_old_records.py
+++ b/tools/tokenserver/purge_old_records.py
@@ -79,7 +79,14 @@ def purge_old_records(
             if rows == previous_list:
                 raise Exception("Loop detected")
             previous_list = rows
-            logger.info("Fetched %d rows at offset %d", len(rows), offset)
+            range_msg = ""
+            if range:
+                range_msg = (
+                    f" within range {range[0] or 'Start'}"
+                    f" to {range[1] or 'End'}"
+                )
+            logger.info(
+                f"Fetched {len(rows)} rows at offset {offset}{range_msg}")
             counter = 0
             for row in rows:
                 # Don't attempt to purge data from downed nodes.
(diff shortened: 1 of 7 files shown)


tools/tokenserver/purge_old_records.py
```diff
@@ -336,17 +341,21 @@ def main(args=None):
...
+    # Secret is the last arg?
+    secret = args[-1]
+    logger.debug(f"Secret: {secret}")
 
     util.configure_script_logging(opts)
 
     uid_range = None
-    if opts.start_range or opts.end_range:
-        uid_range = (opts.start_range, opts.end_range)
+    if opts.range_start or opts.range_start:
```
pjenvey, line 356:
```suggestion
    if opts.range_start or opts.range_end:
```

tools/tokenserver/purge_old_records.py
```diff
@@ -336,17 +341,21 @@ def main(args=None):
...
     opts, args = parser.parse_args(args)
-    if len(args) != 2:
+
+    if len(args) == 0:
         parser.print_usage()
         return 1
 
-    secret = args[1]
+    # Secret is the last arg?
+    secret = args[-1]
```
pjenvey, line 350: I don't understand this change, was it necessary (as this only takes one positional `secret` arg)?

pjenvey (changes requested):

tools/tokenserver/purge_old_records.py
jrconlin: My reasoning is:

We were pulling `arg[1]` and checking that we had two args. This meant that the `usage` [message](https://github.com/mozilla-services/syncstorage-rs/blob/master/tools/tokenserver/purge_old_records.py#L253) was wrong, because `purge_old_records.py  secret` would not work. 

`optparse.OptionsParser()` returns the `args` as a list (e.g. `args = ["secret"]`) meaning that for the script to work correctly, we would have to introduce another argument that we are ignoring. Since `args[1]` equates to the last of the two arguments being passed, it's also equivalent to being the last argument being passed, since previously, we were enforcing that [there be two arguments](https://github.com/mozilla-services/syncstorage-rs/blob/master/tools/tokenserver/purge_old_records.py#L339) passed. 

This fix would allow us to either specify the existing two arguments, or drop the first, ignored argument and still function. 

Is that mistaken?

pjenvey: Got it, sorry I was misreading it as originally being `args[0]` (because why on earth would it be `1` requiring [this sillyness](https://github.com/mozilla-services/cloudops-infra/blob/9dc1f77/projects/sync/k8s/charts/tokenserver/templates/deployment-purge-old-records.yaml#L56) :rage1:)

pjenvey (approved):

PR #1552, opened 2024-05-01
feat: purge log fix (SYNC-4250)

jrconlin:
## Description

Fix up the discrepant logging in `purge_old_records.py` 

## Testing

How should reviewers test?

## Issue(s)

Closes SYNC-4250

### DIFF
--- a/tools/tokenserver/purge_old_records.py
+++ b/tools/tokenserver/purge_old_records.py
@@ -29,14 +29,19 @@ import util
 from database import Database
 from util import format_key_id
 
-LOGGER = "tokenserver.scripts.purge_old_records"
-
-logger = logging.getLogger(LOGGER)
-log_level = os.environ.get("PYTHON_LOG", "INFO").upper()
-logger.setLevel(log_level)
-logger.debug(f"Setting level to {log_level}")
-
+LOGGER = ""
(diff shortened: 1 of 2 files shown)


pjenvey (changes requested):
It seems the original problem is that `purge_old_records` had extra module level logging initialization added at some point instead of solely relying on `util.configure_script_logging` like the other scripts.

We shouldn't need logging setup before the `configure_script_logging` call as the scripts only do one thing beforehand: parse the command line arguments.

Maybe `configure_script_logging` should use `basicConfig` but seems fine as is for e.g. `process_account_events` (note to myself: `logging.getLogger("")` is getting the root logger)? So I'd argue it doesn't need changes and all that's needed here is to remove `purge_old_records` 4 lines of module level initialization code and rely on the verbosity count command line arg to drive the log level.

pjenvey:
> So I'd argue it doesn't need changes and all that's needed here is to remove `purge_old_records` 4 lines of module level initialization code and rely on the verbosity count command line arg to drive the log level.

Err, I guess https://github.com/mozilla-services/syncstorage-rs/pull/1555 pretty much does this. Maybe this one should be closed in favor of it

jrconlin:
Yeah, forgot to close out this PR in favor of that one. Sorry.

PR #1560, opened 2024-05-15
bug: Allow threadpool size to be set.

jrconlin:
It appears that the replacement for setting ACTIX_THREADPOOL would be to call `ServiceBuilder.worker_max_blocking_threads()` This PR introduces the ability to set this value ~either~ by using the `worker_max_blocking_threads` configuration variable ~or `ACTIX_THREADPOOL` environment variable~.

Closes #SYNC-4271

### DIFF
--- a/syncserver-settings/src/lib.rs
+++ b/syncserver-settings/src/lib.rs
@@ -38,6 +38,8 @@ pub struct Settings {
     pub cors_allowed_methods: Option<Vec<String>>,
     pub cors_allowed_headers: Option<Vec<String>>,
 
+    pub actix_threadpool: Option<usize>,
+
     // TOOD: Eventually, the below settings will be enabled or disabled via Cargo features
     pub syncstorage: SyncstorageSettings,
     pub tokenserver: TokenserverSettings,
@@ -205,6 +207,7 @@ impl Default for Settings {
                 .collect(),
             ),
             cors_max_age: Some(1728000),
+            actix_threadpool: Some(512),
             syncstorage: SyncstorageSettings::default(),
             tokenserver: TokenserverSettings::default(),
         }
--- a/syncserver-settings/src/lib.rs
+++ b/syncserver-settings/src/lib.rs
@@ -38,7 +38,7 @@ pub struct Settings {
     pub cors_allowed_methods: Option<Vec<String>>,
     pub cors_allowed_headers: Option<Vec<String>>,
 
-    pub actix_threadpool: Option<usize>,
+    pub worker_max_blocking_threads: Option<usize>,
 
     // TOOD: Eventually, the below settings will be enabled or disabled via Cargo features
     pub syncstorage: SyncstorageSettings,
@@ -207,7 +207,7 @@ impl Default for Settings {
                 .collect(),
             ),
             cors_max_age: Some(1728000),
-            actix_threadpool: Some(512),
+            worker_max_blocking_threads: Some(512),
             syncstorage: SyncstorageSettings::default(),
             tokenserver: TokenserverSettings::default(),
         }
--- a/syncserver-settings/src/lib.rs
+++ b/syncserver-settings/src/lib.rs
@@ -1,8 +1,6 @@
 #[macro_use]
 extern crate slog_scope;
 
-use std::env::{self, VarError};
-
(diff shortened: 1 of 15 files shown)


syncserver-settings/src/lib.rs
```diff
@@ -38,6 +38,8 @@ pub struct Settings {
     pub cors_allowed_methods: Option<Vec<String>>,
     pub cors_allowed_headers: Option<Vec<String>>,
 
+    pub actix_threadpool: Option<usize>,
```
pjenvey, line 41: I think we should move away from this deprecated env var completely as actix-web did, it's confusing to keep it (call the setting something else).

I'm guess I'm not opposed to reading from the existing env var but TBH I'd rather just modify the cloudops config to set a new one alongside the old one and not even bother being backwards compatible, especially now that we have the time #killit

syncserver/src/server/mod.rs
```diff
@@ -358,6 +358,12 @@ impl Server {
         });
 
         let server = server
+            .worker_max_blocking_threads(settings.actix_threadpool.unwrap_or_else(|| {
```
pjenvey, line 361: This value differs from `ACTIX_THREADPOOL`'s which set the size of a global pool across the process, whereas this pool is [per actix-web worker](https://docs.rs/actix-web/latest/actix_web/struct.HttpServer.html#method.worker_max_blocking_threads).

So if we want to continue treating this setting as the overall thread count (wfm) this value should be divided by `available_parallelism()` (see [the default](https://github.com/actix/actix-net/blob/8cf79d3d1364cd91fb933640c7614a40586d1eb6/actix-server/src/worker.rs#L256) )

pjenvey (changes requested):

pjenvey:
FYI: https://github.com/mozilla-services/cloudops-infra/pull/5650

syncserver-settings/src/lib.rs
```diff
@@ -38,6 +36,12 @@ pub struct Settings {
     pub cors_allowed_methods: Option<Vec<String>>,
     pub cors_allowed_headers: Option<Vec<String>>,
 
+    /// The maximum number of blocking threads that can be used by the worker.
+    /// Note, we don't want "None" here because we use this as part of the
+    /// metric periodic reporter. The default value is 512. Setting the value to
+    /// `0` will
+    pub worker_max_blocking_threads: usize,
```
pjenvey, line 43: I think this comment got cut off?

So AFAICT we could still use `Option` here, we would just need to set the default to `None`, but I don't really mind the non Option w/ `0` instead.

syncserver/src/server/mod.rs
```diff
@@ -328,10 +334,18 @@ impl Server {
...
     ) -> Result<dev::Server, ApiError> {
         let settings_copy = settings.clone();
+
         let host = settings.host.clone();
         let port = settings.port;
         let secrets = Arc::new(settings.master_secret.clone());
         let blocking_threadpool = Arc::new(BlockingThreadpool::default());
+        // Adjust the thread count to include FxA blocking threads.
+        let thread_count = settings.worker_max_blocking_threads
+            + settings
```
pjenvey, line 344: Was this a bug on master that this fixes (we didn't seem to be taking `additional_blocking_threads_for_fxa_requests` into account at all unless `ACTIX_THREADPOOL` wasn't set)?

pjenvey (approved):
Really just needs the comment finished/corrected I think

syncserver/src/server/mod.rs
jrconlin: Not really a bug, but more a case that when we the logic worked that if we were running this as a stand-alone TokenServer, we needed to add additional threads to deal with the FxA requests, which are specified in that option. Technically, we could either use just the tokenserver thread count config or the worker_max_blocking_thread count, but I figured we should continue to do the same adjustment.

syncserver-settings/src/lib.rs
jrconlin: Yeah, I'll remove the `0` comment, since it gets really convoluted what happens if we specify 0. 
In short, we should never specify 0.

pjenvey (approved):

PR #1561, opened 2024-05-15
feat: Add timeouts for tokenserver database calls.

jrconlin:
Closes: SYNC-4270


tokenserver-settings/src/lib.rs
```diff
@@ -78,6 +80,7 @@ impl Default for Settings {
             database_pool_max_size: 10,
             database_pool_min_idle: None,
             database_pool_connection_timeout: Some(30),
+            database_request_timeout: Some(30),
```
pjenvey, line 83: I'd like to have the option to completely disable the timeouts and any potential tokio timer overhead along with it -- i.e. when set to disabled, skip creating the timeout entirely.

See deadpool's example along these lines: [deadpool::managed::apply_timeout](https://github.com/bikeshedder/deadpool/blob/2a584a99fa6d80921f6428b7c5e7e6dc06b8e0a9/src/managed/mod.rs#L648)

pjenvey (changes requested):

tokenserver-settings/src/lib.rs
jrconlin: Fair.

pjenvey: Due to that config issue I mentioned the other day, I think we just need to set this to None and apply a value in the cloudops config, otherwise there's no way to force it back to None

```suggestion
            database_request_timeout: None,
```

syncserver/src/tokenserver/handlers.rs
```diff
@@ -250,7 +276,10 @@ pub async fn heartbeat(DbWrapper(db): DbWrapper) -> Result<HttpResponse, Error>
         Value::String(env!("CARGO_PKG_VERSION").to_owned()),
     );
 
-    match db.check().await {
+    match apply_timeout(db.timeout(), db.check())
+        .await
+        .map_err(|_| TokenserverError::elapsed())
+    {
```
pjenvey, line 282: Do we need the `map_err` now that `apply_timeout` does it?
```suggestion
    match apply_timeout(db.timeout(), db.check())
        .await
    {
```

syncserver/src/tokenserver/handlers.rs
```diff
@@ -120,6 +121,24 @@ struct UserUpdates {
...
+    fut: impl std::future::Future<Output = Result<O, E>>,
+) -> Result<O, TokenserverError>
+where
+    E: Into<TokenserverError>,
+{
+    match duration {
+        Some(duration) => match timeout(duration, fut).await {
+            Ok(Ok(result)) => Ok(result),
+            Ok(Err(e)) => Err(e.into()),
+            Err(_) => Err(TokenserverError::elapsed()),
```
pjenvey, line 136: Is this even needed w/ the `From` (maybe it still is since it's the outer exception)? If so we can probably just kill that `From` entirely.

pjenvey (changes requested):

syncserver/src/tokenserver/handlers.rs
jrconlin: Ah, good catch. I was thinking we could keep it for safety reasons, but meh.

pjenvey (approved):

PR #1562, opened 2024-05-18
fix: nix-shell: update `pkgconfig` -> `pkg-config` build input

christoph-heiss:
## Description

The (formerly) `pkgconfig` package in NixOS/nixpkgs has long been renamed to `pkg-config`, to match its upstream name. As of last september, this actually throws a hard error and thus breaks `shell.nix`. (See [here](https://github.com/NixOS/nixpkgs/blob/25ad13412abb8fc5153901cd8695f8ac44cf34a3/pkgs/top-level/aliases.nix#L991)).

## Testing

Run `nix-shell` on a (recent) NixOS system or any system with the Nix package manager installed and a recent nixpkgs channel.

## Issue(s)

Closes N/A.

Thanks for considering!


jrconlin (approved):

jrconlin:
Thanks!

PR #1563, opened 2024-05-20
docs: Remove commented code, unneeded TODO, unneeded collision tracking

jrconlin:
removed commented out code from last PR, as well as spanner collision detection, since we haven't seen any in over 3 months.

### DIFF
--- a/syncserver-common/src/lib.rs
+++ b/syncserver-common/src/lib.rs
@@ -85,7 +85,6 @@ pub trait ReportableError: std::fmt::Display {
     }
 
     /// Experimental: return key value pairs for Sentry Event's extra data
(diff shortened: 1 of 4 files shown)


pjenvey (approved):
Did we ever get any of those collisions (I don't see any at all)?

jrconlin:
Nope, I did a search out for a year and nothing was in metrics (I know, but if it can be killed, it can be over-killed). I figure it's definitely safe to drop.

pjenvey:
> Nope, I did a search out for a year and nothing was in metrics (I know, but if it can be killed, it can be over-killed). I figure it's definitely safe to drop.

👍 Ok, FYI I just removed it from Grafana as well

PR #1567, opened 2024-06-05
chore: Update to debian bookworm / Python 3.12

jrconlin:
Bump to Debian Bookworm from Buster
* Update to Python 3.12 baseline
  * Buster + Python 3.12 complicates python package installs. I've switched to using virtualenv to contain the packages we want.
  * We may wish to move to Poetry or some similar system in the future. 
* Update to Rust 1.78
  * This required some minor Clippy fixes.
  * Added `RUST_VER` tag to easily find instances of rust versions for future updates
* Update to Cargo-Chef  

Closes: SYNC-4289

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,7 +25,9 @@ commands:
       - run:
           name: Setup python
           command: |
-            sudo apt-get update && sudo apt-get install -y python3-dev python3-pip
+            sudo apt-get update && sudo apt-get install -y python3-dev python3-pip python3-venv
+            python3 -mvenv venv
+            . venv/bin/activate
             pip3 install flake8 hawkauthlib konfig pyramid pyramid_hawkauth requests simplejson unittest2 WebTest WSGIProxy2
             pip3 install -r requirements.txt
   rust-check:
@@ -40,6 +42,7 @@ commands:
       - run:
           name: Core Python Checks
           command: |
+            . venv/bin/activate
             flake8 syncserver/src/tokenserver
             flake8 tools/integration_tests
             flake8 tools/tokenserver
@@ -116,6 +119,7 @@ commands:
       - run:
           name: Tokenserver scripts tests
           command: >
+               . venv/bin/activate && 
                pip3 install -r tools/tokenserver/requirements.txt &&
                python3 tools/tokenserver/run_tests.py
           environment:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -165,7 +165,7 @@ commands:
 jobs:
   checks:
     docker:
-      - image: cimg/rust:1.75.0
+      - image: cimg/rust:1.78.0  # RUST_VER
         auth:
           username: $DOCKER_USER
           password: $DOCKER_PASS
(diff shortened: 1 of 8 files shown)


.circleci/config.yml
```diff
@@ -161,7 +165,7 @@ commands:
 jobs:
   checks:
     docker:
-      - image: cimg/rust:1.75.0
+      - image: cimg/rust:1.78.0  # RUST_VER
```
pjenvey, line 168: 👍

pjenvey (approved):
For some cursed reason this branch also fails to build for me locally, even with a fresh checkout of the codebase and w/ `--no-cache`

```
$ time docker build --no-cache -t syncstorage-rs:test-build .
[+] Building 4.2s (13/33)
...
1.460 Get:5 https://repo.mysql.com/apt/debian bookworm InRelease [22.7 kB]
1.709 Get:6 http://deb.debian.org/debian bookworm-updates/main arm64 Packages [13.7 kB]
1.710 Get:7 http://deb.debian.org/debian-security bookworm-security/main arm64 Packages [156 kB]
2.384 Fetched 9131 kB in 1s (6189 kB/s)
2.384 Reading package lists...
2.698 Reading package lists...
3.030 Building dependency tree...
3.136 Reading state information...
3.139 Package libmysqlclient-dev is not available, but is referred to by another package.
3.139 This may mean that the package is missing, has been obsoleted, or
3.139 is only available from another source
3.139 However the following packages replace it:
3.139   libmariadb-dev-compat libmariadb-dev
3.139 
3.142 E: Package 'libmysqlclient-dev' has no installation candidate
------
Dockerfile:13
```

#repeatablebuilds :rage4:

jrconlin:
Huh... 
Looks like `cargo-chef` is the cuplrit. Using maria instead is a reasonable ask. Let me try a test build locally and see.

jrconlin:
Oh, heh, we even comment about that in the README. Let me just change things.

pjenvey:
> Using maria instead is a reasonable ask.

Remember we switched off it for the Oracle client because of maria's basically nonexistent (and or requiring some weird, custom maria configuration that I'm not even sure we're able to configure) TLS support. This may not be as important as it was during the migration from AWS, but it would be preferable to keep TLS enabled -- and I know we even made mysql enforce its usage, that likely still exists over on our GCP db instance.

jrconlin:
Yeah, basically, adding the Maria bits and a big "THIS IS FOR TESTING ONLY" comment in the Dockerfile. (Not that folk will read that anymore than they'll read anything else, but it gives us a sign to tap.)

jrconlin:
Ok, going to roll back the mariadb changes and make that a different ticket since there's more changes needed.

pjenvey (approved):
I don't understand how the reverted mariadb change would have been "testing only" as we use that main `Dockerfile` to build the production images

jrconlin:
Right, since the tokenserver stuff is being included. Hrm. Well, it's rolled back now.

pjenvey:
> For some cursed reason this branch also fails to build for me locally

Note to self: The problem I'm seeing is due to my differing arm64 platform (which I often forget all about) in which there is no libmysqlclient-dev etc packages for.

jrconlin:
I mean, I'm a bit wierded out that there's no libmysqlclient-dev client for arm64 (one would presume that aside from a little bit ordering, it shouldn't be a huge issue). Guessing Oracle has better things to do.

PR #1571, opened 2024-06-12
bug: Revert the venv configuration for python

jrconlin:
* use `--system-break-packages` flag to force install to system.

Closes #SYNC-4298


pjenvey (commented):
There's a few more venv references that can be killed https://github.com/mozilla-services/syncstorage-rs/compare/bug/SYNC-4298_pip...wip/no-venv

pjenvey (changes requested):
See my previous comment https://github.com/mozilla-services/syncstorage-rs/pull/1571#pullrequestreview-2113722309

jrconlin:
> See my previous comment [#1571 (review)](https://github.com/mozilla-services/syncstorage-rs/pull/1571#pullrequestreview-2113722309)

I was fine keeping the apt includes partly because we reference them in things like the Makefile, Readme, etc. and folk are already running our Docker files as stand-alone. meh, whatever.

scripts/start_mock_fxa_server.sh
```diff
@@ -1,5 +1,4 @@
 #!/bin/sh
 
-. /app/venv/bin/activate
 python3 /app/tools/integration_tests/tokenserver/mock_fxa_server.py
```
pjenvey, line 3: Optionally we could also kill this script since it was only needed for the venv? But I don't really care

.circleci/config.yml
```diff
@@ -25,10 +26,10 @@ commands:
       - run:
           name: Setup python
           command: |
-            sudo apt-get update && sudo apt-get install -y python3-dev python3-pip python3-venv
-            python3 -mvenv venv
-            . venv/bin/activate
+            sudo apt-get update && sudo apt-get install -y python3-dev python3-pip
             pip3 install flake8 hawkauthlib konfig pyramid pyramid_hawkauth requests simplejson unittest2 WebTest WSGIProxy2
+            # NOTE: Python3.12 requires `--break-system-packages`.
+            # This command is run on the circleci/mysql:5.7-ram image, which is running python 3.10
```
pjenvey, line 32: Again we don't run anything of ours in the mysql image, these run in the job's primary rust image. You can see that in [this branch's](https://github.com/mozilla-services/syncstorage-rs/compare/bug/SYNC-4298_pip...bug/SYNC-4298_pip-pjenvey) ci run [here](https://app.circleci.com/pipelines/github/mozilla-services/syncstorage-rs/5162/workflows/ce939a49-c623-4407-a81e-60b1d08fb706/jobs/20052/parallel-runs/0/steps/0-103) and [here](https://app.circleci.com/pipelines/github/mozilla-services/syncstorage-rs/5162/workflows/ce939a49-c623-4407-a81e-60b1d08fb706/jobs/20052/parallel-runs/0/steps/0-110)
```suggestion
            # This command is run on the cimg/rust image, which is running python 3.10
```

.circleci/config.yml
```diff
@@ -118,9 +118,10 @@ commands:
     steps:
       - run:
           name: Tokenserver scripts tests
-          command: >
-               . venv/bin/activate && 
-               pip3 install -r tools/tokenserver/requirements.txt &&
+          command: |
+              # NOTE: Python3.12 requires `--break-system-packages`.
+              # This command is run on the circleci/mysql:5.7-ram image, which is running python 3.10
```
pjenvey, line 123:
```suggestion
              # This command is run on the cimg/rust image, which is running python 3.10
```

pjenvey (changes requested):

scripts/start_mock_fxa_server.sh
jrconlin: I'm tempted to leave it, partly because I can see it being useful in the future, at least before we drop the tokenserver.

.circleci/config.yml
jrconlin: Yeah, considering there's a docker image that's created that runs a docker compose that brings up two other docker images, I need a road map to figure out all the dependencies.

.circleci/config.yml
```diff
@@ -25,10 +26,10 @@ commands:
       - run:
           name: Setup python
           command: |
-            sudo apt-get update && sudo apt-get install -y python3-dev python3-pip python3-venv
-            python3 -mvenv venv
-            . venv/bin/activate
+            sudo apt-get update && sudo apt-get install -y python3-dev python3-pip
             pip3 install flake8 hawkauthlib konfig pyramid pyramid_hawkauth requests simplejson unittest2 WebTest WSGIProxy2
```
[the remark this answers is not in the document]
jrconlin: A system level `pip` would rely on the system level `python`. `pip` is a tiny shell wrapper that basically does the same thing as what you're proposing. Unless you've got a broken install, both `pip` and `python` should be the same in the system path, and if they're not, other things would break. 

I know that poetry subverts all of this and uses it's own mechanisms for installing dependencies, so I think it's fine calling `pip` directly for now.

.circleci/config.yml
jrconlin: There's no explicit version of Python, but one is usually built into whatever base image that the docker image is built from. In our case, `cimg/rust` is built off of an image that uses `Ubuntu 22.04`, which pre-installs Python 3.10 (with Python 3.11 available for install). 22.04 was [the prior LTS](https://wiki.ubuntu.com/Releases) for Ubuntu, with 24.04 being the current. I suspect that future versions of `cimg/rust` will change to 24.04 (which may happen 14-Jun-2024). In that case it may just make sense to wait until next week to see if a lot of this complexity just goes away.

.circleci/config.yml
```diff
@@ -42,7 +43,6 @@ commands:
       - run:
           name: Core Python Checks
           command: |
-            . venv/bin/activate
             flake8 syncserver/src/tokenserver
```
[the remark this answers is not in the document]
jrconlin: I think we should not rely on any virtual environment and instead install packages to the system level for whatever dockerfiles we generate. Virtual environments are there to prevent multiple applications from stepping on each other. Docker images encapsulate that by giving applications a sandbox that they can run in. Therefore python's virtualenv concept is redundant in Docker images.

taddes (approved):
LGTM @jrconlin , good addition with the `display-versions` conditionals.

pjenvey (approved):

PR #1581, opened 2024-07-03
docs: clarify the handling of existing expired bsos in writes

### LINKED ISSUE #619: Handle AlreadyExists in bsos errors

pjenvey:
Closes #619

### DIFF
--- a/syncstorage-spanner/src/batch.rs
+++ b/syncstorage-spanner/src/batch.rs
@@ -195,6 +195,8 @@ pub async fn commit_async(
             "timestamp" => as_rfc3339.clone(),
         };
         sqlparam_types.insert("timestamp".to_owned(), as_type(TypeCode::TIMESTAMP));
+        // NOTE: This write treats existing but expired bsos as existing (and
+        // not expired), see the note in SpannerDb::post_bsos_with_mutations
         db.sql(include_str!("batch_commit_update.sql"))?
             .params(sqlparams)
             .param_types(sqlparam_types)
@@ -216,6 +218,8 @@ pub async fn commit_async(
         sqlparam_types.insert("timestamp".to_owned(), as_type(TypeCode::TIMESTAMP));
         let mut timer3 = db.metrics.clone();
         timer3.start_timer("storage.spanner.apply_batch_insert", None);
+        // NOTE: This write treats existing but expired bsos as existing (and
+        // not expired), see the note in SpannerDb::post_bsos_with_mutations
         db.sql(include_str!("batch_commit_insert.sql"))?
             .params(sqlparams)
             .param_types(sqlparam_types)
--- a/syncstorage-spanner/src/batch.rs
+++ b/syncstorage-spanner/src/batch.rs
@@ -195,8 +195,8 @@ pub async fn commit_async(
             "timestamp" => as_rfc3339.clone(),
         };
         sqlparam_types.insert("timestamp".to_owned(), as_type(TypeCode::TIMESTAMP));
-        // NOTE: This write treats existing but expired bsos as existing (and
-        // not expired), see the note in SpannerDb::post_bsos_with_mutations
+        // NOTE: This write treats both expired and non-expired as existing
+        // bsos. See the note in [SpannerDb::post_bsos_with_mutations]
         db.sql(include_str!("batch_commit_update.sql"))?
             .params(sqlparams)
             .param_types(sqlparam_types)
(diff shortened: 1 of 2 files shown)


syncstorage-spanner/src/batch.rs
```diff
@@ -195,6 +195,8 @@ pub async fn commit_async(
             "timestamp" => as_rfc3339.clone(),
         };
         sqlparam_types.insert("timestamp".to_owned(), as_type(TypeCode::TIMESTAMP));
+        // NOTE: This write treats existing but expired bsos as existing (and
+        // not expired), see the note in SpannerDb::post_bsos_with_mutations
```
jrconlin, line 199: I get the idea, but the wording feels awkward.

What about:
```
// NOTE: This write treats both expired and non-expired, existing bsos. See the note in [SpannerDb::post_bsos_with_mutations]
```

syncstorage-spanner/src/models.rs
```diff
@@ -1575,6 +1575,16 @@ impl SpannerDb {
                 .map(|pbso| pbso.id.clone())
                 .collect::<Vec<String>>(),
         };
+        // Determine what bsos already exist (need to be inserted vs updated)
+        // NOTE: Here and in batch commit we match the original Python
+        // syncstorage's behavior: not specifying "AND expiry >
```
jrconlin, line 1580:
```
// syncstorage's behavior:
//  * do not specify `AND expiry > CURRENT_TIMESTAMP()
//  * treat both expired and non-expired, existing bsos
// This simplifies the writes, avoiding the need to
```

syncstorage-spanner/src/models.rs
```diff
@@ -1575,6 +1575,16 @@ impl SpannerDb {
...
+        // CURRENT_TIMESTAMP()", treating existing but expired bsos as existing
+        // (and not expired). This simplifies the writes, avoiding the need to
+        // delete those expired bsos before inserting new ones with the same
+        // id. Unfortunately this means updates may resurrect expired bsos (or
```
jrconlin, line 1584:
```suggestion
        // id. Unfortunately, this means updates may resurrect expired bsos (or
```

syncstorage-spanner/src/models.rs
```diff
@@ -1575,6 +1575,16 @@ impl SpannerDb {
...
+        // at least a subset of their fields), or possibly even write new data
+        // without an associated ttl to an expired record that will be
+        // deleted. This in practice should be a very rare occurence
```
jrconlin, line 1587:
```suggestion
        // deleted. This in practice should be a very rare occurrence.
```

jrconlin (approved):

PR #1592, opened 2024-08-29
fix: revert the python3.10 match statement (for now)

pjenvey:
so the script can operate under bullseye/python3.9

Issue SYNC-4259


taddes (approved):
🚀

PR #1600, opened 2024-09-19
feat: Revert "chore: revert back to bullseye (for now)"

pjenvey:
and also feat: Revert "fix: revert the python3.10 match statement (for now)"

reverts the reverting from debian bookworm -> bullseye commits (minus the added RUSTSEC ignore) as they don't seem to be involved in the recent issues on prod

Issue SYNC-4363
Issue SYNC-4259

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
-# NOTE: Ensure builder's Rust version matches CI's in .circleci/config.yml
-FROM docker.io/lukemathwalker/cargo-chef:0.1.67-rust-1.78-bullseye as chef
+# NOTE: Ensure builder's Rust version matches CI's in .circleci/config.yml # RUST_VER
+FROM docker.io/lukemathwalker/cargo-chef:0.1.67-rust-1.78-bookworm as chef
 WORKDIR /app
 
 FROM chef AS planner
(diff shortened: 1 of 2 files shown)


taddes (approved):
LGTM. Nit: I do think we could convert the logging statements to f-strings instead of using %r, but it’s up to you! Won’t block on that.

jrconlin (commented):
r+ (fwiw)

I agree about swapping to f"", but that's a later fix.

pjenvey:
@taddes @jrconlin Agreed on using f-strings here, but I'll note there's also a rare potential reason (or two) to not use them w/ logging as well:

Specifying the args to the logging function makes the logging impl responsible for the interpolation step -- and it can save a little work by avoiding the interpolation entirely if the message isn't actually logged (e.g. log level isn't enabled). Also, apparently some log implementations even collect the arguments for later querying (see [Google's style guide](https://google.github.io/styleguide/pyguide.html#3101-logging) mentioning this point as a reason why it discourages f-strings for logging).

These aren't very compelling reasons for most code (they certainly don't apply to our scripts) and the downside is you could potentially be stuck with the old style '%' formatting style if you're using a logging [Formatter](https://docs.python.org/3/library/logging.html#logging.Formatter) you can't configure to use a newer `style='{'` style -- the example in the Google style guide suggests the tensorflow logger forces the old '%' style

PR #1601, opened 2024-09-23
fix: correctly read the SYNC_STATSD_HOST/PORT settings

pjenvey:
Closes SYNC-4417


taddes (approved):
LGTM, nice abstraction with the `add_metric_options` function 👍

PR #1604, opened 2024-09-25
fix: ensure the pool counter's always decremented via scopeguard

pjenvey:
also report machine parallelism @ startup

## Testing

stage also shows the metrics issue, so we can verify there first

## Issue(s)

Issue SYNC-4421

### DIFF
--- a/syncserver-common/src/lib.rs
+++ b/syncserver-common/src/lib.rs
@@ -121,6 +121,8 @@ impl BlockingThreadpool {
         E: fmt::Debug + Send + InternalError + 'static,
     {
         self.spawned_tasks.fetch_add(1, Ordering::Relaxed);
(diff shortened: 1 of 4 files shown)


syncserver-common/src/lib.rs
```diff
@@ -121,16 +121,15 @@ impl BlockingThreadpool {
         E: fmt::Debug + Send + InternalError + 'static,
     {
         self.spawned_tasks.fetch_add(1, Ordering::Relaxed);
+        scopeguard::defer! {
```
jrconlin, line 124: Please add a comment here noting why we're waiting until we run the task after the function goes out of scope.

jrconlin (changes requested):

jrconlin (approved):

PR #1605, opened 2024-09-30
feat: track the pool's queued vs actually active tasks

pjenvey:
and consider idle threads as a "max" value as the underlying tokio threadpool will exit threads idle for too long

Closes SYNC-4424

### DIFF
--- a/syncserver/src/server/mod.rs
+++ b/syncserver/src/server/mod.rs
@@ -13,7 +13,10 @@ use actix_web::{
 };
 use cadence::{Gauged, StatsdClient};
 use futures::future::{self, Ready};
-use syncserver_common::{middleware::sentry::SentryWrapper, BlockingThreadpool, Metrics, Taggable};
+use syncserver_common::{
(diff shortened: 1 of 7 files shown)


syncserver/src/server/mod.rs
```diff
@@ -259,16 +262,16 @@ impl Server {
...
+        let blocking_threadpool = Arc::new(BlockingThreadpool::new(
+            settings.worker_max_blocking_threads,
+        ));
         let db_pool = DbPoolImpl::new(
             &settings.syncstorage,
             &Metrics::from(&metrics),
             blocking_threadpool.clone(),
         )?;
         let worker_thread_count =
             calculate_worker_max_blocking_threads(settings.worker_max_blocking_threads);
```
pjenvey, line 274: In retrospect I don't know why we included `worker_` in this setting's name because it's the overall number of threads (not per worker). Whereas the naming of the `calculate_worker_max_blocking_threads` fn makes sense

jrconlin (approved):

PR #1606, opened 2024-09-30
feat: create DAU glean schema and configs

taddes:
## Description

Addition of the metrics.yaml and pings.yaml configurations for Glean metrics. This is to measure the Daily Active User metrics internally.

Noting the open questions doc we created with Glean team [here](https://docs.google.com/document/d/1WDnn49eD0lxdcSfZzaeQZsVaKQ2KqG7VhQrP8vZDdyE/edit#heading=h.i4wnylph1d9x).  We may approach this with the following modification:
- A single event metric with the extra_keys metadata section containing the other data (platform, sync id, collection, etc)
- We may exclude the timestamp as Glean has its own internal timestamp we can use

## Testing
N/A

## Issue(s)

Closes [SYNC-4415](https://mozilla-hub.atlassian.net/browse/SYNC-4415).

[SYNC-4415]: https://mozilla-hub.atlassian.net/browse/SYNC-4415?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/glean/metrics.yaml
+++ b/glean/metrics.yaml
@@ -8,6 +8,83 @@ $schema: moz://mozilla.org/schemas/glean/metrics/2-0-0
 
 # Category
 syncstorage:
-  #Event
-  dau:
-    type: "event"
+  sync_event:
+    type: event
+    description: |
+      Event to record an instance of sync backend activity initiated by client.
+    notification_emails:
+      - jrconlin@mozilla.com
+      - pjenvey@mozilla.com
+      - tkorris@mozilla.com
+    bugs:
+      - https://github.com/mozilla-services/syncstorage-rs/issues
+    data_reviews:
+      # TODO
+    expires: never
+
+  sync_user_id:
+    type: uuid
+    # yamllint disable
+    description: >
+      Sync user identifier. Used to determine which user has initiated sync
+      activity. A single user could make numerous sync actions in a given time
+      and this id is required to ensure only a single count of daily active use
+      is made given a number of actions.
+    # yamllint enable
+    notification_emails:
+      - jrconlin@mozilla.com
+      - pjenvey@mozilla.com
+      - tkorris@mozilla.com
+    bugs:
+      - https://github.com/mozilla-services/syncstorage-rs/issues
+    data_reviews:
+      # TODO
+    expires: never
+
+  user_agent:
+    type: string
+    # yamllint disable
+    description: |
+      User Agent from which sync action was initiated.
+      Firefox Desktop, Fenix, or Firefox iOS.
+    # yamllint enable
+    notification_emails:
+      - jrconlin@mozilla.com
+      - pjenvey@mozilla.com
+      - tkorris@mozilla.com
+    bugs:
+      - https://github.com/mozilla-services/syncstorage-rs/issues
+    data_reviews:
+      # TODO
+    expires: never
+
+  timestamp:
+    type: datetime
+    description: |
+      Timestamp of sync activity.
+    notification_emails:
+      - jrconlin@mozilla.com
+      - pjenvey@mozilla.com
+      - tkorris@mozilla.com
+    bugs:
+      - https://github.com/mozilla-services/syncstorage-rs/issues
+    data_reviews:
+      # TODO
+    expires: never
+
+  collection:
+    type: string
+    # yamllint disable
+    description: |
+      Related individual collection where sync activity took place.
+      Includes bookmarks, history, forms, prefs, tabs, and passwords.
+    # yamllint enable
+    notification_emails:
+      - jrconlin@mozilla.com
+      - pjenvey@mozilla.com
+      - tkorris@mozilla.com
+    bugs:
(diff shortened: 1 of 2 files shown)


glean/metrics.yaml
```diff
@@ -0,0 +1,76 @@
...
+      - https://github.com/mozilla-services/syncstorage-rs/issues
+    data_reviews:
+      # TODO
+    expires: never
+
+  sync_user_id:
+    type: uuid
+    # yamllint disable
+    description: >
+      Sync user identifier. Used to determine which user has initiated sync
```
pjenvey, line 29: "Sync user id" is their numeric id assigned by tokenserver, included in the user's sync endpoint (e.g. `https://sync-1-us-west1-g.sync.services.mozilla.com/1.5/<sync-id>`).

Tokenserver gives users new ids basically whenever their encryption key (represented by `fxa_kid`) changes, such as during password resets without a recovery key.

This makes the sync user id not 100% accurate for counting distinct users: it could count a user multiple times across a time period where their encryption key changed. So a lot like the old fxa query (and I believe the sync client pings) we should use the user's `hashed_fxa_uid` instead.

pjenvey (changes requested):

glean/metrics.yaml
taddes: Excellent point regarding the encryption key change. I'll update to reflect the use of `hashed_fxa_uid`.

taddes: For clarity, I named that metric `hashed_fxa_uid` so there's no confusion on what it is 👍

pjenvey: > Any additional concern with deriving said `hashed_fxa_uid` within the context of the application when we'll emit metrics?

Ah yes I was going to mention a btw for when you implement this: tokenserver already sends this value to syncstorage [in its hawk token payload](https://github.com/mozilla-services/syncstorage-rs/blob/1f0e28d7af9c6f9aea38073a099699897464ceac/tokenserver-auth/src/lib.rs#L44). I guess syncstorage has never actually used it so it's not present as a field in its [parsed HawkPayload](https://github.com/mozilla-services/syncstorage-rs/blob/1f0e28d7af9c6f9aea38073a099699897464ceac/syncserver/src/web/auth.rs#L36). You should be able to simply add it as a new field on this struct and then have it available.

I'll note as an FYI the `salt` field in `HawkPayload` is added for us by the Python code (`make_token`) in case you were wondering why it doesn't exist in tokenserver's `MakeTokenPlaintext`.

Then take a look at how we pass around `fxa_uid/kid` via `UserIdentifier`: I believe you'll want to add the hashed version there so it's accessible when emitting the metric.

taddes:
Lmk if you want any other changes in the meantime @pjenvey , otherwise we can merge in for the Glean folks to review and then do a data review too.

glean/metrics.yaml
```diff
@@ -0,0 +1,96 @@
...
+    data_reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1923967
+    expires: never
+
+  platform:
+    type: string
+    # yamllint disable
+    description: |
+      Platform from which sync action was initiated.
+      Firefox Desktop, Fenix, or Firefox iOS.
```
pjenvey, line 52: Could also mention "Other" here and in device_family

pjenvey (approved):
@taddes I have one more doc suggestion but that can wait for a later commit. Also when you squash this please change the commit message to follow the style as autopush's ("feat: etc"), this repo follows the same format (in `CONTRIBUTING.md`) so `clog` picks it up appropriately for the changelog

glean/metrics.yaml
taddes: @pjenvey  Thx, added it. Will alter the commit for sure when squashing. Sigh... it dismissed your approve after pushing the changes. Any way to avoid dismissing "stale reviews"? That a GitHub setting we can tweak as I recall we didn't have it in merino-py. Otherwise, if you don't mind approving again. Thx

pjenvey: Also for `platform` but again, it can wait

pjenvey (approved):

glean/metrics.yaml
taddes: Yeah that one I hemmed and hawed about, can add it with the next iteration. It's in the enum

pjenvey: @jrconlin Do you recall any requirement for the "Dismiss stale pull request approvals when new commits are pushed" option? I can't find one anywhere now, maybe this was just something our team (you and I) standardized on?

jrconlin: I don't know if it ever was an official company wide policy, but I think our team has been doing it since we started, partly because a new push could impact previously viewed changes.

I _think_ that the the folk that don't use git do something similar (where a new change request basically overrides the prior, so you'd need to re-approve it), but I'm not super familiar with how that side works.

PR #1607, opened 2024-10-07
feat: user agent parsing

taddes:
## Description

Addition of User Agent parsing, consisting of:
Platform (Firefox Desktop, Firefox iOS, Fenix, Other)
Device family (Desktop, Mobile, Tablet, Other)
 OS family (Mac, Windows, Linux, iOS, Android, other)

Note: User agent emission from iOS clients is non-standard. Two observed variants occur:
`Firefox-iOS-FxA/24` or `Firefox-iOS-Sync/115.0b32242 (iPhone; iPhone OS 17.7) (Firefox)`
See the shared spreadsheet in [SYNC-4413](https://mozilla-hub.atlassian.net/browse/SYNC-4413) for details.

As a result of Woothee being unable to parse this non-standard user agent, custom logic was added to fill in the gaps to so that the association logic can work to fill out all the matching enum values.

Subsequent PR will pull this into metarequest and apply the parsing.

## Testing
Added unit tests within `user_agent` to verify new user_agent strings.

## Issue(s)

Closes [SYNC-4414](https://mozilla-hub.atlassian.net/browse/SYNC-4414).

[SYNC-4414]: https://mozilla-hub.atlassian.net/browse/SYNC-4414?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[SYNC-4413]: https://mozilla-hub.atlassian.net/browse/SYNC-4413?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/syncserver/src/server/user_agent.rs
+++ b/syncserver/src/server/user_agent.rs
@@ -97,6 +97,15 @@ impl DeviceInfo {
     }
 }
 
+/// Parses user agents from headers and returns a DeviceInfo struct containing
+/// DeviceFamily, OsFamily, Platform, and Firefox Version.
+///
+/// Intended to handle standard user agent strings but also accomodates the non-standard,
+/// Firefox-specific user agents for iOS and desktop.
+///
+/// Parsing logic for non-standard iOS strings are in the form Firefox-iOS-FxA/24 and
+/// manually modifies WootheeResult to match with correct enums for iOS platform and OS.
+/// FxSync/<...>.desktop result still parses natively with Woothee and doesn't require intervention.
 pub fn get_device_info(user_agent: &str) -> DeviceInfo {
     let mut w_result: WootheeResult<'_> = Parser::new().parse(user_agent).unwrap_or_default();
 
--- a/syncserver/src/server/user_agent.rs
+++ b/syncserver/src/server/user_agent.rs
@@ -1,6 +1,7 @@
 use std::fmt;
 use std::str::FromStr;
 
+use actix_http::header::ACCESS_CONTROL_REQUEST_HEADERS;
 use woothee::parser::{Parser, WootheeResult};
 
 // List of valid user-agent attributes to keep, anything not in this
@@ -55,6 +56,7 @@ pub struct DeviceInfo {
 }
 
 impl DeviceInfo {
+    /// Determine if the device is a desktop device based on either the form factor or OS.
     pub fn is_desktop(&self) -> bool {
         matches!(&self.device_family, DeviceFamily::Desktop)
             || matches!(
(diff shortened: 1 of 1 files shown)


syncserver/src/server/user_agent.rs
```diff
@@ -11,6 +15,124 @@ const VALID_UA_BROWSER: &[&str] = &["Chrome", "Firefox", "Safari", "Opera"];
...
+    /// Determine if the device is an android (Fenix) device based on either the form factor or OS.
+    pub fn is_fenix(&self) -> bool {
+        matches!(&self.device_family, DeviceFamily::Android)
+            && matches!(&self.os_family, OsFamily::Android)
+    }
+}
+
+pub fn get_device_info(user_agent: &str) -> DeviceInfo {
+    let parser = Parser::new();
+    let mut w_result = parser.parse(user_agent).unwrap_or_else(|| WootheeResult {
```
jrconlin, line 90: Since you're only using `parser` once, and it's not a borrow, you could collapse this fairly easily:

```suggestion
    let mut w_result = Parser::new().parse(user_agent).unwrap_or_else(|| WootheeResult {
```

syncserver/src/server/user_agent.rs
```diff
@@ -1,3 +1,7 @@
+use std::fmt;
+use std::str::FromStr;
+
+use actix_http::header::ACCESS_CONTROL_REQUEST_HEADERS;
```
jrconlin, line 4: I don't think we need this, since we don't reference it (kinda surprised that it didn't get flagged by `cargo`
```suggestion
```

syncserver/src/server/user_agent.rs
```diff
@@ -11,6 +14,170 @@ const VALID_UA_BROWSER: &[&str] = &["Chrome", "Firefox", "Safari", "Opera"];
...
+    pub fn is_fenix(&self) -> bool {
+        matches!(&self.device_family, DeviceFamily::Mobile)
+            && matches!(&self.os_family, OsFamily::Android)
+    }
+}
+
+pub fn get_device_info(user_agent: &str) -> DeviceInfo {
+    let mut w_result = Parser::new()
+        .parse(user_agent)
+        .unwrap_or_else(|| WootheeResult {
```
jrconlin, line 104: WootheeResult implements `Default` so this can be a lot easier.
```suggestion
        .unwrap_or_default();
```

jrconlin (changes requested):

syncserver/src/server/user_agent.rs
taddes: Great suggestion, thx 👍

syncserver/src/server/user_agent.rs
```diff
@@ -11,6 +13,164 @@ const VALID_UA_BROWSER: &[&str] = &["Chrome", "Firefox", "Safari", "Opera"];
...
+    /// Determine if the device is iOS based on either the form factor or OS.
+    pub fn is_ios(&self) -> bool {
+        matches!(&self.device_family, DeviceFamily::Mobile)
+            && matches!(&self.os_family, OsFamily::IOS)
+    }
+
+    /// Determine if the device is an android (Fenix) device based on either the form factor or OS.
+    pub fn is_fenix(&self) -> bool {
+        matches!(&self.device_family, DeviceFamily::Mobile)
+            && matches!(&self.os_family, OsFamily::Android)
```
pjenvey, line 96: Let's add a test for `Platform::Other` -- going by the spreadsheet, I'd expect "Mozilla/5.0 (Linux; Android 9; SM-A920F) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/86.0.4216.0 Mobile Safari/537.36" (Android Chrome I believe) to come back as it.

syncserver/src/server/user_agent.rs
```diff
@@ -11,6 +13,164 @@ const VALID_UA_BROWSER: &[&str] = &["Chrome", "Firefox", "Safari", "Opera"];
...
+    }
+}
+
+pub fn get_device_info(user_agent: &str) -> DeviceInfo {
```
pjenvey, line 100: Let's add a note in the docs for this fn that it handles a mix regular user-agents and the sync specific ones. I guess the latter are mostly just the ios "Firefox-iOS-FxA/24" one. Desktop also has its sync specific "FxSync/<...>.desktop" thing but it doesn't end up requiring special handling

pjenvey (changes requested):

syncserver/src/server/user_agent.rs
taddes: I made an implementation change, since the parsing logic was still getting the right OS and device family. I adjusted it to be that if the browser is not Firefox, return the Default implementation of `DeviceInfo` with Other as everything and `FirefoxVersion` of 0, making it easy to screen out, added test too

pjenvey (approved):

jrconlin (approved):

PR #1608, opened 2024-10-10
docs: sync server side metrics adr

taddes:
## Description

Required ADR to validate approach of using Glean for server-side metrics to measure DAU.

The ADR will be created and stored in the syncstorage-rs repo and referenced in our service runbooks and documentation.

## Testing

N/A

## Issue(s)

Closes [SYNC-4434](https://mozilla-hub.atlassian.net/browse/SYNC-4434).

[SYNC-4434]: https://mozilla-hub.atlassian.net/browse/SYNC-4434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/adr/0001-daily-active-use-server-side-metrics-glean.md
+++ b/docs/adr/0001-daily-active-use-server-side-metrics-glean.md
@@ -17,7 +17,6 @@ The goal is to measure DAU (and subsequently WAU & MAU) by emitting metrics from
 * User identifier (hashed_fxa_uid)
 * Timestamp
 * Platform (from UserAgent: Desktop, Fenix, iOS)
-* Collection updated
  
 In researching possible implementation methods, it became clear that many options did not offer us the ease and flexibility to reconcile the data after emission. This is why Glean is recommended as a clear frontrunner. This is not without some drawbacks, but they are minimal compared to other options that would, for example, require considerable data processing and querying difficulties. There is support for this implementation from the Glean team and active support in the process.
 
--- a/docs/adr/0001-daily-active-use-server-side-metrics-glean.md
+++ b/docs/adr/0001-daily-active-use-server-side-metrics-glean.md
@@ -1,0 +1,102 @@
+# Measuring Server-Side Daily Active Use (DAU) With Glean
+
+* Status: proposed
+* Deciders: Taddes Korris, David Durst, JR Conlin, Phillip Jenvey
+* Date: 2024-10-09
+
+Technical Story: 
+[Jira Epic Url](https://mozilla-hub.atlassian.net/browse/SYNC-4389)
+
+
+## Context and Problem Statement
+
+There is a requiement to move away from measuring Sync Daily Active Use (DAU) via FxA.
+The implementation of Relay and additional authentication paths means the metric will no
+longer be accurate. Furthermore, there is a broad movement towards measuring DAU within the service itself. 
+
+The goal is to measure DAU (and subsequently WAU & MAU) by emitting metrics from syncserver-rs itself. This involves 
+
+
+
+## Decision Drivers <!-- optional -->
+
+1. [primary driver, e.g., a force, facing concern, …]
+2. Glean is an internal tool successfully used by many teams at Mozilla.
+3. The Glean team has extensive experience and can provide considerable support in establishing application metrics.
+
+## Considered Options
+
+* A. Glean for server-side measurement of DAU
+* B. StatsD and Grafana
+* C. Sql/Redash
+
+
+## Decision Outcome
+
+Chosen option:
+
+* A. "Glean for server-side measurement of DAU"
+
+[justification. e.g., only option, which meets primary decision driver | which resolves a force or facing concern | … | comes out best (see below)].
+
+### Positive Consequences <!-- optional -->
+
+* [e.g., improvement of quality attribute satisfaction, follow-up decisions required, …]
+* …
+
+### Negative Consequences <!-- optional -->
+
+* [e.g., compromising quality attribute, follow-up decisions required, …]
+* …
+
+## Pros and Cons of the Options <!-- optional -->
+
+### [option A]
+
+[example | description | pointer to more information | …] <!-- optional -->
+
+#### Pros
+
+* [argument for]
+* [argument for]
+* … <!-- numbers of pros can vary -->
+
+#### Cons
+
+* [argument against]
+* … <!-- numbers of cons can vary -->
+
+### [option B]
+
+[example | description | pointer to more information | …] <!-- optional -->
+
+#### Pros
+
+* [argument for]
+* [argument for]
+* … <!-- numbers of pros can vary -->
+
+#### Cons
+
+* [argument against]
+* … <!-- numbers of cons can vary -->
+
+### [option C]
+
+[example | description | pointer to more information | …] <!-- optional -->
+
+#### Pros
+
+* [argument for]
+* [argument for]
+* … <!-- numbers of pros can vary -->
+
+#### Cons
+
+* [argument against]
+* … <!-- numbers of cons can vary -->
+
+## Links 
+
+* [Discussion Document](https://docs.google.com/document/d/1Tk4VIuQZcn8IG-UI38kziZn5e-FMOI0Z-VrvaYTI1SM/edit#heading=h.b0mqx1fng4wa)
+* [FxA DAU Metric in Redash](https://sql.telemetry.mozilla.org/queries/101007/source?p_end%20date=2024-06-26&p_start%20date=2024-05-01#248905)
--- a/docs/adr/0001-daily-active-use-server-side-metrics-glean.md
+++ b/docs/adr/0001-daily-active-use-server-side-metrics-glean.md
@@ -1,6 +1,6 @@
 # Measuring Server-Side Daily Active Use (DAU) With Glean
 
-* Status: proposed
+* Status: approved
 * Deciders: Taddes Korris, David Durst, JR Conlin, Phillip Jenvey
 * Date: 2024-10-09
 
@@ -10,16 +10,15 @@ Technical Story:
 
 ## Context and Problem Statement
 
-There is a requirement to move away from the current measurement of Sync Daily Active Use (DAU), which measures usage via FxA/Mozilla Accounts.
-The addition of Relay, device backups/migrations, through FxA means the metric will no longer an be accurate reflection of Sync usage. This is due to increased browser sign-ins with the potential of overcounting, or not counting those logging into the browser without Sync enabled. Furthermore, there is a broader organizational movement towards measuring DAU within services themselves. 
+There is an organizational requirement for each service to be able to measure its own DAU (Daily Active Users). Sync historically measured DAU via FxA through browser login. With the addition of Relay, device backups/migrations, etc. this will no longer an be accurate reflection of Sync usage, nor a direct measurement originating within Sync. This is due to increased browser sign-ins with the potential of overcounting, or not counting those logging into the browser without Sync enabled. See [Decision Brief - Server-Side Sync Usage Attribution from Mozilla Accounts](https://docs.google.com/document/d/1zD-ia3fP43o-dYpwavDgH5Hb6Xo_fgQzzoWqTiX_wR8/edit?tab=t.0#heading=h.mdoaoiyvqgfo).
 
-The goal is to measure DAU (and subsequently WAU & MAU) by emitting metrics from syncserver-rs itself. This requires the following data:
+The goal is to measure DAU (and subsequently WAU & MAU) by emitting metrics from syncstorage-rs itself. This requires the following data:
 * User identifier (hashed_fxa_uid)
 * Timestamp
-* Platform (from UserAgent: Desktop, Fenix, iOS)
+* Platform (Desktop, Fenix, iOS, Other)
+* Device Family (Desktop, Mobioe, Tablet, Other)
+* Device ID (hashed_device_id) for opt-out/deletion
  
-In researching possible implementation methods, it became clear that many options did not offer us the ease and flexibility to reconcile the data after emission. This is why Glean is recommended as a clear frontrunner. This is not without some drawbacks, but they are minimal compared to other options that would, for example, require considerable data processing and querying difficulties. There is support for this implementation from the Glean team and active support in the process.
-
 ## Decision Drivers
 
 1. Simplicity of implementation, not only for internal metric emission, but for processing and querying.
@@ -30,31 +29,38 @@ In researching possible implementation methods, it became clear that many option
 
 ## Considered Options
 
-* A. Glean
-* B. StatsD and Grafana
-* C. Sql/Redash
+* A. Glean Parser for Rust - Contribute to glean team repo by implementing Rust server
+* B. Custom Glean Implementation - Our own custom implementation, internal only to Sync
+* C. StatsD and Grafana
 
 ## Decision Outcome
 
 Chosen option:
 
-* A. "Glean for server-side measurement of DAU"
+* A. "Glean Parser for Rust: for server-side measurement of DAU"
 
-The use of Glean appears the best choice for measuring internal DAU metrics.  It meets our requirements and provides us with needed support on the data processing side. It also provides considerable support from the Glean team to implement this in a thoughtful manner.  There are some challenges with this implementation (more below), namely in this being a greenfield attempt at Rust server-side metrics, however the pros outweigh the cons. Other metrics implementations like StatsD and Grafana cannot be easily used to measure and aggregate this data. Additionally, it adds considerable overhead in determining how to query the data and reconcile ping emissions of Sync events. Having dedicated organizational support means we establish best practices.
+In researching possible implementation methods, it became clear that many options did not offer us the ease and flexibility to reconcile the data after emission. This is why Glean is recommended as a clear frontrunner, due to its rich tooling in aggregating, querying, and visualizing data.
+
+However, this left an addition decision to either implement our own custom Glean code to emit "Glean-compliant" output, or to contribute to the Glean team's `glean_parser` repository for server-side metrics. The `glean_parser` is used for all server implementations of Glean, since the SDK is only available for client-side applications. Currently, Rust is not supported.
+
+The Glean team does/did not have capacity to implemented the Rust `glean_parser` feature, so we had to decide what is the best solution not only for this use case, but to consider possible future use cases. In consultation with the Glean team, it became clear avoiding our custom implementation and opting for the general-purpose `glean_parser` for Rust was the ideal solution.
 
 ## Pros and Cons of the Options
 
-### A. Glean
+### A. Glean Parser for Rust
 
-Glean is a widely used tool at Mozilla and provides us with a solution to the given issue and possible extensibility in the future. Not without some challenges in initial implementation, but the potential for positive impact is high.
+Glean is a widely used tool at Mozilla and provides us with a solution to the given issue and possible extensibility in the future. Not without some challenges in initial implementation, related to coordination with Glean team and upfront development effort. However, the potential for positive impact within our team and the organization is significant: all Rust server applications will be able to use Glean with full server support, our possible intention to integrate Glean into Push is made easier, and this is done in partnership with the Glean team.
 
 #### Pros
 
-* Satisfaction of requirement to measure internal DAU metrics.
+* Makes Glean compatible for all Rust server applications going forward.
+* Preferred option of the Glean team.
+* Glean team believes, based on FxA metrics volume, that our volume will not be a problem (180-190K per minute).
+* A collection of metrics, emitted as a single "Ping Event" make querying of related data simpler.
 * Core of Glean's purpose is to measure user interactions and the rich metadata that accompanies it.
-* Capacity for future expansion of application metrics within Sync beyond DAU.
+* Capacity for future expansion of application metrics within Sync, beyond DAU.
 * Prepares for implementation of same measurements in autopush, also using Glean.
-* Easier to query.
+* Easier to query, have data team support to set up queries.
 * Use of standardized Mozilla tooling.
 * Establishment of team knowledge of using Glean.
 * Establishment of server-side Rust best practices, leading to easier development for backend Rust applications.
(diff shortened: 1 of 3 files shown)


jrconlin (dismissed):

docs/adr/0001-daily-active-use-server-side-metrics-glean.md
```diff
@@ -0,0 +1,114 @@
...
+2. Reduction of complexity and load on team to process and make sense of data. Ideally there is existing infrastructure to reduce the possible significant work to calculate DAU.
+3. Glean is an internal tool successfully used by many teams at Mozilla.
+4. The Glean team has extensive experience and can provide considerable support in establishing application metrics.
+5. Extensibility: ability to expand scope of metrics to other measurements and apply methodology to other services in Sync ecosystem.
+
+## Considered Options
+
+* A. Glean
+* B. StatsD and Grafana
+* C. Sql/Redash
```
pjenvey, line 36: I'm not sure I understand "Sql/Redash" as an option considered: I see us ultimately using redash to calculate the counts from our new glean metrics, just as it is currently done from FxA's.

docs/adr/0001-daily-active-use-server-side-metrics-glean.md
```diff
@@ -0,0 +1,114 @@
...
+## Context and Problem Statement
+
+There is a requirement to move away from the current measurement of Sync Daily Active Use (DAU), which measures usage via FxA/Mozilla Accounts.
+The addition of Relay, device backups/migrations, through FxA means the metric will no longer an be accurate reflection of Sync usage. This is due to increased browser sign-ins with the potential of overcounting, or not counting those logging into the browser without Sync enabled. Furthermore, there is a broader organizational movement towards measuring DAU within services themselves. 
+
+The goal is to measure DAU (and subsequently WAU & MAU) by emitting metrics from syncserver-rs itself. This requires the following data:
+* User identifier (hashed_fxa_uid)
+* Timestamp
+* Platform (from UserAgent: Desktop, Fenix, iOS)
+* Collection updated
```
pjenvey, line 20: We only came to this conclusion earlier today but I'll note it's looking like we'll wanna remove this line (and the according entry in the schema) at some point

docs/adr/0001-daily-active-use-server-side-metrics-glean.md
taddes: Ah gotcha, I thought it was a different flow and we use Looker/BigQuery to get that result. Is that not the case? Again I've not used Redash enough for observing our metrics.  I'm happy to remove this

pjenvey: Ah, I see where you're coming from: Looker vs Redash

There's a lot of overlap between the two, both are typically pulling from the same BigQuery sources. E.g. this [Sync DAU in looker](https://mozilla.cloud.looker.com/dashboards/1091?Submission%20Date=7%20day&Legacy%20Submission%20Date=2024%2F01%2F01%20to%202024%2F02%2F01) seems similar to the [Redash version](https://sql.telemetry.mozilla.org/queries/101007/source?p_end%20date=2024-06-26&p_start%20date=2024-05-01#248905), they're both reading from the same `account_events` data. I almost think of them as different UIs for I believe largely the same data sources.

(I've even seen it [suggested](https://mozilla-hub.atlassian.net/wiki/spaces/SRE/pages/27922678/STMO+-+Redash+sql.telemetry.mozilla.org#STMO-Redash(sql.telemetry.mozilla.org)-Description) that eventually Redash usage will all move to Looker but I'm not sure that will ever really happen)

I'm fine with building the new dashboard in Looker if that's the plan (or either) but I'd argue the cons section of Sql/Redash here probably don't apply? Our new glean events seem fairly similar to the one's we're relying on from fxa, so I imagine implementing the dashboard in Redash wouldn't be too difficult?

docs/adr/0001-daily-active-use-server-side-metrics-glean.md
```diff
@@ -0,0 +1,113 @@
...
+* Deciders: Taddes Korris, David Durst, JR Conlin, Phillip Jenvey
+* Date: 2024-10-09
+
+Technical Story: 
+[Jira Epic Url](https://mozilla-hub.atlassian.net/browse/SYNC-4389)
+
+
+## Context and Problem Statement
+
+There is a requirement to move away from the current measurement of Sync Daily Active Use (DAU), which measures usage via FxA/Mozilla Accounts.
```
ddurst, line 13: Might be better to say that the requirement is for each service to be able to measure their use directly. FxA will still be able to measure this, and we should be able to compare our measurement with theirs.

docs/adr/0001-daily-active-use-server-side-metrics-glean.md
```diff
@@ -0,0 +1,113 @@
...
+#### Cons
+
+* The `glean_parser` tool used by other Glean applications is currently not supported for Rust. Furthermore, client applications can use the Glean SDK and this is also not an option for us.
+* Server side metrics have not yet been implemented for a Rust server application of this kind, so this is new territory.
+* There is added complexity of data review process and registration of the application to Glean's probe scraper.
+* Potential delays and challenges in new implementation.
+
+### StatsD and Grafana
+
+StatsD and Grafana offer us core application metrics and service health. While we use this frequently, it doesn't neatly fit the measurement requirements we have for DAU and would likely be very difficult to process via queries.
```
ddurst, line 72: We should have here (briefly) _how_ it doesn't fit the measurement requirements and/or query complication.

docs/adr/0001-daily-active-use-server-side-metrics-glean.md
```diff
@@ -0,0 +1,113 @@
...
+
+#### Pros
+
+* Well understood and used.
+* Support for SRE for more complex queries.
+* Already utilized for core metrics.
```
ddurst, line 78: Core metrics of the service?

docs/adr/0001-daily-active-use-server-side-metrics-glean.md
```diff
@@ -0,0 +1,113 @@
...
+
+#### Cons
+
+* StatsD is not a good format for measuring something like user interactions.
+* Somewhat opaque and complicated query logic required.
+* Significant difficulty in aggregation and reconciliation logic. 
+* May not scale well given number of events.
```
ddurst, line 85: Might be good to say more about this: we haven't talked about the expected number of events and where this might hit a limit.

docs/adr/0001-daily-active-use-server-side-metrics-glean.md
```diff
@@ -0,0 +1,113 @@
...
+* Likely considerable overhead in understanding how to make sense of data.
+* Heavier load for team to manage data processing.
```
ddurst, line 87: Why?

docs/adr/0001-daily-active-use-server-side-metrics-glean.md
```diff
@@ -0,0 +1,113 @@
...
+
+### SQL/Redash
+
+The current DAU metric used from FxA uses SQL telemetry and provides the ability to query data.  It is then displayed in a redash panel. While convenient, we do not have the infrastructure in place for this option and it might involve considerable effort to establish.
```
ddurst, line 91: Following Phil's comment, it sounds like we can remove SQL/Redash as an option.

docs/adr/0001-daily-active-use-server-side-metrics-glean.md
```diff
@@ -0,0 +1,113 @@
...
+The addition of Relay, device backups/migrations, through FxA means the metric will no longer an be accurate reflection of Sync usage. This is due to increased browser sign-ins with the potential of overcounting, or not counting those logging into the browser without Sync enabled. Furthermore, there is a broader organizational movement towards measuring DAU within services themselves. 
+
+The goal is to measure DAU (and subsequently WAU & MAU) by emitting metrics from syncserver-rs itself. This requires the following data:
+* User identifier (hashed_fxa_uid)
+* Timestamp
+* Platform (from UserAgent: Desktop, Fenix, iOS)
+ 
+In researching possible implementation methods, it became clear that many options did not offer us the ease and flexibility to reconcile the data after emission. This is why Glean is recommended as a clear frontrunner. This is not without some drawbacks, but they are minimal compared to other options that would, for example, require considerable data processing and querying difficulties. There is support for this implementation from the Glean team and active support in the process.
```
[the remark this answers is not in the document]
ddurst: What I don't see, and I still don't understand, is why the Glean team is interested in this. Nor do I see why Rust is "harder" (wrt to the parser). Those two things probably support this argument, but it's not clear to me yet why or how.

ddurst (changes requested):
I think this can be reduced, focused, and specified a bit more to explain why we want to make this change in this way.

docs/adr/0001-daily-active-use-server-side-metrics-glean.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Context and Problem Statement
+
+There is a requirement to move away from the current measurement of Sync Daily Active Use (DAU), which measures usage via FxA/Mozilla Accounts.
+The addition of Relay, device backups/migrations, through FxA means the metric will no longer an be accurate reflection of Sync usage. This is due to increased browser sign-ins with the potential of overcounting, or not counting those logging into the browser without Sync enabled. Furthermore, there is a broader organizational movement towards measuring DAU within services themselves. 
+
+The goal is to measure DAU (and subsequently WAU & MAU) by emitting metrics from syncserver-rs itself. This requires the following data:
+* User identifier (hashed_fxa_uid)
+* Timestamp
+* Platform (from UserAgent: Desktop, Fenix, iOS)
```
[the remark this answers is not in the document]
taddes: Yeah, it's meant to emulate the FxA query and as we discussed, something we want to know per platform.

docs/adr/0001-daily-active-use-server-side-metrics-glean.md
```diff
@@ -0,0 +1,113 @@
...
+* Date: 2024-10-09
+
+Technical Story: 
+[Jira Epic Url](https://mozilla-hub.atlassian.net/browse/SYNC-4389)
+
+
+## Context and Problem Statement
+
+There is a requirement to move away from the current measurement of Sync Daily Active Use (DAU), which measures usage via FxA/Mozilla Accounts.
+The addition of Relay, device backups/migrations, through FxA means the metric will no longer an be accurate reflection of Sync usage. This is due to increased browser sign-ins with the potential of overcounting, or not counting those logging into the browser without Sync enabled. Furthermore, there is a broader organizational movement towards measuring DAU within services themselves.
```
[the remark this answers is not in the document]
taddes: Sure, can trim. I was going off of the discussion in the Server-Side Attribution Decision Brief to clarify these points.

docs/adr/0001-daily-active-use-server-side-metrics-glean.md
```diff
@@ -0,0 +1,113 @@
...
+* Prepares for implementation of same measurements in autopush, also using Glean.
+* Easier to query.
+* Use of standardized Mozilla tooling.
+* Establishment of team knowledge of using Glean.
+* Establishment of server-side Rust best practices, leading to easier development for backend Rust applications.
+* Transparency in data review process with consideration made to minimum collection of data.
+
+#### Cons
+
+* The `glean_parser` tool used by other Glean applications is currently not supported for Rust. Furthermore, client applications can use the Glean SDK and this is also not an option for us.
```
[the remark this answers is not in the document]
jrconlin: I think we should be more clear here. There is no immediate support for the traditional `glean_parser` code generation for Rust. Rust is more than capable of generating output that is parseable by Glean. Ideally, the Rust code would be generated from the XML declaration file using a combination of jinja2 templates and other tooling which the Sync team is unfamiliar with. The output of which would still need to be validated by the Glean team. (I'll note that if we were to generate the `glean_parser` code generation element, I presume that the Glean team would also want to review that.) 

(Somewhat personally, rust, like java, does not take well to the idea of "automatically generated and inserted code", so the code would need to be pre-generated and made part of the repo before it could be tested and deployed. This would be so that all the additional tooling, such as CVE audit checks, formatting checks, and other code quality checks are performed on the entire code base.)

docs/adr/0001-daily-active-use-server-side-metrics-glean.md
```diff
@@ -0,0 +1,113 @@
...
+* Server side metrics have not yet been implemented for a Rust server application of this kind, so this is new territory.
+* There is added complexity of data review process and registration of the application to Glean's probe scraper.
```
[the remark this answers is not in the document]
jrconlin: I think this is a summation of what I mention above. In essence, this is a bit of "greenfield" development (where the language or platform are "new" or unsupported by existing `glean_parser` generators). Glean does not offer a development suite of test inputs and output validators that folk could use to develop new implementations of the `glean_parser` structures. Any work done, outside of the Glean team, would need to be validated by the Glean team before it was introducted into production code. Without that sort of initial help provided by developmental tooling, it's prudent to presume that initial efforts will be incorrect and require several rounds of refinement.

taddes:
Hey folks, have the updated ADR here for review as we've finalized the approach with Glean team

docs/adr/0001-daily-active-use-server-side-metrics-glean.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Decision Outcome
+
+Chosen option:
+
+* A. "Glean Parser for Rust: for server-side measurement of DAU"
+
+In researching possible implementation methods, it became clear that many options did not offer us the ease and flexibility to reconcile the data after emission. This is why Glean is recommended as a clear frontrunner, due to its rich tooling in aggregating, querying, and visualizing data.
+
+However, this left an addition decision to either implement our own custom Glean code to emit "Glean-compliant" output, or to contribute to the Glean team's `glean_parser` repository for server-side metrics. The `glean_parser` is used for all server implementations of Glean, since the SDK is only available for client-side applications. Currently, Rust is not supported.
```
ddurst, line 44: nit: s/addition/additional/

docs/adr/0001-daily-active-use-server-side-metrics-glean.md
```diff
@@ -0,0 +1,121 @@
...
+
+The Glean team does/did not have capacity to implemented the Rust `glean_parser` feature, so we had to decide what is the best solution not only for this use case, but to consider possible future use cases. In consultation with the Glean team, it became clear avoiding our custom implementation and opting for the general-purpose `glean_parser` for Rust was the ideal solution.
```
ddurst, line 46: nit: s/implemented/implement/

semantic nit: I'm not sure that the lack of capacity in the Glean team informs anything about future use cases.

The last sentence seems off to me. Something learned in consultation with the Glean team steered us to avoid a custom implementation, and that's why the glean parser route is chosen -- it's hinted at in "became clear" and such. I think we should just state plainly why the glean parser route was deemed better.

docs/adr/0001-daily-active-use-server-side-metrics-glean.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Pros and Cons of the Options
+
+### A. Glean Parser for Rust
+
+Glean is a widely used tool at Mozilla and provides us with a solution to the given issue and possible extensibility in the future. Not without some challenges in initial implementation, related to coordination with Glean team and upfront development effort. However, the potential for positive impact within our team and the organization is significant: all Rust server applications will be able to use Glean with full server support, our possible intention to integrate Glean into Push is made easier, and this is done in partnership with the Glean team.
```
ddurst, line 52: nit: second sentence is a phrase; maybe rework it to more simply make the point.

docs/adr/0001-daily-active-use-server-side-metrics-glean.md
```diff
@@ -0,0 +1,113 @@
...
+### A. Glean
+
+Glean is a widely used tool at Mozilla and provides us with a solution to the given issue and possible extensibility in the future. Not without some challenges in initial implementation, but the potential for positive impact is high.
+
+#### Pros
+
+* Satisfaction of requirement to measure internal DAU metrics.
+* Core of Glean's purpose is to measure user interactions and the rich metadata that accompanies it.
+* Capacity for future expansion of application metrics within Sync beyond DAU.
+* Prepares for implementation of same measurements in autopush, also using Glean.
```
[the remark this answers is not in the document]
ddurst: I guess the easy path (to me) is just to lose the bullet "Prepares for implementation of same measurements in autopush, also using Glean", as it's covered by the first bullet.

docs/adr/0001-daily-active-use-server-side-metrics-glean.md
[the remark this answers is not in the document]
ddurst: There's one big "con" here, and that's that we have to write something -- we don't get Glean for free via glean_parser. SDK is not part of the considered option here, so that sentence should go entirely.

Re: clarity, I don't think the first sentence is substantially different from jr's second sentence. We just need to state the big drawback of this option; other options aren't relevant to the discussion here.

docs/adr/0001-daily-active-use-server-side-metrics-glean.md
```diff
@@ -0,0 +1,121 @@
...
+* Establishment of server-side Rust best practices, leading to easier development for backend Rust applications.
+* Transparency in data review process with consideration made to minimum collection of data.
+
+#### Cons
+
+* The `glean_parser` tool used by other Glean applications is currently not supported for Rust. Furthermore, client applications can use the Glean SDK and this is also not an option for us.
+* Server side metrics have not yet been implemented for a Rust server application of this kind, so this is new territory.
+* There is added complexity of data review process and registration of the application to Glean's probe scraper, PubSub topics, ingestion pipeline etc.
+* May require more uptfront development effort to get up to speed with new code and process.
+* Some concerns exist around volume of data emitted from the service and if it is feasible, but we won't know until we try.
```
ddurst, line 75: Is this last bullet still a concern? Earlier we said that the Glean team was confident. If there's a risk here, we should state it -- like if we're wrong about the volume -- but if we are also confident in the volume estimate given to the Glean team, maybe we don't need this bullet.

docs/adr/0001-daily-active-use-server-side-metrics-glean.md
```diff
@@ -0,0 +1,121 @@
...
+
+### B. Custom Glean Implementation 
+
+This was originally appearing to be the desired approach to measuring DAU via Glean.  This was predicated on the ease of prototyping a custom implementation that imitated the Glean team's logic to create "Glean-compliant" output that could be configured to be ingested.  However, in consulting with the Glean team and evaluating the pros and cons of this approach, it became clear this approach had considerably more drawbacks than implementing the `glean_parser` for Rust. These drawbacks were a lack of testing, validation, less support from the Glean team, and the potential problems with maintenance and adding Glean metrics in the future.
```
ddurst, line 79: nit: this is a bit verbose for what it's trying to do, but I'm not gonna hold up the ADR on a revision of it.

docs/adr/0001-daily-active-use-server-side-metrics-glean.md
```diff
@@ -0,0 +1,121 @@
...
+
+#### Pros
+* Gives team control over implementation and allows us to customize the Glean logging as we see fit.
+* Does not require contribution to Glean team's repos, which potentially limits the scope understanding a new codebase.
```
ddurst, line 83: > which potentially limits the scope understanding a new codebase

Missing a word here -- "scope of"? Are you trying to say that the scope / effort is smaller because it doesn't require integrating with an existing product/repo? If so, just say that directly.

docs/adr/0001-daily-active-use-server-side-metrics-glean.md
```diff
@@ -0,0 +1,121 @@
...
+* Easy to prototype and make changes.
+* Doesn't require understanding the templating logic and libraries in the `glean_parser`
```
ddurst, line 85: Seems like this goes along with the "scope" comment before. Maybe they can be combined, with this as an example of the increased scope.

docs/adr/0001-daily-active-use-server-side-metrics-glean.md
```diff
@@ -0,0 +1,121 @@
...
+
+#### Cons
+* There is no built-in testing suite or validation, so this would put a larger development burden on us and require the Glean team's review.
+* Lack of testing and validation means higher likelihood of bugs.
+* If we decide to add new Glean metrics in the future, this may break the custom implementation and impose a greater maintenance overhead.
+* Time required to understand the Glean team's implementations anyways, in order to replicate behavior and data structures.
```
ddurst, line 91: nit: "anyways" bothers me. But I think the larger point here is that we still have the overhead of needing to understand the Glean team's implementations.

docs/adr/0001-daily-active-use-server-side-metrics-glean.md
```diff
@@ -0,0 +1,121 @@
...
+
+#### Cons
+
+* StatsD is not a good format for measuring something like user interactions.
+* Somewhat opaque and complicated query logic required.
+* Significant difficulty in aggregation and reconciliation logic. 
+* May not scale well given number of events.
+* Likely a considerable overhead in understanding how to make sense of data.
+* Heavier load for team to manage data processing, as this approach has not been tried.
+
```
ddurst, line 112: This pros/cons list is great. Straight and to the point.

ddurst (dismissed):
Just a few notes, but overall I like it. r+wc

jrconlin (approved):

docs/adr/0001-daily-active-use-server-side-metrics-glean.md
taddes: I tightened it up a bit nonetheless, good point

docs/adr/0001-daily-active-use-server-side-metrics-glean.md
taddes: Originally it may have been a concern, but from the Glean team's point of view (and the volume related to FxA) we're probably good, however we really won't know until we see the output. I think this warrants removing this point.

jrconlin (approved):

ddurst (approved):

PR #1614, opened 2024-10-22
feat: revert "feat: Revert "chore: revert back to bullseye (for now) 

pjenvey:
Just to be 100% on the safe side, let's go back to the old Debian bullseye until tokenserver is fully stabilized on prod.

We know this bookworm/mysql client upgrade did trigger some differences in establishing connections (they took longer to establish the max 100 we use). Tokenserver should be able to handle this difference, but let's ensure the blocking thread pool changes are successful before reattempting this bookworm/client upgrade.

It could also be useful to see the new blocking thread pool metrics before/after the bookworm upgrade, so it makes sense to revert back to the older debian/client for now.

## Issue(s)

Issue SYNC-4363


taddes (approved):
👍

PR #1616, opened 2024-10-23
wip: feat: add metric values to get_collections

taddes:
## Description

Add required metric values to get_collections call for metric emission

## Testing

N/A

## Issue(s)

Closes [SYNC-4441](https://mozilla-hub.atlassian.net/browse/SYNC-4441).

### DIFF
--- a/syncserver/src/web/handlers.rs
+++ b/syncserver/src/web/handlers.rs
@@ -3,7 +3,9 @@ use std::collections::HashMap;
 use std::convert::Into;
 use std::time::{Duration, Instant};
 
+use crate::server::user_agent::get_device_info;
 use actix_web::{http::StatusCode, web::Data, HttpRequest, HttpResponse, HttpResponseBuilder};
+use http::header;
 use serde::Serialize;
 use serde_json::{json, Value};
 use syncserver_common::{X_LAST_MODIFIED, X_WEAVE_NEXT_OFFSET, X_WEAVE_RECORDS};
@@ -32,6 +34,15 @@ pub async fn get_collections(
     db_pool: DbTransactionPool,
     request: HttpRequest,
 ) -> Result<HttpResponse, ApiError> {
+    let _hashed_fxa_uid = String::from(&meta.user_id.hashed_fxa_uid);
+    let _user_agent = request
+        .headers()
+        .get(actix_http::header::USER_AGENT)
+        .unwrap()
+        .to_str();
+
+    let _device_info = get_device_info(_user_agent?);
+
     db_pool
         .transaction_http(request, |db| async move {
             meta.emit_api_metric("request.get_collections");


syncserver/src/web/handlers.rs
```diff
@@ -32,6 +33,15 @@ pub async fn get_collections(
     db_pool: DbTransactionPool,
     request: HttpRequest,
 ) -> Result<HttpResponse, ApiError> {
+    let _hashed_fxa_uid = String::from(&meta.user_id.hashed_fxa_uid);
+    let _user_agent = request
+        .headers()
+        .get("USER_AGENT")
```
pjenvey, line 39: prefer the constant in `actix_web::http::header`
```suggestion
        .get(header::USER_AGENT)
```

syncserver/src/web/handlers.rs
```diff
@@ -32,6 +33,15 @@ pub async fn get_collections(
     db_pool: DbTransactionPool,
     request: HttpRequest,
 ) -> Result<HttpResponse, ApiError> {
+    let _hashed_fxa_uid = String::from(&meta.user_id.hashed_fxa_uid);
```
pjenvey, line 36: This line likely changing anyway but if a copy's actually needed `.clone()` is more idiomatic

syncserver/src/web/handlers.rs
```diff
@@ -32,6 +33,15 @@ pub async fn get_collections(
...
+    let _user_agent = request
+        .headers()
+        .get("USER_AGENT")
+        .map(|header| header.to_str().unwrap_or("none"))
```
pjenvey, line 40: I'll always argue `and_then` w/ `ok` is worth doing in cases like these so you don't have to duplicate the final `unwrap_or` case
```suggestion
        .and_then(|header| header.to_str().ok())
```

syncserver/src/web/handlers.rs
taddes: I was thinking when I wrote this that I should use the enum 👍

syncserver/src/web/handlers.rs
taddes: Funny enough I talked to @linabutler abut this and she suggested the same, better pattern for sure.

taddes:
There's more I could do here with piping to STDOUT and testing that @pjenvey , but I don't think it makes sense since it would be an intermediate step with the Glean stuff coming up soon. We'll just pass the values to the individual metric payloads.

Let me know if you think this needs anything more in the interim before we add the Glean logic.

syncserver/src/web/handlers.rs
```diff
@@ -32,6 +37,15 @@ pub async fn get_collections(
     db_pool: DbTransactionPool,
     request: HttpRequest,
 ) -> Result<HttpResponse, ApiError> {
+    let _hashed_fxa_uid: String = meta.user_id.hashed_fxa_uid.clone();
```
jrconlin, line 40: A bit of a nit: Since these are prelude values that we're fetching, can we add a comment noting why we're cloning values we're immediately discarding? 
(That or we could add a comment as well as the commented prelude code, but I know that some folk are not fans of commented code blocks.)

taddes: Sure, I can add a comment but this will go away right away as these are prefaced with `_` so clippy doesn't get mad. I want everything in place so that we can pass these into the glean code.

jrconlin: Yep, totally understandable. 
One thing that @pjenvey drilled into my thick head was to make sure that each commit can "stand alone". Basically, if you were to run away to join the circus, could we figure out what you were doing six months from now?

It's surprisingly effective.

taddes: Absolutely, that's a good rule to live by. Done. Let me know if the comment looks good for the interim.

jrconlin (approved):

PR #1618, opened 2024-10-25
Chore/2410 update

jrconlin:
## Description

Update py03

### DIFF
--- a/tokenserver-auth/src/oauth/py.rs
+++ b/tokenserver-auth/src/oauth/py.rs
@@ -21,7 +21,7 @@ use std::{sync::Arc, time::Duration};
 pub struct Verifier {
     // Note that we do not need to use an Arc here, since Py is already a reference-counted
     // pointer
-    inner: Py<PyAny>,
+    inner: Arc<Py<PyAny>>,
     timeout: u64,
     blocking_threadpool: Arc<BlockingThreadpool>,
 }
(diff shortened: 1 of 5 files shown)


tokenserver-auth/src/oauth/py.rs
```diff
@@ -21,7 +21,7 @@ use std::{sync::Arc, time::Duration};
 pub struct Verifier {
     // Note that we do not need to use an Arc here, since Py is already a reference-counted
```
pjenvey, line 22:
```suggestion
```

and note to myself: [Pyo3's migration guide](https://pyo3.rs/main/migration#pyclone-is-now-gated-behind-the-py-clone-feature) recommended moving to Arc<Py<>>

pjenvey (changes requested):
thanks for making this very limited in scope but I might wait till we get the tokenserver sentry client bits sorted out/released before merging this anyway

tokenserver-auth/src/oauth/py.rs
jrconlin: Thanks! forgot to clean up that comment.

jrconlin:
Yeah, zero rush on landing this. I figured I'd just do it quick while it was top of mind.

pjenvey (approved):

PR #1619, opened 2024-10-29
Rename MySQL error to use a more generic name

Eragonfr:
## Description

The MysqlError and related types can be reused by other SQL based backends. See PR #1520

## Testing

Test that everything still works like on master

## Issue(s)

Not really an issue but [here's the review](https://github.com/mozilla-services/syncstorage-rs/pull/1520#discussion_r1817098996) about splitting this from #1520

### DIFF
--- a/syncserver-db-common/src/error.rs
+++ b/syncserver-db-common/src/error.rs
@@ -5,17 +5,17 @@ use http::StatusCode;
 use syncserver_common::{from_error, impl_fmt_display, ReportableError};
 use thiserror::Error;
 
-/// Error specific to any MySQL database backend. These errors are not related to the syncstorage
+/// Error specific to any SQL database backend. These errors are not related to the syncstorage
 /// or tokenserver application logic; rather, they are lower-level errors arising from diesel.
 #[derive(Debug)]
-pub struct MysqlError {
-    kind: MysqlErrorKind,
+pub struct SqlError {
+    kind: SqlErrorKind,
     pub status: StatusCode,
     pub backtrace: Backtrace,
 }
 
 #[derive(Debug, Error)]
-enum MysqlErrorKind {
+enum SqlErrorKind {
     #[error("A database error occurred: {}", _0)]
     DieselQuery(#[from] diesel::result::Error),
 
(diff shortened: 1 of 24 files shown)


jrconlin:
For additional context:

eragonfr is working on the sqlite backend for syncstorage. One of the refactoring changes she would like to make is to qualify the `mysql` errors from general `SQL` errors so that she can introduce `sqlite` errors. Since hers will be the third storage engine, the refactor makes sense to me. 

I'd like to get y'all's opinion as well.

syncserver-db-common/src/error.rs
```diff
@@ -39,22 +39,22 @@ impl From<MysqlErrorKind> for MysqlError {
...
-            MysqlErrorKind::Pool(_) => false,
+            SqlErrorKind::Pool(_) => false,
             _ => true,
         }
     }
 
     fn metric_label(&self) -> Option<String> {
         Some(
             match self.kind {
-                MysqlErrorKind::DieselQuery(_) => "storage.mysql.error.diesel_query",
```
jrconlin, line 54: These metric changes do not impact our dashboards.

jrconlin (approved):

jrconlin:
I'll poke the other two to get another review before I land this. (Sorry, things have been a bit chaotic over here.)

syncserver-db-common/src/error.rs
```diff
@@ -5,17 +5,17 @@ use http::StatusCode;
 use syncserver_common::{from_error, impl_fmt_display, ReportableError};
 use thiserror::Error;
 
-/// Error specific to any MySQL database backend. These errors are not related to the syncstorage
+/// Error specific to any SQL database backend. These errors are not related to the syncstorage
 /// or tokenserver application logic; rather, they are lower-level errors arising from diesel.
```
pjenvey, line 9: This comment still stands and I think suggests a clearer name for this: `DieselError`, how does that sound?

taddes (approved):
👍

syncserver-db-common/src/error.rs
jrconlin: Honestly, I'm fine either way, but I do appreciate the concept of having a more generic `SqlError` for the reason that @Eragonfr notes. I don't really see us replacing diesel any time soon, but the sources of the error being reported are coming from the underlying SQL system. `DieselError` type should be reserved for errors arising from the diesel system directly (e.g. failure to connect to the data store, or some configuration break).

Issue #772, opened 2020-08-10
"Invalid Accept" should be more descriptive


jrconlin:
https://sentry.prod.mozaws.net/operations/syncstorage-stage/issues/9334203/?environment=release&referrer=alert_email

This error should be a bit more descriptive so that future folk understand what it means or at least how to best find it.

e.g. "Invalid value specified for 'Accept` header in Collection request" or something.

(May want to be a bit more clear with other errors as well)

fzzzy:
It appears this was already closed by https://github.com/mozilla-services/syncstorage-rs/pull/784

However, the syntax "Closes #xyz, #abc" is not a syntax that works to automatically close #abc.

PR #1626, opened 2024-11-04
feat: glean metrics logic

taddes:
## Description

Core Glean logic to process metrics and events.

Related to the PR presently in the glean_parser repo [#772](https://github.com/mozilla/glean_parser/pull/772) for permanent rust server metric capabilities, so changes there may propagate to here 

## Issue(s)

Closes [SYNC-4440](https://mozilla-hub.atlassian.net/browse/SYNC-4440).

[SYNC-4440]: https://mozilla-hub.atlassian.net/browse/SYNC-4440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


glean/src/server_events.rs
```diff
@@ -0,0 +1,310 @@
...
+struct LogEnvelope {
+    timestamp: String,
+    logger: String,
+    #[serde(rename = "type")]
+    log_type: String,
+    fields: Ping,
+}
+
+impl GleanEventsLogger {
+    fn create_client_info(&self) -> ClientInfo {
```
jrconlin, line 105: I would strongly recommend converting this to an `impl Default` unless there are more commonly used values to initialize this struct.

glean/src/server_events.rs
```diff
@@ -0,0 +1,309 @@
...
+            os: "Unknown".to_string(),
+            os_version: "Unknown".to_string(),
+            architecture: "Unknown".to_string(),
+            app_build: "Unknown".to_string(),
+            app_display_version: self.app_display_version.clone(),
+            app_channel: self.app_channel.clone(),
+        }
+    }
+
+    fn create_ping_info() -> PingInfo {
```
jrconlin, line 119: Again, I'd strongly recommend converting this to an `impl Default`.

glean/src/server_events.rs
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    fn create_ping(
+        &self,
+        document_type: &str,
+        config: &RequestInfo,
+        payload: &PingPayload,
+    ) -> Ping {
+        let payload_json =
+            serde_json::to_string(payload).expect("unable to marshal payload to json.");
```
jrconlin, line 136: We should consider converting this to a `thiserror` derived struct so that it can be captured by calling systems and retried. An `.expect()` is a glorified panic and will kill whatever thread or process it's in.

glean/src/server_events.rs
```diff
@@ -0,0 +1,309 @@
...
+        let document_id = Uuid::new_v4().to_string();
```
jrconlin, line 137: We can move this to inside the Ping struct declaration. There's no need to create the variable since we're not reusing it and there shouldn't be any lifetime considerations.

glean/src/server_events.rs
```diff
@@ -0,0 +1,309 @@
...
+        // Any `extra_keys` will be output below to be inserted into `extra`.
+        // If there are none, an empty, immutable HashMap is created.
+        let extra: HashMap<String, String> = HashMap::new();
+
+        new_glean_event("syncstorage", "get_collections", extra)
+    }
+}
+
+/// Marker trait for events per ping
+pub trait EventsPingEvent {
```
jrconlin, line 201: I'm not sure we need this trait definition. 
All it does is generate a GleanEvent, which could be done using a Default constructor.

glean/src/server_events.rs
```diff
@@ -0,0 +1,309 @@
...
+        );
+        string_map.insert(
+            "syncstorage.platform".to_string(),
+            serde_json::Value::String(params.syncstorage_platform.to_string()),
+        );
+        metrics.insert("string".to_string(), string_map);
+
+        let mut events: Vec<GleanEvent> = Vec::new();
+        if let Some(event) = &params.event {
+            events.push(event.glean_event());
```
jrconlin, line 242: Unless we have a lot of different `EventsPingEvent` types, this will always return basically the same thing, a default GleanEvent. (Even then, there's a good argument that YAGNI and you could use a default pattern for it as well.)

Could we just do 
```rust
   events.push(GleanEvent{
    category: "syncstorage".to_owned(),
    name: "get_collections".to_owned(),
    ..Default::default()
   })
```
here?

glean/src/server_events.rs
```diff
@@ -0,0 +1,309 @@
...
+
+// Record and submit `sync-dau` ping
+impl GleanEventsLogger {
+    /// General `record_events_ping` function
+    pub fn record_sync_dau_ping(&self, request_info: &RequestInfo, params: &SyncDauPing) {
+        // Define the outer `Metrics` map that holds the metric type.
+        let mut metrics = Metrics::new();
+        // Create the inner metric value map to insert into `Metrics`.
+        metrics.insert(
+            "string".to_string(),
```
jrconlin, line 271: a bit nitty, but we generally call `to_owned()` instead of `to_string()`. Both are equally performant. `to_owned()` reflects that we're converting a static string reference (that's what "string" is) to a own-able String object. Since `insert(...)` essentially is consuming these values, it's semantically more correct.

jrconlin (changes requested):

syncserver/src/web/handlers.rs
```diff
@@ -40,14 +44,35 @@ pub async fn get_collections(
...
+    let hashed_device_id: String = meta.user_id.hashed_device_id.clone();
     let user_agent = request
         .headers()
         .get(header::USER_AGENT)
         .and_then(|header| header.to_str().ok())
         .unwrap_or("none");
-    let _device_info = get_device_info(user_agent);
+    let device_info: DeviceInfo = get_device_info(user_agent);
+
+    let logger: GleanEventsLogger = GleanEventsLogger {
```
pjenvey, line 56: Let's construct this only once as a singleton living on `ServerState`.

syncserver/src/web/handlers.rs
```diff
@@ -40,14 +44,35 @@ pub async fn get_collections(
...
+        // app id will be supplied when added to probe-scraper
+        app_id: "syncstorage".to_string(),
+        app_display_version: "1.0.0".to_string(),
```
pjenvey, line 59: Take a look at the `logging` module on what we use for its similar values. You probably want:
```suggestion
        app_id: env!("CARGO_PKG_NAME").to_owned(),
        app_display_version: env!("CARGO_PKG_VERSION").to_owned(),
```

syncserver/src/web/handlers.rs
```diff
@@ -40,14 +44,35 @@ pub async fn get_collections(
...
+        app_channel: "prod".to_string(),
```
pjenvey, line 60: We have a `SENTRY_ENVIRONMENT` env var which could be used here but we should probably just add a new `SYNC_ENV` var to the helm chart (like merino has).

pjenvey (changes requested):

syncserver/src/web/handlers.rs
taddes: `app_id` is actually what the application is defined as in the `probe-scraper` configuration: https://github.com/mozilla/probe-scraper/pull/834/files , but can use the `app_display_version` as you show above.

syncserver/src/web/handlers.rs
pjenvey: hardcoding the `app_id` to `syncstorage` wfm, I just mean something else for the `app_channel` of "prod"

glean/metrics.yaml
```diff
@@ -53,6 +57,10 @@ syncstorage:
       Platform from which sync action was initiated.
       Firefox Desktop, Fenix, or Firefox iOS.
     # yamllint enable
+    lifetime: application
+    send_in_pings:
+      - sync-dau
+      - events
     notification_emails:
       - jrconlin@mozilla.com
```
pjenvey, line 65: We could probably switch to the new mailing list for these notification_emails?

glean/src/server_events.rs
taddes: Totally, this one makes perfect sense.

glean/src/server_events.rs
taddes: In discussing, and if ok with you, were gonna leave this one as it is closely tied to the `GleanEventsLogger` struct and needs values from the instance of said `GleanEventsLogger`, so wouldn't provide us with what we're looking for.

glean/src/server_events.rs
taddes: This is auto generated for the many possible `metric.yaml` defined `Events` (which in this case really refers to pings, since events plural connotes not the metric type of event, but the ping event). This is to say that the marker trait is generated with the `glean_parser` code which we want to leave in place, just for conformity if that's ok

glean/src/server_events.rs
pjenvey: I was ok with these as I can't imagine what could actually trigger the 2 expects generated in this file, maybe if you really screwed up your glean yaml/schema? In that case I'm ok with panic'ing -- and a nice side benefit is getting away without implementing any actual Error stuff.

glean/src/server_events.rs
```diff
@@ -0,0 +1,308 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// Required imports
+use chrono::Utc;
+use serde::{Deserialize, Serialize};
+use serde_json;
```
jrconlin, line 8: Normally you don't have to specify crate roots like this. You only need to specify the various elements you're importing. Rust will automatically associate crates when you specify the root (like you do later for `Value`.)

If you want, you could specify
```suggestion
use serde_json::{Value, to_string};
```
and then drop the crate prefix from those uses later.

(I'm surprised that clippy didn't yell about this.)

glean/src/server_events.rs
```diff
@@ -0,0 +1,308 @@
...
+    // Application Id to identify application per Glean standards
+    pub app_id: String,
+    // Version of application emitting the event
+    pub app_display_version: String,
+    // Channel to differentiate logs from prod/beta/staging/devel
+    pub app_channel: String,
+}
+
+// Exported type for public method parameters
+// Default impl empty values will be omitted in json from ping struct definition
```
jrconlin, line 26: FWIW: serde offers a flag for this: https://serde.rs/field-attrs.html#skip_serializing_if

glean/src/server_events.rs
jrconlin: Ah, because the `app_display_version` and `app_channel` are defined from the values set in `GleanEventsLogger`. 

Technically, a more rust way of doing that would be to create a
```rust
impl From<GleanEventsLogger> for ClientInfo {
   fn from(gel: GleanEventsLogger>) -> ClientInfo {
      ClientInfo {
             app_display_version: gel.app_display_version,
             app_channel: gel.app_channel, 
             ..Default::default()
      }
  }
}
```
(this is where you'd have a `impl Default for ClientInfo` that specifies the other values.)

You could then do 
```rust
let telemetry_payload: PingPayload = PingPayload {
            client_info: self.clone().into(),
```

(This also lends itself nicely to template building, because then you can just define the non-standard stuff you want to specify and trust that rust will backfill with the standard bits.)

But if this is the way that the Glean team is advising you create this, so be it.

glean/src/server_events.rs
```diff
@@ -0,0 +1,308 @@
...
+use serde::{Deserialize, Serialize};
+use serde_json;
+use std::collections::HashMap;
+use uuid::Uuid;
+
+// log type string used to identify logs to process in the Moz Data Pipeline
+const GLEAN_EVENT_MOZLOG_TYPE: &str = "glean-server-event";
+
+// Code below is static, regardless of what is defined in `metrics.yaml`:
+pub struct GleanEventsLogger {
```
jrconlin, line 16: Since we specify a docstring for `GleanEventsLogger::record()` let's add one here to explain what's special about this struct.

glean/src/server_events.rs
```diff
@@ -0,0 +1,308 @@
...
+        println!("{}", envelope_json);
+    }
+}
+
+// Code below is generated based on the provided `metrics.yaml` file:
+
+// Metrics of the `event` type. Anything defined in `extra_keys` has it's own struct field.
+// The appended `Event` term to any metric of the event type implies the ping event.
+
+pub struct SyncstorageGetCollectionsEvent {
```
jrconlin, line 182: Wondering if these should be in a different file. The stuff above this feel very much like generics, where this is more focused on the `syncstorage` product. 

I'm fine keeping them in place for now, though.

glean/src/server_events.rs
```diff
@@ -0,0 +1,308 @@
...
+
+pub struct EventsPing {
+    pub syncstorage_device_family: String, // Device family from which sync action was initiated. Desktop PC, Tablet, Mobile, and Other.
+    pub syncstorage_hashed_device_id: String, // Hashed device id that is associated with a given account.
+    pub syncstorage_hashed_fxa_uid: String, // User identifier. Uses `hashed_fxa_uid` for accurate count of sync actions.
+    pub syncstorage_platform: String, // Platform from which sync action was initiated. Firefox Desktop, Fenix, or Firefox iOS.
+    pub event: Option<Box<dyn EventsPingEvent>>, // valid event of `EventsPingEvent` for this ping.
+}
+
+// Record and submit `events` ping
```
jrconlin, line 212: We should make this a doc string, since we've got one for the `record_events_ping()` function.

glean/src/server_events.rs
```diff
@@ -0,0 +1,308 @@
...
+}
+
+pub struct SyncDauPing {
+    pub syncstorage_device_family: String, // Device family from which sync action was initiated. Desktop PC, Tablet, Mobile, and Other.
+    pub syncstorage_hashed_device_id: String, // Hashed device id that is associated with a given account.
+    pub syncstorage_hashed_fxa_uid: String, // User identifier. Uses `hashed_fxa_uid` for accurate count of sync actions.
+    pub syncstorage_platform: String, // Platform from which sync action was initiated. Firefox Desktop, Fenix, or Firefox iOS.
+}
+
+// Record and submit `sync-dau` ping
```
jrconlin, line 262: Again, probably should be a docstring

syncserver/src/server/mod.rs
```diff
@@ -266,6 +273,14 @@ impl Server {
             &Metrics::from(&metrics),
             blocking_threadpool.clone(),
         )?;
+        let glean_logger = Arc::new(GleanEventsLogger {
```
jrconlin, line 276: Future optimization would be to have `glean_logger` be an `Option<GleanEventsLogger>` and only instantiate it if `settings.syncstorage.glean_enabled`. That saves us a byte in the `ServerState`.

syncserver/src/web/handlers.rs
```diff
@@ -30,24 +30,42 @@ use crate::{
...
-    // We'd like for the data to be ready and in place to pass to that logic.
-    let _hashed_fxa_uid: String = meta.user_id.hashed_fxa_uid.clone();
-    let _hashed_device_id: String = meta.user_id.hashed_device_id.clone();
-    let user_agent = request
-        .headers()
-        .get(header::USER_AGENT)
-        .and_then(|header| header.to_str().ok())
-        .unwrap_or("none");
-    let _device_info = get_device_info(user_agent);
+    if state.glean_enabled {
```
jrconlin, line 43: if you used `Option<GleanEventLogger>` you'd do something like:

```rust
if let Some(glean_logger) = state.glean_logger { 
   ...
   glean_logger.record_events_ping( ...
```

Cargo.toml
```diff
@@ -90,6 +90,7 @@ slog-stdlog = "4.1"
 slog-term = "2.6"
 tokio = "1"
 thiserror = "1.0.26"
+uuid = { version = "1.6", features = ["serde", "v4"] }
```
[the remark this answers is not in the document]
taddes: Good call, we can go with 1.11. 1.6 is what was in syncstorage-spanner

glean/src/server_events.rs
pjenvey: Sure but this also gets a bit in to the weeds of what you want for default values of `app_display_version/channel` -- I'm wondering if glean might never want those to be empty strings or "Unknown" (unlike the other fields) so you might want to discourage that from potentially ever happening by not having such a default as a footgun 🤷‍♂️

taddes: All the other parsers (JS, Go, Ruby, Pythonm, etc) generate the string "Unknown" by default, so I think we want to keep it. The comment in one of the server implementations stated: `"Unknown" fields below are required in the Glean schema, however they are not useful in server context.` Therefore, we don't get any benefit but it is expected that the string "Unknown" is present.

syncserver/src/web/handlers.rs
taddes: Got the PR in place to add this, for now we will keep this in place and when the cloudops-infra changes land, we'll modify.

taddes:
Made requested changes and really tightened up the documentation. As far as the From/Into impl stuff, I definitely want to implement it, but if ok with you we'll keep it the same as the generated Rust code for validating everything works and I'll open another short ticket for those refinements. Just makes it simple to validate our approach in case there's something standing out.  Thanks!

jrconlin (approved):

syncserver/src/server/mod.rs
```diff
@@ -266,6 +273,14 @@ impl Server {
...
+            // app_id corresponds to probe-scraper entry.
+            // https://github.com/mozilla/probe-scraper/blob/main/repositories.yaml
+            app_id: "syncstorage".to_owned(),
+            app_display_version: env!("CARGO_PKG_VERSION").to_owned(),
+            app_channel: "prod".to_owned(),
```
pjenvey, line 281: (or whatever the new env var is named?)
```suggestion
            app_channel:  env!("SYNC_ENVIRONMENT").to_owned(),
```

syncserver/src/web/handlers.rs
```diff
@@ -30,24 +30,42 @@ use crate::{
...
+
+        state.glean_logger.record_events_ping(
+            &RequestInfo {
+                user_agent: user_agent.to_owned(),
+                ip_address: "".to_owned(),
+            },
+            &EventsPing {
+                syncstorage_device_family: device_info.device_family.to_string(),
+                syncstorage_hashed_device_id: hashed_device_id.to_string(),
+                syncstorage_hashed_fxa_uid: hashed_fxa_uid.to_string(),
```
pjenvey, line 63: you already have ownership of these Strings here so I don't think they need to_string'ing (or probably even the cloning as well?)
```suggestion
        let user_agent = request
            .headers()
            .get(header::USER_AGENT)
            .and_then(|header| header.to_str().ok())
            .unwrap_or("none");
        let device_info: DeviceInfo = get_device_info(user_agent);

        state.glean_logger.record_events_ping(
            &RequestInfo {
                user_agent: user_agent.to_owned(),
                ip_address: "".to_owned(),
            },
            &EventsPing {
                syncstorage_device_family: device_info.device_family.to_string(),
                syncstorage_hashed_device_id: meta.user_id.hashed_device_id,
                syncstorage_hashed_fxa_uid: meta.user_id.hashed_fxa_uid,
```

pjenvey (changes requested):

syncserver/src/web/handlers.rs
pjenvey: Also I wonder if an empty string for the lack of user agent would be preferable to "none"

taddes:
LMK @pjenvey if the env var impl looks good to you. Should be overwritten when the matching `SYNC_ENVIRONMENT` var is read in, like the others. Lives on the base `Settings` struct, not `SyncstorageSettings`

pjenvey (approved):

PR #1628, opened 2024-11-05
fix: avoid underflow of the queued_tasks metric

pjenvey:
I almost just caught the specific underflow case `active_threads.load(Ordering::Relaxed).min(spawned_tasks)` but I think `SeqCst` should also avoid this.

Closes SYNC-4475


jrconlin (approved):
I agree that [SeqCst](https://doc.rust-lang.org/std/sync/atomic/enum.Ordering.html#variant.SeqCst) is preferable in this case at this point. I think the `.min()` wouldn't really fix things because then we'd run the risk of over-counts.

PR #1633, opened 2024-11-21
chore: tag 0.17.15

pjenvey:


taddes (approved):
LGTM, though we may want to add the Glean stuff if we're gonna push this out to stage? That or if you want it separate, fine by me.

PR #1640, opened 2024-12-05
Feat: add task to release unused db conns 

taddes:
## Description

By default, Deadpool releases unused connections pretty lazily, causing connections that may otherwise be available to be tied up unnecessarily. This could be part of what’s causing pods to run out of available connections to Spanner in certain cases. To resolve this, we should pursue a solution like the one described [here](https://github.com/bikeshedder/deadpool/issues/178#issuecomment-1036953054), where we spawn a background task that periodically releases connections that haven’t been used within a given amount of time (this amount of time should be configurable in a setting).

To do this, we’ll need to make use of the [Pool::retain](https://docs.rs/deadpool/latest/deadpool/managed/struct.Pool.html#method.retain) method, which was introduced in deadpool 0.9.3 (we’re currently on version 0.5, so we’ll need to upgrade).

## Testing

How should reviewers test?

## Issue(s)

Closes [SYNC-4542](https://mozilla-hub.atlassian.net/browse/SYNC-4542).

[SYNC-4542]: https://mozilla-hub.atlassian.net/browse/SYNC-4542?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/syncstorage-mysql/src/pool.rs
+++ b/syncstorage-mysql/src/pool.rs
@@ -99,6 +99,13 @@ impl MysqlDbPool {
         })
     }
 
+    /// Spawn a task to periodically evict idle connections. Calls wrapper sweeper fn
+    ///  to use pool.retain, retaining objects only if they are shorter in duration than
+    ///  defined max_idle. Noop for mysql impl.
+    pub fn spawn_sweeper(&self, _interval: Duration) {
+        sweeper()
+    }
+
     pub fn get_sync(&self) -> DbResult<MysqlDb> {
         Ok(MysqlDb::new(
             self.pool.get()?,
@@ -109,6 +116,13 @@ impl MysqlDbPool {
         ))
     }
 }
+
+/// Sweeper to retain only the objects specified within the closure.
+/// In this context, if a Spanner connection is unutilized, we want it
+/// to release the given connections.
+/// See: https://docs.rs/deadpool/latest/deadpool/managed/struct.Pool.html#method.retain
+/// Noop for mysql impl
+fn sweeper() {}
 
 #[async_trait]
 impl DbPool for MysqlDbPool {
--- a/syncstorage-settings/src/lib.rs
+++ b/syncstorage-settings/src/lib.rs
@@ -78,6 +78,8 @@ pub struct Settings {
     pub database_pool_connection_lifespan: Option<u32>,
     /// Max time a connection should sit idle before being dropped.
     pub database_pool_connection_max_idle: Option<u32>,
+    /// Interval for sweeper task releasing unused connections.
+    pub database_pool_sweeper_task_interval: Option<u32>,
     #[cfg(debug_assertions)]
     pub database_use_test_transactions: bool,
     #[cfg(debug_assertions)]
@@ -115,6 +117,7 @@ impl Default for Settings {
             database_pool_min_idle: None,
             database_pool_connection_lifespan: None,
             database_pool_connection_max_idle: None,
+            database_pool_sweeper_task_interval: Some(30),
             database_pool_connection_timeout: Some(30),
             #[cfg(debug_assertions)]
             database_use_test_transactions: false,
--- a/syncstorage-settings/src/lib.rs
+++ b/syncstorage-settings/src/lib.rs
@@ -79,7 +79,7 @@ pub struct Settings {
     /// Max time a connection should sit idle before being dropped.
     pub database_pool_connection_max_idle: Option<u32>,
     /// Interval for sweeper task releasing unused connections.
-    pub database_pool_sweeper_task_interval: Option<u32>,
+    pub database_pool_sweeper_task_interval: u32,
     #[cfg(debug_assertions)]
     pub database_use_test_transactions: bool,
     #[cfg(debug_assertions)]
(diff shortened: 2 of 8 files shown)


syncstorage-spanner/src/manager/session.rs
```diff
@@ -64,6 +64,8 @@ pub struct SpannerSessionSettings {
     pub max_lifespan: Option<u32>,
     /// Max idle time of a Session
     pub max_idle: Option<u32>,
+    /// Interval for sweeper task releasing unused connections.
+    pub pool_max_idle: Option<u32>,
```
pjenvey, line 68: We already have a `database_pool_connection_max_idle` setting, let's use that instead -- the pool implementation uses it to timeout connections as it checks them out, but I believe the sweeper's going to do a better job of reducing the number than that code

pjenvey (changes requested):

syncstorage-spanner/src/manager/session.rs
taddes: You bet @pjenvey , I wasn't 100% sure about reusing some of those settings so opted to just make a new one. Can I also pull this out of the Option and set a default?

syncstorage-settings/src/lib.rs
```diff
@@ -78,6 +78,8 @@ pub struct Settings {
     pub database_pool_connection_lifespan: Option<u32>,
     /// Max time a connection should sit idle before being dropped.
     pub database_pool_connection_max_idle: Option<u32>,
+    /// Interval for sweeper task releasing unused connections.
+    pub database_pool_sweeper_task_interval: Option<u32>,
```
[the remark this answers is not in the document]
taddes: @pjenvey you ok with 30 s as a sweeper interval?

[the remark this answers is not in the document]
pjenvey: Yea, whatever works for autopush should be fine for syncstorage

syncstorage-mysql/src/pool.rs
```diff
@@ -99,6 +99,13 @@ impl MysqlDbPool {
         })
     }
 
+    /// Spawn a task to periodically evict idle connections. Calls wrapper sweeper fn
+    ///  to use pool.retain, retaining objects only if they are shorter in duration than
+    ///  defined max_idle. Noop for mysql impl.
+    pub fn spawn_sweeper(&self, _interval: Duration) {
```
jrconlin, line 105: If this is a function that all implementations of `DbPool` should perform, we should add it to the `DbPool` trait (and move this to the `impl DbPool` block.

I wonder if we should rename this to `spawn_maintenance` (`spawn_janitor`?) 

While we don't need to sweep for pool connections for SQL libs, folk may want to do other maintenance things like run `vacuum` or the like.

syncstorage-spanner/src/pool.rs
```diff
@@ -84,6 +87,30 @@ impl SpannerDbPool {
...
+    ///  to use pool.retain, retaining objects only if they are shorter in duration than
+    ///  defined max_idle.
+    pub fn spawn_sweeper(&self, interval: Duration) {
+        let Some(max_idle) = self.pool.manager().settings.max_idle else {
+            return;
+        };
+        let pool = self.pool.clone();
+        rt::spawn(async move {
+            loop {
+                sweeper(&pool, Duration::from_secs(max_idle.into()));
```
pjenvey, line 101: in general, prefer constructing something like this `Duration` outside of the loop -- but it's a small, `Copy` type anyway so it doesn't really matter at all here

pjenvey (approved):

syncstorage-mysql/src/pool.rs
```diff
@@ -99,6 +99,13 @@ impl MysqlDbPool {
...
+        sweeper()
```
jrconlin, line 106: I think we can drop this and make `spawn_sweeper(..)` the noop in MysqlDb.

syncstorage-mysql/src/pool.rs
```diff
@@ -110,6 +117,13 @@ impl MysqlDbPool {
     }
 }
 
+/// Sweeper to retain only the objects specified within the closure.
+/// In this context, if a Spanner connection is unutilized, we want it
+/// to release the given connections.
+/// See: https://docs.rs/deadpool/latest/deadpool/managed/struct.Pool.html#method.retain
+/// Noop for mysql impl
+fn sweeper() {}
```
jrconlin, line 125: Yeah, let's drop this. Less functions, less problems.

syncstorage-spanner/src/pool.rs
```diff
@@ -84,6 +87,30 @@ impl SpannerDbPool {
             self.quota,
         ))
     }
+
+    /// Spawn a task to periodically evict idle connections. Calls wrapper sweeper fn
+    ///  to use pool.retain, retaining objects only if they are shorter in duration than
+    ///  defined max_idle.
+    pub fn spawn_sweeper(&self, interval: Duration) {
```
jrconlin, line 94: If you do move this to the `impl DbPool` trait, you can keep the `fn sweeper()` as part of the general `impl` block.

jrconlin (changes requested):

jrconlin:
Let's add  a new trait method for `spawn_sweeper` (or rename it to `spawn_maintenance`) in DbPool. We're essentially requiring it for all DbPool implementations, so might as well make it official.

syncstorage-mysql/src/pool.rs
chenba: this never happened.  @taddes any reason we didn't drop it?  and can we do the same in https://github.com/mozilla-services/syncstorage-rs/pull/1875 ?

taddes: We just decided to keep the noop when we talked for some reason, can't recall, there was some back and forth on keeping it due to a bug we had. I think we can look at dropping now.

PR #1647, opened 2025-01-09
chore: Upgrade to Diesel 2.2

tommie:
## Description

Upgrades Diesel to 2.2.

This uncovered some bugs in the MySQL backend, e.g. there can be at most one batch ID per user every 10 ms. It seems Diesel 2 is faster (and libmysqlclient is faster than libmariadb), causing the E2E batch tests to generate batches faster than before. Nothing ever updated the session timestamp in the backend, which suggests there could only really be one batch per user.

Split off from #1644.

## Testing

Existing integration tests.

## Issue(s)

Closes [SYNC-4568](https://mozilla-hub.atlassian.net/browse/SYNC-4568) (?)

[SYNC-4568]: https://mozilla-hub.atlassian.net/browse/SYNC-4568?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jrconlin:
Ugh, Looks like these commits aren't signed, so this PR can't be merged.
You can [sign using the same SSH key](https://stackoverflow.com/a/76511638) that you set up in github, which can be easier (and more secure) than dealing with GPG signatures.
FWIW, a really stupid way to do this would be to:

* [set up git signing](https://stackoverflow.com/a/76511638)
* `git diff origin/master > diesel.patch`
* `git checkout -b diesel3`
* `patch -p1 < diesel.patch`
* `git commit -a` -- If you're not running an ssh-agent, you should be prompted for your passphrase. 
* `git push ...`

That's basically what I'd do to duplicate your PR as signed.

tommie:
Oh, you actually did mean a digital signature. Now I'm curious what this gains you? The code merged is what's reviewed in the PR, so it doesn't seem to protect against anything?

jrconlin:
It's required by our Security Operations (SecOps) folk for all production run code. Basically, it's to ensure that this and future code comes from "you", and can be trusted. [This article](https://dlorenc.medium.com/should-you-sign-git-commits-f068b07e1b1f) presents a fairly good reason for why we require them, even though we're open source and public.

jrconlin:
beh, So, in these cases I usually just pull master,  drop `Cargo.lock`, and rebuild. 

I'll approve after that, but I'll want to give at least one other team member a chance before I land anything.

syncstorage-mysql/src/batch.rs
```diff
@@ -32,18 +37,16 @@ pub fn create(db: &MysqlDb, params: params::CreateBatch) -> DbResult<results::Cr
...
     // skew the sharding distribution.
     //
-    // So we mix in the lowest digit of the uid to improve the distribution
-    // while still letting us treat these ids as millisecond timestamps.  It's
-    // yuck, but it works and it keeps the weirdness contained to this single
-    // line of code.
-    let batch_id = db.timestamp().as_i64() + (user_id % 10);
+    // We mix in a per-process counter to make batch IDs (more) unique within
+    // a timestamp.
+    let batch_id = db.timestamp().as_i64() + COUNTER.fetch_add(1, Ordering::Relaxed) as i64 % 10;
```
pjenvey, line 42: The original `user_id % 10` is a bit of a kludge (there was some old sharding algorithm of data that apparently took this last digit into account).

I'm going to change this into a limited retry on inserts w/ the specific conflict error with a simple increment

syncstorage-mysql/src/models.rs
```diff
@@ -212,24 +213,36 @@ impl MysqlDb {
...
                 .coll_modified_cache
                 .insert((user_id as u32, collection_id), modified);
-        }
+            now
+        } else {
+            let result = sql_query("SELECT UNIX_TIMESTAMP(UTC_TIMESTAMP(2))*1000 AS timestamp")
+                .get_result::<TimestampResult>(&mut *self.conn.write()?)?;
+            SyncTimestamp::from_i64(result.timestamp)?
+        };
+        self.set_timestamp(timestamp);
```
pjenvey, line 244: I don't understand why this is needed: we should get an update to `user_collections::modified` whenever writes are made (via `update_collection`)

pjenvey:
Also @tommie if you could fix the lack of the signing verification, I'll attempt to maintain your exact commits: it's pretty easy to setup with an ssh key (then you may need to rebase it, or possibly force push? for the "verified" badge to show).

pjenvey:
superseded by https://github.com/mozilla-services/syncstorage-rs/pull/1778

thanks @tommie!

PR #1651, opened 2025-02-26
chore: migrate to nextest and llvm-cov for unit tests and coverage

nshirley:
## Description

Describe these changes.

This migrates unit test to use [nextest](https://nexte.st/) and [cargo-llvm-cov](https://github.com/taiki-e/cargo-llvm-cov) to support integration with the [ete test pipeline](https://mozilla.github.io/ecosystem-test-scripts/reference-guides/project_onboarding_procedure.html)
- In addition to moving to nextest, the workflows now use the `makefile` test commands for consistency running locally vs. in ci
- There are a few extra variables in the `Makefile` in preparation for integration tests reporting as well

## Testing

How should reviewers test?
- Unit tests should still pass in CI ([successful build](https://app.circleci.com/pipelines/github/mozilla-services/syncstorage-rs/5560/workflows/35ff6756-46f0-4739-8a08-bb3816b91924/jobs/22334))
- There should also be an output/captured test results in CI ([test artifacts](https://app.circleci.com/pipelines/github/mozilla-services/syncstorage-rs/5560/workflows/35ff6756-46f0-4739-8a08-bb3816b91924/jobs/22334/tests))

## Issue(s)

Closes [SYNC-4611](https://mozilla-hub.atlassian.net/browse/SYNC-4611).

[SYNC-4611]: https://mozilla-hub.atlassian.net/browse/SYNC-4611?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.circleci/config.yml
```diff
@@ -60,11 +60,6 @@ commands:
           name: Rust Clippy Spanner
           command: |
             cargo clippy --workspace --all-targets --no-default-features --features=syncstorage-db/spanner --features=py_verifier -- -D warnings
-  cargo-build:
```
nshirley, line 63: This was only being used in the `build-and-test` job, so it was an extra step now. Running `cargo nextest` will compile a different output (target/nextest) and with the added install steps of `nextest` and `llvm` it was just pushing the time to build and test into 15 minute range.

.config/nextest.toml
```diff
@@ -0,0 +1,5 @@
+[profile.ci.junit]
```
nshirley, line 1: These are obviously the same, but if we need to add additional configuration specific to local (default) or CI then it's already configured! 🎉

README.md
```diff
@@ -241,7 +241,18 @@ We use [env_logger](https://crates.io/crates/env_logger): set the `RUST_LOG` env
 
 ### Unit tests
 
-`make test` - open the Makefile to adjust your `SYNC_SYNCSTORAGE__DATABASE_URL` as needed.
+You'll need [`nextest`](https://nexte.st/docs/installation/from-source/) and [`llvm-cov`](https://github.com/taiki-e/cargo-llvm-cov?tab=readme-ov-file#installation) installed for full unittest and test coverage.
```
nshirley, line 244: This was the only reference to unit-tests that I could find, but please let me know if I missed any!

test-results/.gitignore
nshirley, line 1: The output for llvm-cov needs a place to write but it can't make the folder. So, instead of adding a step to make the folder in CI, I included it so that running tests locally is as close to the same as they run in CI.

README.md
```diff
@@ -241,7 +241,18 @@ We use [env_logger](https://crates.io/crates/env_logger): set the `RUST_LOG` env
...
+
+    $ cargo install cargo-nextest --locked
+    $ cargo +stable install cargo-llvm-cov --locked
```
pjenvey, line 247: nit: I'm not sure why their install instructions explicitly mention `+stable` but that's the default toolchain for most users. it's less confusing without it IMHO
```suggestion
    $ cargo install cargo-llvm-cov --locked
```

test-results/.gitignore
pjenvey: I'm not in love with checking it in but indeed it saves the needed mkdir in CI, or needing to run it yourself locally so `make test_with_coverage` doesn't break, so I'm not opposed to it

pjenvey (changes requested):

test-results/.gitignore
pjenvey: Though I guess alternatively make test_with_coverage could do a mkdir, which'll be ignored when it already exists

Trinaa: I hope i'm understanding the problem correctly... I don't think we should have .gitignore or .gitkeep outside of the root. CI often needs working directories created ahead of time, it is typical to add a step for that, example: https://github.com/mozilla-services/autopush-rs/blob/master/.circleci/config.yml#L134-L138

Trinaa (changes requested):

Makefile
```diff
@@ -10,6 +10,22 @@ PATH_TO_SYNC_SPANNER_KEYS = `pwd`/service-account.json
...
+# TODO: update workflow name appropriately
+WORKFLOW := run-tests
+EPOCH_TIME := $(shell date +"%s")
+TEST_RESULTS_DIR ?= test-results
+TEST_PROFILE := $(if $(CIRCLECI),ci,default)
+TEST_FILE_PREFIX := $(if $(CIRCLECI),$(CIRCLE_BUILD_NUM)__$(EPOCH_TIME)__$(CIRCLE_PROJECT_REPONAME)__$(WORKFLOW)__)
+UNIT_JUNIT_XML := $(TEST_RESULTS_DIR)/$(TEST_FILE_PREFIX)unit__results.xml
+UNIT_COVERAGE_JSON := $(TEST_RESULTS_DIR)/$(TEST_FILE_PREFIX)unit__coverage.json
+INTEGRATION_JUNIT_XML := $(TEST_RESULTS_DIR)/$(TEST_FILE_PREFIX)integration__results.xml
+INTEGRATION_JUNIT_XML_LEGACY := $(TEST_RESULTS_DIR)/$(TEST_FILE_PREFIX)integration__legacy-results.xml
```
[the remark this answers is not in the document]
nshirley: We'll need them for the python integration tests - happy to remove since they're not used _yet_.

.circleci/config.yml
```diff
@@ -185,7 +191,9 @@ jobs:
 
   build-and-test:
     docker:
-      - image: cimg/rust:1.78.0  # RUST_VER
+      # nextest is required for testing to get junit output
+      # however, it requires >1.81.0, so we use a different version here
+      - image: cimg/rust:1.81.0  # RUST_VER
```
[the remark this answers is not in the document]
nshirley: Sounds good! I'll bump the other references. Are there additional testing considerations with other teams by bumping the other jobs? Mostly, I'm thinking about the final spanner build

.config/nextest.toml
nshirley: Ah, thanks; I didn't see the config in there - I'll update to stay consistent

Makefile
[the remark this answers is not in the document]
Trinaa: Yes, but also the integration tests also need to be renamed to e2e tests to be consistent when the time comes.

.circleci/config.yml
```diff
@@ -91,16 +86,27 @@ commands:
             "$CIRCLE_PROJECT_REPONAME" \
             "$CIRCLE_BUILD_URL" > syncserver/version.json
 
+  install-test-deps:
+    steps:
+      - run: 
+          name: Install cargo-nextest
+          command: curl -LsSf https://get.nexte.st/latest/linux | tar zxf - -C ${CARGO_HOME:-~/.cargo}/bin
```
jrconlin, line 93: While I have nothing against the fine residents of São Tomé and Príncipe, I do wonder if we should cache copies of stuff like this to prevent side channel attacks. (This goes double for things like `rustup.rs`)

.circleci/config.yml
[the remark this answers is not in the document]
jrconlin: Not sure I follow.
The version of rust we use generally is isolated to our system. We are generally more focused on any API change since that impacts external systems more. 
We update our base requirement version of rust about once a quarter, whenever we do a general sweep of our dependencies, or if there's some other requirement to bump the version of rust (e.g. new feature, CVE, etc.)
If nextest has a base requirement of 1.81, then I can see us bumping up to that. 
(We also may wish to include [rust-version](https://doc.rust-lang.org/cargo/reference/rust-version.html) in these cases so that there's a hard baseline.)

.config/nextest.toml
```diff
@@ -0,0 +1,23 @@
...
+fail-fast = false
+slow-timeout = { period = "300s" }
+
+[profile.ci]
+fail-fast = false
+
+[profile.ci.junit]
+path = "junit.xml"
+
+report-name = "autopush-unit-tests"
```
nshirley, line 21: lol, copy pasta error, I'll fix it

README.md
```diff
@@ -250,7 +261,7 @@ default, we use the diesel test_transaction functionality to ensure test data
 is not committed to the database. Therefore, there is an environment variable
 which can be used to turn off test_transaction.
 
-        SYNC_SYNCSTORAGE__DATABASE_USE_TEST_TRANSACTIONS=false cargo test [testname]
+        SYNC_SYNCSTORAGE__DATABASE_USE_TEST_TRANSACTIONS=false make test ARGS="[testname]"
```
jrconlin, line 264: This is one of those weird cases where we're using a pseudo-variable `[testname]`, and I wonder if that might screw up folk. 

`cargo test` (and presumably `cargo nextest`) allow you to run tests that match a name pattern you provide. So something like `cargo nextest test_uaid` would run any test that matches `.*test_uaid.*`  Passing the value as `cargo nextest [test_uaid]` would generate an error.

We probably don't need to address it here, but we should have some consistent way to express variables in documentation like this. Maybe use `example_test_name` or something.

jrconlin (dismissed):

pjenvey (dismissed):

Trinaa (changes requested):
I see that there are 2 sets of artifact files for the unit tests. Are there 2 suites or is this a result of parallelism?

image

If it is a result of 2 suites we may need to make that distinction in the `test_suite` portion of the file names. 

Coverage files will need to be merged before they are uploaded to the bucket or stored as an artifact.

~The easiest thing may be not to run them in parallel, if they can run in under 5 minutes.~ I see it's taking more than 5 minutes

pjenvey (approved):

.circleci/config.yml
```diff
@@ -106,6 +108,16 @@ commands:
           command: make test_with_coverage
           environment:
               SYNC_SYNCSTORAGE__ENFORCE_QUOTA: 1
+
+  merge-unit-test-coverage:
+    steps:
+      - run:
+          name: Merge llvm-cov results
+          command: |
+            make merge_coverage_results
```
Trinaa, line 117: If the tests fail, a coverage file is still produced usually, but will the merge-unit-test-coverage step execute?
I think we may need a `when: always` on this step.
A good test would be to force a test fail on CI.

nshirley: Okay, I've got it figured out!

[Here](https://app.circleci.com/pipelines/github/mozilla-services/syncstorage-rs/5572/workflows/525b4ae9-4df6-41cd-8a83-779fb4ba72fe/jobs/22408) is a run where I forced a failure and the merge still ran. 

And the fix for the make step was relatively simple, but you were right that `.ONESHELL` was causing problems. I had to use the `|| {}` to capture the output of the llvm command step output.

Trinaa (dismissed):
Thanks for all the hard work @nshirley! 🚢 

Make sure that you squash these commits and format the message to comply with the contribution guidelines before merging.

jrconlin (dismissed):

Trinaa (approved):

PR #1657, opened 2025-03-14
chore: migrate tokenserver tests to pytest with junit output

nshirley:
## Description

Describe these changes.

The original ticket was scoped to move all python tests (tokenserver, mysqle2e and spannere2e) to utilize pytest, but for the sake of speed the latter two will be tackled in another ticket. 

Interestingly, by using pytest it can discover the existing `unittest` tests and run them without extra work at this time, and in doing so, it's also running _more_ tests that were not being run before! Here are the tests that additionally run:
```
TestProcessAccountEventsForceSpanner::test_delete_user_force_spanner
TestMigrationRecords::test_purging_no_override
TestMigrationRecords::test_purging_override_null_keys_changed_at
TestMigrationRecords::test_purging_override_with_migrated
TestMigrationRecords::test_purging_override_with_migrated_password
TestMigrationRecords::test_purging_replaced_at
```
Should we keep these tests in? Do they need to run?

## Testing

How should reviewers test?
- Tokenserver integration tests still run, pass, and now capture results in CircleCI
  - [Passing tests](https://app.circleci.com/pipelines/github/mozilla-services/syncstorage-rs/5595/workflows/aea11911-2d79-4b23-a1e9-0587e6bfbc6d/jobs/22550?invite=true#step-112-9726_80)
  - [Artifacts](https://app.circleci.com/pipelines/github/mozilla-services/syncstorage-rs/5595/workflows/aea11911-2d79-4b23-a1e9-0587e6bfbc6d/jobs/22550/artifacts)

## Issue(s)

Closes [SYNC-4516](https://mozilla-hub.atlassian.net/browse/SYNC-4612).

[SYNC-4516]: https://mozilla-hub.atlassian.net/browse/SYNC-4516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.circleci/config.yml
```diff
@@ -240,11 +240,11 @@ jobs:
       - make-test-dir
       - run-unit-tests
       - merge-unit-test-coverage
-      - store-unit-test-results
       # if the above tests don't run tokenserver-db tests (i.e. using --workspace)
       # then run-tokenserver-scripts-tests will fail. These tests expect the db to be
       # configured already, and it appears unit-tests modify the db to the expected state
-      - run-tokenserver-scripts-tests
+      - run-tokenserver-integration-tests
+      - store-test-results
```
nshirley, line 247: Since unit and tokenserver integration tests are running in the same job, just moving the storage step to the end captures all tests.

tools/tokenserver/test_database.py
```diff
@@ -467,23 +467,23 @@ def test_first_seen_at(self):
     def test_build_old_range(self):
         params = dict()
         sql = self.database._build_old_user_query(None, params)
-        self.assert_(sql.text.find("uid > :start") < 0)
-        self.assert_(sql.text.find("uid < :end") < 0)
+        self.assertTrue(sql.text.find("uid > :start") < 0)
```
nshirley, line 470: **note**: these were throwing a deprecation warning when running with pytest, so I updated to their suggestion `assertTrue()`

Trinaa (approved):
LGTM 🚢. Don't forget to squash before merging.

What is the ticket number for the remaining work for e2e tests? I don't see it in the epic atm.

tools/tokenserver/conftest.py
```diff
@@ -0,0 +1,4 @@
+import sys
+import os
+
+sys.path.insert(0, os.path.abspath(os.path.dirname(__file__)))
```
pjenvey, line 4: I think we could remove the need for this by switching to relative imports in this package when needed, but let's leave that for a [separate issue](https://mozilla-hub.atlassian.net/browse/SYNC-4676)

Makefile
```diff
@@ -100,4 +99,9 @@ test_with_coverage:
 	exit $$exit_code
 
 merge_coverage_results:
-	cargo llvm-cov report --summary-only --json --output-path ${UNIT_COVERAGE_JSON}
\ No newline at end of file
+	cargo llvm-cov report --summary-only --json --output-path ${UNIT_COVERAGE_JSON}
+
+.ONESHELL:
```
pjenvey, line 104: nit: more of a note to @Trinaa and @jrconlin as we have this in autopush too -- it seems `.ONESHELL` is a Makefile wide [special target](https://www.gnu.org/software/make/manual/html_node/Special-Targets.html) that only needs to be [set once](https://www.gnu.org/software/make/manual/html_node/One-Shell.html), _"If the .ONESHELL special target appears anywhere in the makefile then all recipe lines for each target.."_

pjenvey (approved):

pjenvey:
@nshirley and don't forget to add a "chore:" prefix to the commit when you squash

tools/tokenserver/conftest.py
nshirley: That's good to know - the import pathing still throws me off. We have a follow up epic for adding python linting and setting up poetry, this might be an AC we can add to a ticket there too!

PR #1681, opened 2025-04-03
docs: add tokenserver documentation to sync

taddes:
## Description

Describe these changes.
This PR adds documentation relevant to Tokenserver, as part of addressing concerns of such information not being present during an incident.

## Testing

N/A

## Issue(s)

Closes [SYNC-4183)](https://mozilla-hub.atlassian.net/browse/SYNC-4183).

### DIFF
--- a/docs/tokenserver/tokenserver.md
+++ b/docs/tokenserver/tokenserver.md
@@ -19,6 +19,8 @@ and [here](https://mozilla-services.readthedocs.io/en/latest/token/index.html).
 
 In practice today, it is only used for connecting to Sync.
 
+## Tokenserver API
+
 
 ## How Tokenserver Handles Failure Cases
 
@@ -33,3 +35,10 @@ When a user forgets and resets their password, their Sync keys change. The Token
 
 ### User Changes Their Password
 This is similar to the case where users revoke their access tokens. Any devices with a not-expired access token will continue to sync until expiry, but clients will likely disconnect those clients faster than the 1 hour - however, a malicious user might be able to sync upwards of 1 hour.
+
+## Utilities
+Tokenserver has two regular running utility scripts:
+1 - [Process Account Events](../tools/process_account_events.md)
+2 - [Purge Old Records](../tools/purge_old_records_tokenserver.md)
+
+For context on these processes, their purpose, and how to run them, please review their documentation pages.
--- a/docs/tokenserver/tokenserver.md
+++ b/docs/tokenserver/tokenserver.md
@@ -19,9 +19,6 @@ and [here](https://mozilla-services.readthedocs.io/en/latest/token/index.html).
 
 In practice today, it is only used for connecting to Sync.
 
-## Tokenserver API
-
-
 ## How Tokenserver Handles Failure Cases
 
 ### Token Expiry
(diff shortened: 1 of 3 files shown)


docs/tokenserver/tokenserver_api.md
```diff
@@ -0,0 +1,157 @@
...
+    - **id** -- a signed authorization token, containing the
+      user's id for the application and the node.
+    - **key** -- a secret derived from the shared secret
+    - **uid** -- the user id for this service
+    - **api_endpoint** -- the root URL for the user for the service.
+    - **duration** -- the validity duration of the issued token, in seconds.
+
+    Example::
+
+        HTTP/1.1 200 OK
```
jrconlin, line 42: I wonder if we should put this in a code block so that it's visually distinct?

docs/tokenserver/tokenserver_api.md
```diff
@@ -0,0 +1,157 @@
...
+
+
+**X-Client-State**
+
+    An optional string that can be sent to identify a unique configuration
+    of client-side state.  It may be up to 32 characters long, and must
+    contain only characters from the urlsafe-base64 alphabet (i.e.
+    alphanumeric characters, underscore and hyphen) and the period.
+
+    A change in the value of this header will cause the user's node
```
jrconlin, line 69: Well, _may_ cause the user's node allocation to be reset. 

Sync currently has only one node.

jrconlin (approved):

docs/tokenserver/tokenserver.md
```diff
@@ -0,0 +1,64 @@
...
+
+### User revoking access token
+The user could revoke the access token by signing out using the Mozilla Account’s Manage Account settings. In that case, clients continue to sync up to the expiry time, which is one hour. To mitigate against this case, Firefox clients currently receive push notifications from FxA instructing them to disconnect. Additionally, any requests done against FxA itself (for example to get the user’s profile data, connected devices, etc) will also trigger the client to disconnect.
+
+### User Forgetting Their Password
+When a user forgets and resets their password, their Sync keys change. The Tokenserver request includes the key ID (which is a hash of the sync key). Thus, on the next sync, Tokenserver recognizes that the password changed, and ensures that the tokens it issues point users to a new location on Sync Storage. In practice, it does that by including the Key ID itself in the Tokenserver token, which is then sent to Sync Storage.
```
pjenvey, line 54: Let's mention recovery keys here. I believe this flow w/ a recovery key acts similar to the password change/revoke access tokens, let's add a mention of that in a new paragraph at the end of this session -- maybe also moving "User Changes Their Password" above this one?

```suggestion
### User Forgetting Their Password (without a recovery key)
When a user forgets and resets their password without an account recovery key, their Sync keys change. The Tokenserver request includes the key ID (which is a hash of the sync key). Thus, on the next sync, Tokenserver recognizes that the password changed, and ensures that the tokens it issues point users to a new location on Sync Storage. In practice, it does that by including the Key ID itself in the Tokenserver token, which is then sent to Sync Storage.
```

pjenvey (changes requested):

docs/tokenserver/tokenserver.md
taddes: Good suggestions

pjenvey (approved):

PR #1682, opened 2025-04-03
fix: resolve pyo3 vuln deprecations

taddes:
## Description

Given a recent vulnerability [RUSTSEC-2025-0020](https://rustsec.org/advisories/RUSTSEC-2025-0020.html), we are required to upgrade to 0.24 of PyO3 due to Risk of buffer overflow in PyString::from_object.

Several methods are deprecated and require change. Will update PyO3 and resolve deprecation. 
PyO3 0.23 is a significant rework of PyO3's internals for two major improvements:

Support of Python 3.13's new freethreaded build (aka "3.13t")

Rework of to-Python conversions with a new IntoPyObject trait.

 [Migration guide](https://pyo3.rs/v0.23.0/migration) 

  

## Issue(s)

Closes [SYNC-4707](https://mozilla-hub.atlassian.net/browse/SYNC-4707).

[SYNC-4707]: https://mozilla-hub.atlassian.net/browse/SYNC-4707?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


taddes:
Thx again for the assistance in the brain-bender overflow error described in https://github.com/PyO3/pyo3/issues/4702.  Posting [your gist](https://gist.github.com/pjenvey/138177df67b576a407aaf944d8a48292) for reference too.

I implimented the `From<PyErr>` trait and called internal_error, though I do see one at the bottom that could pass a message context: https://github.com/mozilla-services/syncstorage-rs/blob/6f0c7b28db3f8a2701c4af4dfe7a2d691fc079ef/tokenserver-common/src/error.rs#L322. Not sure if that's preferred. Need to look at calling that appropriately for that type.

Also, I still avoided putting `pyo3` in the root `Cargo.toml` to avoid bloat, but lmk if we prefer it to be in the root since it's now called in 2 places. Compiler didn't want me implementing in a trait in a place where there error isn't defined.

tokenserver-common/src/error.rs
```diff
@@ -1,12 +1,12 @@
...
-
 use actix_web::{HttpResponse, ResponseError};
 use backtrace::Backtrace;
 use http::StatusCode;
+use pyo3::prelude::PyErr;
 use serde::{
     ser::{SerializeMap, Serializer},
     Serialize,
 };
+use std::{cmp::PartialEq, error::Error, fmt};
```
pjenvey, line 9: nit: Was this an IDE automatic rearrange? Let's put it back.

FYI (and I just noticed this, so I'm happy to write it down somewhere) `rustfmt` nightly [has a group_imports feature](https://rust-lang.github.io/rustfmt/?version=v1.8.0&search=#group_imports). I don't know if it's stabilizing any time soon but I'm looking forward to using its `StdExternalCrate` setting as it's the format we've attempted to adopt.

(it'd be nice if it even grouped separate crates that live in the same workspace into their own block..)

.cargo/audit.toml
```diff
@@ -2,5 +2,6 @@
 ignore = [
        "RUSTSEC-2024-0365", # Bound by diesel 1.4 (4GB limit n/a to tokenserver)
        "RUSTSEC-2024-0421", # Bound by diesel 1.4, `idna`  < 0.1.5, Upgrade to >=1.0.0
-       "RUSTSEC-2024-0437", # Bound by grpcio 0.13
+       "RUSTSEC-2024-0437", # Bound by grpcio 0.13,
+       "RUSTSEC-2025-0024", # Have to wait on slog-async >2.8.0 and cadence 1.5.0
```
pjenvey, line 6: It looks like this isn't needed after a: `cargo update -p crossbeam-channel`

pjenvey (changes requested):
> Also, I still avoided putting `pyo3` in the root `Cargo.toml` to avoid bloat, but lmk if we prefer it to be in the root since it's now called in 2 places. Compiler didn't want me implementing in a trait in a place where there error isn't defined.

Oops, I forgot the orphan rule would force this. I agree on keeping pyo3 out of the root so it's only built when necessary.

tokenserver-common/src/error.rs
taddes: I'll check out that feature, that would really be handy for sure!

tokenserver-common/src/error.rs
```diff
@@ -328,6 +328,12 @@ impl InternalError for TokenserverError {
     }
 }
 
+impl From<PyErr> for TokenserverError {
+    fn from(_message: PyErr) -> Self {
+        TokenserverError::internal_error()
```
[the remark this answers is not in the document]
taddes: Ah, I just noticed it's actually calling `InternalError::internal_error` since that impl is called for `TokenserverError`. Lmk if that looks right, getting into some OOP-ish chains that are kinda crappy, but it's the right method, since we want that context passed to it. https://github.com/mozilla-services/syncstorage-rs/blob/6f0c7b28db3f8a2701c4af4dfe7a2d691fc079ef/tokenserver-common/src/error.rs#L322

pjenvey (approved):

pjenvey (approved):

PR #1690, opened 2025-04-18
chore: update python cryptography

pjenvey:
## Description

bump cryptography and remove apparently no longer needed build flags


taddes (approved):
That's great! Maybe we can then look at bumping to newest Python version after this then too, since I know this was a blocker. 🐍

pjenvey:
Yup, we should do a full crate/python/rust dependency bump soon (early in the quarter) -- I just want to get a new release out for next week -- we can do all that afterwards

PR #1695, opened 2025-05-02
feat: build docker w/ Oracle's libmysqlclient

pjenvey:
## Description

instead of libmariadb as we still need the Oracle client's seamless TLS support (for now)

## Issue(s)

Closes STOR-250


taddes (approved):
Thx @pjenvey , made a note on the ticket with a question.

PR #1696, opened 2025-05-02
fix: re-enable tokensever e2e tests

pjenvey:
## Description

w/ the pyfxa 0.8.1 fix

## Issue(s)

Closes STOR-204

### DIFF
--- a/syncserver/src/web/middleware/rejectua.rs
+++ b/syncserver/src/web/middleware/rejectua.rs
@@ -37,7 +37,7 @@ pub fn reject_user_agent<B>(
           + 'static),
 ) -> LocalBoxFuture<'static, Result<ServiceResponse<EitherBody<B>>, actix_web::Error>> {
     match request.headers().get(USER_AGENT).cloned() {
-        Some(header) if header.to_str().map_or(false, should_reject) => Box::pin(async move {
+        Some(header) if header.to_str().is_ok_and(should_reject) => Box::pin(async move {
             trace!("Rejecting User-Agent: {:?}", header);
             let (req, payload) = request.into_parts();
             MetricsWrapper::extract(&req)
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -225,7 +225,7 @@ commands:
 jobs:
   checks:
     docker:
-      - image: cimg/rust:1.81.0  # RUST_VER
+      - image: cimg/rust:1.86  # RUST_VER
         auth:
           username: $DOCKER_USER
           password: $DOCKER_PASS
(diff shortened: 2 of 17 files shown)


taddes (dismissed):
Thx for this @pjenvey 👍

pjenvey:
@taddes I just pushed a bump to RUST_VER to fix CI (also breaking master)

jrconlin:
FWIW, yeah, I noticed that we need to specify the rust version with Cargo chef now.

syncserver/src/web/middleware/rejectua.rs
```diff
@@ -37,7 +37,7 @@ pub fn reject_user_agent<B>(
           + 'static),
 ) -> LocalBoxFuture<'static, Result<ServiceResponse<EitherBody<B>>, actix_web::Error>> {
     match request.headers().get(USER_AGENT).cloned() {
-        Some(header) if header.to_str().map_or(false, should_reject) => Box::pin(async move {
```
jrconlin, line 40: stupid nit, but I wonder if we should set `rust-version="1.70.0"` to indicate the base version we need? (might need higher)

jrconlin (dismissed):

pjenvey:
> I saw that master was breaking too, while installing dependencies for test & coverage. I added the `--locked` flag to the step on my branch and that resolved it (and tests still ran). I noticed we use that in a few places but don't know if that's something to avoid

Where was `--locked` added that helped? The issue here was dependencies of `nextest` ultimately required a later rust version -- I could have alternatively pinned the `nextest` version to solve this `cargo install nextest@0.9.94` but I think such a dependency issue with nextest/cargo-llvm-cov is probably going to be pretty rare.

We should consider pinning these anyway, but since they're solely a specified in the CI config it's probably going to easy for us to forget to upgrade them..

jrconlin (approved):

nshirley:
I added the `--locked` here:
```yaml
# .circleci/config.yml

  install-test-deps:
    steps:
      - run: 
          name: Install test dependencies
          command: cargo install --locked cargo-nextest cargo-llvm-cov
```

Admittedly I don't know if that's best practice but it solved the build issue on my branch so I was unblocked. I'm not opinionated one way or another on the solution here though. 

Is there a way to better include them so they're not defined just in the circle config?

pjenvey:
> Admittedly I don't know if that's best practice but it solved the build issue on my branch so I was unblocked. I'm not opinionated one way or another on the solution here though.
> 
> Is there a way to better include them so they're not defined just in the circle config?

Ah, my mistake -- cargo install must be taking nextest's `Cargo.lock` into account here (TIL), which is locking the dependency in question on an older version which doesn't have the issue, thanks!

PR #1697, opened 2025-05-06
test(e2e): Migrate e2e to pytest

nshirley:
## Description

This moves the e2e tests from using a custom runner and `unittest` to use pytest. In doing so there are a few additional things:
- New conftest fixtures to manage server and variables for test runner
- Cleanup for circle steps to reuse one step for both sets of e2e tests
- Moved the docker-compose from circle config to Makefile 
- Update tests to use appropriate asserts with the move from `unittest` -> `pytest` (i.e. `assertEqual` over `assertEquals`)

## Testing

I've confirmed that the tests run in CI, produce the appropriate `.xml` results file, and the file is saved to the artifacts!

- [Here](https://app.circleci.com/pipelines/gh/mozilla-services/syncstorage-rs/5752/workflows/8459f3d3-98f8-45d2-9364-9894f68e4a72) is an example of a fully successful run with both `spanner` and `mysql` runs passing with saved results
- [Here](https://app.circleci.com/pipelines/gh/mozilla-services/syncstorage-rs/5751/workflows/b3314720-86be-4880-a5f7-6477bdbb378e/jobs/23333) is also an example of failed run where, when tests failed, it properly propagated up

[Here is the latest run](https://app.circleci.com/pipelines/github/mozilla-services/syncstorage-rs/5726/workflows/d3c6cee7-94bd-4a0d-b66d-eda77fe7178d)

## Issue(s)

Closes [STOR-235](https://mozilla-hub.atlassian.net/browse/STOR-235).

[STOR-235]: https://mozilla-hub.atlassian.net/browse/STOR-235?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.circleci/config.yml
```diff
@@ -108,6 +108,9 @@ commands:
       - run:
           name: Create test-results directory
           command: mkdir -p workflow/test-results
+      - run:
```
nshirley, line 111: @nshirley - need to clean this up

.circleci/config.yml
```diff
@@ -445,10 +447,17 @@ jobs:
...
+          command: cp /home/circleci/cache/Makefile .
       - run:
           name: Restore docker-compose config
           command: cp /home/circleci/cache/docker-compose*.yaml .
-      - run-e2e-spanner-tests
+      - make-test-dir
+      - run-e2e-tests:
+          db: spanner
+          use_oauth: true
+      - store-test-results
```
nshirley, line 460: @nshirley - need to add the step to push results to GCP

.circleci/config.yml
```diff
@@ -145,34 +148,22 @@ commands:
...
-               -f docker-compose.spanner.yaml
-               -f docker-compose.e2e.spanner.yaml
-               up
-               --exit-code-from spanner-e2e-tests
-               --abort-on-container-exit
+            make run_<< parameters.db >>_e2e_tests
           environment:
-              SYNCSTORAGE_RS_IMAGE: app:build
+            SYNCSTORAGE_RS_IMAGE: app:build
+            USE_OAUTH: << parameters.use_oauth >>
```
nshirley, line 166: This is passed into the Conftest - in the original setup there was a step that would delete all the oauth vars and and then re-run a subset of the tests so this is here so we can achieve the same

Dockerfile
```diff
@@ -25,7 +25,9 @@ RUN \
     apt-get -q install -y --no-install-recommends $MYSQLCLIENT_PKG cmake
 
 COPY --from=planner /app/recipe.json recipe.json
-RUN cargo chef cook --release --no-default-features --features=syncstorage-db/$DATABASE_BACKEND --features=py_verifier --recipe-path recipe.json
+# TODO: add back in the --release flag
```
nshirley, line 28: @nshirley - need to clean this up

Makefile
```diff
@@ -21,6 +21,7 @@ TEST_FILE_PREFIX := $(if $(CIRCLECI),$(CIRCLE_BUILD_NUM)__$(EPOCH_TIME)__$(CIRCL
 UNIT_JUNIT_XML := $(TEST_RESULTS_DIR)/$(TEST_FILE_PREFIX)unit__results.xml
 UNIT_COVERAGE_JSON := $(TEST_RESULTS_DIR)/$(TEST_FILE_PREFIX)unit__coverage.json
 INTEGRATION_JUNIT_XML := $(TEST_RESULTS_DIR)/$(TEST_FILE_PREFIX)integration__results.xml
+LOCAL_INTEGRATION_JUNIT_XML := $(TEST_RESULTS_DIR)/$(TEST_FILE_PREFIX)local_integration__results.xml
```
nshirley, line 24: I know there are some different names as far as e2e vs integration. I'm a bit unclear if all of the tests fall into one bucket or another so it might be good to update to e2e everywhere as a catch all and then split out integration tests individually at some point

docker-compose.e2e.mysql.yaml
```diff
@@ -2,26 +2,21 @@ version: "3"
...
-      - mock-fxa-server
-      - syncserver
+      mock-fxa-server:
+        condition: service_started
+      tokenserver-db:
+        condition: service_healthy
     image: app:build
     privileged: true
     user: root
+    volumes:
```
nshirley, line 15: @nshirley - note to self, I don't think this is necessary with the docker cp. need to remove and test

docker-compose.e2e.spanner-old.yaml
nshirley, line 1: @nshirley - this can be deleted

syncserver/src/main.rs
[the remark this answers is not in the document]
nshirley: CC: @pjenvey @taddes @jrconlin - tagging you guys directly on this one since it's the thing I'm most worried about. Rust code and all that

tools/integration_tests/test_support.py
```diff
@@ -870,15 +869,16 @@ def run_live_functional_tests(TestCaseClass, argv=None):
 
     # Now use the unittest2 runner to execute them.
     suite = unittest.TestSuite()
-    import test_storage
+    from tokenserver.test_storage import TestStorage
```
nshirley, line 872: @nshirley - note to self: I think most of this can be removed now that it's not being used. The `run.py` files can also be deleted

tools/integration_tests/tokenserver/conftest.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+@pytest.fixture(scope="session")
+def setup_server_local_testing():
+    """
+    Fixture to set up the server for local testing.
+    This fixture sets the necessary environment variables and
+    starts the server.
+    """
+    print("Using setup_server_local_testing fixture")
```
nshirley, line 113: @nshirley - note to self: remove comment

tools/integration_tests/tokenserver/conftest.py
```diff
@@ -0,0 +1,163 @@
...
+    """
+    Fixture to set up the server for end-to-end testing.
+    This fixture sets the necessary environment variables and
+    starts the server.
+    """
+    _set_local_test_env_vars()
+    # debatable if this should ONLY be here since it was only
+    # done against the "run_end_to_end_tests" prior, of if we
+    # just do it in _set_local_test_env_vars...
+    use_oauth = os.environ.get("USE_OAUTH")
```
nshirley, line 145: @nshirley - note to self: remove this print

tools/integration_tests/tokenserver/test_authorization.py
```diff
@@ -1,12 +1,18 @@
...
+import pytest
 import unittest
 from tokenserver.test_support import TestCase
 
 
+@pytest.mark.local_integration
+@pytest.mark.usefixtures('setup_server_local_testing_with_oauth')
 class TestAuthorization(TestCase, unittest.TestCase):
     def setUp(self):
+        # print("Environment Variables in setup...:")
```
nshirley, line 13: @nshirley - note to self: remove comment

tools/integration_tests/tokenserver/test_authorization.py
```diff
@@ -125,6 +133,12 @@ def test_keys_changed_at_less_than_equal_to_generation(self):
                                            client_state='aaaa')
         self.app.get('/1.0/sync/1.5', headers=headers)
 
+    # In testing for the pytest migration, this test is flaky
```
nshirley, line 136: @nshirley - note to self: remove comment

tools/integration_tests/tokenserver/test_storage.py
```diff
@@ -836,57 +835,30 @@ def test_quota(self):
         self.retry_put_json(self.root + "/storage/xxx_col2/12345", bso)
         res = self.app.get(self.root + "/info/quota")
         used = res.json[0]
-        self.assertEquals(used - old_used, len(_PLD) / 1024.0)
-
-    def test_overquota(self):
-        # This can't be run against a live server.
-        raise unittest.SkipTest
```
nshirley, line 843: This test was _always_ skip because of this so I removed it. Happy to add it back if necessary

docker-compose.e2e.mysql.yaml
nshirley: deleted, need to test

Makefile
```diff
@@ -104,4 +105,29 @@ merge_coverage_results:
...
+		-f docker-compose.e2e.mysql.yaml \
+	 	up \
+	 	--exit-code-from mysql-e2e-tests \
+	 	--abort-on-container-exit;
+	exit_code=$$?
+	docker cp mysql-e2e-tests:/mysql_integration_results.xml ${LOCAL_INTEGRATION_JUNIT_XML}
+	exit $$exit_code
+
+.ONESHELL:
+run_spanner_e2e_tests:
```
pjenvey, line 123: I'd like to make it clear that these invoke docker, let's prefix them with `docker_` similar to the target `docker_start_mysql` and the like (and maybe they should live in that section of the file too).

While you're there could you also change them to invoke `docker compose` (w/ a space) instead of the old `docker-compose`?
```suggestion
docker_run_mysql_e2e_tests:
	docker compose \
		-f docker-compose.mysql.yaml \
		-f docker-compose.e2e.mysql.yaml \
	 	up \
	 	--exit-code-from mysql-e2e-tests \
	 	--abort-on-container-exit;
	exit_code=$$?
	docker cp mysql-e2e-tests:/mysql_integration_results.xml ${LOCAL_INTEGRATION_JUNIT_XML}
	exit $$exit_code

.ONESHELL:
docker_run_spanner_e2e_tests:
```

docker-compose.e2e.spanner.yaml
```diff
@@ -11,17 +11,26 @@ services:
     # inside its docker that would do this for us.
     entrypoint: >
       /bin/sh -c "
-        sleep 15;
```
pjenvey, line 14: this whole syncserver section should be killed like in `e2e.mysql`.

I'll note that syncserver is still invoked even when this section doesn't exist -- I'm not sure why we've included empty `sync-db/sync-db-setup` /etc sections here.

syncserver/src/main.rs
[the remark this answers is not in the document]
pjenvey: let's kill this, I prefer the service to fail fast and any probing logic like this be the responsibility of the caller, especially now that we have docker/docker-compose's help with that.

tools/integration_tests/tokenserver/pytest.ini
```diff
@@ -0,0 +1,7 @@
+[pytest]
+markers =
+    local_integration: marks tests for running as local integration tests
+    live_integration: marks tests for running as live integration tests
+
+adopts =
```
pjenvey, line 6: I noticed this due to a warning in the log: `/usr/local/lib/python3.9/dist-packages/_pytest/config/__init__.py:1441: PytestConfigWarning: Unknown config option: adopts`
```suggestion
addopts =
```

docker-compose.spanner.yaml
```diff
@@ -29,12 +29,21 @@ services:
     restart: always
     ports:
       - "3306"
+    command: --explicit_defaults_for_timestamp
```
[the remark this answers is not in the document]
nshirley: oops, no! I'll remove it

tools/integration_tests/tokenserver/test_storage.py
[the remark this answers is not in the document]
nshirley: oh no, I don't remember moving this! I'll move it back and make sure it still runs though. 

This does open up the question too - should I make sure we have separate runs for tokenserver and syncstorage since the point here is we're generating test results to import to the test pipeline? I suppose we could just say they're all e2e and lump together, but we could also separate them I think. So we could see metrics on "synstorage e2e test results" or "tokenserver e2e test results" (I'd need to double check how that's setup/configured in the test pipeline)

docker-compose.e2e.spanner.yaml
nshirley: Apologies, I'm not sure I follow. We want to remove the `  sync-db:` service from `e2e.mysql.yaml`, and then the `syncserver` and `sync-db` sections from `mysql.yaml`?

nshirley: ah, nvm, I think I'm understanding what you're saying - I shouldn't review pr comments late at night

tools/integration_tests/tokenserver/test_storage.py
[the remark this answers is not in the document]
pjenvey: In the long term we want to further unify tokenserver with syncstorage (we refer to the combination as "syncserver"), so I think it's fine keeping them together as one single project.

tools/integration_tests/tokenserver/test_authorization.py
```diff
@@ -75,6 +77,8 @@ def test_invalid_client_state_in_x_client_state(self):
 
         res = self.app.get('/1.0/sync/1.5', headers=headers, status=400)
 
+        print("Response: ", res.json)
```
nshirley, line 80: @nshirley - remove this

docker-compose.spanner.yaml
```diff
@@ -53,7 +59,11 @@ services:
       - "8000:8000"
     depends_on:
       - sync-db-setup
+    volumes:
```
nshirley, line 62: This and the `RUST_LOG` probably aren't needed, right?

.circleci/config.yml
```diff
@@ -453,10 +456,21 @@ jobs:
...
+          name: Restore Makefile from save_cache
+          command: cp /home/circleci/cache/Makefile .
       - run:
           name: Restore docker-compose config
           command: cp /home/circleci/cache/docker-compose*.yaml .
-      - run-e2e-spanner-tests
+      - make-test-dir
+      - run-e2e-tests:
+          db: spanner
+          JWK_CACHE_ENABLED: true
```
pjenvey, line 468: Does it make sense to keep this env var here now that the compose test runner always toggles it? I think it should be killed

tools/integration_tests/conftest.py
```diff
@@ -0,0 +1,162 @@
...
+    """
+    Fixture to set up the server for end-to-end testing.
+    This fixture sets the necessary environment variables and
+    starts the server.
+    """
+    _set_local_test_env_vars()
+    # debatable if this should ONLY be here since it was only
+    # done against the "run_end_to_end_tests" prior, of if we
+    # just do it in _set_local_test_env_vars...
+    if not JWK_CACHE_ENABLED:
```
pjenvey, line 146: nit: my bad, you were probably correct initially to call this `"_DISABLED"` since it triggers the deletion, but I won't block the PR on the naming :)

pjenvey (changes requested):

.circleci/config.yml
nshirley: That's a good point - I can just set it in the docker-compose env vars, then in execution it's left as is (set to false in the entrypoint between runs)

tools/integration_tests/conftest.py
nshirley: Since I was in there for the circle updates I fixed this too, pretty easy to swap it out too.

pjenvey (changes requested):
just a couple more `JWK_CACHE_ENABLED` in the entrypoints

PR #1698, opened 2025-05-09
fix(infra): configure gcp utils before upload

nshirley:
## Description

I broke the build on master because the final step for e2e tests tries to upload to GCS without being authenticated with GCP first. This just moves the authentication into the same command so we don't have to remember to add the auth step. It also cleans up other references to the step

- Move authentication setup for GCP into upload command
- Update name to be more descriptive for gcs upload
- Update references to use new name and remove redundant GCP authentication setup

Closes STOR-255

## Testing

[Here](https://app.circleci.com/pipelines/github/mozilla-services/syncstorage-rs/5761/workflows/a35d04fe-65db-41ad-a3b1-a9ee7b2d1da8/jobs/23402) is a run where I tested with my branch name to see it work. The interesting caveat with this change is that we now see duplicate steps for configuring if upload needs to happen multiple times, here's an example: 

image

If we don't like this I can revert it and make sure the configure step is explicitly called everywhere it's needed!

## Issue(s)

Closes [STOR-255](https://mozilla-hub.atlassian.net/browse/STOR-255).

[STOR-255]: https://mozilla-hub.atlassian.net/browse/STOR-255?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -158,7 +158,7 @@ commands:
           environment:
             SYNCSTORAGE_RS_IMAGE: app:build
 
-  upload-to-gcs:
+  gcs-configure-and-upload:
     parameters:
       source:
         type: string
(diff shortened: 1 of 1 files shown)


pjenvey (approved):

taddes (approved):
Looks good, though I do see a duplication of the `Install latest gcloud CLI version, if not available` & `Initialize gcloud CLI to connect to Google Cloud` steps. Perhaps it's not necessary to have those run one after another? At present, it's between the `workflow/test-results xml files`. Not a big deal in the end since it doesn't appear to be consuming time, but something to look at splitting out so it only runs once.

PR #1701, opened 2025-06-03
refactor: quiet dbg calls

taddes:
## Description
Removed `dbg!` call and annotated some commented out calls in testing that indicate it should only be used in testing.

Added a command in the two clippy commands, in the Makefile and CI, to deny the use of `dbg!` macro calls.

See clippy docs here for [dbg_macro](https://rust-lang.github.io/rust-clippy/master/index.html#dbg_macro)

## Testing

Run clippy

## Issue(s)

Closes [STOR-79](https://mozilla-hub.atlassian.net/browse/STOR-79).

[STOR-79]: https://mozilla-hub.atlassian.net/browse/STOR-79?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,14 +63,12 @@ commands:
     steps:
       - run:
           name: Rust Clippy MySQL
-          command: |
-            cargo clippy --workspace --all-targets --no-default-features --features=syncstorage-db/mysql --features=py_verifier -- -D clippy::dbg_macro -D warnings
+          command: make clippy_mysql
   rust-clippy-spanner:
     steps:
       - run:
           name: Rust Clippy Spanner
-          command: |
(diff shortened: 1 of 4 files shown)


pjenvey (approved):

.circleci/config.yml
```diff
@@ -64,13 +64,13 @@ commands:
       - run:
           name: Rust Clippy MySQL
           command: |
-            cargo clippy --workspace --all-targets --no-default-features --features=syncstorage-db/mysql --features=py_verifier -- -D warnings
+            cargo clippy --workspace --all-targets --no-default-features --features=syncstorage-db/mysql --features=py_verifier -- -D clippy::dbg_macro -D warnings
```
pjenvey: nit: I think we could just replace these with the Makefile target equivalents at this point?

taddes: Good call, one less thing to maintain 👍

PR #1706, opened 2025-06-13
feat: use poetry for dependency mangement

taddes:
## Description

Describe these changes.

To align with other Python projects, the syncserver project should use poetry for Python dependency management. 

This means the incorporation of the `pyproject.toml` files with matching `poetry.lock` files that pin dependencies. `README.md` files are updated with new conventions and Dockerflow files that used to call builds using pip are replaced with Poetry.

## Testing

Running individual scripts with new poetry instructions. All `README.md` files are updated with Poetry and pyenv virtualenv instructions. 

## Issue(s)

Closes [link](link).


Dockerfile
```diff
@@ -47,10 +53,14 @@ RUN \
...
     apt-get -q update && \
-    apt-get -q install -y --no-install-recommends $MYSQLCLIENT_PKG cmake golang-go python3-dev python3-pip python3-setuptools python3-wheel && \
+    apt-get -q install -y --no-install-recommends $MYSQLCLIENT_PKG cmake golang-go python3-dev python3-pip python3-setuptools python3-wheel pkg-config && \
+    curl -sSL https://install.python-poetry.org | python3 - && \
+    ln -s /root/.local/bin/poetry /usr/local/bin/poetry && \
     pip3 install -r requirements.txt && \
     rm -rf /var/lib/apt/lists/*
 
+
+
```
pjenvey, line 70:
```suggestion
```

pjenvey (approved):

PR #1707, opened 2025-06-16
test: make StreamedResultSet's stream generic

pjenvey:
## Description

adds a test for STOR-212 (you can see the change outside of the refactor in the second commit)

## Issue(s)

Issue STOR-212


taddes (approved):
LGTM, thanks for going over this today and discussing the context related to the bug. We can land this and I'll build over it for the `next_async` / `one_or_none` patch

PR #1708, opened 2025-06-18
bug : fix stream one or none

taddes:
## Description

The one_or_none method occasionally emits errors on SQL that can only produce at most one result.

Seeing this occur from update_user_collection_quotas, get_storage_timestamp and lock_for_write_async

https://sentry.io/organizations/mozilla/issues/3517020595/?environment=prod&project=6316464&query=is%3Aunresolved

https://sentry.io/organizations/mozilla/issues/3514687669/?environment=prod&project=6316464&query=is%3Aunresolved

https://sentry.io/organizations/mozilla/issues/3516548273/?environment=prod&project=6316464&query=is%3Aunresolved

We surmise the issue here is a bug in one_or_none: it doesn’t notice its second call to next_async could potentially be an Error. It should always propagate errors from next_async before validating its result (always call next_async().transpose()?). 

This was easy to miss because next_async returns Option<Result> (vs Result<Option>) to match poll_next (which it may rewritten into at some point).

[TryStream::try_next](https://docs.rs/futures/latest/futures/stream/trait.TryStreamExt.html#method.try_next) as the model for changing the implementation of `next_async`
## Issue(s)

Closes [STOR-212](https://mozilla-hub.atlassian.net/browse/STOR-212).

[STOR-212]: https://mozilla-hub.atlassian.net/browse/STOR-212?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/syncstorage-spanner/src/models.rs
+++ b/syncstorage-spanner/src/models.rs
@@ -1412,8 +1412,7 @@ impl SpannerDb {
 
         let mut ids = vec![];
         let mut modifieds = vec![];
-        while let Some(row) = stream.try_next().await? {
-            let mut row = row;
+        while let Some(mut row) = stream.try_next().await? {
             ids.push(row[0].take_string_value());
             modifieds.push(sync_timestamp_from_rfc3339(row[1].get_string_value())?.as_i64());
         }
@@ -1589,8 +1588,7 @@ impl SpannerDb {
             .param_types(sqlparam_types)
             .execute_async(&self.conn)?;
         let mut existing = HashSet::new();
-        while let Some(row) = streaming.try_next().await? {
-            let mut row = row;
+        while let Some(mut row) = streaming.try_next().await? {
             existing.insert(row[0].take_string_value());
         }
         let mut inserts = vec![];
--- a/syncstorage-spanner/src/models.rs
+++ b/syncstorage-spanner/src/models.rs
@@ -604,7 +604,7 @@ impl SpannerDb {
             .param_types(sqlparam_types)
             .execute_async(&self.conn)?;
         let mut results = HashMap::new();
-        while let Some(row) = streaming.next_async().await? {
+        while let Some(row) = streaming.try_next().await? {
             let collection_id = row[0]
                 .get_string_value()
                 .parse::<i32>()
@@ -658,8 +658,7 @@ impl SpannerDb {
                 )?
                 .params(params)
                 .execute_async(&self.conn)?;
-            while let Some(row) = rs.next_async().await? {
-                let mut row = row;
+            while let Some(mut row) = rs.try_next().await? {
                 let id = row[0]
                     .get_string_value()
                     .parse::<i32>()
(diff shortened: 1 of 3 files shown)


syncstorage-spanner/src/models.rs
```diff
@@ -659,8 +658,8 @@ impl SpannerDb {
                 )?
                 .params(params)
                 .execute_async(&self.conn)?;
-            while let Some(row) = rs.next_async().await {
-                let mut row = row?;
+            while let Some(row) = rs.next_async().await? {
+                let mut row = row;
```
pjenvey: These extra assignments can be killed via:
```suggestion
            while let Some(mut row) = rs.next_async().await? {
```

syncstorage-spanner/src/stream.rs
```diff
@@ -144,19 +144,19 @@ where
...
             match self.consume_next().await {
-                Ok(true) => (),
-                Ok(false) => return None,
+                Ok(true) => {}
+                Ok(false) => return Ok(None),
                 // Note: Iteration may continue after an error. We may want to
                 // stop afterwards instead for safety sake (it's not really
                 // recoverable)
-                Err(e) => return Some(Err(e)),
+                Err(e) => return Err(e),
```
pjenvey: We can kill this `Err` arm of the match by propagating it earlier via `self.consume_next().await?` -- and once you do that, you'll notice the match statement itself is no longer necessary (and killing it is a nice cleanup)

syncstorage-spanner/src/stream.rs
```diff
@@ -144,19 +144,19 @@ where
     }
 
     // We could implement Stream::poll_next instead of this, but
-    // this is easier for now and we can refactor into the trait later
-    pub async fn next_async(&mut self) -> Option<DbResult<Vec<Value>>> {
+    // this is easier for now and we can refactor into the trait later.
+    pub async fn next_async(&mut self) -> DbResult<Option<Vec<Value>>> {
```
pjenvey: Let's also rename this to more closely mimic [TryStream's naming](https://docs.rs/futures/latest/futures/prelude/stream/trait.TryStreamExt.html#method.try_next)
```suggestion
    pub async fn try_next_async(&mut self) -> DbResult<Option<Vec<Value>>> {
```

pjenvey (changes requested):
Also don't forget to uncomment the new unit test which will now pass

syncstorage-spanner/src/models.rs
taddes: Snap, good call, didn't even think of that

syncstorage-spanner/src/models.rs
```diff
@@ -1417,8 +1412,8 @@ impl SpannerDb {
 
         let mut ids = vec![];
         let mut modifieds = vec![];
-        while let Some(row) = stream.next_async().await {
-            let mut row = row?;
+        while let Some(row) = stream.try_next().await? {
+            let mut row = row;
```
pjenvey: there's a couple more of these here and below
```suggestion
        while let Some(mut row) = stream.try_next().await? {
```

pjenvey (changes requested):

pjenvey (approved):

PR #1711, opened 2025-06-20
test: add spanner db tests to ci

taddes:
## Description

Historically we never ran the db unit tests under Spanner in circleci due to lacking an easy means of doing so – but we can do so now via the Spanner emulator.

It looks like doing so might only add a short additional amount of time to our CI build which makes it worth doing.

## Testing

These tests follow the same local setup of a spanner emulator and testing against the emulator. The behavior should be akin to that setup, however utilizing the spanner emulator in CI.

## Issue(s)

Closes [STOR-61](https://mozilla-hub.atlassian.net/browse/STOR-61).

[STOR-61]: https://mozilla-hub.atlassian.net/browse/STOR-61?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scripts/prepare-spanner-unit-tests.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/sh
```
pjenvey: I attempted a branch with the entrypoint alternative I suggested, it [seems to work](https://app.circleci.com/pipelines/github/mozilla-services/syncstorage-rs?branch=test%2Fspanner-db-tests-ci-STOR-61-entrypoint-pjenvey) and negates the need for separate scripts:

https://github.com/mozilla-services/syncstorage-rs/compare/test/spanner-db-tests-ci-STOR-61...test/spanner-db-tests-ci-STOR-61-entrypoint-pjenvey

Want to merge this in and kill this file?

.circleci/config.yml
```diff
@@ -156,6 +156,39 @@ commands:
...
+          environment:
+            SYNC_SYNCSTORAGE__DATABASE_URL: spanner://projects/test-project/instances/test-instance/databases/test-database
+            SYNC_SYNCSTORAGE__SPANNER_EMULATOR_HOST: 127.0.0.1:9020
+          command: scripts/prepare-spanner-unit-tests.sh
+      - run:
+          name: cargo test spanner
+          environment:
+            SYNC_SYNCSTORAGE__DATABASE_URL: spanner://projects/test-project/instances/test-instance/databases/test-database
+            SYNC_SYNCSTORAGE__SPANNER_EMULATOR_HOST: 127.0.0.1:9020
+          command: cargo test --no-default-features -p syncstorage-spanner
```
pjenvey: I'm just now realizing that we should be capturing the junit output XML of this run and collecting them for our test metrics -- essentially a new version of `make test_with_coverage` that runs through nextest with a specific coverage filename for spanner unit tests that then gets uploaded.

Let's merge this as is though and log another issue for making that happen, we may need to coordinate this w/ @nshirley on its setup.

pjenvey (changes requested):

scripts/prepare-spanner-unit-tests.sh
taddes: Totally, that's a better solution for sure. Will add that to the Docker quiver for the future.

.circleci/config.yml
pjenvey: I think we should `cargo test` basically the entire test suite vs just the `syncstorage-spanner` crate.

There's a number of tests that wouldn't utilize the spanner code at all (like tokenserver*) but just `syncstorage-spanner` isn't enough: our larger db unit test suite lives in the `syncstorage-db` crate, and additionally `syncserver` itself has a [integration test suite](https://github.com/mozilla-services/syncstorage-rs/blob/master/syncserver/src/server/test.rs) that utilizies the database.

pjenvey (changes requested):

.circleci/config.yml
```diff
@@ -156,6 +156,39 @@ commands:
...
+            if ! nc -z 127.0.0.1 9020; then
+              echo "ERROR: Cannot connect to Spanner emulator at 127.0.0.1:9020. Terminating."
+              exit 1
+            fi
+      - run:
+          name: setup spanner
+          environment:
+            SYNC_SYNCSTORAGE__DATABASE_URL: spanner://projects/test-project/instances/test-instance/databases/test-database
+            SYNC_SYNCSTORAGE__SPANNER_EMULATOR_HOST: 127.0.0.1:9020
+          command: scripts/prepare-spanner-unit-tests.sh
```
pjenvey:
```suggestion
          command: scripts/prepare-spanner.sh
```

.circleci/config.yml
pjenvey: from nick re: "do we simply need to create new unit test xml file, upload it to the bucket?"

> Yep, it should be that simple. As long as the output .xml results file has a unique name during upload, it'll get picked up and processed. MIGHT need to tweak the dashboard, but I don't think so. You can use the gcs-configure-and-upload command after running tests, and it'll take care of the upload.

..so we can just follow the pattern of having a separate Makefile target with its own unique filename for this

taddes:
Thx for confirming some of the convoluted things going on with cargo on this one @pjenvey, I'll file an issue to look at those in the future and also one for the nexttest reporting 👍

PR #1713, opened 2025-07-01
chore: strip actix-web features

pjenvey:
## Description

while providing an optional actix-compress feature

also disables actix-web's cookies and unicode features

(note I'm not seeing a significant improvement in dev build times with this so far but we'll keep an eye on CI)

## Issue(s)

Closes STOR-256


taddes (approved):
👍 looks good, nice to get all of the 3 compression features. Perhaps we'll see the benefits in some other places and can still see how it affects CI down the road.

PR #1714, opened 2025-07-07
feat: spanner scripts parse gcp project

taddes:
## Description
The `SYNC_SYNCSTORAGE__DATABASE_URL` environment variable defines a spanner DSN URL that is used for parsing out the instance_id & database_id from. However, it was not parsing GCP project, which defaults to the project determined from the environment when not specified. These are defined within webservices-infra in the sync project.

This PR adds the parsing logic to pull out the project_id. Several modernizations and improvements are made to the scripts. It also adds unit tests for better maintainability. 

We’re currently working around this by specifying a GOOGLE_CLOUD_PROJECT env var, which the Spanner client recognizes, but the script should really do it itself since the project is always included in the database url setting.

## Testing

Visit webservices-infra configurations and Kubernetes workflow runs to verify.

New unit tests have been added to validate logic and several bugs were found and now fixed from the addition of tests.

## Issue(s)

Closes [STOR-290](https://mozilla-hub.atlassian.net/browse/STOR-290).

[STOR-290]: https://mozilla-hub.atlassian.net/browse/STOR-290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tools/spanner/purge_ttl.py
+++ b/tools/spanner/purge_ttl.py
@@ -28,6 +28,16 @@ client = spanner.Client()
 
 
 def use_dsn(args):
+    """
+    Function that extracts the instance, project, and database ids from the DSN url.
+    It is defined as the SYNC_SYNCSTORAGE__DATABASE_URL environment variable.
+    The defined defaults are in webservices-infra/sync and can be configured there for
+    production runs. 
+
+    For reference, an example spanner url passed in is in the following format:
+    `spanner://projects/moz-fx-sync-prod-xxxx/instances/sync/databases/syncdb`
+    database_id = `syncdb`, instance_id = `sync`, project_id = `moz-fx-sync-prod-xxxx`
+    """
     try:
         if not args.sync_database_url:
             raise Exception("no url")
@@ -40,7 +50,7 @@ def use_dsn(args):
             args.database_id = path[-1]
     except Exception as e:
         # Change these to reflect your Spanner instance install
-        print("Exception {}".format(e))
+        print(f"Exception {e}")
     return args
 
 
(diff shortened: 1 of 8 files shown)


tools/spanner/purge_ttl.py
```diff
@@ -25,22 +27,32 @@
 
 # Change these to match your install.
 client = spanner.Client()
+# DSN_URL = "SYNC_SYNCSTORAGE__DATABASE_URL"
```
jrconlin: Commented code can be removed?

tools/spanner/purge_ttl.py
```diff
@@ -225,17 +273,26 @@ def get_args():
 
     # override using the DSN URL:
     if args.sync_database_url:
-        args = use_dsn(args)
+        (instance_id, database_id, project_id) = ids_from_env(args.sync_database_url)
+        args.instance_id = instance_id
+        args.database_id = database_id
+        args.project_id = project_id
+        import pdb
```
jrconlin: remove the debugger.

tools/spanner/test_count_expired_rows.py
```diff
@@ -0,0 +1,85 @@
+import os
+import types
+from unittest.mock import MagicMock
+import pytest
+import logging
+
+import count_expired_rows
+
+@pytest.fixture(autouse=True)
```
jrconlin: Fetch this from `test_utils.py`? ;)

jrconlin (changes requested):

tools/spanner/utils.py
```diff
@@ -0,0 +1,68 @@
...
+        if mode == Mode.ENV_VAR:
+            url = os.environ.get(dsn)
+            if not url:
+                raise Exception(f"No env var found for provided DSN: {dsn}")
+        elif mode == Mode.URL:
+            url = dsn
+            if not url:
+                raise Exception(f"No valid url found: {url}")
+        parsed_url = parse.urlparse(url)
+        if parsed_url.scheme == "spanner":
```
pjenvey, line 53: FYI in general I'd prefer failing fast if a bad/invalid url was provided, but this matches the original behavior before this PR so this is fine.

tools/spanner/utils.py
```diff
@@ -0,0 +1,68 @@
...
+Environment variable that stores Sync database URL
+Depending on deployment, can be MySQL or Spanner.
+In this context, should always point to spanner for these scripts.
+"""
+
+class Mode(Enum):
+    URL = auto()
+    ENV_VAR = auto()
+
+def ids_from_env(dsn=DSN_URL, mode=Mode.ENV_VAR) -> Tuple[str, str, str]:
```
pjenvey, line 24: nit: with the limited usage of this function I'd argue it could be simplified a bit (albeit being less flexible) -- by making the dsn arg `dsn=None` and treating `None` as the `Mode.ENV_VAR` case with a hardcoded `SYNC_SYNCSTORAGE__DATABASE_URL` env name, without the need for a `Mode` enum.

taddes: Yeah, I gotcha. This came to me because each case of this almost identical function used the env var, but in one case it used the url for the dsn. Either way, maybe we can keep the mode to be really explicit about what's going on? If I hadn't written the tests to go with this, it might have taken me longer to sort out the bug.

pjenvey: I'm fine with what you've got here (I approved the PR) -- just figured I'd add the feedback, probably should have made it clear it's not blocking this

PR #1715, opened 2025-07-10
refactor: convert db extract calls to async

taddes:
## Description

The default implementation of Db::extract_resource doesn’t use async/await. The core of this method should be converted to an {{async {}}} block so it can utilize async/await for better readability (as we’ve already done for e.g. [many of the extractors](https://github.com/mozilla-services/syncstorage-rs/issues/544)).
## Issue(s)

Closes [STOR-101](https://mozilla-hub.atlassian.net/browse/STOR-101).

[STOR-101]: https://mozilla-hub.atlassian.net/browse/STOR-101?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/syncstorage-db-common/src/lib.rs
+++ b/syncstorage-db-common/src/lib.rs
@@ -189,45 +189,42 @@ pub trait Db: Debug {
         collection: Option<String>,
         bso: Option<String>,
     ) -> DbFuture<'_, SyncTimestamp, Self::Error> {
-        // If there's no collection, we return the overall storage timestamp
-        let collection = match collection {
-            Some(collection) => collection,
-            None => return Box::pin(async move { self.get_storage_timestamp(user_id).await }),
-        };
-        // If there's no bso, return the collection
-        let bso = match bso {
-            Some(bso) => bso,
-            None => {
-                return Box::pin(async move {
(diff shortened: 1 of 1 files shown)


pjenvey (changes requested):
For these conversion to async/await changes, it's best to follow what we did with most of the extractors: basically convert the entire block of the method into an async block (this often just works if there's no borrowing needed) e.g. https://github.com/mozilla-services/syncstorage-rs/blob/b0c8ac50a0ec4ebb3e19ecb9a37eb7b9501da507/syncserver/src/web/extractors.rs#L747.

FYI there's also an alternative conversion for this specific method (since this is our own trait): switching the entire trait to either use the `async-trait` crate or potentially using the limited [async fn in trait](https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits/#async-fn-in-public-traits) support added in Rust 1.75 (at some point the `DbPool` trait in this module was switched to `async-trait`) -- but I think either of those become large changes.

taddes:
Sounds good @pjenvey , opted for the simpler approach, instead of switching to using the `async-trait`

pjenvey (approved):

PR #1717, opened 2025-07-15
chore: require minimum mysql 8

pjenvey:
## Description

BREAKING CHANGE: now depending on >= mysql 8 features

## Issue(s)

Closes STOR-160

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -271,7 +271,7 @@ jobs:
           RUST_BACKTRACE: 1
           # XXX: begin_test_transaction doesn't play nice over threaded tests
           RUST_TEST_THREADS: 1
-      - image: cimg/mysql:5.7
+      - image: cimg/mysql:8.0
         auth:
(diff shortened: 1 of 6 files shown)


pjenvey:
> Was there a compat concern with some newer versions and that informs going with 8?

@taddes A couple reasons:

- We're now using Cloud SQL MySQL with a compatibility level of 8.0.x ([8.0 is still the current default](https://cloud.google.com/sql/docs/mysql/db-versions#database-version-support))
- I think it's best to stay as flexible as possible with at least a few self hosting users out there  -- I wasn't even that eager to get off 5 (but probably should have been), but it's been long deprecated and it's best to test against what's closest to what we're using on production. So it was a good opportunity to bump the minimum version now, it has a couple other benefits:
  - kills the `LockInShareMode` code
  - there was no arm64 mysql 5 images on docker hub (but 8 has them) so the test suite should finally work out of the box on e.g. macos

PR #1740, opened 2025-08-07
feat: quiet pool timeout events (emit as metrics)

pjenvey:
## Issue(s)

Closes STOR-295

### DIFF
--- a/syncstorage-spanner/src/error.rs
+++ b/syncstorage-spanner/src/error.rs
@@ -144,21 +144,24 @@ impl ReportableError for DbError {
     }
 
     fn metric_label(&self) -> Option<&str> {
-        Some(match &self.kind {
-            DbErrorKind::Common(e) => return e.metric_label(),
+        match &self.kind {
+            DbErrorKind::Common(e) => e.metric_label(),
 
             DbErrorKind::Grpc(grpcio::Error::RpcFailure(status)) => {
                 match status.code() {
                     // Code 14 - UNAVAILABLE
-                    RpcStatusCode::UNAVAILABLE => "storage.spanner.grpc.unavailable",
+                    RpcStatusCode::UNAVAILABLE => Some("storage.spanner.grpc.unavailable"),
                     // Code 3 - INVALID_ARGUMENT
-                    RpcStatusCode::INVALID_ARGUMENT => "storage.spapner.grpc.invalid_argument",
-                    _ => return None,
+                    RpcStatusCode::INVALID_ARGUMENT => {
+                        Some("storage.spanner.grpc.invalid_argument")
+                    }
+
+                    _ => None,
                 }
             }
-            DbErrorKind::PoolTimeout(_) => "storage.spanner.pool.timeout",
-            _ => return None,
-        })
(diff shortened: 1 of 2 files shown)


syncstorage-spanner/src/error.rs
```diff
@@ -131,22 +138,35 @@ impl ReportableError for DbError {
...
 
             DbErrorKind::Grpc(grpcio::Error::RpcFailure(status)) => {
                 match status.code() {
-                    RpcStatusCode::UNAVAILABLE => Some("grpc.unavailable"), // Code 14 - UNAVAILABLE
-                    RpcStatusCode::INVALID_ARGUMENT => Some("grpc.invalid_argument"), // Code 3 - INVALID_ARGUMENT
-                    _ => None,
+                    // Code 14 - UNAVAILABLE
+                    RpcStatusCode::UNAVAILABLE => "storage.spanner.grpc.unavailable",
+                    // Code 3 - INVALID_ARGUMENT
+                    RpcStatusCode::INVALID_ARGUMENT => "storage.spapner.grpc.invalid_argument",
```
taddes:
```suggestion
                    RpcStatusCode::INVALID_ARGUMENT => "storage.spanner.grpc.invalid_argument",
```

syncstorage-spanner/src/error.rs
```diff
@@ -131,22 +138,35 @@ impl ReportableError for DbError {
                     _ => true,
                 }
             }
+            DbErrorKind::PoolTimeout(_) => false,
             _ => true,
         }
     }
 
     fn metric_label(&self) -> Option<&str> {
```
taddes: I couldn't help but wonder if it's best to keep the previous return format by wrapping the match arm results in `Some(...)`, instead of the whole match statement. The early returns in the match arms threw me a little bit. I'll still defer to you on this if you want to keep it. I just felt it was a little harder to read, namely with the `return` for `None` and `e.metric_label()`
```suggestion
    fn metric_label(&self) -> Option<&str> {
        match &self.kind {
            DbErrorKind::Common(e) => e.metric_label(),
            DbErrorKind::Grpc(grpcio::Error::RpcFailure(status)) => {
                match status.code() {
                    // Code 14 - UNAVAILABLE
                    RpcStatusCode::UNAVAILABLE => Some("storage.spanner.grpc.unavailable"),
                    // Code 3 - INVALID_ARGUMENT
                    RpcStatusCode::INVALID_ARGUMENT => {
                        Some("storage.spanner.grpc.invalid_argument")
                    }
                    _ => None,
                }
            }
            DbErrorKind::PoolTimeout(_) => Some("storage.spanner.pool.timeout"),
            _ => None,
        }
    }
```

pjenvey: That's fair, I've switched it back. I'll note that we do have this other pattern with early returns in at least one other place (across sync/push) -- I'd argue it's preferable when there's a large list of labels to define as it negates the need to repeat `Some` for every entry.

taddes: Cool, yeah, I can see it would make sense if there’s a whole big list of it.

PR #1778, opened 2025-08-29
feat: Upgrade to Diesel 2.2 (take 2)

pjenvey:
## Description

This is https://github.com/mozilla-services/syncstorage-rs/pull/1647 with @tommie's commits signed (by me recommitting them which switches the `committer` field to me but keeps him as author)

I've reverted https://github.com/mozilla-services/syncstorage-rs/commit/d4511cc746d8c2e991c8faf55ef608e8ff7895bc for now as it's not clear if it's necessary and changed https://github.com/mozilla-services/syncstorage-rs/commit/2a1268ed0440de4c5709b150b37c440369bf2a2e to become a retry w/ incr for creating new batches.

One of the major changes here is diesel 2 requiring mutability, which is quickly solved by surrounding the connection with a `RwLock`. Switching to `diesel-async` and `&mut self` trait methods will let us remove the need for the lock.

## Issue(s)

Closes STOR-324


syncstorage-mysql/src/batch.rs
```diff
@@ -74,7 +83,7 @@ pub fn validate(db: &MysqlDb, params: params::ValidateBatch) -> DbResult<bool> {
         .filter(batch_uploads::batch_id.eq(&batch_id))
         .filter(batch_uploads::user_id.eq(&user_id))
         .filter(batch_uploads::collection_id.eq(&collection_id))
-        .get_result::<i32>(&db.conn)
+        .get_result::<i32>(&mut *db.conn.write()?)
```
taddes, line 86: I think I recall you mentioned this is the part that acquires the write lock before inserting and we'll be able to change once we land some of the other patches, namely `&mut` for connection references, right?

syncstorage-mysql/src/diesel_ext.rs
```diff
@@ -1,36 +1,70 @@
+use std::{fmt::Debug, marker::PhantomData};
```
taddes, line 1: Not blocking or necessary, but given our goal to make Sync more comprehensible and documented, perhaps a bit of docs in this crate to explain it's to construct the custom wrapper struct for `ON DUPLICATE KEY UPDATE` ORM code for Diesel. To the uninitiated it might be a bit vague.

tokenserver-db/Cargo.toml
```diff
@@ -14,7 +14,6 @@ slog-scope.workspace = true
 thiserror.workspace = true
 
 async-trait = "0.1.88"
-# diesel 1.4
```
taddes, line 17: nit: also got one of these in the syncstorage-db-common [Cargo.toml](https://github.com/mozilla-services/syncstorage-rs/blob/19b6176d5b9d7848b57954a489d494394c175798/syncstorage-db-common/Cargo.toml#L19)

tokenserver-db/src/models.rs
```diff
@@ -26,6 +26,11 @@ const MAX_GENERATION: i64 = i64::MAX;
 
 type Conn = PooledConnection<ConnectionManager<MysqlConnection>>;
 
+#[cfg(not(debug_assertions))]
+type InternalConn = Conn;
+#[cfg(debug_assertions)]
+type InternalConn = LoggingConnection<Conn>; // display SQL when RUST_LOG="diesel_logger=trace"
```
taddes, line 32: we might wanna document this somewhere as it could be very useful when debugging.

taddes (approved):
🚀 LGTM! Just have to watch out for the RwLock concern, but the subsequent changes should address that. Just minor stuff. Thanks for this!

tokenserver-db/src/models.rs
pjenvey: It's unfortunately going away with the move to `diesel-async`

syncstorage-mysql/src/diesel_ext.rs
pjenvey: https://mozilla-hub.atlassian.net/browse/STOR-348

taddes (approved):
👍

PR #1782, opened 2025-09-02
fix: kill unnecessary transactions (savepoints)

pjenvey:
## Description

These methods are only called within `lock_for_write` based transactions created by `transaction_http` (all sync writes get this transaction implicitly), so there's no need for them to initiate a transaction (which actually is a savepoint since a transaction already exists). This was old sqlite impl code (possibly inherited from go-syncstorage).

This also simplifies borrowing for `&mut self`.

## Issue(s)

Issue STOR-327

### DIFF
--- a/syncstorage-mysql/src/models.rs
+++ b/syncstorage-mysql/src/models.rs
@@ -327,16 +327,14 @@ impl MysqlDb {
             return Ok(id);
         }
 
-        let id = self.conn.write()?.transaction(|tx| {
-            diesel::insert_or_ignore_into(collections::table)
-                .values(collections::name.eq(name))
-                .execute(tx)?;
+        diesel::insert_or_ignore_into(collections::table)
+            .values(collections::name.eq(name))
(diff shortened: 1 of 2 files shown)


syncstorage-mysql/src/models.rs
```diff
@@ -917,23 +914,18 @@ impl MysqlDb {
...
             .filter(bso::expiry.gt(self.timestamp().as_i64()))
-            .filter(bso::collection_id.eq(collection_id));
-        let (total_bytes, count): (i64, i32) = if let Some(conn) = conn {
-            q.get_result(conn)
-        } else {
-            q.get_result(&mut *self.conn.write()?)
-        }
-        .optional()?
-        .unwrap_or_default();
+            .filter(bso::collection_id.eq(collection_id))
```
taddes, line 925: Nice improvement to not have the conditional check to run the query 👍

taddes (approved):
lgtm

PR #1788, opened 2025-09-03
refactor: resolve vulns and upgrade

taddes:
## Description
Resolves a number of security vulnerabilities raised by `cargo audit` and several outdated packages that introduced breaking changes, therefore we didn't update. 

Will need some context from @pjenvey on deadpool and protobuf, as the effects of updating those may be larger. grpcio warnings remain.

Regarding the historical issue holding us back with `config`:

The historical comment next to the dep said: "pin to 11, 12+ introduces a breaking change for env vars".

Turns out in config 0.12, the prefix is separated from the key using the same separator we configure. So with separator` __`, the env var `SYNC_TOKENSERVER__ENABLED` would need to become `SYNC__TOKENSERVER__ENABLED`. The old `SYNC_` prefix format no longer worked.

This means in CI, `SYNC_TOKENSERVER__DATABASE_URL` and `SYNC_SYNCSTORAGE__DATABASE_URL` were never parsed and the settings fell back to defaults (empty), causing the test failures.

The fix is to restore the _ prefix delimiter explicitly using the newer`prefix_separator` method.

## Testing
compilation

## Issue(s)

Closes [STOR-352](https://mozilla-hub.atlassian.net/browse/STOR-352).

[STOR-352]: https://mozilla-hub.atlassian.net/browse/STOR-352?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/syncstorage-spanner/Cargo.toml
+++ b/syncstorage-spanner/Cargo.toml
@@ -22,7 +22,7 @@ form_urlencoded = "1.2"
 # syncserver to either fail to either compile, or start. In those cases, try
 # `cargo build --features grpcio/openssl ...`
 grpcio = { version = "0.13.0", features = ["openssl"] }
-protobuf = { version = "2.28.0" } # must match what's used by google-cloud-rust-raw
+protobuf = { version = "=2.28.0" } # must match what's used by google-cloud-rust-raw
 syncserver-common = { path = "../syncserver-common" }
 syncserver-db-common = { path = "../syncserver-db-common" }
 syncstorage-db-common = { path = "../syncstorage-db-common" }


syncserver-settings/Cargo.toml
```diff
@@ -9,7 +9,7 @@ edition.workspace = true
 serde.workspace = true
 slog-scope.workspace = true
 
-config = "0.11"                                             # pin to 11, 12+ introduces a breaking change for env vars.
+config = "0.15"                                             # pin to 11, 12+ introduces a breaking change for env vars.
```
pjenvey:
```suggestion
config = "0.15"
```

pjenvey:
> Will need some context from @pjenvey on deadpool and protobuf, as the effects of updating those may be larger. grpcio warnings remain.

I don't recall any issues with moving to a new deadpool.

Are there grpcio warnings on this branch currently? Or when upgrading protobuf?

We should be able to go to max protobuf 2.28 per the `google-cloud-rust-raw` package we depend on, which locks it to that version. It looks like we have a protobuf req in the root Cargo.toml but I'm not sure why, it's only used by syncstorage-spanner which specifies its own dep there. Can we kill it from the root?

taddes:
> > Will need some context from @pjenvey on deadpool and protobuf, as the effects of updating those may be larger. grpcio warnings remain.
> 
> I don't recall any issues with moving to a new deadpool.
> 
> Are there grpcio warnings on this branch currently? Or when upgrading protobuf?
> 
> We should be able to go to max protobuf 2.28 per the `google-cloud-rust-raw` package we depend on, which locks it to that version. It looks like we have a protobuf req in the root Cargo.toml but I'm not sure why, it's only used by syncstorage-spanner which specifies its own dep there. Can we kill it from the root?

Removing protobuf from the root, though funny `cargo machete` never saw it? Perhaps a re-declaration in a child crate doesn't show it, since it's not using the workspace default.

I’ve looked at https://lib.rs/crates/google-cloud-rust-raw to check the max protobuf version, which appears to be 3.4.0. Looks like we can bump that and as per your comment earlier in that PR, but not higher than what is a dep in google-cloud-rust-raw, right?

Ah, nevermind about deadpool now... I remember. We can't upgrade yet since there are some compat issues between it and `diesel-async`, gotta wait a little longer before bumping.

syncstorage-spanner/Cargo.toml
```diff
@@ -22,7 +22,7 @@ form_urlencoded = "1.2"
 # syncserver to either fail to either compile, or start. In those cases, try
 # `cargo build --features grpcio/openssl ...`
 grpcio = { version = "0.13.0", features = ["openssl"] }
-protobuf = { version = "2.28.0" } # must match what's used by googleapis-raw
+protobuf = { version = "2.28.0" } # must match what's used by google-cloud-rust-raw
```
chenba: `"=2.28.0"`?

taddes: I don't think we were getting issues with this, but makes sense that we pin the exact version. Thx

Cargo.toml
```diff
@@ -61,19 +61,18 @@ log = { version = "0.4", features = [
     "max_level_debug",
     "release_max_level_info",
 ] }
-protobuf = "=2.25.2" # pin to 2.25.2 to prevent side updating
-rand = "0.8"
+rand = "0.10"
 regex = "1.12"
-reqwest = { version = "0.12", default-features = false, features = [
-  "rustls-tls",
+reqwest = { version = "0.13.2", default-features = false, features = [
```
pjenvey, line 66: FYI I usually use `cargo upgrade` from the `cargo edit` suite to update deps in Cargo.toml (it would have upgraded to "0.13" here, leaving off the patch version because the original value lacked a patch version)

PR #1799, opened 2025-09-10
refactor: move tokenserver's Db to async-trait

pjenvey:
Closes STOR-142

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ SYNC_SYNCSTORAGE__DATABASE_URL ?= mysql://sample_user:sample_password@localhost/
 SYNC_TOKENSERVER__DATABASE_URL ?= mysql://sample_user:sample_password@localhost/tokenserver_rs
 
 SRC_ROOT = $(shell pwd)
-PYTHON_SITE_PACKGES = $(shell $(SRC_ROOT)/venv/bin/python -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())")
+PYTHON_SITE_PACKGES = $(shell poetry run python -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())")
 
 clippy_mysql:
 	# Matches what's run in circleci
@@ -56,7 +56,6 @@ clippy_spanner:
 
 clean:
 	cargo clean
-	rm -r venv
 
 docker_start_mysql:
 	docker compose -f docker-compose.mysql.yaml up -d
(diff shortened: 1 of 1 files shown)


tokenserver-db/src/models.rs
```diff
@@ -685,156 +684,220 @@ impl TokenserverDb {
...
+        params: params::ReplaceUser,
+    ) -> Result<results::ReplaceUser, DbError> {
+        TokenserverDb::replace_user(self, params).await
+    }
+
+    async fn replace_users(
+        &mut self,
+        params: params::ReplaceUsers,
+    ) -> Result<results::ReplaceUsers, DbError> {
+        TokenserverDb::replace_users(self, params).await
```
pjenvey, line 700: I removed the macro usage because it didn't work with async-trait's attribute macro. This can be a stop gap, we can clean this up a little bit more next, having the actual trait impls have all the implementation themselves (instead of `TokenserverDb`)

tokenserver-db/src/mock.rs
```diff
@@ -49,112 +48,125 @@ impl MockDb {
     }
 }
 
+#[async_trait(?Send)]
 impl Db for MockDb {
-    fn replace_user(&mut self, _params: params::ReplaceUser) -> DbFuture<'_, results::ReplaceUser> {
-        Box::pin(future::ok(()))
+    async fn replace_user(
```
pjenvey, line 53: Maybe we can replace this with automock..

pjenvey: Here's [where I used it in autoconnect](https://github.com/mozilla-services/autopush-rs/blob/180fa5f254e13e1013248ecaaadb6a8c3053b357/autoconnect/autoconnect-ws/src/session.rs#L14), where it created a `MockSession` type for us

taddes: Sure, we could roll that in at some point. I'm not super familiar with that, will look into how that works there.

tokenserver-db/src/models.rs
taddes: Sure, wfm. Honestly the way it is here is clean enough anyway so that's totally fine.

tokenserver-db/src/models.rs
```diff
@@ -685,156 +684,220 @@ impl TokenserverDb {
...
-    async_db_method!(post_service, TokenserverDb::post_service, PostService);
+    async fn post_service(
+        &mut self,
+        params: params::PostService,
+    ) -> Result<results::PostService, DbError> {
+        TokenserverDb::post_service(self, params).await
+    }
 }
 
+#[async_trait(?Send)]
```
taddes, line 814: This was that weird negation macro thing you mentioned, right? So the compiler doesn't break if the trait bound is not satisfied?

taddes (approved):
Very nice to not have the `DbFuture` any longer! Just a few comments, nothing blocking.

tokenserver-db/src/models.rs
pjenvey: Yea we briefly discussed this but for a different reason, I switched `DbPool` to begin using it -- which it really  should have always used, but didn't.

An example of the code generated by async_trait is here: https://docs.rs/async-trait/latest/async_trait/#explanation

So it's adding that `Send` requirement by default to futures. This makes a lot of sense, the default tokio executor requires futures to be `Send` as it runs a single executor across multiple threads, potentially sending your futures across those threads.

Whereas we're built against actix-web, which chooses the alternate `current_thread` tokio executor: https://docs.rs/actix-web/latest/actix_web/rt/struct.Runtime.html -- which runs on a single thread, thus not requiring `Send` (however actix-web spawns multiple of these executors on separate threads. so we use all of our cores, just in a different fashion from the default executor).

Here's a little more info about these executors in the tokio docs: https://docs.rs/tokio/latest/tokio/runtime/index.html#runtime-configurations

So we should generally omit the Send requirement to make our lives easier since actix-web doesn't need `Send` on our futures, that `?Send` tells async-trait to disable it. If you look at the original definition of this future it is based on [this type](https://github.com/mozilla-services/syncstorage-rs/blob/1dd7dab55ec3aaf2658019dde4b1ac8e70c668a4/syncserver-db-common/src/lib.rs#L8): "`Local`" is used in various places to refer to the lack of `Send` https://docs.rs/futures/latest/futures/future/type.LocalBoxFuture.html

Long story short using `?Send` here is just maintaining the `LocalBoxFuture` we're already using

PR #1801, opened 2025-09-12
refactor: move shared tokenserver db into tokenserver-db-common

pjenvey:
Closes STOR-354

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ members = [
   "tokenserver-db",
   "tokenserver-settings",
   "syncserver",
-  "tokenserver-db-postgres",
+  "tokenserver-db-postgres", "tokenserver-db-common",
 ]
 default-members = ["syncserver"]
 
@@ -31,7 +31,7 @@ license = "MPL-2.0"
 
 [workspace.dependencies]
 actix-web = { version = "4", default-features = false, features = ["compat", "http2", "macros"] }
-
+async-trait = "0.1.88"
 docopt = "1.1"
 base64 = "0.22"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,8 @@ members = [
   "tokenserver-db",
   "tokenserver-settings",
   "syncserver",
-  "tokenserver-db-postgres", "tokenserver-db-common",
+  "tokenserver-db-postgres",
+  "tokenserver-db-common",
 ]
 default-members = ["syncserver"]
 
--- a/tokenserver-db-common/src/error.rs
+++ b/tokenserver-db-common/src/error.rs
@@ -7,7 +7,7 @@ use syncserver_db_common::error::SqlError;
 use thiserror::Error;
 use tokenserver_common::TokenserverError;
 
-pub(crate) type DbResult<T> = Result<T, DbError>;
+pub type DbResult<T> = Result<T, DbError>;
 
 /// An error type that represents any database-related errors that may occur while processing a
 /// tokenserver request.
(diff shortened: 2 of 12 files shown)


tokenserver-db-common/src/error.rs
```diff
@@ -68,9 +68,9 @@ enum DbErrorKind {
 impl From<DbErrorKind> for DbError {
     fn from(kind: DbErrorKind) -> Self {
         match kind {
-            DbErrorKind::Sql(ref mysql_error) => Self {
-                status: mysql_error.status,
-                backtrace: Box::new(mysql_error.backtrace.clone()),
+            DbErrorKind::Sql(ref sqle) => Self {
```
taddes, line 71: Good semantic distinction, then it can be agnostic for Postgres SQL

Cargo.toml
```diff
@@ -14,7 +14,7 @@ members = [
   "tokenserver-db",
   "tokenserver-settings",
   "syncserver",
-  "tokenserver-db-postgres",
+  "tokenserver-db-postgres", "tokenserver-db-common",
```
taddes: newline? I notice when cargo building it often puts it on the same line as the last...

tokenserver-db/Cargo.toml
```diff
@@ -6,14 +6,11 @@ authors.workspace = true
 edition.workspace = true
 
 [dependencies]
-backtrace.workspace = true
+async-trait.workspace = true
 futures.workspace = true
```
taddes:
```suggestion
```
Can remove this one too

tokenserver-db/src/models.rs
```diff
@@ -760,49 +759,49 @@ impl Db for TokenserverDb {
...
+    #[cfg(debug_assertions)]
     async fn set_user_created_at(
         &mut self,
         params: params::SetUserCreatedAt,
     ) -> Result<results::SetUserCreatedAt, DbError> {
         TokenserverDb::set_user_created_at(self, params).await
     }
 
-    #[cfg(test)]
+    #[cfg(debug_assertions)]
```
taddes, line 770: Should we be annotating more tests with `debug_assertions`? Given it can be used for testing and debug builds. Just curious why the change and a broader comment on expanding this, given it seems to have some utility.

taddes (changes requested):
Just a couple little things. Then I can build from the models and methods for Postgres.

Cargo.toml
pjenvey: yup, `cargo init` did this 👍

tokenserver-db/src/models.rs
pjenvey: These are switched to `debug_assertions` because `cfg(test)` will only be enabled when testing the crate itself. Thus it doesn't work across multiple crates, e.g. `cargo test` in `tokenserver-db` wouldn't enable `cfg(test)` in `tokenserver-db-common`.

Alternatively we could add an explicit cargo feature exposing such test code but that's tedious. IIRC it even required running cargo test with a specific `--features` command line arg enabling said feature. We had a `db_test` feature a long time ago and even the annotations became more complicated to support not needing to specify the feature in the owning crate: `#[cfg(any(test, feature = "db_test"))]`.

So `debug_assertions` isn't ideal but much easier. It's used for this numerous times in syncstorage-rs and a few in autopush-rs.

taddes (approved):
👍

tokenserver-db/src/models.rs
taddes: ugh yeah, way easier.

PR #1805, opened 2025-09-16
chore: use environment value from syncserver settings for Sentry

chenba:
## Description

Use `environment` value from syncserver settings for Sentry's environment config value.

## Testing

sync server should start without error and sentry should continue to function.

## Issue(s)

Closes STOR-90

### DIFF
--- a/syncserver/src/main.rs
+++ b/syncserver/src/main.rs
@@ -45,10 +45,12 @@ async fn main() -> Result<(), Box<dyn Error>> {
         // Note: set "debug: true," to diagnose sentry issues
         transport: Some(Arc::new(curl_transport_factory)),
         release: sentry::release_name!(),
+        environment: Some(settings.environment.clone().into()),
         ..sentry::ClientOptions::default()
     });
     opts.integrations.retain(|i| i.name() != "debug-images");
     opts.default_integrations = false;
+
     let _sentry = sentry::init(opts);
 
     // Setup and run the server


chenba:
@pjenvey or @taddes this PR doesn't have any test changes since it doesn't looks like there are tests on the Sentry initialization.  Is that correct?

pjenvey:
> @pjenvey or @taddes this PR doesn't have any test changes since it doesn't looks like there are tests on the Sentry initialization. Is that correct?

Yup -- no test needed here (and it's a difficult thing to test anyway)

syncserver/src/main.rs
```diff
@@ -45,10 +45,12 @@ async fn main() -> Result<(), Box<dyn Error>> {
         // Note: set "debug: true," to diagnose sentry issues
         transport: Some(Arc::new(curl_transport_factory)),
         release: sentry::release_name!(),
+        environment: Some(Cow::Owned(settings.environment.clone())),
```
pjenvey: I'll just note there's numerous ways to construct a `Cow`, including via various [From (and thus Into)](https://doc.rust-lang.org/rust-by-example/conversion/from_into.html) implementations for `Cow`, so another common way of doing this would be:
```suggestion
        environment: Some(settings.environment.clone().into()),
```

But the explicit way here is totally fine, I've often written it this way.

pjenvey (approved):
Thanks Barry, and then at some point in the future we can remove the old env var from [here](https://github.com/mozilla/webservices-infra/blob/dd28ef480b0e5ff3788ea4ce17a99339bc2b4c27/sync/k8s/sync/templates/configmap.yaml#L15) (and noted that [here](https://mozilla-hub.atlassian.net/browse/STOR-314?focusedCommentId=1139744)). We'll wait a bit before doing that just in case we need to rollback to an older version.

syncserver/src/main.rs
chenba: Thanks for the protip.  Is there a preference for the repo/project?

pjenvey: Nope, I think it's fine either way, probably more depending on the context

PR #1808, opened 2025-09-17
feat: adapt the test suite to the generic dyn DbPool

pjenvey:
## Description

and moved the couple generic Db methods into the trait itself

## Issue(s)

Closes STOR-360


tokenserver-db/src/models.rs
```diff
@@ -337,175 +333,13 @@ impl TokenserverDb {
...
-            })
-            .await?
-            .id;
-
-        Ok(results::AllocateUser {
-            uid,
-            node: node.node,
-            created_at,
-        })
+        tokenserver_db_common::allocate_user(self, params).await
```
pjenvey: ~~We really want `allocate_users` shared between impls, but we don't have `self.metrics` in the trait. So I ended up with this as we can't call into `Db::allocate_user(self, params)` like as if it was a "super" trait we're inheriting from (calling that just calls back into ourselves w/ a stack overflow).~~

Maybe I'll just add a `metrics()` method to the trait to avoid all this..

pjenvey: > Maybe I'll just add a metrics() method to the trait to avoid all this..

done

tokenserver-db-common/src/lib.rs
```diff
@@ -1,19 +1,27 @@
...
 
-use std::time::Duration;
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
 use async_trait::async_trait;
+use syncserver_common::Metrics;
 use syncserver_db_common::{GetPoolState, PoolState};
 
-use crate::error::DbError;
+pub use crate::error::{DbError, DbResult};
```
taddes: I took a look at the other generic type aliases for `DbResult<T>` and those reside in the lib.rs modules for the syncstorage-mysql and syncstorage-spanner crates, so I think we should move this definition of `DbResult` into lib.rs here in tokenserver-db-common, or even tokenserver-db.

taddes (changes requested):
I like the improvement of moving into `pool_from_settings`. Just one comment on moving the generic type alias for `DbResult` into a common lib for tokenserver. Additionally, I think I know the issue you had when enabling postgres: on [line 5](https://github.com/mozilla-services/syncstorage-rs/blob/8d8b79e1b5849aa27d4eb4f57452f635012decc8/tokenserver-db-postgres/migrations/2025-09-02-195720_create_schema/up.sql#L5) for the `services`, we just need to remove the comma after `VARCHAR(128)`, then enabling migrations should work

tokenserver-db-common/src/lib.rs
pjenvey: :thumbsup: also made this error mod private

PR #1809, opened 2025-09-18
feat: create tokenserver postgres db trait

taddes:
## Description
Create an implementation of the `Db` trait for the `TokenserverPgDb`

## Issue(s)
Closes [STOR-355](https://mozilla-hub.atlassian.net/browse/STOR-355).


tokenserver-db-common/src/lib.rs
```diff
@@ -87,34 +101,41 @@ pub trait Db {
...
     async fn post_node(&mut self, params: params::PostNode) -> Result<results::PostNode, DbError>;
 
+    /// Get complete node entry based on passed id.
     #[cfg(debug_assertions)]
     async fn get_node(&mut self, params: params::GetNode) -> Result<results::GetNode, DbError>;
 
+    /// Based on Node ID, unassign node from `users`.
     #[cfg(debug_assertions)]
-    async fn unassign_node(
+    async fn unassign_node_from_users(
```
pjenvey: This is slightly better naming but the original name correlates with all the `unassign_node` references in [the python code/script](https://github.com/mozilla-services/syncstorage-rs/tree/82e84743027f3a2a703c540d055f147f00c2babe/tools/tokenserver) -- so I'd argue it's not worth changing here (as the python should also be changed) -- especially with the added docs here making it clearer (thanks for the docs btw)

pjenvey (changes requested):

tokenserver-db-common/src/lib.rs
taddes: aha ok makes sense, the documentation for the method should be sufficient, as would the raw query. Good call

pjenvey (approved):

PR #1811, opened 2025-09-19
feat: table models for orm schemas

taddes:
## Description

The use of raw queries as we have used for Tokenserver are advantageous from a readability perspective, but favor should be given to the ORM query constructor. This is especially important as a performance consideration for Postgres, where raw queries needing sql_query will deserialize its data by name, not by index. That means that you cannot deserialize into a tuple, and structs which you deserialize from this function will need to have #[derive(QueryableByName)].

This PR introduces the model structs to be used in cases where we want to move away from raw queries in favor of diesel ORM methods. However, given the readability advantage, it is easy to use a Rust doc comment, which when indented one point, allows one to paste the raw query as part of the method. We should do this because it makes the ORM queries instantaneously understandable. 

See [Diesel Getting Started](https://diesel.rs/guides/getting-started.html)

Note: will rebase upon upcoming changes

## Issue(s)

Closes [STOR-363](https://mozilla-hub.atlassian.net/browse/STOR-363).

[STOR-363]: https://mozilla-hub.atlassian.net/browse/STOR-363?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tokenserver-db-postgres/src/orm_models.rs
+++ b/tokenserver-db-postgres/src/orm_models.rs
@@ -1,0 +1,35 @@
+use crate::schema::{nodes, services, users};
+use diesel::{Identifiable, Insertable, Queryable};
+
+#[derive(Queryable, Debug, Identifiable, Insertable)]
+pub struct Service {
+    pub id: i32,
+    pub service: Option<String>,
+    pub pattern: Option<String>,
+}
+
+#[derive(Queryable, Debug, Identifiable, Insertable)]
+#[diesel(primary_key(uid))]
+pub struct User {
+    pub uid: i64,
+    pub service: i32,
+    pub email: String,
+    pub generation: i64,
+    pub client_state: String,
+    pub created_at: i64,
+    pub replaced_at: Option<i64>,
+    pub nodeid: i64,
+    pub keys_changed_at: Option<i64>,
+}
+
+#[derive(Queryable, Debug, Identifiable, Insertable)]
+pub struct Node {
+    pub id: i64,
+    pub service: i32,
+    pub node: String,
+    pub available: i32,
+    pub current_load: i32,
+    pub capacity: i32,
+    pub downed: i32,
+    pub backoff: i32,
+}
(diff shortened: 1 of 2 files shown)


tokenserver-db-postgres/src/orm_models.rs
```diff
@@ -0,0 +1,35 @@
+use crate::schema::{nodes, services, users};
+use diesel::{Identifiable, Insertable, Queryable};
+
+#[derive(Queryable, Debug, Identifiable, Insertable)]
+pub struct Service {
```
pjenvey, line 5: How do these differ from the types defined in `tokenserver-db-common::results`? It seems like those could be utilized instead -- unless we begin straying the postgres schema from the mysql one (which we haven't, yet anyway, I do wonder if we should revisit that, like e.g. using postgres `TIMESTAMP` types, but we can change our minds a little later while its still a WIP).

taddes: Diesel uses these models in relation to the `table!` march that's defined in `schema.rs` to construct table aliases and make ORM operations possible. You have to match them up as the compiler gets upset if the plural schema def does not match the singular model name. Then, within a function to make the actual call, you pull in Ex. `use self::schema::users::dsl::*;` and that allows for various ORM calls within said method. This can always move into another crate, doesn't have to be its own. It can be a part of results, or rather even be renamed models.rs when we change the current models.rs (which we agreed isn't the greatest name currently)

pjenvey: Let's keep an eye on these as we progress. The mysql version gets away with [this](https://github.com/mozilla-services/syncstorage-rs/blob/f2c7f9444085c9772da35797de9b6614635c7d88/tokenserver-db-common/src/results.rs#L133) for its service calls, so I'm not sure we would need anything different. `Queryable` could probably be added to the db-common struct if needed, there's at least one struct in syncstorage-db-common that includes both `Queryable/ByName`

Issue #1807, opened 2025-09-17
[mysql] Don't report db failures from post_bsos


data-sync-user:
Copied from the old [https://github.com/mozilla-services/syncstorage-rs/issues/679](https://github.com/mozilla-services/syncstorage-rs/issues/679|smart-link) 

This involves a couple parts:

1. `MysqlDb` should match both the Spanner backend and the Python behavior here: any errors from db calls should not be reported as part of the "failed" bucket, they should propagate up instead as normal.
1. After 1., the `Db` layer won't utilize `failed` buckets at all. So we should cleanup both `params::PostBsos` and `results::PostBsos` to no longer include `failed`. In fact `result::PostBsos` doesn't need a `success` bucket either: kill the whole struct and just return a `SyncTimestamp` from `post_bsos`

This contributes partly to the effort needed for [#418](https://github.com/mozilla-services/syncstorage-rs/issues/418).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/STOR-362)

data-sync-user:
➤ Barry Chen commented:

Philip Jenvey doing part of 2 of this Jira would break POST https://<endpoint-url>/storage/<collection> as documented in https://github.com/mozilla-services/docs/blob/master/source/storage/apis-1.5.rst ( https://github.com/mozilla-services/docs/blob/master/source/storage/apis-1.5.rst|smart-link ) .  Would that break some clients?

data-sync-user:
➤ Philip Jenvey commented:

Barry Chen only if we change the handlers response -- let’s keep it the same -- this change will just make the handler itself responsible for creating the success array (I think it will need to copy the bso ids before passing it to db.post_collection) and the empty failed array.

data-sync-user:
➤ Barry Chen commented:

In that case we cannot remove the struct altogether correct?  Since the handler is still using it?

data-sync-user:
➤ Philip Jenvey commented:

We could keep the struct for the response, or alternatively use the json! macro: https://docs.rs/serde_json/latest/serde_json/macro.json.html ( https://docs.rs/serde_json/latest/serde_json/macro.json.html|smart-link ) (the struct might be a little more efficient and the macro always adds a slight more needed compilation time).

data-sync-user:
➤ Philip Jenvey commented:

We already use json! in the batch handler (it returns a similar response) so it would be fine.

data-sync-user:
[...]

PR #1812, opened 2025-09-19
refactor(db): return only a SyncTimestamp from post_bsos db fns

chenba:
Update the `post_bsos` db functions to return a SyncTimestamp only.  No more "success" or "failed" fields in the db results.  The http responses are unchanged.

Closes https://github.com/mozilla-services/syncstorage-rs/issues/1807

### DIFF
--- a/syncserver/src/web/handlers.rs
+++ b/syncserver/src/web/handlers.rs
@@ -291,19 +291,29 @@ pub async fn post_collection(
                 }
             }
 
-            let result = db
+            let (success_ids, bsos): (Vec<_>, Vec<_>) = coll
+                .bsos
+                .valid
+                .into_iter()
+                .map(|x| (x.id.clone(), x.into()))
+                .unzip();
+
+            let modified = db
                 .post_bsos(params::PostBsos {
                     user_id: coll.user_id,
                     collection: coll.collection,
-                    bsos: coll.bsos.valid.into_iter().map(From::from).collect(),
+                    bsos,
                     for_batch: false,
-                    failed: coll.bsos.invalid,
                 })
                 .await?;
 
             Ok(HttpResponse::build(StatusCode::OK)
-                .insert_header((X_LAST_MODIFIED, result.modified.as_header()))
-                .json(result))
+                .insert_header((X_LAST_MODIFIED, modified.as_header()))
+                .json(json!({
+                    "modified": modified,
+                    "success": success_ids,
+                    "failed": coll.bsos.invalid,
+                })))
         })
         .await
 }
@@ -464,7 +474,6 @@ pub async fn post_collection_batch(
                     })
                     .collect(),
                 for_batch: false,
-                failed: Default::default(),
             })
             .await
             .map(|_| ());
--- a/syncserver/src/server/test.rs
+++ b/syncserver/src/server/test.rs
@@ -18,13 +18,13 @@ use http::StatusCode;
 use lazy_static::lazy_static;
 use rand::{thread_rng, Rng};
 use serde::de::DeserializeOwned;
-use serde_json::json;
+use serde_json::{json, Value};
 use sha2::Sha256;
 use syncserver_common::{self, X_LAST_MODIFIED};
 use syncserver_settings::{Secrets, Settings};
 use syncstorage_db::{
     params,
-    results::{DeleteBso, GetBso, PostBsos, PutBso},
+    results::{DeleteBso, GetBso, PutBso},
     DbPoolImpl, SyncTimestamp,
 };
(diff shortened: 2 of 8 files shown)


pjenvey:
👍

chenba:
Rebased on master and tests are passing; calling this ready for reviews.

syncserver/src/web/handlers.rs
```diff
@@ -291,19 +292,25 @@ pub async fn post_collection(
...
+                coll.bsos.valid.into_iter().map(From::from).collect();
+            let modified = db
                 .post_bsos(params::PostBsos {
                     user_id: coll.user_id,
                     collection: coll.collection,
-                    bsos: coll.bsos.valid.into_iter().map(From::from).collect(),
+                    bsos: bsos.clone(),
                     for_batch: false,
-                    failed: coll.bsos.invalid,
+                    failed: coll.bsos.invalid.clone(),
```
pjenvey: The dbs are no longer utilizing `failed` with this change, so we can remove it from `params::PostBsos` altogether (and of course remove the need to clone it).

syncserver/src/web/handlers.rs
```diff
@@ -291,19 +292,25 @@ pub async fn post_collection(
...
 
-            let result = db
+            let bsos: Vec<PostCollectionBso> =
+                coll.bsos.valid.into_iter().map(From::from).collect();
+            let modified = db
                 .post_bsos(params::PostBsos {
                     user_id: coll.user_id,
                     collection: coll.collection,
-                    bsos: coll.bsos.valid.into_iter().map(From::from).collect(),
+                    bsos: bsos.clone(),
```
pjenvey: I see the need for cloning as `bsos` is needed for `success`  later but we should especially avoid it as clients can `POST` a large amount of data per each request (their payload data are the `bsos`). FYI clients adjust various limits per the server's `/info/configuration` [endpoint](https://mozilla-services.readthedocs.io/en/latest/storage/apis-1.5.html#general-info): [prod's](https://sync-1-us-west1-g.sync.services.mozilla.com/1.5/123456789/info/configuration) `max_post_bytes` allows up to 2.5mb.

Luckily it's easily avoided here as we can instead move the `success` `collect` on line 311 up front, just before the call to `db.post_bsos`, then remove the `clone()`.

Though we can do even better than that: since we're already looping through `coll.bsos` to create `bsos`, we can piggy back onto that to collect the `bso.id`s at the same time. A few ways to do that:

- simplest to understand: have a for loop create both vecs instead of 2 `collect()`s
- populate a vec in the existing `into_iter()`:
```rust
            let mut bso_ids = vec![];
            let bsos: Vec<PostCollectionBso> = coll
                .bsos
                .valid
                .into_iter()
                .map(|bso| {
                    bso_ids.push(bso.id.clone());
                    bso.into()
                })
                .collect();
```
(note `bso.into()` is the same as `From::from(bso)`, just more idiomatic)
- more advanced! `collect()` can convert iterators into other common collections such as `HashMaps` and even do fancy things like [help w/ handling collections of Results](https://doc.rust-lang.org/rust-by-example/error/iter_result.html#fail-the-entire-operation-with-collect). There's even a way to have it produce multiple `Vec`s given an `Iterator::Item` of a tuple (like is used for collecting into a HashMap). You can see how this works [here](https://doc.rust-lang.org/std/iter/trait.FromIterator.html#impl-FromIterator%3C(EA,)%3E-for-(A,)) but there is also a simpler version of this handling only 2 items: [unzip](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.unzip)
 
I like avoiding for loops in rust when possible -- though not terribly important here, there are optimizations of such transforms where they'll perform the same (and sometimes even better) than for loops

syncserver/src/web/handlers.rs
```diff
@@ -291,19 +292,25 @@ pub async fn post_collection(
                 }
             }
 
-            let result = db
+            let bsos: Vec<PostCollectionBso> =
```
pjenvey: FYI/nit: You often need to inform `collect` of what collection type it's producing, but in this case with the assignment of it to `PostBsos::bsos` later in the fn, the compiler knows it's a `Vec<PostCollectionBso>`, so this could be omitted.
```suggestion
            let bsos =
```

I'm not against leaving it, some find it more readable but IMHO it's not needed.

syncserver/src/web/handlers.rs
chenba: > I see the need for cloning as bsos is needed for success later but we should especially avoid it as clients can POST a large amount of data per each request

Great point.  Thanks.

PR #1818, opened 2025-09-22
fix: switch check to SELECT 1 to fix it on diesel-async

pjenvey:
## Description

which is sufficient vs reading uptime as we weren't reading it correctly. execute returns number of rows affected and diesel returns 1 here whereas diesel-async returns 0

Note this is against the last release's branch to avoid changes from master from going out for now (0.21.0 includes the tokenserver move to diesel-async but not much else after that)

## Issue(s)

Closes STOR-365


taddes (approved):
lgtm 👍

PR #1828, opened 2025-09-30
feat: postgres node methods

taddes:
## Description

Implementation of all node-related methods in Postgres. This includes all `GET`, `POST`, `PUT`, and `DELETE` methods defined in our schema required for Tokenserver operations in Postgres.

This requires a number of operations:
- `get_node` : get the full node with metadata
- `get_node_id` : get a specific node_id by passing in service string and node string.
- `post_node` : create a new node and return the most recently created record
- `get_best_node` : get the node with least load. Requires a base query as well as queries to check release capacity. This query may need to run again a few times.
- `add_user_to_node` : based on current load, add a user by the passed service_id and node string identifier.
- `remove_node` : delete a node.

## Issue(s)

Closes [STOR-345](https://mozilla-hub.atlassian.net/browse/STOR-345).

[STOR-345]: https://mozilla-hub.atlassian.net/browse/STOR-345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tokenserver-db-postgres/src/models.rs
+++ b/tokenserver-db-postgres/src/models.rs
@@ -299,7 +299,7 @@ impl TokenserverPgDb {
 
     /**
     Update the current load count of a node, passing in the service string and node string.
-    This represents the addition of a user to a note, while not defining which user specifically.
+    This represents the addition of a user to a node, while not defining which user specifically.
     Does not return anything.
 
         UPDATE nodes


tokenserver-db-postgres/src/models.rs
```diff
@@ -104,6 +106,263 @@ impl TokenserverPgDb {
...
+                        "Tokenserver get_best_node query - Unable to get Spanner node: {}",
+                        e
+                    ));
+                    db_error.status = StatusCode::SERVICE_UNAVAILABLE;
+                    db_error
+                })
+        } else {
+            // This loop allows for a maximum of 5 retries before stopping.
+            // This allows for query retries if more capacity needs to be released.
+            for _ in 0..5 {
```
chenba, line 233: Why 5?

tokenserver-db-postgres/src/models.rs
```diff
@@ -104,6 +106,263 @@ impl TokenserverPgDb {
...
+
+        diesel::sql_query(Self::LAST_INSERT_ID_QUERY)
+            .get_result::<results::PostNode>(&mut self.conn)
+            .await
+            .map_err(Into::into)
+    }
+
+    /**
+    Update the current load count of a node, passing in the service string and node string.
+    This represents the addition of a user to a note, while not defining which user specifically.
```
chenba:
```suggestion
    This represents the addition of a user to a node, while not defining which user specifically.
```

chenba:
Not all that knowledgeable about the token server db schema but the code reads fine to me.  👍

tokenserver-db-postgres/src/models.rs
taddes: It predates me, but I think it was just a sensible threshold given possible network issues or capacity changes on the node. More than likely it's better served as a config value, so maybe we can tweak that in the future, but it's just a gut check.

tokenserver-db-postgres/src/models.rs
```diff
@@ -104,6 +106,263 @@ impl TokenserverPgDb {
...
+            QUERY
+        };
+
+        diesel::sql_query(query)
+            .bind::<Integer, _>(params.service_id)
+            .bind::<Text, _>(params.node)
+            .execute(&mut self.conn)
+            .await
+            .map(|_| ())
+            .map_err(Into::into)
```
pjenvey, line 346: nit: I know this is from the mysql version but I'll usually prefer `Ok(call?)` instead of `map_err(Into::into)` such as in `post_node`, but of course that's no big deal (and even offers potentially less indenting of a multiline call)

However the former is just plain simpler than a combined `map+map_err` here/elsewhere:
```suggestion
        diesel::sql_query(query)
            .bind::<Integer, _>(params.service_id)
            .bind::<Text, _>(params.node)
            .execute(&mut self.conn)
            .await?;
        Ok(())
```

taddes: Great point, I'll make that consistent in the next iteration 👍

PR #1830, opened 2025-09-30
refactor: move syncstorage's Db to async-trait

pjenvey:
## Issue(s)

Closes STOR-144


syncstorage-mysql/src/models.rs
```diff
@@ -951,109 +949,205 @@ impl MysqlDb {
...
-        MysqlDb::get_batch,
-        GetBatch,
-        Option<results::GetBatch>
-    );
-    async_db_method!(commit_batch, MysqlDb::commit_batch, CommitBatch);
-
-    fn get_collection_id(&mut self, name: String) -> DbFuture<'_, i32, Self::Error> {
-        Box::pin(async move { self.get_collection_id(&name).map_err(Into::into).await })
+    async fn commit(&mut self) -> Result<(), Self::Error> {
+        MysqlDb::commit(self).await
```
pjenvey, line 957: As w/ the other backends (tokenserver's and spanner) at some point we can get rid of this extra level of indirection by moving the actual implementation of these `Db` impl methods from `MysqlDb` into this `Db` method itself, maybe in a separate `db_impl.rs` module

syncstorage-db/src/mock.rs
```diff
@@ -50,83 +47,232 @@ impl MockDb {
...
-    mock_db_method!(delete_bso, DeleteBso);
-    mock_db_method!(get_bso, GetBso, Option<results::GetBso>);
-    mock_db_method!(get_bso_timestamp, GetBsoTimestamp);
-    mock_db_method!(put_bso, PutBso);
-    mock_db_method!(create_batch, CreateBatch);
-    mock_db_method!(validate_batch, ValidateBatch);
-    mock_db_method!(append_to_batch, AppendToBatch);
-    mock_db_method!(get_batch, GetBatch, Option<results::GetBatch>);
-    mock_db_method!(commit_batch, CommitBatch);
+    async fn delete_storage(
```
pjenvey, line 133: Note one downside of the `async-trait` attribute macro is [it breaks](https://github.com/dtolnay/async-trait/issues/112) the `mock_db_method!` macro that was concisely defining methods for us.

chenba (dismissed):
👍

syncstorage-mysql/src/models.rs
taddes: Yeah, totally, that said though part of me is happy and fine with this indirection, at least for now. Makes the signatures clear, though I know it becomes redundant.

syncserver-db-common/src/lib.rs
```diff
@@ -28,15 +24,3 @@ impl From<deadpool::Status> for PoolState {
         }
     }
 }
-
-#[macro_export]
```
taddes, line 32: Delighted that this can go.

syncstorage-db-common/src/lib.rs
```diff
@@ -68,190 +68,189 @@ impl<E> Clone for Box<dyn DbPool<Error = E>> {
...
+    ) -> Result<results::CommitBatch, Self::Error>;
 
-    fn check(&mut self) -> DbFuture<'_, results::Check, Self::Error>;
+    async fn check(&mut self) -> Result<results::Check, Self::Error>;
 
     fn get_connection_info(&self) -> results::ConnectionInfo;
 
     /// Retrieve the timestamp for an item/collection
     ///
     /// Modeled on the Python `get_resource_timestamp` function.
```
taddes: We still need this here? If so, maybe a direct ref to the archived repo in question, otherwise maybe we don't need this comment.

syncstorage-db-common/Cargo.toml
```diff
@@ -6,6 +6,7 @@ authors.workspace = true
 edition.workspace = true
 
 [dependencies]
+async-trait.workspace = true
 backtrace.workspace = true
 chrono.workspace = true
 diesel.workspace = true
```
taddes, line 12: Can also get rid of `futures` here

syncstorage-db/Cargo.toml
```diff
@@ -6,13 +6,13 @@ authors.workspace = true
 edition.workspace = true
 
 [dependencies]
+async-trait.workspace = true
 env_logger.workspace = true
 futures.workspace = true
```
taddes:
```suggestion

```

syncstorage-mysql/Cargo.toml
```diff
@@ -6,6 +6,7 @@ authors.workspace = true
 edition.workspace = true
 
 [dependencies]
+async-trait.workspace = true
 backtrace.workspace = true
 base64.workspace = true
 deadpool.workspace = true
```
taddes, line 12: Can also get rid of `futures` here

taddes (approved):
Won't block, but a few minor comments. Also noticed we can remove `futures` from:
- syncserver-db-common
-syncstorage-db
-syncstorage-mysql

taddes (approved):

Issue #1820, opened 2025-09-22
Split up syncstorage (syncserver's) extractors into separate modules


data-sync-user:
All of syncstorage extractors reside in one very large file. Each extractor and associated tests should be separated into their own individual sub module within an extractors module (as [autoendpoint does](https://github.com/mozilla-services/autopush-rs/tree/180fa5f254e13e1013248ecaaadb6a8c3053b357/autoendpoint/src/extractors)).

This could probably be done piecemeal with an initial `git mv extractors.rs extractors/mod.rs` then moving each extractor out individually.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/STOR-367)

PR #1831, opened 2025-10-01
chore(syncserver): break up extractors.rs

chenba:
## Description

sync server's extractors.rs was quite large.  This patch simply break it up.  There are no type or functionality changes.

## Testing

Existing tests should pass.

## Issue(s)

Closes STOR-367 / #1820


syncserver/src/web/extractors/mod.rs
```diff
@@ -0,0 +1,53 @@
+//! Request header/body/query extractors
+//!
+//! Handles ensuring the header's, body, and query parameters are correct, extraction to
+//! relevant types, and failing correctly with the appropriate errors if issues arise.
+
+mod constants;
+pub use constants::*;
+mod utils;
+pub use utils::*;
```
chenba: Couldn't come up with good name for this.  :-/

syncserver/src/web/extractors/tests.rs
```diff
@@ -0,0 +1,701 @@
...
+
+    // Not sure why but sending req through *::extract loses the body.
+    // Compose a payload here and call the *::from_request
+    let (_sender, mut payload) = h1::Payload::create(true);
+    payload.unread_data(Bytes::from(bod_str.to_owned()));
+    CollectionPostRequest::from_request(&req, &mut payload.into()).await
+}
+
+#[test]
+fn test_invalid_query_args() {
```
pjenvey: We could also move these tests into their specific modules (e.g. bso_query_params) but totally fine for now.

syncserver/src/web/extractors/batch_bso_body.rs
```diff
@@ -0,0 +1,56 @@
+use crate::web::extractors::{
```
pjenvey: Couple nitpicks here that I won't block on:

- we're not consistent but we try to organize imports by a grouping of std/external/crate separated by a newline. rustfmt doesn't enforce this but I'll note it has an option for this grouping in nightly, "StdExternalCrate". So this `crate` block would go below the last import, separated by a newline like in batch_request.rs
- You could also replace `crate::web::extractors::` with `super::` (or more specifically `super::validation::` here) -- not just because it's shorter but I think it could help avoid wildcard imports in mod.rs. Ideally mod.rs would only `pub use` what's exposed outside of `extractors` vs internal -- e.g. everything in `validation.rs` is internal and doesn't need `pub` in `mod.rs`. But I'm not complaining, still beats 2k lines of extractors.rs

syncserver/src/web/extractors/offset.rs
```diff
@@ -0,0 +1,50 @@
+use serde::Deserialize;
```
pjenvey: Forgot my only actual request: could you move the contents of this module into `bso_query_params.rs`? As it's specific to there and doesn't have its own `FromRequest`

syncserver/src/web/extractors/tests.rs
chenba: > move these tests into their specific modules

Oh yeah, my bad.  I should've done that in the first place; that's my preference as well.

chenba:
@pjenvey Thank you for the feedback.  I think I've made all the suggested changes.  They are in a separate commit: https://github.com/mozilla-services/syncstorage-rs/pull/1831/commits/7678e88c691a9f0581a3aa9de05a2c80872a2fb6.  (I'll squash and rebase master once approved.)

pjenvey (approved):

PR #1839, opened 2025-10-14
feat: postgres user methods

taddes:
## Description

Creation of all Postgres user methods for Tokenserver

## Issue(s)

Closes [STOR-346](https://mozilla-hub.atlassian.net/browse/STOR-346).

[STOR-346]: https://mozilla-hub.atlassian.net/browse/STOR-346?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tokenserver-db-postgres/src/models.rs
+++ b/tokenserver-db-postgres/src/models.rs
@@ -53,13 +53,7 @@ impl TokenserverPgDb {
 
     // Services Table Methods
 
-    /**
-    Acquire service_id through passed in service string.
-
-        SELECT id
-          FROM services
-         WHERE service = <service String>
-     */
+    /// Acquire service_id through passed in service string.
     pub async fn get_service_id(
         &mut self,
         params: params::GetServiceId,
@@ -81,13 +75,8 @@ impl TokenserverPgDb {
         }
     }
 
-    /**
-    Create a new service, given a provided service string and pattern.
-    Returns a service_id.
-
-        INSERT INTO services (service, pattern)
-        VALUES (<service String >, <pattern String>)
-     */
+    // Create a new service, given a provided service string and pattern.
+    // Returns a service_id.
     #[cfg(debug_assertions)]
     pub async fn post_service(
         &mut self,
@@ -114,15 +103,9 @@ impl TokenserverPgDb {
 
     // Nodes Table Methods
 
-    /**
-    Get Node with complete metadata, given a provided Node ID.
-    Returns a complete Node, including id, service_id, node string identifier
-    availability, and current load.
-
-        SELECT *
-          FROM nodes
-         WHERE id = <id i64>
-     */
+    /// Get Node with complete metadata, given a provided Node ID.
+    /// Returns a complete Node, including id, service_id, node string identifier
+    /// availability, and current load.
     #[cfg(debug_assertions)]
     async fn get_node(&mut self, params: params::GetNode) -> DbResult<results::GetNode> {
         const QUERY: &str = r#"
@@ -138,15 +121,8 @@ impl TokenserverPgDb {
             .map_err(Into::into)
     }
 
-    /**
-    Get the specific Node ID, given a provided service string and node.
-    Returns a node_id.
-
-        SELECT id
-          FROM nodes
-         WHERE service = <service String>
-           AND node = <node String>
-     */
+    /// Get the specific Node ID, given a provided service string and node.
+    /// Returns a node_id.
     async fn get_node_id(&mut self, params: params::GetNodeId) -> DbResult<results::GetNodeId> {
         const QUERY: &str = r#"
             SELECT id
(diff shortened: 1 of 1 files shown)


tokenserver-db-postgres/src/models.rs
```diff
@@ -363,6 +368,313 @@ impl TokenserverPgDb {
...
+            <replaced_at NULL>)
+        RETURNING id;
+
+    */
+    #[cfg(debug_assertions)]
+    async fn post_user(&mut self, params: params::PostUser) -> DbResult<results::PostUser> {
+        const QUERY: &str = r#"
+            INSERT INTO users (service, email, generation, client_state, created_at, nodeid, keys_changed_at, replaced_at)
+            VALUES ($1, $2, $3, $4, $5, $6, $7, NULL)
+            RETURNING id;
```
pjenvey:
```suggestion
            RETURNING uid;
```

tokenserver-db-postgres/src/models.rs
```diff
@@ -363,6 +368,313 @@ impl TokenserverPgDb {
...
+    Update several user records with the given email and service id
+    marking them as 'replaced'. This is through updating the `replaced_at` field.
+    The `replaced_at` field should be null AND the `created_at` field should be earlier
+    than the `replaced_at`.
+
+        UPDATE users
+            SET replaced_at = <replaced_at i64>
+            WHERE service = <service i32>
+            AND uid = <uid i64>
+
```
pjenvey: I don't think we need queries in docstrings, especially since we have raw queries in the code. This one's out of sync with the actual query in code -- I think that's always a risk when duplicating them.

The parameter type information can be useful but we get a form of that from diesel's `bind` calls.

tokenserver-db-postgres/src/models.rs
```diff
@@ -363,6 +368,313 @@ impl TokenserverPgDb {
...
+        metrics.start_timer("storage.get_users", None);
+
+        const QUERY: &str = r#"
+            SELECT uid, nodes.node, generation, keys_changed_at, client_state, created_at, replaced_at
+            FROM users
+            LEFT OUTER JOIN nodes ON users.nodeid = nodes.id
+                WHERE email = $1
+                AND users.service = $2
+            ORDER BY created_at DESC, uid DESC
+                LIMIT 20
```
pjenvey: FYI we've attempted to follow https://www.sqlstyle.guide/ for formatting queries. You can use the tokenserver mysql queries as a guide, basically align the keywords on the left side. I'm attempting to line them up here but I think github is being weird about it (see it [here instead](https://github.com/mozilla-services/syncstorage-rs/blob/0dd2c8179d44a7b8db2036f373f320602715c28f/tokenserver-db/src/models.rs#L315)):
```suggestion
                 SELECT uid, nodes.node, generation, keys_changed_at, client_state, created_at, replaced_at
                    FROM users
LEFT OUTER JOIN nodes ON users.nodeid = nodes.id
                  WHERE email = $1
                       AND users.service = $2
             ORDER BY created_at DESC, uid DESC
                     LIMIT 20
```

tokenserver-db-postgres/src/models.rs
```diff
@@ -363,6 +368,313 @@ impl TokenserverPgDb {
...
+            WHERE uid = $2
+        "#;
+
+        diesel::sql_query(QUERY)
+            .bind::<BigInt, _>(params.created_at)
+            .bind::<BigInt, _>(params.uid)
+            .execute(&mut self.conn)
+            .await
+            .map(|_| ())
+            .map_err(Into::into)
```
pjenvey, line 520: did you want to kill these [here or later](https://github.com/mozilla-services/syncstorage-rs/pull/1828#discussion_r2399602909) (later is fine)

taddes: I can do this in the next PR, I had it down for STOR-376 (last insert ID replacement) 👍

tokenserver-db-postgres/src/models.rs
taddes: Ok cool, can do. Always awkward trying to line queries up in code... I'm gonna see if there's an auto-formatter somewhere we could integrate.

tokenserver-db-postgres/src/models.rs
taddes: Ok, fixed the one in question. I liked how the SQL auto formatted in the IDE autodocs when you hover over the method (see below), so I included it in case we ever changed any to the ORM queries. Still happy to take them out if you think they're just an added burden 👍 
Screenshot 2025-10-15 at 11 03 36 AM

pjenvey: let's kill it anyway (sorry) -- I don't think the `TokenserverPgDb` specific methods would pop up in the IDE very often, if at all, as most usage of these methods is on the `Db` trait versions (and when they do you're just a hop to definition away to see the sql albeit w/ no formatting)

Apparently there's a [code plugin that syntax highlights sql](https://marketplace.visualstudio.com/items?itemName=qufiwefefwoyn.inline-sql-syntax) -- unfortunately it requires a `--sql` prefix to enable it per string, which is just an extra enough of annoyance to discourage its usage IMHO 🤷‍♂️

tokenserver-db-postgres/src/models.rs
pjenvey: Agreed, an auto formatter would be nice -- probably going to be extra difficult to apply it to only to our strings, though ☹️

tokenserver-db-postgres/src/models.rs
taddes: You bet 👍 makes sense. Updated @pjenvey

pjenvey (approved):

Issue #1791, opened 2025-09-05
Enable CI testing, unit & integration (postgres)


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/STOR-353)

data-sync-user:
➤ Barry Chen commented:

Taddes Korris can you add a list of high level things to be done in this issue just so I won’t miss anything?  Thanks!

data-sync-user:
➤ Taddes Korris commented:

Can do Barry Chen,  let me peek at the mysql and spanner impl and make a list, since we want to make the shared generic tests run for both sql and postgres in Tokenserver, we might need to emulate a bit of how syncserver does this between spanner and mysqlmeanwhile  made the bare bones description a bit better. 

We can pair a bit on this one too as it might be a little funky. More soon!

data-sync-user:
➤ Barry Chen commented:

At this point I’ve spent more time trying to get all the tests pass than I did setting up the workflow, but the e2e tests are still not passing with Tokenserver running on Postgres.  There seems to be a stack overflow issue somewhere; no such problem with the tests when Tokenserver is running on MySQL, however.

I’ll file an issue to fix the bug/tests.

PR #1840, opened 2025-10-15
 chore(ci): configure CircleCI for postgres builds

chenba:
## Description

This commit enables Postgres builds and tests in CircleCI.  Only
Tokenserver supports Postgres at the moment, and the Python based
integration tests have an implicit dependency on MySQL, so the CI jobs only
use Postgres where possible.

The jobs have been split up and named more explicitly.  The hope is to
simply delete the mysql jobs in the future.

## Testing

The Tokenserver unit tests running on Postgres should pass.

## Issue(s)

Closes STOR-353 / #1791

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,7 +9,8 @@ GitHub.
 ## Bug Reports
 
 You can file issues here on GitHub. Please try to include as much information as
-you can and under what conditions you saw the issue.
+you can and under what conditions you saw the issue. We will do our best to triage
+the issue as soon as possible.
 
 ## Sending Pull Requests
 
@@ -17,9 +18,9 @@ Patches should be submitted as pull requests (PR).
 
 Before submitting a PR:
 - Your code must run and pass all the automated tests before you submit your PR
-  for review. "Work in progress" pull requests are allowed to be submitted, but
-  should be clearly labeled as such and should not be merged until all tests
-  pass and the code has been reviewed.
+  for review. "Work in progress" or "Draft" pull requests are allowed to be submitted,
+  but should be clearly labeled as such and should not be merged until all tests pass and the code
+  has been reviewed.
 - Your patch should include new tests that cover your changes. It is your and
   your reviewer's responsibility to ensure your patch includes adequate tests.
 
@@ -42,11 +43,11 @@ installing, running and testing.
 
 ## Code Review
 
-This project is production Mozilla code and subject to our [engineering practices and quality standards](https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Committing_Rules_and_Responsibilities). Every patch must be peer reviewed.
+This project is production Mozilla code and subject to the contributing guidelines established in this documentation. Every patch must be peer reviewed by a member of the official Sync team. 
 
 ## Git Commit Guidelines
 
-We loosely follow the [Angular commit guidelines](https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#type)
+We loosely follow the [Angular commit guidelines][angular_commit_guidelines]
 of `<type>: <subject>` where `type` must be one of:
 
 * **feat**: A new feature
@@ -59,6 +60,27 @@ of `<type>: <subject>` where `type` must be one of:
 * **test**: Adding missing tests
 * **chore**: Changes to the build process or auxiliary tools and libraries such as documentation
   generation
+
+For Mozilla engineers:
+
+If associated with a Jira ticket, synchronization with Jira and GitHub is possible by appending the suffix of the Jira ticket to the branch name (`STOR-1234` in the example below). Name the branch using the appropriate `<type>` above followed by a forward slash, followed by a dash-separated description of the task and then by the Jira ticket and 
+. Ex. `feat/add-sentry-sdk-STOR-1234` or `add-sentry-sdk-STOR-1234` 
+
+Note: the Jira ticket project and number can be added anywhere in the
+branch name, but adding to the beginning is ideal. You can also include the Jira issue at the end of
+commit messages to keep the task up to date. See Jira Docs for referencing issues [here][jira].
+
+For Community Contributors:
+
+You won't have access to Jira, so just add the GitHub Issue number at the end of the PR branch name:
+Ex. `feat/add-sentry-sdk-1112`
+
+For Everyone:
+Make sure that the title of your pull request in GitHub has the `type` followed by a colon so that it will be automatically added in the changelog when Sync is published for release.
+Ex. `feat: add sentry sdk`
+
+[angular_commit_guidelines]: https://github.com/angular/angular/blob/main/CONTRIBUTING.md
+[jira]: https://support.atlassian.com/jira-software-cloud/docs/reference-issues-in-your-development-work/
 
 ### Subject
 


.circleci/config.yml
```diff
@@ -77,13 +77,24 @@ commands:
       - run:
           name: Install MySQL client
           command: sudo apt-get update && sudo apt-get install -y default-mysql-client
-  create-tokenserver-database:
+  create-tokenserver-database-mysql:
```
chenba, line 80: Naming the jobs more explicitly since we are introducing another db into the mix.

.circleci/config.yml
```diff
@@ -162,6 +181,42 @@ commands:
           name: Merge llvm-cov results
           command: make merge_coverage_results
 
+  persist-coverage-to-workspace:
+    parameters:
+      job_name:
+        type: string
+    steps:
+      - run:
+          name: Compress and copy coverage data
```
chenba, line 190: This doesn't save time when saving data but does save a little bit of time when attaching workspace later.

.circleci/config.yml
```diff
@@ -284,26 +339,101 @@ jobs:
...
+        auth:
+          username: $DOCKER_USER
+          password: $DOCKER_PASS
+        environment:
+          SYNC_SYNCSTORAGE__DATABASE_URL: postgres://test:test@127.0.0.1/syncstorage
+          SYNC_TOKENSERVER__DATABASE_URL: postgres://test:test@127.0.0.1/tokenserver
+          RUST_BACKTRACE: 1
+          # XXX: begin_test_transaction doesn't play nice over threaded tests
+          RUST_TEST_THREADS: 1
+      - image: cimg/postgres:18.0
```
chenba, line 372: I don't remember any discussion on postgres versions.

Dockerfile
```diff
@@ -12,24 +13,39 @@ COPY . .
...
+    apt-get -q install -y --no-install-recommends $MYSQL_PKG $POSTGRES_DEV_PKG cmake
 
 COPY --from=planner /app/recipe.json recipe.json
-RUN cargo chef cook --release --no-default-features --features=syncstorage-db/$DATABASE_BACKEND --features=py_verifier --recipe-path recipe.json
+RUN set -x && \
+    TOKENSERVER_FEATURES="" && \
+    if [ "$TOKENSERVER_DATABASE_BACKEND" = "postgres" ]; then \
+        TOKENSERVER_FEATURES="--features=tokenserver-db/postgres"; \
+    fi && \
+    cargo chef cook --release --no-default-features --features=syncstorage-db/$SYNCSTORAGE_DATABASE_BACKEND $TOKENSERVER_FEATURES --features=py_verifier --recipe-path recipe.json
```
chenba, line 47: Temporarily support the permutation of Spanner Syncserver and Postgres Tokenserver while we await for Syncserver Postgres support.

chenba, line 47: Temporarily support the permutation of Spanner Syncserver and Postgres Tokenserver while we await for Syncserver Postgres support.

Dockerfile
```diff
@@ -91,18 +121,22 @@ COPY --from=builder /app/pyproject.toml /app/poetry.lock /app/
...
+        echo "deb https://repo.mysql.com/apt/debian/ bookworm mysql-8.0" >> /etc/apt/sources.list && \
+        wget -qO- https://repo.mysql.com/RPM-GPG-KEY-mysql-2023 > /etc/apt/trusted.gpg.d/mysql.asc ; \
+    fi && \
+    if [ "$TOKENSERVER_DATABASE_BACKEND" = "postgres" ]; then \
+        POSTGRES_PKG="libpq5"; \
     fi && \
     # update again now that we trust repo.mysql.com
     apt-get -q update && \
-    apt-get -q install -y build-essential $MYSQLCLIENT_PKG libssl-dev libffi-dev libcurl4 python3-dev python3-pip python3-setuptools python3-wheel python3-venv cargo curl jq pkg-config && \
+    apt-get -q install -y build-essential $MYSQLCLIENT_PKG $POSTGRES_PKG libssl-dev libffi-dev libcurl4 python3-dev python3-pip python3-setuptools python3-wheel python3-venv cargo curl jq pkg-config && \
```
chenba: Both MySQL and Postgres packages are installed because the Python based integration tests have an implicit dependency on MySQL.

tokenserver-db-postgres/src/models.rs
```diff
@@ -155,7 +155,7 @@ impl TokenserverPgDb {
                AND capacity > current_load
                AND downed = 0
                AND backoff = 0
-             ORDER BY LOG(current_load) / LOG(capacity)
+             ORDER BY CASE WHEN current_load = 0 THEN -1 ELSE LN(current_load) / LN(capacity) END
```
chenba, line 158: (The changes in this file are not directly CI related but they are needed for the Tokenserver unit tests on Postgres to pass.)

When LOG(0) is called in MySQL it returns a NULL along with a warning, and that behavior along with how MySQL's ORDER BY treats NULL made that query work.  It does not work in Postgres.  Postgres throws an error when the query tries to take the LOG of 0.

tokenserver-db-postgres/src/models.rs
```diff
@@ -386,7 +386,7 @@ impl TokenserverPgDb {
              WHERE service = $3
                AND email = $4
                AND  generation <= $5
-               AND COALESCE(keys_changed_at, 0) <= COALESCE(<keys_changed_at i64>, keys_changed_at, 0)
+               AND COALESCE(keys_changed_at, 0) <= COALESCE($6, keys_changed_at, 0)
```
chenba, line 389: I think it was just a good ol' bug.

tokenserver-db-postgres/src/models.rs
```diff
@@ -618,13 +618,6 @@ impl Db for TokenserverPgDb {
         TokenserverPgDb::get_users(self, params).await
     }
 
-    async fn get_or_create_user(
-        &mut self,
-        params: params::GetOrCreateUser,
-    ) -> Result<results::GetOrCreateUser, DbError> {
-        TokenserverPgDb::get_or_create_user(self, params).await
-    }
```
chenba, line 626: Infinite call loop from trait impl delegation.

chenba:
I'm considering this ready since the workflow is there.  We'll update the jobs and commands as we progress with the Postgres work.

tools/integration_tests/pyproject.toml
```diff
@@ -18,6 +18,7 @@ cryptography = "44.0.2"
 hawkauthlib = "^2.0.0"
 konfig = "^1.1"
 mysqlclient = "^2.2.7"
+psycopg2-binary = "^2.9.11"
```
chenba, line 21: Added this after https://github.com/mozilla-services/syncstorage-rs/pull/1855 was merged into master and this branch was rebased on master.  Might not be necessary if it gets added in https://github.com/mozilla-services/syncstorage-rs/pull/1857 and that PR's merged first.

.circleci/config.yml
taddes: We can do the same thing with pinning a Postgres version like we will with Spanner, but explicitly defining 18 like you have here is perfect

Dockerfile
taddes: Yeah, same issue with Spanner needing SQL, as we unfortunately learned with the public key fiasco...

tokenserver-db-postgres/src/models.rs
taddes: Good call. This was missed when we moved the `get_or_create_user` call into `tokenserver-db-common` since it was identical for MySQL and Postgres.

tools/integration_tests/pyproject.toml
taddes: Either works for me, I'll add them regardless in https://github.com/mozilla-services/syncstorage-rs/pull/1857 along with the poetry lockfile, so both should end up looking the same.

taddes (approved):
👍  lgtm, I'll review again after the rebase for `Dockerfile` we discussed.

tools/integration_tests/tokenserver/test_support.py
```diff
@@ -30,7 +30,11 @@ def setUpClass(cls):
         cls._build_auth_headers = cls._build_oauth_headers
 
     def setUp(self):
-        engine = create_engine(os.environ["SYNC_TOKENSERVER__DATABASE_URL"])
+        db_url = os.environ["SYNC_TOKENSERVER__DATABASE_URL"]
+        # SQLAlchemy 1.4+ wants postgresql
+        if db_url.startswith("postgres://"):
+            db_url = db_url.replace("postgres://", "postgresql://", 1)
+        engine = create_engine(db_url)
```
chenba, line 37: Thanks to @taddes for catching this.

taddes: Here’s hoping it works 😅🤞🍀

taddes (approved):

taddes (approved):
👍 Thx for this @chenba !

PR #1841, opened 2025-10-15
refactor: rmv last insert id, simplify queries, add check

taddes:
## Description

Replace the LAST_INSERT_ID calls that run after raw queries and instead favor of RETURNING.
See [A Note on RETURNING clauses](https://diesel.rs/guides/getting-started.html) in the Diesel documentation. More at the [SQLite RETURNING Overview](https://www.sqlite.org/lang_returning.html)

This requires the modification of several queries, the removal of the utility constant, and an alteration of the `execute` calls to `get_result`, passing in the correct type struct and returning either `id`, in the case of `nodes` and `services` and the `uid` in the case of `users`.

Additionally:
- This adds `IF NOT EXISTS` as an additional gut check in our migrations.
- remove instances of map+map_err to favor Ok(()):
- Used pattern of `let result = <QUERY>` and returned in `Ok(result)` instead of wrapping whole block in Ok for readability. Can go either way on that.
        
```rust
 .map(|_| ())
  .map_err(Into::into)
```

`   Ok(())`
- implements the `check` function in `TokenserverPgDb` so the default signature impl does not recursively call `check`

## Issue(s)

Closes [STOR-376](https://mozilla-hub.atlassian.net/browse/STOR-376).

[STOR-376]: https://mozilla-hub.atlassian.net/browse/STOR-376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tokenserver-db/src/models.rs
+++ b/tokenserver-db/src/models.rs
@@ -29,6 +29,7 @@ impl TokenserverDb {
     // requests, using this function would introduce a race condition, as we could potentially
     // get IDs from records created during other requests.
     const LAST_INSERT_ID_QUERY: &'static str = "SELECT LAST_INSERT_ID() AS id";
+    const LAST_INSERT_UID_QUERY: &'static str = "SELECT LAST_INSERT_ID() AS uid";
 
     pub fn new(
         conn: Conn,
@@ -172,7 +173,7 @@ impl TokenserverDb {
             .execute(&mut self.conn)
             .await?;
 
-        let result = diesel::sql_query(Self::LAST_INSERT_ID_QUERY)
+        let result = diesel::sql_query(Self::LAST_INSERT_UID_QUERY)
             .get_result::<results::PostUser>(&mut self.conn)
             .await?;
         Ok(result)


tokenserver-db-postgres/src/models.rs
```diff
@@ -85,20 +83,18 @@ impl TokenserverPgDb {
...
-
-        diesel::sql_query(Self::LAST_INSERT_ID_QUERY)
             .get_result::<results::LastInsertId>(&mut self.conn)
             .await
             .map(|result| results::PostService {
                 id: result.id as i32,
-            })
-            .map_err(Into::into)
+            })?;
+        Ok(result)
```
pjenvey: I was about to suggest killing this `map` to simply construct `result::PostService` afterwards, but I think we can remove `LastInsertId` entirely:

```rust
        let result = diesel::sql_query(INSERT_SERVICE_QUERY)
            .bind::<Text, _>(&params.service)
            .bind::<Text, _>(&params.pattern)
            .get_result::<results::PostService>(&mut self.conn)
            .await?;
        Ok(result)
```

tokenserver-db-postgres/src/models.rs
```diff
@@ -371,7 +362,7 @@ impl TokenserverPgDb {
         let mut metrics = self.metrics.clone();
         metrics.start_timer("storage.post_user", None);
 
-        diesel::sql_query(QUERY)
+        let result = diesel::sql_query(QUERY)
```
pjenvey, line 362: FYI I just noticed the last commit breaking this (you can fix this later, I'm sure the tests will show it):

We switched from `SELECT LAST_INSERT_ID() AS id` to `RETURNING uid` but `results::PostUser` still assumes the old column name of `id`.

A quick fix would be `RETURNING uid as id` but you might want to consider renaming the struct field to `uid` which would be more accurate

tokenserver-db-postgres/src/models.rs
taddes: K, good call, replaced with `result::PostService`. UPDATED: I'm removing `LastInsertId` in `results.rs` then

tokenserver-db-postgres/src/models.rs
taddes: Ah ok, disregard my comment above. I am just going to get rid of the type alias altogether since it only references two structs, in which one is `uid` as you helpfully pointed out, and the other `id`. The type alias isn't doing us any favors then

tokenserver-db-postgres/src/models.rs
pjenvey: mysql will still need it

pjenvey: (or maybe it doesn't? 👍 if so)

pjenvey (approved):

PR #1845, opened 2025-10-17
feat: add syncstorage-postgres

pjenvey:
## Description

Adds a barebones syncstoage-postgres.

This should be able to share its error type w/ the mysql impl, so that was also moved into syncstorage-db-common

(this is on top of https://github.com/mozilla-services/syncstorage-rs/pull/1844, [diff](https://github.com/mozilla-services/syncstorage-rs/compare/refactor/syncstorage-db-impl-STOR-330...feat/syncstorage-postgres-STOR-330))

And here's some additional tl;dr on code sharing/feature flags w/ all our impls: https://mozilla-hub.atlassian.net/browse/STOR-330?focusedCommentId=1159106 (sigh, this isn't synced to github. if any one out side of Mozilla wants to read this let me know)

## Issue(s)

Closes STOR-330


syncstorage-db-common/src/diesel.rs
```diff
@@ -4,10 +4,11 @@ use backtrace::Backtrace;
 use http::StatusCode;
```
pjenvey, line 4: I'm not in love with keeping this `DbError` here in `syncstorage-db-common::diesel` -- it almost deserves to live in a `syncstorage-diesel` crate, but last thing I want is a new crate for a small module. We can always move this around later -- in the long term we should try to consolidate the spanner `DbError` with it anyways: (more about this [here](https://mozilla-hub.atlassian.net/browse/STOR-330?focusedCommentId=1159106))

felfert:
@pjenvey
Just out of curiosity: How does one get access to Jira?
It offered me to register and I did that already, but when I follow the STOR-330 link, it accepts the login and then immediately says I don't have access

taddes:
Sorry @felfert , Jira is an internal tool that requires you to be a Mozilla employee. It's liked to our LDAP and access control systems. However, we keep issues synced with our Jira tickets in the issues for syncstorage, so you see the various issues we're working on, submitted PRs, etc.

PR #1855, opened 2025-10-21
 test: refactor of tokenserver python tests to support postgres

taddes:
## Description

The integration test’s base TestCase writes directly to the database with raw mysql queries. 

These have been adapted to work with postgres, while maintaining MySQL support. This is through:
- Added a `db_mode` attribute that allows for checking if the `SYNC_TOKENSERVER__DATABASE_URL` env var resolves to a `postgres` or `mysql` prefix, to determine the logic paths.
-  Updating all queries to use the `text` method (aliased as `sqltext`) so that agnostic bind parameters work in both implementations. This also removes the clunky use of %s params.
- Removing helper method that would call `LAST_ROW_ID` which only works for SQL. There were only 3 references to this method anyway, so it was not a particularly beneficial abstraction.
- Refactored an instance of a query being built by string concatenation. While less verbose, it's a little harder to understand. Instead, created explicit queries and parameters for both versions, and then Postgres specific option.
- General refactors for improved readability, a couple old style format strings replaced.

## Testing

Running Tokenserver integration tests

## Issue(s)

Closes [STOR-384](https://mozilla-hub.atlassian.net/browse/STOR-384).

[STOR-384]: https://mozilla-hub.atlassian.net/browse/STOR-384?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tools/integration_tests/tokenserver/mock_fxa_server.py
+++ b/tools/integration_tests/tokenserver/mock_fxa_server.py
@@ -42,5 +42,5 @@ if __name__ == "__main__":
     port = os.environ.get("MOCK_FXA_SERVER_PORT", 6000)
 
     with make_server(host, int(port)) as httpd:
-        print("Running mock FxA server on %s:%s" % (host, port))
+        print(f"Running mock FxA server on {host}:{port}")
         httpd.serve_forever()
--- a/tools/integration_tests/tokenserver/test_e2e.py
+++ b/tools/integration_tests/tokenserver/test_e2e.py
@@ -143,7 +143,7 @@ class TestE2e(TestCase, unittest.TestCase):
         }
         self.assertEqual(res.json, expected_error_response)
         token = self._get_bad_token()
-        headers = {"Authorization": "Bearer %s" % token, "X-KeyID": "1234-qqo"}
+        headers = {"Authorization": f"Bearer {token}", "X-KeyID": "1234-qqo"}
         # Bad token -> 'invalid-credentials'
         res = self.app.get("/1.0/sync/1.5", headers=headers, status=401)
         expected_error_response = {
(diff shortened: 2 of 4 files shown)


tools/integration_tests/tokenserver/test_node_assignment.py
```diff
@@ -5,6 +5,7 @@
 import unittest
 
 from integration_tests.tokenserver.test_support import TestCase
+from sqlalchemy.sql import text as sqltext
```
taddes, line 8: We use this `sqltext` alias elsewhere, so keeping it consistent.

tools/integration_tests/tokenserver/test_support.py
```diff
@@ -31,7 +32,7 @@ def setUpClass(cls):
     def setUp(self):
         engine = create_engine(os.environ["SYNC_TOKENSERVER__DATABASE_URL"])
         self.database = engine.execution_options(isolation_level="AUTOCOMMIT").connect()
-
+        self.db_mode = os.environ["SYNC_TOKENSERVER__DATABASE_URL"].split(":")[0]
```
taddes, line 35: Enables us to check Postgres or SQL. This ended up being the best way to check.

tools/integration_tests/tokenserver/test_support.py
```diff
@@ -109,32 +110,68 @@ def _add_node(
...
-            capacity,
-            current_load,
-            backoff,
-            downed,
-        )
-
-        if id:
-            query += ", id) VALUES(%s, %s, %s, %s, %s, %s, %s, %s)"
-            data += (id,)
+        if not id:
```
taddes, line 113: I know this is verbose, but I have a personal dislike of constructing queries as strings and using concatenation, which in my experience have let to annoying bugs. I thought for our purposes, it's just simpler to have the different explicit queries, and then their Postgres variants. If/when we ever just move to a single DB for support, this can go, but for now, it's much less a headache.

tools/integration_tests/tokenserver/test_support.py
```diff
@@ -151,58 +188,74 @@ def _get_node(self, id):
...
     ):
-        query = """
-            INSERT INTO users (service, email, generation, client_state, \
-                created_at, nodeid, keys_changed_at, replaced_at)
-            VALUES (%s, %s, %s, %s, %s, %s, %s, %s);
-        """
+        if self.db_mode == "postgres":
+            insert_sql = sqltext("""\
+            insert into users (service, email, nodeid, generation, keys_changed_at, client_state, created_at, replaced_at)
+            values (:service, :email, :nodeid, :generation, :keys_changed_at, :client_state, :created_at, :replaced_at)
```
chenba, line 224: (very much nit on formatting) I guess we don't have linter for this python code?  These lines are long but also inconsistent with those in the `else` block.

chenba (approved):
LGTM.   I like the explicit separation of MySQL and Postgres so we can delete the MySQL specific code easily.

tools/integration_tests/tokenserver/test_support.py
taddes: We do have a linter (ruff) but it doesn't work/apply well to the multi-line comments, which can be annoying. I think this can be made to be similar to the SQL standard comment guide where the SQL statements can line up. No argument on it looking inconsistent...

PR #1856, opened 2025-10-21
fix: mariadb compatibility

### LINKED ISSUE #1753: FOR SHARE mariadb sql error
Hello

Since this commit (https://github.com/mozilla-services/syncstorage-rs/commit/69005091a5d9e3adca246d95ebee97b44d241dce#diff-d926ef1a79db49dbd19bbed08d1b43c6a01b9387bf07a7ffe21aadd21e497fe4), my syncstorage server setup is broken.

Logs:

    Aug 18 09:33:16.645 DEBG Query ran in 0.1 ms: SELECT `user_collections`.`last_modified` FROM `user_collections` WHERE `user_collections`.`userid` = ? AND `user_collections`.`collection` = ? LIMIT ? FOR SHARE -- binds: [4, 6, 1]
    Aug 18 09:33:16.646 DEBG Reporting error to Sentry (response error): A database error occurred: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'SHARE' at line 1With debian12 mariadb setup:

```
ii  mariadb-server                     1:10.11.11-0+deb12u1                amd64        MariaDB database server binaries
ii  mariadb-server-core                1:10.11.11-0+deb12u1                amd64        MariaDB database core server files
```

Can you tell me if I'm the only one seeing this ?

Regards

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/STOR-320)

felfert:
Partially reverting #1717
According to MySQL docs (even in latest MySQL 9.4) FOR SHARE and LOCK IN SHARE MODE are equivalent. MariaDB implements the latter only. The additional options mentioned in the docs are not used here.

Closes #1753

## Description

See commit message.

## Testing

Test already exists against MySQL, test for MariaDB done manually. in Light of the current effort to support postgres in the future, extending the existing tests to include MariaDB is probably a waste of time .

## Issue(s)

Closes #1753

### DIFF
--- a/syncstorage-mysql/src/diesel_ext.rs
+++ b/syncstorage-mysql/src/diesel_ext.rs
@@ -3,10 +3,43 @@ use std::{fmt::Debug, marker::PhantomData};
 use diesel::{
     backend::Backend,
     insertable::CanInsertInSingleQuery,
+    mysql::Mysql,
     query_builder::{AstPass, InsertStatement, QueryFragment, QueryId},
+    query_dsl::methods::LockingDsl,
     result::QueryResult,
     Expression, QuerySource, RunQueryDsl,
 };
+
+/// Emit MySQL <= 5.7's `LOCK IN SHARE MODE`
+///
+/// MySQL 8 supports `FOR SHARE` as an alias (which diesel natively supports)
+/// This is required for MariaDB which does not provide that alias.
+pub trait LockInShareModeDsl {
+    type Output;
+
+    fn lock_in_share_mode(self) -> Self::Output;
+}
+
+impl<T> LockInShareModeDsl for T
+where
+    T: LockingDsl<LockInShareMode>,
+{
+    type Output = <T as LockingDsl<LockInShareMode>>::Output;
+
+    fn lock_in_share_mode(self) -> Self::Output {
+        self.with_lock(LockInShareMode)
+    }
+}
+
+#[derive(Debug, Clone, Copy, QueryId)]
+pub struct LockInShareMode;
+
+impl QueryFragment<Mysql> for LockInShareMode {
+    fn walk_ast<'b>(&'b self, mut out: AstPass<'_, 'b, Mysql>) -> QueryResult<()> {
+        out.push_sql(" LOCK IN SHARE MODE");
+        Ok(())
+    }
+}
 
 #[allow(dead_code)] // Not really dead, Rust can't see it.
 #[derive(Debug, Clone)]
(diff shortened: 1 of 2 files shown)


felfert:
Tests failed because of some google cloud error (some key missing)?!

chenba:
Thank you for the patch, @felfert.  The code looks fine to me.  But let's see what other team members think.

taddes:
> Tests failed because of some google cloud error (some key missing)?!

That was an annoying problem with Oracle not keeping their key up to date! Rebasing on `master` should fix it!

chenba:
> Tests failed because of some google cloud error (some key missing)?!

Hey @felfert, CI won't work on PRs from contributors as those workflows cannot access secrets.  That's why I created https://github.com/mozilla-services/syncstorage-rs/pull/1861.  The CI status of that PR is reflected here since they are on the same commit.

felfert:
Ok, rebased on `master` as suggested by @taddes 
Apparently that did not help so it must be the permission problem mentioned by @chenba 
with his PR #1861 Tests did run fine.

felfert:
Regardless, this is a circleCI problem. I don't know circleCI, but on github workflows, if you configure a test that runs for PR's, those runs are performed in the repository owner's context, not in the context of the PR-submitter.
Strange concept that circleCI has there.

chenba:
I've pushed the latest commit from this PR to https://github.com/mozilla-services/syncstorage-rs/pull/1861 so we can run it through CI.  However, I don't think that is going to work since we've made changes to the CI workflow itself _and_ the project is configured to wait for specific CI jobs to complete successfully; some of those CI jobs do not exist in the revision for this PR.  @felfert unfortunately we'll need you to rebase at least one more time.

felfert:
@chenba Ok, rebase done

chenba:
@felfert Thanks.  CI all green now.

chenba:
@pjenvey what do you think of this patch?

syncstorage-mysql/src/diesel_ext.rs
```diff
@@ -3,11 +3,43 @@ use std::{fmt::Debug, marker::PhantomData};
...
+    mysql::Mysql,
     query_builder::{AstPass, InsertStatement, QueryFragment, QueryId},
+    query_dsl::methods::LockingDsl,
     result::QueryResult,
     Expression, QuerySource, RunQueryDsl,
 };
 
+/// Emit MySQL <= 5.7's `LOCK IN SHARE MODE`
+///
+/// MySQL 8 supports `FOR SHARE` as an alias (which diesel natively supports)
```
pjenvey, line 15: Could you add an additional comment here that this is needed for MariaDB compat?

pjenvey (changes requested):

syncstorage-mysql/src/diesel_ext.rs
felfert: done

PR #1858, opened 2025-10-21
feat: support systemd-journal logging

felfert:
## Description

This PR adds support for native systemd-journal logging if running under systemd.

### Benefits
- Eliminates duplicated timestams and useles loglevel texts showing with journalctl
- Reduces journal log record size
- As a side effect, enables default loglevel colors in journalctl

## Testing

Done manually by running syncserver in a systemd-service and then observing the log with journalctl

### DIFF
--- a/tools/tokenserver/loadtests/poetry.lock
+++ b/tools/tokenserver/loadtests/poetry.lock
@@ -117,7 +117,7 @@ version = "1.2.0"
 description = "Python bindings for the Brotli compression library"
 optional = false
 python-versions = "*"
-groups = ["main", "dev"]
+groups = ["main"]
 files = [
     {file = "brotli-1.2.0-cp27-cp27m-macosx_10_9_x86_64.whl", hash = "sha256:99cfa69813d79492f0e5d52a20fd18395bc82e671d5d40bd5a91d13e75e468e8"},
     {file = "brotli-1.2.0-cp27-cp27m-manylinux1_i686.whl", hash = "sha256:3ebe801e0f4e56d17cd386ca6600573e3706ce1845376307f5d2cbd32149b69a"},
@@ -999,7 +999,7 @@ version = "3.0.2"
 description = "Safely add untrusted strings to HTML/XML markup."
 optional = false
 python-versions = ">=3.9"
-groups = ["main"]
+groups = ["main", "dev"]
 files = [
     {file = "MarkupSafe-3.0.2-cp310-cp310-macosx_10_9_universal2.whl", hash = "sha256:7e94c425039cde14257288fd61dcfb01963e658efbc0ff54f5306b06054700f8"},
     {file = "MarkupSafe-3.0.2-cp310-cp310-macosx_11_0_arm64.whl", hash = "sha256:9e2d922824181480953426608b81967de705c3cef4d1af983af849d7bd619158"},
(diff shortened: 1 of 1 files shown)


clouserw:
Hi @felfert 

Thanks for the patch.

We're currently focused on a roadmap to provide docker images for releases which should make deployment and maintenance a lot easier for folks.  That would mean the logging would typically be taken care of by docker/k8s automatically.  I'm expecting that to cover most of the self-hosting crowd and would let us avoid adding these additional dependencies.  Do you have a use case that wouldn't be able to use a container and would need this?  

Cheers

felfert:
> Do you have a use case that wouldn't be able to use a container and would need this?

Well, IMHO it's more like a matter of personal taste. I have packaged syncstorage-rs as rpm for [Fedora and RHEL](https://copr.fedorainfracloud.org/coprs/felfert/syncstorage-rs/) (Ubuntu&Debian in the works) and regarding ease of setup, those are even simpler than dealing with containers. Especially on cheap vhosts with LAMP-stacks where you have  mysql/mariadb and apache available by default.

If you worry about the dependencies, I could add a feature. So:
Let me know, if adding a feature for selecting/deselecting  would lead to acceptance of this PR.

Regards
 -Fritz

clouserw:
Sorry for the delay here.  I'm inclined to find a way to land this.  We have it on our list to review this sprint.

syncserver/src/logging.rs
```diff
@@ -4,6 +4,17 @@ use crate::error::ApiResult;
...
 use slog_mozlog_json::MozLogJson;
+use std::os::fd::AsFd;
+
+fn connected_to_journal() -> bool {
+    rustix::fs::fstat(std::io::stderr().as_fd())
+        .map(|stat| format!("{}:{}", stat.st_dev, stat.st_ino))
+        .ok()
+        .and_then(|stderr| {
+            std::env::var_os("JOURNAL_STREAM").map(|s| s.to_string_lossy() == stderr.as_str())
+        })
```
jalil-salame: You can compare the bytes directly instead of doing `to_string_lossy()`

syncserver/src/logging.rs
```diff
@@ -4,6 +4,17 @@ use crate::error::ApiResult;
 
 use slog::{self, slog_o, Drain};
 use slog_mozlog_json::MozLogJson;
+use std::os::fd::AsFd;
+
+fn connected_to_journal() -> bool {
+    rustix::fs::fstat(std::io::stderr().as_fd())
```
jalil-salame: Since systemd is linux only you could drop `rustix` and use [`MetadataExt`](https://doc.rust-lang.org/std/os/linux/fs/trait.MetadataExt.html) to get the information from `std::fs::metadata("/dev/stderr")`. It's mildly annoying that you can't stat the `fd` directly, but AFAIK they should be equivalent.

jalil-salame (changes requested):

syncserver/src/logging.rs
pjenvey: 👍 and without rustix, there's really no reason for needing to wrap this under a feature flag

felfert: Strange.. I got no notification about this... just accidentially saw this right now.
Ok, will have a look at MetaDataExt ...

felfert: Ok, rebased against latest master and eliminated dependency on rustix by using MetadataExt now.

syncserver/src/logging.rs
felfert: done

syncserver/src/logging.rs
```diff
@@ -1,10 +1,22 @@
+use std::fs;
 use std::io;
+use std::os::linux::fs::MetadataExt;
```
jalil-salame, line 7: Since this is linux only, you do need to gate it with `#[cfg(target_os = "linux")]`or it won't compile in other OSs

felfert: right

felfert: Ok, added cfg conditionals now, what I am not shure about, is the dependency on the slog-journald crate itself. Does that need to be conditional as well? Problem is: I don't have a non-Linux system to verify that.

jalil-salame: Looking at the source code, `libsystemd` (a dependency of `slog-journald`) is `unix` only, so it should be gated behind linux, or at-least unix.

felfert: Ok, now the dependeny on slog-journald is conditional. Some more conditionals added in the code to get rid of the reference to slog_journald::JournaldDrain.

chenba:
it failed `cargo fmt -- --check`

```diff
Diff in /***/syncstorage-rs/syncserver/src/logging.rs:16:
     fs::metadata("/dev/stderr")
         .map(|meta| format!("{}:{}", meta.st_dev(), meta.st_ino()))
         .ok()
-        .and_then(|stderr| {
-            std::env::var_os("JOURNAL_STREAM").map(|s| s == stderr.as_str())
-        })
+        .and_then(|stderr| std::env::var_os("JOURNAL_STREAM").map(|s| s == stderr.as_str()))
         .unwrap_or(false)
 }
```

chenba:
The patch compiled without error on macOS.

felfert:
Yep that fmt check was just a silly missing fmt after implementing suggestions from @jalil-salame.
Fixed now.

jalil-salame (approved):

chenba (approved):

pjenvey (approved):

pjenvey:
Thanks for the contribution @felfert and also @jalil-salame!

PR #1866, opened 2025-10-23
chore(CI): fetch MySQL public key from ubuntu

chenba:
This fixes MySQL lib sig verification due to expired pub key.


chenba:
the commit message is wrong, sorry.  but I don't think it's worth more CI time just to amend it.

taddes (approved):

PR #1872, opened 2025-10-29
WIP: feat create stync db models and schema structs

taddes:
## Description

The use of raw queries as we have used for Sync and Tokenserver are advantageous from a readability perspective, but favor should be given to the ORM query constructor in certain circumstances. Diesel requires there to be instantiated models that represent the schema data structures.

This is especially important as a performance consideration for Postgres, where raw queries needing sql_query will deserialize its data by name, not by index. That means that you cannot deserialize into a tuple, and structs which you deserialize from this function will need to have #[derive(QueryableByName)].

This PR introduces the model structs to be used in cases where we want to move away from raw queries in favor of diesel ORM methods. However, given the readability advantage, it is easy to use a Rust doc comment, which when indented one point, allows one to paste the raw query as part of the method. We should do this because it makes the ORM queries instantaneously understandable.

## Issue(s)

Closes [STOR-398](https://mozilla-hub.atlassian.net/browse/STOR-398).

[STOR-398]: https://mozilla-hub.atlassian.net/browse/STOR-398?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/syncstorage-postgres/src/lib.rs
+++ b/syncstorage-postgres/src/lib.rs
@@ -1,5 +1,5 @@
 #![allow(non_local_definitions)]
-#![allow(unused_imports)] // XXX:
+#![allow(unused_imports)]
 #[macro_use]
 extern crate diesel;
 extern crate diesel_migrations;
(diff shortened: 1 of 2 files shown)


taddes:
renamed branch. #1873

Issue #1869, opened 2025-10-27
Failing e2e (integration) Python tests when Tokenserver is using Postgres


data-sync-user:
In [https://mozilla-hub.atlassian.net/browse/STOR-353](https://mozilla-hub.atlassian.net/browse/STOR-353|smart-link) we set up the CI workflow to run existing unit and integration tests with Tokenserver using Postgres.  Three bugs were found and fixed in the course of fixing the unit tests.

The integration tests are failing due to what appears to be a stack overflow bug.  I’ve added `|| true` to the command to unblock CI.  (Note that the unit tests on Spanner is also bypassing CI failure with `|| true`.)  We need to fix the stack overflow bug or whatever the source of the failures is.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/STOR-397)

data-sync-user:
➤ Barry Chen commented:

In the quest to get more debug info, I’ve found that the stack overflow problem does not occur when testing with a debug build, i.e. cargo install --deubg ... in Dockerfile.

data-sync-user:
➤ Barry Chen commented:

The issue appears to be a #[cfg(debug_assertions)] flag leading to TokenserPgDb::post_user being excluded from the release build, causing an infinite recursion from trait delegation.

PR #1876, opened 2025-10-31
Fix Postgres post_user Infinite Recursion

chenba:
The `postgres-e2e-tests` job was failing in CircleCI due to a stack overflow error.  The problem was that `TokenserverPgDb::post_user` had `#[cfg(debug_assertions)]` thus was not in the release build against which the integration tests were executed, resulting in an infinite recursion in the Db train impl of `post_user`.

Closes STOR-397 / https://github.com/mozilla-services/syncstorage-rs/issues/1869

### DIFF
--- a/tokenserver-db-postgres/src/models.rs
+++ b/tokenserver-db-postgres/src/models.rs
@@ -347,7 +347,6 @@ impl TokenserverPgDb {
     }
 
     /// Method to create a new user, given a `PostUser` struct containing data regarding the user.
-    #[cfg(debug_assertions)]
     async fn post_user(&mut self, params: params::PostUser) -> DbResult<results::PostUser> {
         const QUERY: &str = r#"
             INSERT INTO users (service, email, generation, client_state, created_at,
--- a/Makefile
+++ b/Makefile
@@ -98,7 +98,7 @@ docker_run_postgres_e2e_tests:
 		-f docker-compose.e2e.postgres.yaml \
 	 	up \
 	 	--exit-code-from postgres-e2e-tests \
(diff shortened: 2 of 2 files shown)


tokenserver-db-postgres/src/models.rs
```diff
@@ -347,7 +347,6 @@ impl TokenserverPgDb {
     }
 
     /// Method to create a new user, given a `PostUser` struct containing data regarding the user.
-    #[cfg(debug_assertions)]
```
chenba, line 350: @taddes Why did we have this flag here?

taddes: That's a conditional compilation attribute, which tells the compiler to only include the annotated code when compiling in debug mode (`debug_assertions`). This was for testing purposes and matched the interface for MySQL, but it looks like it might have needed to be annotated  the `impl Db for TokenserverPgDb `. There is a mix of cases where some are annotated with `#[cfg(debug_assertions)]` at the interface definition at the bottom, and some when the traits are actually implemented. I don't know why it's one of the few methods in both postgres and mysql to not have that annotation. Needless to say, I'm sorry if that's the goose you've been chasing and that was seemingly enough to break the test run.

chenba: > This was for testing purposes and matched the interface for MySQL, but it looks like it might have needed to be annotated the impl Db for TokenserverPgDb .

Why don't we want the actual impl in the release build?

taddes: That's a good question, this applies to a lot of them and existing ones in MySQL. Let's tag @pjenvey

Issue #1859, opened 2025-10-22
Investigate Removing host name from "syncstorage_storage_pool_connections_idle" metric


data-sync-user:
The telegraf node is consuming excess memory due to a highly cardinal value in the metrics. This was an issue that was discovered in autopush and may well apply to syncserver.
[~accountid:62fffa83d75f523a60ef5653]  identified the `hostname` tag as being problematic, where the `hostname` corresponds with a pod name - so it's basically an infinitely cardinal value over time. The `hostname` label will have different values depending on the pod and because the values for that field are unbound, we'll keep multiplying our metrics count any time we spin up new pods.

A lot of the metrics look something like:

```
# HELP syncstorage_storage_pool_connections_idle Telegraf collected metric
# TYPE syncstorage_storage_pool_connections_idle gauge
syncstorage_storage_pool_connections_idle{app="sync",environment="prod",host="telegraf-polling-service",hostname="debug-syncapp-76cbb94554-pk48l",metric_type="gauge",realm="prod"} 11
syncstorage_storage_pool_connections_idle{app="sync",environment="prod",host="telegraf-polling-service",hostname="syncapp-785bccc775-22v97",metric_type="gauge",realm="prod"} 20
...
syncstorage_storage_pool_connections_idle{app="sync",environment="prod",host="telegraf-polling-service",hostname="syncapp-785bccc775-jblp2",metric_type="gauge",realm="prod"} 21
syncstorage_storage_pool_connections_idle{app="sync",environment="prod",host="telegraf-polling-service",hostname="syncapp-785bccc775-jphxh",metric_type="gauge",realm="prod"} 11
```

Slack convo: [https://mozilla.slack.com/archives/C4E35LBBR/p1758812024402119](https://mozilla.slack.com/archives/C4E35LBBR/p1758812024402119|smart-link) 
PR in Autopush: [https://github.com/mozilla-services/autopush-rs/pull/1029](https://github.com/mozilla-services/autopush-rs/pull/1029|smart-link) 
Jira Ticket in PUSH: [https://mozilla-hub.atlassian.net/browse/PUSH-578](https://mozilla-hub.atlassian.net/browse/PUSH-578|smart-link) 
[...]

PR #1880, opened 2025-11-05
refactor(metrics): make hostname in metrics optional

chenba:
## Description

The hostname in the metrics could significantly increase the cardinality.  This patch make the inclusion of the hostname optional, with a default of false.

## Testing

All existing tests should pass.

## Issue(s)

Closes STOR-393 / https://github.com/mozilla-services/syncstorage-rs/issues/1859


syncserver/src/server/mod.rs
```diff
@@ -489,46 +491,69 @@ fn spawn_metric_periodic_reporter<T: GetPoolState + Send + 'static>(
...
+        ) -> MetricBuilder<'a, 'b, M>
+        where
+            M: Metric + From<String>,
+        {
+            if let Some(hostname) = hostname {
+                metric.with_tag("hostname", hostname)
+            } else {
+                metric
+            }
+        }
```
chenba: The alternative would be a macro.  I think a macro reads better and it's my preference.  @pjenvey do you have a preference?

pjenvey: I agree with all the lifetimes involved makes a quick macro justified.

Though I think there's potentially another less generic option since they're all (currently) `gauge` calls w/ `u64`s: a closure that does the entire gauge+send call: `let send_gauge_with_maybe_hostname = |key, value| { ... }` (with it capturing metrics/hostname from the outer scope so they don't need to be passed as args)

chenba: Made the change to a closure.  Thank you for the suggestion.

taddes (approved):
👍

syncserver/src/server/mod.rs
```diff
@@ -489,46 +491,54 @@ fn spawn_metric_periodic_reporter<T: GetPoolState + Send + 'static>(
...
-        .expect("Couldn't get hostname");
+    let hostname = include_hostname_tag.then(|| {
+        hostname::get()
+            .expect("Couldn't get hostname")
+            .into_string()
+            .expect("Couldn't get hostname")
+    });
+
     tokio::spawn(async move {
+        let send_gauage_with_maybe_hostname = |key, value| {
```
pjenvey:
```suggestion
        let send_gauge_with_maybe_hostname = |key, value| {
```

pjenvey (changes requested):

pjenvey (approved):

PR #1883, opened 2025-11-06
 refactor: tokenserver db crates models/etc -> db

pjenvey:
## Description

Refactors tokenserver db crates to the new layout -- this was pretty quick, so just getting it out of the way now

## Issue(s)

Closes STOR-405

(Note I git mv'd tokenserver-db `models.rs` to tokenserver-mysql's `db_impl.rs`, github just doesn't render it as such here. You can see an easier diff of it in the second commit [here](https://github.com/mozilla-services/syncstorage-rs/pull/1883/commits/51e487d1db0942a42489d018f10d23201a8b7da7))

### DIFF
--- a/tokenserver-db-postgres/src/db/db_impl.rs
+++ b/tokenserver-db-postgres/src/db/db_impl.rs
@@ -15,45 +15,14 @@ use http::StatusCode;
 use syncserver_common::Metrics;
 use tokenserver_db_common::{params, results, Db, DbError, DbResult};
 
-use super::pool::Conn;
+use super::TokenserverPgDb;
 
-/// Struct containing connection and related metadata to a Tokenserver
-/// Postgres Database.
-pub struct TokenserverPgDb {
-    /// Async PgConnection handle.
-    conn: Conn,
-    /// Syncserver_common Metrics object.
-    metrics: Metrics,
-    /// Optional Service Identifier.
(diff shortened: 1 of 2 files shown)


tokenserver-db/src/lib.rs
```diff
@@ -39,3 +35,6 @@ pub fn pool_from_settings(
         }
     })
 }
+
+#[cfg(not(any(feature = "mysql", feature = "postgres")))]
+compile_error!("at least one of the \"mysql\" or \"postgres\" features must be enabled");
```
chenba, line 40: 👍

taddes (approved):
👍 good stuff, will be nice to have consistency across all implementations. Think the CI failures are just the intermittent flakes, just watching those but approving 👍

chenba:
LGTM but I think FxA stage is breaking the integration tests.

taddes:
> LGTM but I think FxA stage is breaking the integration tests.

Yeah, I'm seeing the Oauth test fail due to a missing `verified` key. I imagine there was a recent change that resulted in this?

pjenvey:
> I imagine there was a recent change that resulted in this?

It's due to [this](https://github.com/mozilla/fxa/pull/19621) having gone out, but Lauren's aiming to roll out a backwards compat fix for us probably end of day Friday

PR #1887, opened 2025-11-12
feat: impl get collection and collection id methods

taddes:
## Description
In order to implement the read/write lock logic, we need `get_collection_id` defined for `lock_for_read` and `get_or_create_collection_id` for `lock_for_write`, so that must be completed first.

These are base query methods that are used by other method calls, so they need to be implemented first

This also updated the `collection_id` type to match the other signatures, which are of type `INTEGER` / `Int4` / `i32`

## Issue(s)

Closes [STOR-401](https://mozilla-hub.atlassian.net/browse/STOR-401).

[STOR-401]: https://mozilla-hub.atlassian.net/browse/STOR-401?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/syncstorage-postgres/src/db/db_impl.rs
+++ b/syncstorage-postgres/src/db/db_impl.rs
@@ -1,6 +1,10 @@
 #![allow(unused_variables)] // XXX:
 use async_trait::async_trait;
-use diesel::{sql_query, sql_types::Text, OptionalExtension};
+use diesel::{
+    sql_query,
+    sql_types::{Integer, Text},
+    OptionalExtension,
+};
 use diesel_async::{AsyncConnection, RunQueryDsl, TransactionManager};
 use syncstorage_db_common::{params, results, util::SyncTimestamp, Db};
 use syncstorage_settings::Quota;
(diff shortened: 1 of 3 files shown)


syncstorage-db-common/src/results.rs
```diff
@@ -86,6 +86,14 @@ pub struct ConnectionInfo {
 
 pub type GetCollectionId = i32;
 
+pub type GetOrCreateCollectionId = i32;
```
pjenvey: let's get rid of the bits in here for a couple reasons:

- results/params only include the types needed for the trait method params/results, whereas both these types are more internal details. Especially for `GetOrCreateCollectionId`, if we really want a standard type to reference we could probably just reuse the existing `GetCollectionId`
- At some point it probably make sense to put shared diesel things like `IdResult` somewhere (maybe this justifies a `syncserver-diesel` crate? but I'm avoiding that for now as we have a lot of crates!). There is a `diesel.rs` module in syncstorage-db-common where this could live instead for the sake of sharing (`syncstorage-mysql` has an exact copy of this `IdResult`)

taddes: Sounds good, makes sense. Moved the `IdResult` into the local impl and we can certainly file a ticket to create that `syncserver-diesel` crate down the road, that makes sense to me.

PR #1891, opened 2025-11-13
feat: postgres read and write locks

taddes:
## Description

Introduces the logic to lock the db for reads and writes for the Postgres implementation. 

Will merge in and rebase based on changes from #1896 

## Issue(s)

Closes [STOR-355](https://mozilla-hub.atlassian.net/browse/STOR-335).

[STOR-355]: https://mozilla-hub.atlassian.net/browse/STOR-355?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/syncstorage-postgres/src/db/db_impl.rs
+++ b/syncstorage-postgres/src/db/db_impl.rs
@@ -13,7 +13,10 @@ use syncstorage_db_common::{
 use syncstorage_settings::Quota;
 
 use super::PgDb;
-use crate::{bsos_query, pool::Conn, schema::bsos, schema::user_collections, DbError, DbResult};
+use crate::{
+    bsos_query, db::CollectionLock, pool::Conn, schema::bsos, schema::user_collections, DbError,
+    DbResult,
+};
 
 #[async_trait(?Send)]
 impl Db for PgDb {


syncstorage-postgres/syncstorage_postgres_db.md
```diff
@@ -14,10 +14,10 @@ Stores per-user, per-collection metadata.
 
 | Column          | Type        | Description                                                           |
 | --------------- | ----------- | --------------------------------------------------------------------- |
-| `fxa_uid`       | `UUID`      | Firefox Account UID PK (part 1)                              |
+| `fxa_uid`       | `BIGINT`      | Firefox Account UID PK (part 1)                              |
 | `fxa_kid`       | `TEXT`      | Key identifier; part of the sync crypto context. PK (part 2) |
 | `collection_id` | `INTEGER`    | Maps to a named collection. PK (part 3)                       |
-| `modified`      | `TIMESTAMP` | Last modification time (server-assigned, updated on writes)           |
+| `modified`      | `BIGINT` | Last modification time (server-assigned, updated on writes)           |
```
pjenvey: Is this too bothersome to support as a `TIMESTAMP`?

syncstorage-postgres/migrations/2025-10-20-155711_create_schema/up.sql
```diff
@@ -9,10 +9,10 @@
 
 -- user_collections table
 CREATE TABLE user_collections (
-    fxa_uid UUID NOT NULL,
+    fxa_uid BIGINT NOT NULL,
```
chenba: I don't understand this change.  Should the column also be renamed?  iirc the last time the team chatted about this type we want it to be TEXT.

pjenvey: Agreed, but I'm killing `fxa_uid/kid` shortly, so it doesn't matter

syncstorage-postgres/src/db/db_impl.rs
```diff
@@ -50,11 +50,63 @@ impl Db for PgDb {
         Ok(true)
     }
 
+    /// Explicitly lock the matching row in the user_collections table. Read
+    /// locks do `SELECT ... LOCK IN SHARE MODE` and write locks do `SELECT
+    /// ... FOR UPDATE`.
+    ///
+    /// In theory it would be possible to use serializable transactions rather
+    /// than explicit locking, but our ops team have expressed concerns about
+    /// the efficiency of that approach at scale.
```
chenba, line 65: Cool.  Are there written notes on that somewhere?

syncstorage-postgres/migrations/2025-10-20-155711_create_schema/up.sql
taddes: It'll be in another PR when we change the whole field. Also the PR is still WIP

syncstorage-postgres/syncstorage_postgres_db.md
taddes: Not necessarily, may opt for a raw query over the ORM methods, as we've discussed, given those can be a pain with the chain methods. Do we prefer this to be a timestamp? I know in Spanner it's a timestamp and in MySQL a BIGINT

taddes: For now, @pjenvey  I say let's keep it consistent as an integer, given the integrated testing for MySQL and Postgres, much like in Tokenserver. We have a ticket on the books to standardize all datetime ops to `chrono`, which also happens to include making sure any ops that should move to datetime ones, opposed to using large numbers will update. This way we can update MySQL and Postgres at the same time. Thoughts?

syncstorage-postgres/src/db/db_impl.rs
taddes: I will look around and see regarding the ops team stuff, though the first portion is what I took from Diesel's internal docs on these lock methods.

syncstorage-postgres/syncstorage_postgres_db.md
chenba: > given the integrated testing for MySQL and Postgres

Sorry I don't have enough context to understand this.  Can you link to a test as an example?  Thanks!

taddes: A good example is like the work you did for the Postgres tests that ran in CI for Tokenserver. The reason they were general purpose and could work for both SQL and Postgres was that the interfaces were quite similar. My thought here is that we try and keep true-ish to MySQL and as we near the end of integrating the tests, look at making adjustments if need be. Changing a data type from an `i64` to a `timestamp` for instance results in pretty different logic that would check it. This way we're not duplicating things before we need to.

syncstorage-postgres/src/db/db_impl.rs
```diff
@@ -50,18 +50,105 @@ impl Db for PgDb {
...
+        if let Some(modified) = modified {
+            let modified = SyncTimestamp::from_i64(modified)?;
+            self.session
+                .coll_modified_cache
+                .insert((user_id as u32, collection_id), modified);
+        }
+        self.session
+            .coll_locks
+            .insert((user_id as u32, collection_id), CollectionLock::Read);
+        Ok(())
```
chenba, line 116: This function is identical to the one in `syncstorage-mysql/src/db/db_impl.rs`.  Do we need two copies of this function?

(The MySQL copy also has [two](https://github.com/mozilla-services/syncstorage-rs/blob/70a7854b8fe1257320656efa933b21445e3ba7a3/syncstorage-mysql/src/db/db_impl.rs#L79) [questions](https://github.com/mozilla-services/syncstorage-rs/blob/70a7854b8fe1257320656efa933b21445e3ba7a3/syncstorage-mysql/src/db/db_impl.rs#L81) in the comments I'm very interested in.)

syncstorage-postgres/src/db/db_impl.rs
```diff
@@ -50,18 +50,105 @@ impl Db for PgDb {
...
+            }
+            self.session
+                .coll_modified_cache
+                .insert((user_id as u32, collection_id), modified);
+        }
+
+        self.session
+            .coll_locks
+            .insert((user_id as u32, collection_id), CollectionLock::Write);
+        Ok(())
```
chenba, line 160: Same with this function.  It's functionally identical to the MySQL one.  (There's a `to_string` here and a `to_owned` there for the error message so it's not syntactically 100% identical.)

syncstorage-postgres/src/db/db_impl.rs
taddes: Yes! I am curious about those comments as well. It's likely we can move said function into a shared crate at some point if both of them use it.

syncstorage-postgres/src/db/db_impl.rs
taddes: Certainly can swap it, since those `to_string` and `to_owned` do much of the same thing.

chenba:
Related to this PR but more of a general question.  When are entries in those process-local hashmaps of locks updated?  I see inserts but not removals.  I feel like I'm missing something very fundamental on db level locks vs what's in those hashmaps.

chenba:
> When are entries in those process-local hashmaps of locks 

Oh the hashmaps are per request look like.  I _think_ I understand now.

pjenvey:
> Related to this PR but more of a general question. When are entries in those process-local hashmaps of locks updated? I see inserts but not removals. I feel like I'm missing something very fundamental on db level locks vs what's in those hashmaps.

These were lazily carried over from the Python version: they're never removed, and only really used as a sanity check in case the lock methods are called more than once (which never happens).

This could probably be improved/simplified -- we could remove them upon commit/rollback, but if we're doing that we might consider making commit/rollback consume `self` (if our usage of those methods allows that).

syncstorage-postgres/src/db/db_impl.rs
pjenvey: There's some small opportunities to share more common code between the different backends but unfortunately we can't share general diesel ORM usage (utilizing diesel-async) between different backends easily (if at all), at least at the moment.

Basically because diesel's very strict: its base API (and its more internal associated types/trait bounds that come along with it) is typed to specific db backends. This has the benefit of enforcing specific backend dialect API usage at compile time, but with the downside of making it difficult to write code against its generic traits like `diesel::Connection` -- it just wasn't designed for such usage.

However plain diesel added a kind of layer on top of its core called [MultiConnection](https://diesel.rs/news/2_1_0_release.html#multiconnection-support) that enables this kind of usage, but unfortunately it's [not yet supported by diesel-async](https://github.com/weiznich/diesel_async/discussions/108).

The original plan was to prefer raw sql queries for most cases anyway, moving away from the mysql ORM styled code like tokenserver does. Part of building postgres "from scratch" is we're able to make a sort of audit of non spanner impls as the mysql version was never fully finalized (granted self hosters seemed to have had success with it).

I'm not opposed to the ORM usage though, if it's easy enough to use -- in fact I intended for raw sql in *most cases* but not necessarily all -- I definitely prefer the ORM for at least one case: the `get_bsos/_ids` call where we dynamically build a query is much easier/cleaner/less bug ridden with an ORM building it for us vs manually constructing it.

pjenvey: The second question I've covered here: https://github.com/mozilla-services/syncstorage-rs/pull/1891#issuecomment-3556052449

The first question sounds like more of a TODO/bug, we shouldn't be casting a u64 to a u32 here (or in mysql)! Let's change the type in the `coll_` maps to match `legacy_id`'s. We could even put the entire `UserIdentifier` in there instead, as long as we're sure it eq/hashes correctly for such usage.

syncstorage-postgres/src/db/db_impl.rs
chenba: (I'm pretty sure I responded but I guess the packets leaked into the ocean or something.)

The `to_string` vs `to_owned` was just an aside.  My point was the Postgres and MySQL versions of the fn are essentially identical.

syncstorage-postgres/syncstorage_postgres_db.md
chenba: I think I understand your explanation, @taddes, thanks.  But I don't agree or maybe simply don't understand why the existing tests should stop us from using the correct data type(s).

> My thought here is that we try and keep true-ish to MySQL and as we near the end of integrating the tests, look at making adjustments if need be. 

How do we decide if adjustments are needed?

pjenvey: The tests are against the `Db`/`DbPool` interfaces, whereas the column type here is an internal impl. detail. By the time our `Box<dyn Db>` returns e.g. a `results::GetBso`, it would have converted whatever internal representation of `modified` (whether `BIGINT` or `TIMESTAMP`) to a `SyncTimestamp`.

syncstorage-postgres/src/db/db_impl.rs
taddes: Hey @pjenvey & @chenba , I filed [STOR-414](https://mozilla-hub.atlassian.net/browse/STOR-414) to take care of the aforementioned casting, just posting here for tracking purposes.

Otherwise, imported the chrono feature and made changes to support timestamp.

syncstorage-postgres/src/db/db_impl.rs
taddes: Yeah, as @pjenvey , we unfortunately can't make a shared version given Diesel's API

syncstorage-postgres/syncstorage_postgres_db.md
taddes: Updated to use timestamp 👍

syncstorage-postgres/migrations/2025-10-20-155711_create_schema/up.sql
```diff
@@ -102,4 +87,4 @@ CREATE TABLE batch_bsos (
         collection_id,
         batch_id
     ) ON DELETE CASCADE
-);
+);
```
chenba, line 90: Were these formatting changes from an automated tool?

syncstorage-postgres/src/db/db_impl.rs
```diff
@@ -50,18 +51,105 @@ impl Db for PgDb {
...
+        // use it as-is.
+        if self
+            .session
+            .coll_locks
+            .contains_key(&(user_id as u32, collection_id))
+        {
+            return Ok(());
+        }
+
+        // Lock db.
```
chenba: This is a bit misleading since it's not locking the db.

syncstorage-postgres/src/db/db_impl.rs
```diff
@@ -50,18 +51,105 @@ impl Db for PgDb {
...
+            .session
+            .coll_locks
+            .get(&(user_id as u32, collection_id))
+        {
+            return Err(DbError::internal(
+                "Can't escalate read-lock to write-lock".to_string(),
+            ));
+        }
+
+        // Lock DB
```
chenba: Same as above.  This is misleading.

syncstorage-postgres/src/db/mod.rs
```diff
@@ -53,9 +53,9 @@ struct PgDbSession {
     /// The "current time" on the server used for this session's operations.
     timestamp: SyncTimestamp,
     /// Cache of collection modified timestamps per (HawkIdentifier, collection_id).
-    coll_modified_cache: HashMap<(UserIdentifier, i32), SyncTimestamp>,
+    coll_modified_cache: HashMap<(u32, i32), SyncTimestamp>,
     /// Currently locked collections.
-    coll_locks: HashMap<(UserIdentifier, i32), CollectionLock>,
+    coll_locks: HashMap<(u32, i32), CollectionLock>,
```
chenba: I like to learn the reason behind this change.

syncstorage-postgres/migrations/2025-10-20-155711_create_schema/up.sql
taddes: VS Code format on save.

syncstorage-postgres/src/db/db_impl.rs
taddes: The call to `begin` sets the session `in_transaction` value to `true`, which for our purposes is setting a flag that it's "locked". It's the same in the MySQL impl with the db lock comment. We can remove them if you want.

syncstorage-postgres/src/db/mod.rs
taddes: Simplified it to match how it worked in SQL, without the `UserIdentifer` struct, since all we need in the context is the `u32` value to use the `user_id`. In Spanner, a number of `clone` calls are made on the `UserIdentifier`, which is more costly. We can decide which we want, I changed it back to `UserIdentifier`.

syncstorage-postgres/migrations/2025-10-20-155711_create_schema/up.sql
chenba: Nice.  What's the extension?  I'll see if I can find something in nvim that'll match it for consistency sake.

syncstorage-postgres/src/db/db_impl.rs
chenba: > for our purposes is setting a flag that it's "locked"

Maybe we are not talking about same "it".  The entire db is not locked; we are locking some records for read or write, sure.  I guess if the comment in the MySQL impl hasn't confused anyone all these years, couldn't hurt here either.

taddes: Yeah, maybe this warrants a discussion offline if there's another bit of logic we want for more explicit "locking"

syncstorage-postgres/migrations/2025-10-20-155711_create_schema/up.sql
taddes: There's an included formatting with the official Postgres support for VSCode and also [PgFormatter](https://marketplace.visualstudio.com/items?itemName=bradymholt.pgformatter) which I think is the same one.

syncstorage-postgres/src/db/db_impl.rs
pjenvey: Maybe "Lock the users' rows (FOR SHARE)" or something along those lines

pjenvey (dismissed):
looks good but let's also clarify the old comments

syncstorage-postgres/src/db/db_impl.rs
taddes: Cool, I just rewrote them to give some context to the `FOR SHARE` vs `FOR UPDATE` lock types.

PR #1892, opened 2025-11-14
feat: add script to purge expired items in Postgres

chenba:
## Description

Add a simple util Python script to purge expired items from Postgres.  

## Testing

`poetry -C tools/postgres run pytest test_purge_ttl.py -v`

## Issue(s)

closes STOR-54


tools/postgres/pyproject.toml
```diff
@@ -0,0 +1,24 @@
...
+    {name = "Mozilla"}
+]
+license = "Mozilla Public License Version 2.0"
+requires-python = ">=3.9.2"
+
+[tool.poetry]
+package-mode = false
+
+[tool.poetry.dependencies]
+sqlalchemy = "^1.4.46"
```
chenba, line 15: I didn't use 2.x here to be consistent with the rest of the repo.

.circleci/config.yml
```diff
@@ -778,6 +790,44 @@ jobs:
...
+          password: $DOCKER_PASS
+    steps:
+      - setup_remote_docker
+      - checkout
+      - display-versions
+      - write-version
+      - run:
+          name: Build and deploy to Dockerhub
+          command: |
+            export UTILS_DOCKERHUB_REPO=mozilla/sync-postgres-py-utils
```
chenba: This Docker Hub repo does not exist at the moment.  I've filed a support ticket to have it created.

chenba: We cannot merge this PR until that Docker Hub repo is created.

tools/postgres/pyproject.toml
taddes: Yeah, good call. The refactors required didn't warrant the commitment given our other priorities a while ago.

.circleci/config.yml
taddes: Yeah, since we only have the `sync-spanner-py-utils` now after this change. I can't help but wonder if we should rename that Docker Hub repo to something more generic and they all sit there, opposed to a separate one still. But for now it's probs easiest to just still make a new repo like you've requested. I guess it's also the idea of shipping only one or the other, which makes separation sensible.

chenba: I learned that we are moving away from Docker Hub.  So I'll remove this part of the PR and file a jira for moving to Github Actions for publishing images to GAR and GHCR.

chenba:
The Dockfile has been updated so that the script (along with the new dir) is now copied into the Postgres variant of the docker images.  However, `docker build` with `--build-arg SYNCSTORAGE_DATABASE_BACKEND=postgres` isn't supported at this time as we are still working on the Postgres impl of Sync server, and, that Docker image isn't published anywhere at this point.  (Only the Spanner variant is pushed to Docker Hub.)

taddes:
> The Dockfile has been updated so that the script (along with the new dir) is now copied into the Postgres variant of the docker images. However, `docker build` with `--build-arg SYNCSTORAGE_DATABASE_BACKEND=postgres` isn't supported at this time as we are still working on the Postgres impl of Sync server, and, that Docker image isn't published anywhere at this point. (Only the Spanner variant is pushed to Docker Hub.)

Makes sense, that'll work. We'll just make note to update the docs once that's in place and when we publish the Docker image. Thx!

tools/postgres/purge_ttl.py
```diff
@@ -0,0 +1,196 @@
...
+    """Create a SQLAlchemy engine from a database url."""
+    parsed = urlparse(database_url)
+    if parsed.scheme not in ['postgresql', 'postgres']:
+        raise ValueError(f"Unsupported database scheme: {parsed.scheme}")
+
+    # newer versions of SQLAlchemy want 'postgresql' instead of 'postgres'
+    if parsed.scheme == 'postgres':
+        database_url = database_url.replace('postgres://', 'postgresql://', 1)
+
+    return sqlalchemy.create_engine(database_url)
```
pjenvey: parsed url has a nice method for this
```suggestion
    # newer versions of SQLAlchemy want 'postgresql' instead of 'postgres'
    if parsed.scheme == 'postgres':
        parsed = parsed._replace(scheme='postgresql')

    return sqlalchemy.create_engine(parsed.geturl())
```

pjenvey (changes requested):

PR #1899, opened 2025-11-24
feat: add get_bsos/bso_ids

pjenvey:
- switch sortindex to 32 bit: the API defines it as 9 digits max
- switch payload to TEXT

Closes STOR-336

### DIFF
--- a/syncserver/Cargo.toml
+++ b/syncserver/Cargo.toml
@@ -57,6 +57,9 @@ validator = "0.20"
 validator_derive = "0.20"
 woothee = "0.13"
 
+[target.'cfg(target_os = "linux")'.dependencies]
+slog-journald = "2.2.0"
+
 [features]
 default = ["mysql", "py_verifier"]
 no_auth = []
--- a/syncserver/src/logging.rs
+++ b/syncserver/src/logging.rs
@@ -1,9 +1,29 @@
+#[cfg(target_os = "linux")]
+use std::fs;
+
 use std::io;
+
+#[cfg(target_os = "linux")]
+use std::os::linux::fs::MetadataExt;
 
 use crate::error::ApiResult;
 
 use slog::{self, slog_o, Drain};
 use slog_mozlog_json::MozLogJson;
+
+#[cfg(target_os = "linux")]
+fn connected_to_journal() -> bool {
+    fs::metadata("/dev/stderr")
+        .map(|meta| format!("{}:{}", meta.st_dev(), meta.st_ino()))
+        .ok()
+        .and_then(|stderr| std::env::var_os("JOURNAL_STREAM").map(|s| s == stderr.as_str()))
+        .unwrap_or(false)
+}
+
+#[cfg(not(target_os = "linux"))]
+fn connected_to_journal() -> bool {
+    false
+}
 
 pub fn init_logging(json: bool) -> ApiResult<()> {
     let logger = if json {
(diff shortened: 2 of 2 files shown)


syncstorage-postgres/src/db/db_impl.rs
```diff
@@ -135,14 +136,41 @@ impl Db for PgDb {
...
+            bsos::expiry,
+        );
+        let (bsos, next_offset) = bsos_query!(self, params, selection, GetBso);
+        let items = bsos
+            .into_iter()
+            .map(|bso| results::GetBso {
+                id: bso.bso_id,
+                modified: SyncTimestamp::from_milliseconds(
+                    bso.modified.and_utc().timestamp_millis() as u64,
+                ),
```
pjenvey: We can cleanup these to/from conversions later, especially since we're still discussing use of `TIMESTAMP` types in the other PR.

syncstorage-postgres/src/db/mod.rs
```diff
@@ -162,3 +162,79 @@ impl PgDb {
             .collect()
     }
 }
+
+#[macro_export]
```
pjenvey, line 166: I'm back and forth on having a macro for this.

I attempted a macro as it could be potentially be utilized for mysql as well, but there are some differences that need addressing first: the different naming of the fields in schema.rs (which is easy to change via `#[sql_name=]`) but also differing types, probably just the use of `TIMESTAMP`s

syncstorage-postgres/src/db/db_impl.rs
```diff
@@ -241,3 +252,31 @@ struct IdResult {
...
+    pub sortindex: Option<i32>,
+    #[diesel(sql_type = Timestamp)]
+    pub expiry: NaiveDateTime,
+}
+
+impl From<GetBso> for results::GetBso {
+    fn from(pg: GetBso) -> Self {
+        Self {
+            id: pg.bso_id,
+            modified: SyncTimestamp::from_milliseconds(
```
pjenvey, line 274: We can cleanup these to/from conversions later, especially since we're still discussing use of TIMESTAMP types in the other PR.

taddes: Yeah, about that: should we just keep the integer types across the board for conversion purposes? If we end up just having to cast timestamps to `i64` in the end, I wonder if it's even worth it.

chenba: What's wrong with conversions?

syncstorage-postgres/src/db/mod.rs
taddes: Yeah, there's the whole situation of constants remapping the field values from their legacy values in my MySQL implementation.

syncstorage-postgres/src/db/db_impl.rs
taddes: > What's wrong with conversions?

This is more a general comment, but there have been times we've gotten into the idea of converting or changing a type, but the signature of other methods or external crates that accept that changed type require a specific type (say `i64`). Then there are a tons of rewrites and conversion impls only to have to convert back to the same original type it was. At those points we've just abandoned the idea, or just added a ton of code with little benefit. This is not always the case, but has applied in the past with Sync and Push when dealing with some datetime values.

taddes (approved):

syncstorage-postgres/src/db/db_impl.rs
chenba: Sounds like you are referring to casting within Rust?  If an external crate need negative values to represent the past before the unix epoch but our values do not have that requirement, e.g., then casting is required.  And that seems fine?  Sure I can understand the cost/benefit tradeoff in updating existing, working code.  But this is not that though?

I was referring converting between Postgres and the service(s).  Storing timestamps as timestamps has its advantages.  We can use the Postgres functions and operators that work with timestamps.  Or maybe use extensions or new features or whatever that operates on timestamps.

We shouldn't assume we'd only ever work with these values through Rust and/or Python, and rely on whatever language to assign meaning to some numbers.  Storing the type along with the data is always preferable to me.

chenba:
This also closes STOR-402 no?

PR #1900, opened 2025-11-25
feat: impl for update collection

taddes:
## Description

To support a number of the bso methods, we need to first implement the update method, since it is a dependency for those methods.

PostgreSQL does not have the UPSERT statement as we would see in Spanner, nor is the MySQL `ON DUPLICATE KEY UPDATE`, but it supports the upsert operation via the `INSERT...ON CONFLICT` statement, making use of the `EXCLUDED` keyword.

Note: there is a stub for quota enforcement with a note. We have this down to work on later should we wish to support this functionality. At present, a standard `GetQuotaUsage` struct is created and passed. Just wanted to note this @pjenvey as we had discussed in last meeting

## Issue(s)

Closes [STOR-407](https://mozilla-hub.atlassian.net/browse/STOR-407).

[STOR-407]: https://mozilla-hub.atlassian.net/browse/STOR-407?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/syncstorage-postgres/src/db/db_impl.rs
+++ b/syncstorage-postgres/src/db/db_impl.rs
@@ -4,7 +4,7 @@ use async_trait::async_trait;
 use chrono::NaiveDateTime;
 use diesel::{
     sql_query,
-    sql_types::{BigInt, Integer, Nullable, Text, Timestamp},
+    sql_types::{Integer, Nullable, Text, Timestamp},
     ExpressionMethods, OptionalExtension, QueryDsl,
 };
 use diesel_async::{AsyncConnection, RunQueryDsl, TransactionManager};
@@ -15,8 +15,11 @@ use syncstorage_settings::Quota;
 
 use super::PgDb;
 use crate::{
-    bsos_query, db::CollectionLock, pool::Conn, schema::bsos, schema::user_collections, DbError,
-    DbResult,
+    bsos_query,
+    db::CollectionLock,
+    pool::Conn,
+    schema::{bsos, user_collections},
+    DbError, DbResult,
 };
 
 #[async_trait(?Send)]
@@ -328,31 +331,26 @@ impl Db for PgDb {
             total_bytes: 0,
             count: 0,
         };
-
-        // This method is an upsert operation, which allows the update of an existing row
-        // or inserts a new one if it doesn’t exist. Postgres does not have `UPSERT` but
-        // achieves this though `INSERT...ON CONFLICT`.
-        let upsert = r#"
-                INSERT INTO user_collections (user_id, collection_id, modified, count, total_bytes)
-                VALUES ($1, $2, $3, $4, $5)
-                    ON CONFLICT (user_id, collection_id)
-                    DO UPDATE SET
-                         modified = EXCLUDED.modified,
-                         total_bytes = EXCLUDED.total_bytes,
-                         count = EXCLUDED.count
-        "#;
         let total_bytes = quota.total_bytes as i64;
         let sync_ts = self.timestamp();
-
         // Convert SyncTimestamp -> NaiveDateTime using the new method
-        let modified: NaiveDateTime = sync_ts.as_naive_datetime().map_err(DbError::from)?; // adjust if DbError conversion is different
+        let modified: NaiveDateTime = sync_ts.as_naive_datetime()?;
 
-        sql_query(upsert)
-            .bind::<BigInt, _>(params.user_id.legacy_id as i64)
-            .bind::<Integer, _>(&params.collection_id)
-            .bind::<Timestamp, _>(&modified)
-            .bind::<Integer, _>(&quota.count)
-            .bind::<BigInt, _>(&total_bytes)
+        diesel::insert_into(user_collections::table)
+            .values((
+                user_collections::user_id.eq(params.user_id.legacy_id as i64),
+                user_collections::collection_id.eq(params.collection_id),
+                user_collections::modified.eq(modified),
+                user_collections::count.eq(quota.count as i64),
+                user_collections::total_bytes.eq(total_bytes),
+            ))
+            .on_conflict((user_collections::user_id, user_collections::collection_id))
+            .do_update()
+            .set((
+                user_collections::modified.eq(modified),
+                user_collections::total_bytes.eq(total_bytes),
+                user_collections::count.eq(quota.count as i64),
+            ))
             .execute(&mut self.conn)
             .await?;
         Ok(self.timestamp())


taddes:
Lmk if this looks ok @pjenvey, given the context of our convo with the quota enforcement, or if there needs to be support for it. Thx!

syncstorage-postgres/src/db/db_impl.rs
```diff
@@ -1,8 +1,9 @@
-#![allow(unused_variables)] // XXX:
+#![allow(unused_variables)]
+// XXX:
```
chenba, line 2: ...okay time for me to admit that I don't know this comment convention.  what action is needed?

syncstorage-postgres/src/db/db_impl.rs
```diff
@@ -207,11 +208,44 @@ impl Db for PgDb {
...
+                         modified = EXCLUDED.modified,
+                         total_bytes = EXCLUDED.total_bytes,
+                         count = EXCLUDED.count
+        "#;
+        let total_bytes = quota.total_bytes as i64;
+        let timestamp = self.timestamp().as_i64();
+        sql_query(upsert)
+            .bind::<BigInt, _>(params.user_id.legacy_id as i64)
+            .bind::<Integer, _>(&params.collection_id)
+            .bind::<BigInt, _>(&timestamp)
```
chenba: `modified` is a timestamp so we cannot bind an i64 here.

syncstorage-db-common/src/util.rs
```diff
@@ -132,6 +132,19 @@ impl From<SyncTimestamp> for u64 {
...
 }
 
+impl From<SyncTimestamp> for NaiveDateTime {
+    fn from(ts: SyncTimestamp) -> NaiveDateTime {
+        let millis = ts.as_i64();
+        let secs = millis / 1000;
+        let nsecs = ((millis % 1000) * 1_000_000) as u32;
+
+        // chrono 0.5 deprecation-free path (straight NaiveDateTime doesn't work):
+        DateTime::<Utc>::from_timestamp(secs, nsecs)
```
chenba: There is a [`from_timestamp_millis`](https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.from_timestamp_millis)

syncstorage-db-common/src/util.rs
```diff
@@ -132,6 +132,19 @@ impl From<SyncTimestamp> for u64 {
...
+            .expect("timestamp out of range")
```
chenba: I think this panics?  Probably safer if we can return a `Result`.

taddes:
Went for a somewhat simpler approach here given most all the method conversion calls for the `SyncTimestamp` are util methods that convert. Made it a little less obtuse, I think. Created a `as_naive_datetime` method to just convert. That way the err handling is a little less crappy. @pjenvey @chenba

syncstorage-db-common/src/util.rs
```diff
@@ -112,6 +112,29 @@ impl SyncTimestamp {
...
+        let millis = self.as_i64();
+        let secs = millis / 1000;
+        let nsecs: u32 = ((millis % 1000) * 1_000_000).try_into().map_err(|e| {
+            SyncstorageDbError::internal(format!(
+                "Invalid timestamp (nanoseconds) {}: {}",
+                millis, e
+            ))
+        })?;
+
+        let ts = Utc.timestamp_opt(secs, nsecs);
```
chenba: I think that's the older fn @pjenvey mentioned in our meeting?  Maybe he can remind us why we shouldn't use it.

syncstorage-db-common/src/util.rs
```diff
@@ -112,6 +112,29 @@ impl SyncTimestamp {
...
+
+    /// Convert this SyncTimestamp into a chrono::NaiveDateTime (UTC) for use with Diesel's `Timestamp`.
+    pub fn as_naive_datetime(self) -> Result<NaiveDateTime, SyncstorageDbError> {
+        let millis = self.as_i64();
+        let secs = millis / 1000;
+        let nsecs: u32 = ((millis % 1000) * 1_000_000).try_into().map_err(|e| {
+            SyncstorageDbError::internal(format!(
+                "Invalid timestamp (nanoseconds) {}: {}",
+                millis, e
+            ))
```
chenba: some int mod 1000 then multiply 1,000,000 should fit into a `u32`.  It can be a simple cast.

syncstorage-db-common/src/util.rs
```diff
@@ -112,6 +112,29 @@ impl SyncTimestamp {
...
+
+        if let Some(dt) = ts.single() {
+            Ok(dt.naive_utc())
+        } else {
+            Err(SyncstorageDbError::internal(format!(
+                "Invalid or ambiguous timestamp {}: {:?}",
+                millis, ts
+            )))
+        }
+    }
```
chenba, line 122: I think `from_timestamp_millis` is still simpler, unless there's something I'm just not understanding here.

chenba: @pjenvey also suggested implementing `to_sql` but I have not looked into that.

syncstorage-postgres/src/db/db_impl.rs
pjenvey: It's a general TODO here or a warning/heads up. We initialized these new modules with some structs largely copied from the existing backends, so there was a lot of dead/unused code, requiring `#![allow(unused_variables)]` etc. to quiet the compiler warnings those produce. We should kill these as soon as it's possible to

syncstorage-postgres/src/db/db_impl.rs
```diff
@@ -313,11 +314,48 @@ impl Db for PgDb {
...
+                    DO UPDATE SET
+                         modified = EXCLUDED.modified,
+                         total_bytes = EXCLUDED.total_bytes,
+                         count = EXCLUDED.count
+        "#;
+        let total_bytes = quota.total_bytes as i64;
+        let sync_ts = self.timestamp();
+
+        // Convert SyncTimestamp -> NaiveDateTime using the new method
+        let modified: NaiveDateTime = sync_ts.as_naive_datetime().map_err(DbError::from)?; // adjust if DbError conversion is different
```
pjenvey: Does this already convert for us?
```suggestion
        let modified: NaiveDateTime = sync_ts.as_naive_datetime()?;
```

syncstorage-postgres/src/db/db_impl.rs
```diff
@@ -313,11 +314,48 @@ impl Db for PgDb {
...
+            count: 0,
+        };
+
+        // This method is an upsert operation, which allows the update of an existing row
+        // or inserts a new one if it doesn’t exist. Postgres does not have `UPSERT` but
+        // achieves this though `INSERT...ON CONFLICT`.
+        let upsert = r#"
+                INSERT INTO user_collections (user_id, collection_id, modified, count, total_bytes)
+                VALUES ($1, $2, $3, $4, $5)
+                    ON CONFLICT (user_id, collection_id)
```
pjenvey: If we're going to prefer the ORM this is actually supported by it on postgres, here's some examples: https://docs.diesel.rs/2.1.x/diesel/upsert/struct.IncompleteOnConflict.html#method.do_update

syncstorage-db-common/src/util.rs
pjenvey: I haven't used `ToSql` but it's basically the inverse of `FromSql` which we use below. I'm not 100% sure it would work as an argument to bind (e.g. `.bind<SyncTimestamp>()`) but in theory it should work for that as well. There aren't a lot of great examples of its usage other than [use within a struct](https://github.com/diesel-rs/diesel/issues/1783).

 Feel free to *not* use it, just throwing it out there as a potential option.

syncstorage-db-common/src/util.rs
pjenvey: I think that was a different case when converting in the opposite direction, where chrono forces usage of `and_utc` when converting from a `NaiveDateTime` before calling e.g. `timestamp_millis` (where `NaiveDateTime::timestamp_millis` is deprecated)

syncstorage-db-common/src/util.rs
```diff
@@ -112,6 +112,29 @@ impl SyncTimestamp {
     pub fn as_rfc3339(self) -> Result<String, SyncstorageDbError> {
         to_rfc3339(self.as_i64())
     }
+
+    /// Convert this SyncTimestamp into a chrono::NaiveDateTime (UTC) for use with Diesel's `Timestamp`.
+    pub fn as_naive_datetime(self) -> Result<NaiveDateTime, SyncstorageDbError> {
+        let millis = self.as_i64();
```
pjenvey: I think we can use chrono for all this: basically `DateTime::from_timestamp_millis(self.as_i64())` then `naive_utc()`.

syncstorage-postgres/src/db/db_impl.rs
chenba: Why not literally writing `TODO`?  What's the `XXX` for?  Some tool?

pjenvey: Good question: I guess I use XXX as more immediate/higher priority warnings, that this should be taken care of sooner rather than later, if even committed. Whereas TODOs, which shouldn't live forever either, unfortunately end up doing just that sometimes :)

taddes: I've always liked this as it's easy to use the IDE to find stuff that we've forgotten about. Already took a handful of these out as we've gone along. They do indeed sometimes stay... forever 😆

chenba: > I've always liked this as it's easy to use the IDE to find stuff

Interesting.  I know that some tools (e.g. Visual Studio (but not VS Code)) has built-in support to build a task list from TODOs in the code base.  But I was not aware of the XXX convention.

syncstorage-db-common/src/util.rs
taddes: Yes, easier indeed.

syncstorage-db-common/src/util.rs
taddes: Yeah, going with the simpler approach using `from_timestamp_millis` with a map, like you did so in #1901

syncstorage-db-common/src/util.rs
taddes: Yeah, opting not to in this case as it's easy enough to call the conversion and it gets used a bunch, plus that the `SyncTimestamp` has a lot of handy utility methods, keeps its interface somewhat consistent.

syncstorage-db-common/src/util.rs
taddes: Yeah, just went for the following approach instead:

```rust
    pub fn as_naive_datetime(self: SyncTimestamp) -> Result<NaiveDateTime, SyncstorageDbError> {
        chrono::DateTime::from_timestamp_millis(self.as_i64())
            .map(|dt| dt.naive_utc())
            .ok_or_else(|| SyncstorageDbError::internal("Invalid timestamp".to_owned()))
    }
}
```

syncstorage-postgres/src/db/db_impl.rs
taddes: Yeah, I wouldn't say we're going all in on the ORM methods, since I still like the raw queries like you do. I guess keeping it consistent amongst some of the implementations, where it makes sense. This still seems simple enough as a raw query, right?

syncstorage-postgres/src/db/db_impl.rs
chenba: @taddes wanna accept this suggestion?

syncstorage-postgres/src/db/db_impl.rs
chenba: > This still seems simple enough as a raw query, right?

I think we'd want to use raw query when the query is not simple or even impossible to express in Diesel.  Since this is simple enough then let's use ORM methods.

taddes: Got it, using the orm query.

pjenvey (approved):

chenba (approved):

Issue #1766, opened 2025-08-20
Create delete methods (postgres)


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/STOR-338)

PR #1901, opened 2025-11-25
feat: add delete fns for Postgres db impls

chenba:
## Description

Add the `delete_*` fns for Postgres.

## Testing

`RUST_TEST_THREADS=1 cargo test --no-default-features --features=syncstorage-db/postgres --features=tokenserver-db/postgres --features=py_verifier` should only show failures from functions with `todo!()`

## Issue(s)

Closes STOR-338 / https://github.com/mozilla-services/syncstorage-rs/issues/1766

### DIFF
--- a/syncstorage-postgres/src/db/db_impl.rs
+++ b/syncstorage-postgres/src/db/db_impl.rs
@@ -1,7 +1,7 @@
 #![allow(unused_variables)]
 // XXX:
 use async_trait::async_trait;
-use chrono::{DateTime, NaiveDateTime};
+use chrono::NaiveDateTime;
 use diesel::{
     delete, sql_query,
     sql_types::{Integer, Nullable, Text, Timestamp},
@@ -301,9 +301,7 @@ impl Db for PgDb {
     async fn delete_bso(&mut self, params: params::DeleteBso) -> DbResult<results::DeleteBso> {
         let user_id = params.user_id.legacy_id;
         let collection_id = self.get_collection_id(&params.collection).await?;
-        let naive_datetime = DateTime::from_timestamp(self.timestamp().as_i64() / 1000, 0)
-            .ok_or_else(|| DbError::internal("Invalid timestamp".to_owned()))?
-            .naive_utc();
+        let naive_datetime = self.timestamp().as_naive_datetime()?;
         let affected_rows = delete(bsos::table)
             .filter(bsos::user_id.eq(user_id as i64))
             .filter(bsos::collection_id.eq(&collection_id))


syncstorage-postgres/src/db/mod.rs
```diff
@@ -161,6 +164,24 @@ impl PgDb {
             })
             .collect()
     }
+
+    async fn erect_tombstone(&mut self, user_id: i64) -> DbResult<()> {
+        let naive_datetime = DateTime::from_timestamp(self.timestamp().as_i64() / 1000, 0)
```
pjenvey: these should move to the more accurate `as_naive_datetime` from the other PR when that lands

syncstorage-postgres/src/db/mod.rs
```diff
@@ -161,6 +164,24 @@ impl PgDb {
             })
             .collect()
     }
+
+    async fn erect_tombstone(&mut self, user_id: i64) -> DbResult<()> {
```
taddes, line 167: I see in this one that we're not using the raw query like in MySQL, which is fine as I think this remains clear. 

My concern here is calling `DateTime::from_timestamp` as we pass in seconds, opposed to miliseconds, so we have reduced precision. Perhaps we can use the `from_timestamp_millis` like in the other PR you mentioned it in.

syncstorage-postgres/src/db/mod.rs
```diff
@@ -161,6 +164,28 @@ impl PgDb {
...
+            .on_conflict((user_collections::user_id, user_collections::collection_id))
+            .do_update()
+            .set(user_collections::modified.eq(naive_datetime))
+            .execute(&mut self.conn)
+            .await?;
+        Ok(())
+    }
+}
+
+pub fn timestamp_to_naive_datetime(timestamp_ms: i64) -> DbResult<NaiveDateTime> {
```
chenba: This function is now in both here and https://github.com/mozilla-services/syncstorage-rs/pull/1903.  I can update whichever PR gets merged second to ensure we don't have a dup.

We could also consider moving the fn to syncstorage-db-common/src/util.rs.  Although the `DbResult` here is a crate level type.  I don't know why syncstorage-mysql and syncstorage-postgres each have their own `DbResult` type though; maybe we can consolidate that as well.

chenba: Right, Taddes added a fn to `SyncTimestamp` for this conversion.  I like that approach so we'll do that.

taddes: Sounds good @chenba , and like your suggestion to have the function live in syncstorage-db-common/src/util.rs., that's where the conversion function in #1900 lives. I'll make the updates there and we can land this and that one around the same time. 

`DbResult` is just a generic type alias defined in `lib.rs` that captures the common `syncstorage_db_common::diesel::DbError` used by `DbResult`, whether in Spanner, Postgres or MySQL. It's defined in the common db crate already. We have to create the type alias for  the `DbResult` at the crate level in `lib.rs` for each of those implementations to access, so it's already consolidated technically. But good to observe where we can consolidate, regardless!

taddes (approved):
We can coordinate between this and #1900 landing, but either works for me! Just let me know your preference.

syncstorage-postgres/src/db/db_impl.rs
```diff
@@ -262,11 +298,28 @@ impl Db for PgDb {
...
-        &mut self,
-        params: params::DeleteBso,
-    ) -> Result<results::DeleteBso, Self::Error> {
-        todo!()
+    async fn delete_bso(&mut self, params: params::DeleteBso) -> DbResult<results::DeleteBso> {
+        let user_id = params.user_id.legacy_id;
+        let collection_id = self.get_collection_id(&params.collection).await?;
+        let naive_datetime = DateTime::from_timestamp(self.timestamp().as_i64() / 1000, 0)
+            .ok_or_else(|| DbError::internal("Invalid timestamp".to_owned()))?
+            .naive_utc();
```
pjenvey:
```suggestion
        let naive_datetime = self.timestamp().as_naive_datetime()?;
```

pjenvey (changes requested):

syncstorage-postgres/src/db/db_impl.rs
chenba: 🤦 I thought I reviewed the changes after rebasing and resolving conflicts.  Thanks for catching that.

pjenvey (dismissed):

pjenvey (approved):

Issue #1769, opened 2025-08-20
Create Batch get and delete (postgres)


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/STOR-341)

Issue #1768, opened 2025-08-20
Create Batch create, commit and append logic (postgres)


data-sync-user:
The `BatchDb` requires the following trait bounds to be satisfied with the batch logic for:
`create_batch`
`commit_batch`
`append_to_batch`

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/STOR-340)

PR #1903, opened 2025-12-01
feat: impl {create,commit,append_to,validate,get,delete}_batch for Postgres

chenba:
## Description

Add the `{create,commit,append_to,validate,get,delete}_batch` `BatchDb` impl for Postgres.  And change the `batch_id` Postgres columns to `uuid`. 

## Testing

No new test failures as a result of this patch.

## Issue(s)

Closes STOR-340 / https://github.com/mozilla-services/syncstorage-rs/issues/1768
Closes STOR-341 / https://github.com/mozilla-services/syncstorage-rs/issues/1769

### DIFF
--- a/syncstorage-postgres/src/orm_models.rs
+++ b/syncstorage-postgres/src/orm_models.rs
@@ -12,8 +12,8 @@ pub struct BatchBso {
     pub collection_id: i32,
     pub batch_id: Uuid,
     pub batch_bso_id: String,
-    pub sortindex: Option<i64>,
-    pub payload: Option<Vec<u8>>,
+    pub sortindex: Option<i32>,
+    pub payload: Option<String>,
     pub ttl: Option<i64>,
 }
 
@@ -33,8 +33,8 @@ pub struct Bso {
     pub user_id: i64,
     pub collection_id: i32,
     pub bso_id: String,
-    pub sortindex: Option<i64>,
-    pub payload: Vec<u8>,
+    pub sortindex: Option<i32>,
+    pub payload: String,
     pub modified: NaiveDateTime,
     pub expiry: NaiveDateTime,
 }
--- a/syncstorage-postgres/syncstorage_postgres_db.md
+++ b/syncstorage-postgres/syncstorage_postgres_db.md
@@ -33,7 +33,7 @@ Stores actual records being synced — Basic Storage Objects.
 | `collection_id` | `INTEGER`    | Maps to a named collection. PK (part 2) & FK (part 2) to `user_collections`                           |
 | `bso_id`        | `TEXT`      | Unique ID within a collection. PK (part 4) |
 | `sortindex`     | `BIGINT`    | Indicates record importance for syncing (optional) |
-| `payload`       | `BYTEA`     | Bytes payload (e.g. JSON blob)                     |
+| `payload`       | `TEXT`      | Bytes payload (e.g. JSON blob)                     |
 | `modified`      | `TIMESTAMP` | Auto-assigned modification timestamp               |
 | `expiry`        | `TIMESTAMP` | TTL as absolute expiration time (optional)         |
 
@@ -63,8 +63,8 @@ Temporary table for staging batch uploads before final commit.
 | --------------- | ----------- | ------------------------------------------------- |
 | `user_id`       | `BIGINT`    | The user id (assigned by Tokenserver), FK (part 1) to `user_collections` |
 |
-| `collection_id` | `INTEGER`    | Maps to a named collection. PK (part 2) & FK (part 2) to `user_collections` |
-| `batch_id`      | `TEXT`      | Client-generated or server-assigned batch ID. PK (part 3)  |
+| `collection_id` | `INTEGER`   | Maps to a named collection. PK (part 2) & FK (part 2) to `user_collections` |
+| `batch_id`      | `UUID`      | Client-generated or server-assigned batch ID. PK (part 3)  |
 | `expiry`        | `TIMESTAMP` | Time at which batch is discarded if not committed |
 
 Indexes:
@@ -75,12 +75,12 @@ Stores BSOs during a batch upload, not yet committed to bsos.
 
 | Column          | Type     | Description                 |
 | --------------- | -------- | --------------------------- |
-| `user_id`       | `BIGINT` | FK to `batches`          |
-| `collection_id` | `INTEGER` | FK to `batches`             |
-| `batch_id`      | `TEXT`   | FK to `batches`             |
+| `user_id`       | `BIGINT` | FK to `batches`             |
+| `collection_id` | `INTEGER`| FK to `batches`             |
+| `batch_id`      | `UUID`   | FK to `batches`             |
 | `batch_bso_id`  | `TEXT`   | Unique ID within batch      |
 | `sortindex`     | `BIGINT` | Optional, for sort priority |
-| `payload`       | `BYTEA`  | Bytes payload               |
+| `payload`       | `TEXT`   | Payload                     |
 | `ttl`           | `BIGINT` | Time-to-live in seconds     |
 
 ## Database Diagram and Relationship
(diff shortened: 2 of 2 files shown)


syncstorage-postgres/src/db/batch_impl.rs
```diff
@@ -40,14 +121,129 @@ impl BatchDb for PgDb {
...
+        let default_ttl_seconds = DEFAULT_BSO_TTL as i64;
+
+        sql_query(
+            "INSERT INTO bsos (user_id, collection_id, bso_id, sortindex, payload, modified, expiry)
+             SELECT
+                 $1::BIGINT,
+                 $2::INTEGER,
+                 batch_bso_id,
+                 sortindex,
+                 COALESCE(payload, ''::BYTEA),
```
chenba: I'll update the patch once https://github.com/mozilla-services/syncstorage-rs/pull/1899 is merged; that patch updated the payload to TEXT.

syncstorage-postgres/src/db/batch_impl.rs
```diff
@@ -12,42 +27,234 @@ impl BatchDb for PgDb {
...
         &mut self,
         params: params::CreateBatch,
-    ) -> Result<results::CreateBatch, Self::Error> {
-        todo!()
+    ) -> DbResult<results::CreateBatch> {
+        let batch_id = Uuid::new_v4();
+        let user_id = params.user_id.legacy_id as i64;
+        let collection_id = self.get_or_create_collection_id(&params.collection).await?;
+        let timestamp_ms = self.timestamp().as_i64();
+        let expiry = timestamp_to_naive_datetime(timestamp_ms + BATCH_LIFETIME)?;
```
pjenvey: I've added `SyncTimestamp::as_naive_datetime` in https://github.com/mozilla-services/syncstorage-rs/pull/1912, when that lands I'd say it'd be more expicit/preferable to shift time via chrono
```suggestion
        let expiry = self.timestamp().as_naive_datetime() + chrono::TimeDelta::milliseconds(BATCH_LIFETIME)?;
```

syncstorage-postgres/src/db/batch_impl.rs
```diff
@@ -12,42 +27,234 @@ impl BatchDb for PgDb {
...
+                id: params.id.clone(),
+            })
+            .await?;
+        let batch = if is_valid {
+            Some(results::GetBatch { id: params.id })
+        } else {
+            None
+        };
+
+        Ok(batch)
```
pjenvey: FYI Rust's bool added some nice helpers a little while back:

```suggestion
        Ok(is_valid.then_some(results::GetBatch { id: params.id }))
```

syncstorage-postgres/src/db/batch_impl.rs
```diff
@@ -12,42 +27,234 @@ impl BatchDb for PgDb {
...
+        Ok(timestamp)
     }
 
     async fn delete_batch(
         &mut self,
         params: params::DeleteBatch,
     ) -> Result<results::DeleteBatch, Self::Error> {
-        todo!()
+        let batch_id = Uuid::parse_str(&params.id)
+            .map_err(|e| DbError::internal(format!("Invalid batch_id: {}", e)))?;
```
pjenvey: Can you move this into a `batch_validate_id` fn and fill in the `todo!` call in `PgDbPool` to call it?

(syncstorage-spanner also has the exact same validation function -- we should probably share them, but we can clean that up later)

pjenvey (changes requested):

chenba:
I rebased and pushed before I saw pjenvey's comments.  I'll address those issue in the next push.

syncstorage-postgres/src/db/batch_impl.rs
chenba: Yeah, I can do that.  But there is a small fly in the ointment.  Here we need the batch id returned; the `DbPool` trait's return type is `Result<(), Self::Error>`.  So it wont' be a direct call thru in `PgDbPool` as we'd need to drop the batch id and return `()` instead.

That's how I have it at moment in my working copy.  Let me know what you think.

syncstorage-postgres/src/orm_models.rs
```diff
@@ -9,7 +10,7 @@ use diesel::{Identifiable, Queryable};
 pub struct BatchBso {
     pub user_id: i64,
     pub collection_id: i32,
-    pub batch_id: String,
+    pub batch_id: Uuid,
     pub batch_bso_id: String,
     pub sortindex: Option<i64>,
     pub payload: Option<Vec<u8>>,
```
taddes: This currently won't break anything since the module isn't being used for verifying the structs or queries but the migration defines batch_bsos.payload as `TEXT`, but in `orm_models.rs`, `BatchBso.payload` is Option<Vec<u8>> (bytea). The `commit_batch` method selects `COALESCE(payload, ''::TEXT)` to insert into bsos.payload — if bsos.payload is BYTEA this will fail, so we want to change this to encapsulate a string type. This doesn't have to block, but we should probs change it.

syncstorage-postgres/src/orm_models.rs
```diff
@@ -9,7 +10,7 @@ use diesel::{Identifiable, Queryable};
 pub struct BatchBso {
     pub user_id: i64,
     pub collection_id: i32,
-    pub batch_id: String,
+    pub batch_id: Uuid,
     pub batch_bso_id: String,
     pub sortindex: Option<i64>,
```
taddes: Think here we have to update to 
```suggestion
    pub sortindex: Option<i32>,
```
Since it uses an Int4 in `schema.rs` 
Also not a huge deal

syncstorage-postgres/migrations/2025-10-20-155711_create_schema/up.sql
```diff
@@ -67,7 +67,7 @@ CREATE INDEX batch_expiry_idx ON batches (
 CREATE TABLE batch_bsos (
     user_id BIGINT NOT NULL,
     collection_id INTEGER NOT NULL,
-    batch_id TEXT NOT NULL,
+    batch_id UUID NOT NULL,
```
pjenvey, line 70: FYI this should also be updated in `syncstorage_postgres_db.md`. I forgot to update it on a recent change, so I'm fine fixing it all up later

syncstorage-postgres/src/db/batch_impl.rs
```diff
@@ -16,55 +27,231 @@ impl BatchDb for PgDb {
...
+                user_id: params.user_id.clone(),
+                collection_id,
+                collection: params.collection.clone(),
+            })
+            .await?;
+        let naive_datetime = timestamp.as_naive_datetime()?;
+        let default_ttl_seconds = DEFAULT_BSO_TTL as i64;
+
+        sql_query(
+            "INSERT INTO bsos (user_id, collection_id, bso_id, sortindex, payload, modified, expiry)
```
pjenvey, line 147: I'll note I've been keeping these larger more complicated queries in separate files in the other impls, but I guess this one isn't too bad

syncstorage-postgres/src/orm_models.rs
chenba: Neither `sortindex` nor `payload` was changed in this patch but yeah I can definitely make changes to ensure they match whatever the schema is in main at the moment.  Thanks for catching them.

syncstorage-postgres/migrations/2025-10-20-155711_create_schema/up.sql
chenba: I'll update it.  Thanks.

syncstorage-postgres/src/orm_models.rs
taddes: Yeah totally, I think when we were working on the initial PRS to set up the migrations, we had been thinking we’d use the byte type and whatnot so it’s likely just and artifact from that.

Issue #1703, opened 2025-06-08
Insufficient defaults regarding payload


jdeus:
The defaults regarding payload limits are way too small. As it stands, I can't even sync my bookmarks (~180MB) :

`1749395910820	Sync.ErrorHandler	DEBUG	bookmarks failed: Error: Single record too large to submit to server(resource://services-sync/record.sys.mjs:1201:16) JS Stack trace: enqueue@record.sys.mjs:1201:16
_uploadOutgoing@engines.sys.mjs:1922:53`

Increasing these values to sensible ones would be great.
Also I think we're missing the option to set them via environment variables, a must when running as a container.
And also, a u32 is too small for DEFAULT_MAX_QUOTA_LIMIT, it'd like to specify 10's of GB for this limit. 

Reference : https://github.com/mozilla-services/syncstorage-rs/blob/master/syncstorage-settings/src/lib.rs#L15

Best

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/STOR-294)

jdeus:
Project considered abandoned

data-sync-user:
➤ Philip Jenvey commented:

{quote}As it stands, I can't even sync my bookmarks (~180MB) :{quote}

I believe bumping the limits higher should accommodate this larger size. I would try the batch limits first (prefixed w/ max_total_, as these should be able to go higher than the per request limits and clients will automatically split requests that exceed the individual request limit across multiple requests (via the batch API).

{quote}Also I think we're missing the option to set them via environment variables, a must when running as a container.{quote}

This should be supported already, e.g. SYNC_SYNCSTORAGE__LIMITS__MAX_QUOTA_LIMIT=<value>

Switching max_quota_limit to u64 seems reasonable, thanks.

data-sync-user:
➤ Philip Jenvey commented:

To close this out we’ll move the quota field to u64 and reconsider the default limits for the new postgres backend (keeping in mind mysql’s).

PR #1906, opened 2025-12-02
refactor: use u64 for DEFAULT_MAX_QUOTA_LIMIT

chenba:
## Description

Update the type of DEFAULT_MAX_QUOTA_LIMIT to u64 to allow larger limits

The issue comments also mentioned reconsidering the default limits for Postgres.  But I don't think I have anything to base those values on at this point.  So I did not make any changes for that.

Closes STOR-294 / https://github.com/mozilla-services/syncstorage-rs/issues/1703


taddes (approved):
👍

PR #1911, opened 2025-12-04
feat: create post/put bsos postgres

taddes:
## Description

Implement logic to `put` and `post` to `bsos` table.
## Issue(s)

Closes [STOR-339](https://mozilla-hub.atlassian.net/browse/STOR-339).

[STOR-339]: https://mozilla-hub.atlassian.net/browse/STOR-339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/syncstorage-postgres/src/db/db_impl.rs
+++ b/syncstorage-postgres/src/db/db_impl.rs
@@ -1,7 +1,7 @@
 #![allow(unused_variables)]
 // XXX:
 use async_trait::async_trait;
-use chrono::NaiveDateTime;
+use chrono::{NaiveDateTime, TimeDelta};
 use diesel::{
     delete,
     dsl::{count, max, sql},
@@ -442,7 +442,6 @@ impl Db for PgDb {
     async fn put_bso(&mut self, bso: params::PutBso) -> DbResult<results::PutBso> {
         let collection_id = self.get_or_create_collection_id(&bso.collection).await?;
         let user_id: u64 = bso.user_id.legacy_id;
-        let timestamp = self.timestamp().as_i64();
         if self.quota.enabled {
             let usage = self
                 .get_quota_usage(params::GetQuotaUsage {
@@ -467,20 +466,15 @@ impl Db for PgDb {
         let sortindex = bso.sortindex;
         let ttl = bso.ttl.map_or(DEFAULT_BSO_TTL, |ttl| ttl);
 
-        // original required millisecond conversion
-        let expiry_ts = SyncTimestamp::from_i64(timestamp + (i64::from(ttl) * 1000))?;
-        let modified_ts = SyncTimestamp::from_i64(timestamp)?;
-
-        let expiry_dt = expiry_ts.as_naive_datetime()?;
-        let modified_dt = modified_ts.as_naive_datetime()?;
-
         let modified = self.timestamp().as_naive_datetime()?;
-        let expiry = modified_dt + TimeDelta::seconds(ttl as i64);
+        // Expiry originally required millisecond conversion
+        let expiry = modified + TimeDelta::seconds(ttl as i64);
+
         // The changeset utilizes Diesel's `AsChangeset` trait.
         // This allows selective updates of fields if and only if they are `Some(<T>)`
         let changeset = BsoChangeset {
             sortindex: if bso.sortindex.is_some() {
-                Some(sortindex)
+                sortindex // sortindex is already an Option of type `Option<i32>`
             } else {
                 None
             },
@@ -490,12 +484,12 @@ impl Db for PgDb {
                 None
             },
             expiry: if bso.ttl.is_some() {
-                Some(expiry_dt)
+                Some(expiry)
             } else {
                 None
             },
             modified: if bso.payload.is_some() || bso.sortindex.is_some() {
-                Some(modified_dt)
+                Some(modified)
             } else {
                 None
             },
@@ -507,8 +501,8 @@ impl Db for PgDb {
                 bsos::bso_id.eq(&bso.id),
                 bsos::sortindex.eq(sortindex),
                 bsos::payload.eq(payload),
-                bsos::modified.eq(modified_dt),
-                bsos::expiry.eq(expiry_dt),
+                bsos::modified.eq(modified),
+                bsos::expiry.eq(expiry),
             ))
             .on_conflict((bsos::user_id, bsos::collection_id, bsos::bso_id))
             .do_update()
--- a/syncstorage-postgres/src/orm_models.rs
+++ b/syncstorage-postgres/src/orm_models.rs
@@ -42,7 +42,7 @@ pub struct Bso {
 #[derive(AsChangeset)]
 #[diesel(table_name = bsos)]
 pub struct BsoChangeset<'a> {
-    pub sortindex: Option<Option<i32>>,
+    pub sortindex: Option<i32>,
     pub payload: Option<&'a str>,
     pub modified: Option<NaiveDateTime>,
     pub expiry: Option<NaiveDateTime>,


syncstorage-postgres/src/db/db_impl.rs
```diff
@@ -291,11 +293,29 @@ impl Db for PgDb {
...
+                ttl: pbso.ttl,
+            })
+            .await?;
+        }
+        self.update_collection(params::UpdateCollection {
+            user_id: params.user_id,
+            collection_id,
+            collection: params.collection,
+        })
+        .await?;
```
chenba, line 380: This is already called in `put_bso`.

syncstorage-postgres/src/db/db_impl.rs
```diff
@@ -291,11 +293,29 @@ impl Db for PgDb {
         Ok(results::GetBsoIds { items, offset })
     }
 
-    async fn post_bsos(
-        &mut self,
-        params: params::PostBsos,
-    ) -> Result<results::PostBsos, Self::Error> {
-        todo!()
+    async fn post_bsos(&mut self, params: params::PostBsos) -> DbResult<results::PostBsos> {
+        let collection_id = self.get_or_create_collection_id(&params.collection).await?;
```
chenba, line 361: Since we have the collection id here, let's pass it to `put_bso` instead of repeatedly call `get_or_create_collection` there.

syncstorage-postgres/src/db/db_impl.rs
```diff
@@ -353,7 +355,28 @@ impl Db for PgDb {
     }
 
     async fn post_bsos(&mut self, params: params::PostBsos) -> DbResult<results::PostBsos> {
-        todo!()
+        let collection_id = self.get_or_create_collection_id(&params.collection).await?;
+        let modified = self.timestamp();
+
```
chenba, line 363: Why couldn't we have a quota check here and error out early?

syncstorage-postgres/src/db/db_impl.rs
```diff
@@ -413,8 +436,87 @@ impl Db for PgDb {
...
+            sortindex: if bso.sortindex.is_some() {
+                Some(sortindex)
+            } else {
+                None
+            },
+            payload: if bso.payload.is_some() {
+                Some(payload)
+            } else {
+                None
+            },
```
pjenvey, line 485:
```suggestion
            payload: bso.payload.as_deref(),
```

syncstorage-postgres/src/db/db_impl.rs
```diff
@@ -413,8 +436,87 @@ impl Db for PgDb {
...
+        let payload = bso.payload.as_deref().unwrap_or_default();
+        let sortindex = bso.sortindex;
+        let ttl = bso.ttl.map_or(DEFAULT_BSO_TTL, |ttl| ttl);
+
+        // original required millisecond conversion
+        let expiry_ts = SyncTimestamp::from_i64(timestamp + (i64::from(ttl) * 1000))?;
+        let modified_ts = SyncTimestamp::from_i64(timestamp)?;
+
+        let expiry_dt = expiry_ts.as_naive_datetime()?;
+        let modified_dt = modified_ts.as_naive_datetime()?;
```
pjenvey: I think this is simpler/more explicit going through chrono:

```suggestion
        let modified = self.timestamp().as_naive_datetime()?;
        let expiry = modified_dt + TimeDelta::seconds(ttl as i64);
```

syncstorage-postgres/src/orm_models.rs
```diff
@@ -38,6 +38,15 @@ pub struct Bso {
     pub expiry: NaiveDateTime,
 }
 
+#[derive(AsChangeset)]
+#[diesel(table_name = bsos)]
+pub struct BsoChangeset<'a> {
+    pub sortindex: Option<Option<i32>>,
```
pjenvey: I don't think you're able to actually NULL out `sortindex` on update
```suggestion
    pub sortindex: Option<i32>,
```

syncstorage-postgres/src/db/db_impl.rs
```diff
@@ -413,8 +436,87 @@ impl Db for PgDb {
...
+            expiry: if bso.ttl.is_some() {
+                Some(expiry_dt)
+            } else {
+                None
+            },
+            modified: if bso.payload.is_some() || bso.sortindex.is_some() {
+                Some(modified_dt)
+            } else {
+                None
+            },
```
pjenvey, line 495: nit: you could use map and/or then_some for these
```suggestion
            modified: (bso.payload.is_some() || bso.sortindex.is_some()).then_some(modified),
            expiry: bso.ttl.map(|_| expiry_dt),
```

syncstorage-postgres/src/db/db_impl.rs
taddes: Little confused by this one. A few lines up there's
`let payload = bso.payload.as_deref().unwrap_or_default();` which I wrap in `Some` and return. If we don't call `unwrap_or_default`, we get a compiler error since it expects the `&str`. Speaking of: I see in the SQL version we use a string slice, opposed to a `String`. Might we just want to use a `String` here too? Curious why it's a `&str`. I see in the Spanner impl it's using a `String` type for the option. Any reason not to change to `String`?

syncstorage-postgres/src/db/db_impl.rs
taddes: Makes sense. Will take out the timestamp defined at start of function too since it won't be needed then.

syncstorage-postgres/src/db/db_impl.rs
taddes: I think since it uses `put_bso` checks the quotas immediately as it iterates though the param that's passed to it, it removes the step of having to check?

chenba: > put_bso checks the quotas immediately as it iterates 

Precisely.  That's the issue.  It repeatedly queries the db to check quotas.  Why not do it here, once?

taddes: I'd be in favor of this, but we should look at filing another issue for this one, since it requires modifying the shared `params` struct for `PutBso` that MySQL and Postgres use, so making a change would alter both and require a little extra thought, and might warrant changing the params object for both since . I know we've got some things down to alter across the two implementations and it'd be better to address that in one separate issue imo.

I was also thinking, can't `put_bso` just be called on its own outside of the context of being called by `post_bsos`? In which case, we'd not be passed it. Or do you mean  something else?

chenba: > I know we've got some things down to alter across the two implementations

Are those in an epic?  I want to take a quick look to just get some context.

> since it requires modifying the shared params struct for PutBso 
> can't put_bso just be called on its own outside of the context of being called by post_bsos?

Temporarily set pgdb.quota.enabled to false for `put_bso` during the execution of `post_bsos` solves both.  We don't need to alter `put_bso`.

chenba: @taddes and I chatted about this.  The complicating factor for calculating the new usage is the updates.  So we'll table this for now.

syncstorage-postgres/src/db/db_impl.rs
chenba: We don't need the `update_collection` call here since `put_bso` already call it.

taddes: I think we do actually, otherwise this would be divergent from the MySQL and Postgres implementations.

chenba: How?  Please explain how the state of db is affected by omitting this call.

taddes: [Here](https://github.com/mozilla-services/syncstorage-rs/blob/75af89c0245576d9ecf8a1c628d00335fa32f615/syncstorage-mysql/src/db/db_impl.rs#L545) it is in the MySQL impl. My understanding is you update a user_collection reference when there is a single BSO update (which are facilitated by individual puts in `post_bsos`), and also for the total operation. The reason this matters is updating the `modified` timestamp that the server and client use to know what the last update was and to what collection.

pjenvey: Let's punt on this and the optimization to `get_collection_id` for now as we should try to optimize `post_bsos` first: https://github.com/mozilla-services/syncstorage-rs/issues/1930 which would negate those issues.

syncstorage-postgres/src/db/db_impl.rs
pjenvey: Right, you initially get a `&str` from the `unwrap_or_default` for the insert value. 

Since you still have the original `bso.payload` (and you're even calling `is_some` on it) and you need an `Option<&str>` here for `BsoChangeset`, `bso.payload.as_deref()` is basically equivalent to what you have here but I'd argue simpler, being a one liner.

So it's a nit, but I figured you'd produce a similar line for the `Option<sortindex>` change -- that is, both these `BsoChangeset` assignments could be one liners:

```rust
            sortindex: bso.sortindex,
            payload: bso.payload.as_deref(),
```

diesel accepts both `&str`/`String` in such places. In this specific case, `&str` is preferable with needing `payload` used in two places: switching to `String` would require a clone for the second usage. `payload` is the largest chunk of data we're processing, so we should strive to never clone it.

(I'll note that this query could probably be written to avoid the need for passing the `payload` in twice via `EXCLUDED.payload` on the `ON CONFLICT DO UPDATE SET`, but it's fine for now.. we'll look at that for https://github.com/mozilla-services/syncstorage-rs/issues/1930)

PR #1907, opened 2025-12-02
feat: added mozcloud-publish workflow to create and push container images to GAR when a preview label is added to a pull request

dlactin:
## Description

New GitHub Actions workflow to build and push `syncstorage-rs` container images to GAR when a preview label is added to a pull request. 

Will restrict paths to sync-related directories after testing. 

No changes to the existing CircleCI workflows. 

## Testing

We can add a `preview` label to this pull request to validate the build process.

PR #1924, opened 2025-12-09
[MZCLD-1233][MZCLD-1382] Dockerfile changes to reduce build cache

grahamalama:
## Description
In https://github.com/mozilla-services/syncstorage-rs/pull/1907, we saw the GHA runner running out of disk space due to a a large Docker build cache (~16GB). The goal of this PR is to reduce cache size while maintaining fast successive build times.

  Changes:

  - Consolidated build stages: Merged cacher and builder stages to eliminate duplicate work
  - Added [BuildKit cache mounts](https://docs.docker.com/build/cache/optimize/#use-cache-mounts): Added `--mount=type=cache` for Cargo registry, git, and target directories, keeping dependency caches persistent across builds without inflating layer sizes
  - Pre-built Python wheels: Generate all Python dependency wheels in the builder stage and install from local cache in final stage, eliminating need for build tools in final image
  - Streamlined dependencies: Removed unnecessary packages (golang-go, build-essential, cargo, cmake, libpq-dev) from final stage, keeping only essential runtime libraries

In addition to reducing the build cache, these changes also made the final images smaller

I ran image builds for each variant (MySQL, Postgres, Spanner) before and after the changes in this PR from an empty Docker builder, and these were the results I found:

  | Variant  | Build Cache (Before) | Build Cache (After) |Image Size (Before) | Image Size (After) |
  | -------- | -------------------- | ------------------- | ------------------- | ------------------ |
  | MySQL    | 9.7 GB               | 5.8 GB              |646 MB              | 146 MB             |
  | Postgres | 16.9 GB              | 9.8 GB              | 741 MB              | 240 MB             |
  | Spanner  | 16.9 GB              | 9.2 GB              | 755 MB              | 235 MB             |

I also tried to validate these changes by running the e2e tests locally. When I ran the tests, everything seemed to pass. But we'll see what Circle says.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -473,6 +473,7 @@ jobs:
           name: Build MySQL Docker image
           command: >
             docker build -t app:build
+            --progress=plain
             --build-arg SYNCSTORAGE_DATABASE_BACKEND=mysql
             --build-arg TOKENSERVER_DATABASE_BACKEND=mysql
             .
@@ -511,6 +512,7 @@ jobs:
           name: Build Postgres Docker image
           command: >
             docker build -t app:build
+            --progress=plain
             --build-arg SYNCSTORAGE_DATABASE_BACKEND=spanner
             --build-arg TOKENSERVER_DATABASE_BACKEND=postgres
             .
@@ -550,6 +552,7 @@ jobs:
           # Build w/ the Oracle libmysqlclient-dev for TLS support
           command: >
             docker build -t app:build
+            --progress=plain
             --build-arg SYNCSTORAGE_DATABASE_BACKEND=spanner
             --build-arg MYSQLCLIENT_PKG=libmysqlclient-dev
             .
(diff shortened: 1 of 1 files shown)


chenba (approved):
I'm certainly not qualified to review the GH action, but thank you for improving the Dockerfile!

taddes (approved):
Hey @grahamalama , those are impressive results! Thank you so much for doing this and making all these improvements. You know your way around Docker optimization for sure. This is a huge win for us, especially considering we're going to launch a Docker build for our Postgres support.

Dockerfile
```diff
@@ -12,93 +12,90 @@ FROM chef AS planner
 COPY . .
 RUN cargo chef prepare --recipe-path recipe.json
 
-FROM chef AS cacher
+FROM chef AS builder
```
pjenvey, line 15: I was concerned about removing this cacher step as it was recommended by `cargo chef`, but it seems that was older guidance that was removed a while ago, thanks for catching this! https://github.com/LukeMathWalker/cargo-chef/pull/90

grahamalama:
Thanks folks! I'll merge this in, but you may see a follow up -- I want to do a few successive runs of the build (in the GHA workflow, at least) to make sure caching is working as expected. Though we reduced the amount of disk space used for the build cache, I want to make sure I didn't disrupt efficient layer caching and trade less disk usage for longer build times.

PR #1925, opened 2025-12-10
feat: support query logging via diesel's instrumentation

pjenvey:
- upgrades to latest diesel-async/diesel for diesel-async's new migration harness
- removes diesel_logger

## Issue(s)

Closes STOR-417

### DIFF
--- a/syncstorage-postgres/src/db/db_impl.rs
+++ b/syncstorage-postgres/src/db/db_impl.rs
@@ -1,7 +1,7 @@
 #![allow(unused_variables)]
 // XXX:
 use async_trait::async_trait;
-use chrono::NaiveDateTime;
+use chrono::{NaiveDateTime, TimeDelta};
 use diesel::{
     delete,
     dsl::{count, max, sql},
@@ -11,8 +11,9 @@ use diesel::{
 };
 use diesel_async::{AsyncConnection, RunQueryDsl, TransactionManager};
 use futures::TryStreamExt;
+use std::collections::HashMap;
 use syncstorage_db_common::{
-    error::DbErrorIntrospect, params, results, util::SyncTimestamp, Db, Sorting,
+    error::DbErrorIntrospect, params, results, util::SyncTimestamp, Db, Sorting, DEFAULT_BSO_TTL,
 };
 use syncstorage_settings::Quota;
 
@@ -20,6 +21,7 @@ use super::PgDb;
 use crate::{
     bsos_query,
     db::CollectionLock,
+    orm_models::BsoChangeset,
     pool::Conn,
     schema::{bsos, user_collections},
     DbError, DbResult,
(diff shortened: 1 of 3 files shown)


syncstorage-postgres/src/pool.rs
```diff
@@ -159,8 +145,8 @@ impl DbPool for PgDbPool {
     type Error = DbError;
 
     async fn init(&mut self) -> Result<(), Self::Error> {
-        let database_url = self.database_url.clone();
-        spawn_blocking(move || run_embedded_migrations(&database_url))
+        let conn = self.get_conn().await?;
```
pjenvey: Note: we previously copied the mysql migration setup where it established a connection outside of the pool, which was necessary on mysql because its DDL statements implicitly commit (even when ran within a transaction) which ended up breaking diesel's special `test_transaction` setup by the pool.

I changed this because Postgres didn't need it, but note that its migrations now run in a rolled back transaction, in case you were expecting tests to leave you with a setup db. You might also notice each unit test potentially redefining the schema if it isn't already defined.

pjenvey:
I'll also note this doesn't log the query bind variables as diesel_logger did, this seems to be a limitation of the instrumentation hook

tokenserver-mysql/src/pool.rs
```diff
@@ -1,48 +1,31 @@
...
 use tokenserver_settings::Settings;
-use tokio::task::spawn_blocking;
 
 use crate::db::TokenserverDb;
 
 pub const MIGRATIONS: EmbeddedMigrations = embed_migrations!();
 
 pub(crate) type Conn = Object<AsyncMysqlConnection>;
 
-/// Run the diesel embedded migrations
```
felfert, line 30: Removing the migrations completely is - IMHO - a bad idea. Even if they are currently empty.
Why are they empty anyway? They were not empty at some earlier point in time.
In my packaging, I re-populated them in order to make syncserver startup by its own without manually creating the tables and then added a function that updates the nodes table (node and capacity) from env variables. This is probably the most common usecase for people who want to run their own personal syncserver. This eliminates the need for externally populating the DB before first startup.

See [this patch](https://github.com/felfert/syncstorage-rs-rpm/blob/main/syncstorage-rs-populate-tokendb.patch)  and [this patch](https://github.com/felfert/syncstorage-rs-rpm/blob/main/syncstorage-rs-updatenodes.patch).

felfert (commented):
Please consider keeping migrations of tokenserver-db

tokenserver-mysql/src/pool.rs
pjenvey: This doesn't remove the migrations run, the call is still in `DbPool::init`, the implementation of `run_embedded_migrations` just became generic across backends and moved elsewhere (but they still need to be explicitly enabled in Tokenserver, that hasn't changed).

Thanks for the links and great question, we also just recently noticed the lack of bootstrapping of the `services` table, we're going to address that shortly (the migrations should really do it, they previously did but then this was reversed in a later migration). Setting up the `node` table will likely still be a manual process for now.

taddes (approved):
👍

PR #1927, opened 2025-12-11
feat: UserIdentifier modification in lock and timestamp methods

taddes:
## Description
Update implementation to use `UserIdentifier` within the key of the HashMaps when making comparisons `coll_modified_cache` & `coll_locks` 

## Issue(s)

Closes [STOR-414](https://mozilla-hub.atlassian.net/browse/STOR-414).

[STOR-414]: https://mozilla-hub.atlassian.net/browse/STOR-414?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/syncstorage-mysql/src/db/db_impl.rs
+++ b/syncstorage-mysql/src/db/db_impl.rs
@@ -51,13 +51,12 @@ impl Db for MysqlDb {
                     Err(e)
                 }
             })?;
+
+        let user_id = params.user_id.legacy_id as i64;
+        let key = (params.user_id, collection_id);
         // If we already have a read or write lock then it's safe to
         // use it as-is.
-        if self
-            .session
-            .coll_locks
-            .contains_key(&(params.user_id.clone(), collection_id))
-        {
+        if self.session.coll_locks.contains_key(&key) {
             return Ok(());
         }
 
@@ -65,7 +64,7 @@ impl Db for MysqlDb {
         self.begin(false).await?;
         let modified = user_collections::table
             .select(user_collections::modified)
-            .filter(user_collections::user_id.eq(params.user_id.legacy_id as i64))
+            .filter(user_collections::user_id.eq(user_id))
             .filter(user_collections::collection_id.eq(collection_id))
             .lock_in_share_mode()
             .first(&mut self.conn)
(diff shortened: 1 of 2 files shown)


syncstorage-mysql/src/db/db_impl.rs
```diff
@@ -76,12 +75,13 @@ impl Db for MysqlDb {
...
                 .coll_modified_cache
-                .insert((user_id as u32, collection_id), modified); // why does it still expect a u32 int?
+                .insert((params.user_id.clone(), collection_id), modified); // why does it still expect a u32 int?
         }
         // XXX: who's responsible for unlocking (removing the entry)
-        self.session
-            .coll_locks
-            .insert((user_id as u32, collection_id), CollectionLock::Read);
+        self.session.coll_locks.insert(
+            (params.user_id.clone(), collection_id),
```
pjenvey: You can avoid the final clones since `params` is owned, could you apply that to postgres as well?
```suggestion
            (params.user_id, collection_id),
```

syncstorage-mysql/src/db/db_impl.rs
```diff
@@ -116,11 +116,12 @@ impl Db for MysqlDb {
...
             self.session
                 .coll_modified_cache
-                .insert((user_id as u32, collection_id), modified);
+                .insert((params.user_id.clone(), collection_id), modified);
         }
-        self.session
-            .coll_locks
-            .insert((user_id as u32, collection_id), CollectionLock::Write);
+        self.session.coll_locks.insert(
+            (params.user_id.clone(), collection_id),
```
pjenvey:
```suggestion
            (params.user_id, collection_id),
```

syncstorage-mysql/src/db/db_impl.rs
```diff
@@ -76,12 +75,13 @@ impl Db for MysqlDb {
             let modified = SyncTimestamp::from_i64(modified)?;
             self.session
                 .coll_modified_cache
-                .insert((user_id as u32, collection_id), modified); // why does it still expect a u32 int?
+                .insert((params.user_id.clone(), collection_id), modified); // why does it still expect a u32 int?
```
pjenvey: Very much a nit (cloning `user_id` a couple times is no biggie, and even its strings might be empty on these backends): we can also avoid the clone here because the first `contains_key` usage takes a ref -- if we construct the key up front, then reuse it:
```rust
let key = (params.user_id.clone(), collection_id);
contains_key(&key);
insert(key, modified);
filter(user_id.eq(params.user_id.legacy_id as i64));
insert((params.user_id.clone(), collection_id), Read);
```

Since we need `legacy_id` in the middle of all this, we can go even further, pulling it out early as well then reusing `key` in the last insert (killing another clone I think):

```rust
let user_id = user.params.legacy_id as i64;
let key = (params.user_id, collection_id);
contains_key(&key);
insert(key.clone(), modified);
filter(user_id.eq(user_id));
insert(key, Read);
```

syncstorage-mysql/src/db/db_impl.rs
taddes: Applied to both, good catch

syncstorage-mysql/src/db/db_impl.rs
taddes: Good call. Made the change to both the read and write methods in postgres and sql.

PR #1935, opened 2025-12-15
feat: update collections for standard collections

taddes:
## Description
As we have set up in Spanner and MySQL, we want to explicitly create the standard collections in the `collections` table. 

This includes:
clients
crypto
forms
history
keys
meta
bookmarks
prefs
tabs
passwords
addons
addresses
creditcards

This also ensures that he first 100 slots are reserved for possible future collections, so that all following 100 are user-defined or custom. 

## Issue(s)

Closes [STOR-424](https://mozilla-hub.atlassian.net/browse/STOR-424).

[STOR-424]: https://mozilla-hub.atlassian.net/browse/STOR-424?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/syncstorage-postgres/migrations/2025-10-20-155711_create_schema/up.sql
+++ b/syncstorage-postgres/migrations/2025-10-20-155711_create_schema/up.sql
@@ -43,6 +43,27 @@ CREATE TABLE collections (
     name VARCHAR(32) NOT NULL UNIQUE
 );
 
+-- Insert Standard Collections.
+-- These are the 13 standard collections that are expected to exist by clients.
+-- The IDs are fixed.
+-- Reserved spaces for additions to the standard collections begin after 100.
+INSERT INTO collections (collection_id, name) VALUES
+    ( 1, 'clients'),
+    ( 2, 'crypto'),
+    ( 3, 'forms'),
+    ( 4, 'history'),
+    ( 5, 'keys'),
+    ( 6, 'meta'),
+    ( 7, 'bookmarks'),
+    ( 8, 'prefs'),
+    ( 9, 'tabs'),
+    (10, 'passwords'),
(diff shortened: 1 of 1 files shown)


syncstorage-postgres/migrations/2025-10-20-155711_create_schema/up.sql
```diff
@@ -43,6 +43,28 @@ CREATE TABLE collections (
...
+    ( 8, 'prefs'),
+    ( 9, 'tabs'),
+    (10, 'passwords'),
+    (11, 'addons'),
+    (12, 'addresses'),
+    (13, 'creditcards');
+
+-- Reserve space for additions to the standard collections.
+-- User defined or custom collections should begin after 100.
+INSERT INTO collections (collection_id, name) VALUES (100, '');
```
pjenvey: I had forgotten until now but we never did this with the Spanner impl, instead it incorporates the `FIRST_CUSTOM_COLLECTION_ID` const (which was carried over from the original python impl).

Let's do that instead of writing this bogus collection value to the db that our mysql impl's doing.

taddes: You bet, good call. Looked at the old Python impl as we've discussed and finding a middle ground. Spanner impl differs a bit with some wrapper code for executing the query. I do see Diesel can support a [ sql_function! proc macro](https://diesel.rs/guides/extending-diesel.html) that could be used to define a `coalesce` call, if that's how we do this. Some wiggle room between `create_collection` and `get_or_create_collection_id`

pjenvey (approved):

PR #1938, opened 2025-12-16
Fix: use PRETOUCH_TS when inserting new user collection in Postgres

chenba:
When creating a new user collection, use the "pretouch" timestamp.  And also filter these new collections out appropriately when querying collections.

Closes STOR-425

### DIFF
--- a/syncstorage-postgres/src/db/mod.rs
+++ b/syncstorage-postgres/src/db/mod.rs
@@ -1,4 +1,5 @@
 #![allow(dead_code)] // XXX:
+use chrono::{DateTime, NaiveDate, Utc};
 use diesel::{
     dsl::{now, sql},
     sql_types::BigInt,
@@ -25,6 +26,12 @@ mod db_impl;
 pub use batch_impl::validate_batch_id;
 
 const TOMBSTONE: i32 = 0;
+
+const PRETOUCH_DT: DateTime<Utc> = NaiveDate::from_ymd_opt(1, 1, 1)
+    .unwrap()
+    .and_hms_opt(0, 0, 0)
+    .unwrap()
+    .and_utc();
 
 #[derive(Debug, Eq, PartialEq)]
 enum CollectionLock {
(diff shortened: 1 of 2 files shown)


chenba:
@pjenvey I can update the datetime types here once you merged the TIMESTAMPTZ patch.

syncstorage-postgres/src/db/mod.rs
```diff
@@ -25,6 +27,12 @@ pub use batch_impl::validate_batch_id;
 
 const TOMBSTONE: i32 = 0;
 
+lazy_static! {
+    static ref PRETOUCH_NDT: NaiveDateTime = chrono::DateTime::parse_from_rfc3339(PRETOUCH_TS)
+        .expect("PRETOUCH_TS must be a valid RFC3339 datetime string")
+        .naive_utc();
+}
+
```
pjenvey: Nit: apparently we can construct this as a `const` (here's the one w/ TIMESTAMPTZ).
```suggestion
const PRETOUCH_DT: DateTime<Utc> = NaiveDate::from_ymd_opt(1, 1, 1)
        .unwrap()
        .and_hms_opt(0, 0, 0)
        .unwrap()
        .and_utc();
```
which seems worth it to avoid `lazy_static`/its needed dereferencing. Unfortunately we can't spit out `to_rfc3339` as a const, we could just leave it as a string as is in the spanner code, which is all it needs anyway.

syncstorage-db-common/src/lib.rs
```diff
@@ -48,6 +48,9 @@ pub const DEFAULT_BSO_TTL: u32 = 2_100_000_000;
 /// Non-standard collections will be allocated IDs beginning with this value
 pub const FIRST_CUSTOM_COLLECTION_ID: i32 = 101;
 
+/// Timestamp value to indicate a new user collection that hasn't been used/updated
+pub const PRETOUCH_TS: &str = "0001-01-01T00:00:00.00Z";
+
```
chenba: eh, I can move this back now that we aren't using it for PRETOUCH_DT.

PR #1943, opened 2025-12-17
fix: preserve existing value on batch append if new value is null/empty

chenba:
## Description

When updating a bso in a batch, preserve the existing value if the new value in the payload is null or empty.

## Testing

`RUST_TEST_THREADS=1 RUST_BACKTRACE=1 cargo test --workspace --no-default-features --features=syncstorage-db/postgres --features=tokenserver-db/postgres --features=py_verifier` should be pass after the patch

## Issue(s)

Fixes STOR-428

### DIFF
--- a/syncstorage-postgres/migrations/2025-10-20-155711_create_schema/up.sql
+++ b/syncstorage-postgres/migrations/2025-10-20-155711_create_schema/up.sql
@@ -43,6 +43,27 @@ CREATE TABLE collections (
     name VARCHAR(32) NOT NULL UNIQUE
 );
 
+-- Insert Standard Collections.
+-- These are the 13 standard collections that are expected to exist by clients.
+-- The IDs are fixed.
+-- Reserved spaces for additions to the standard collections begin after 100.
+INSERT INTO collections (collection_id, name) VALUES
+    ( 1, 'clients'),
+    ( 2, 'crypto'),
+    ( 3, 'forms'),
+    ( 4, 'history'),
+    ( 5, 'keys'),
+    ( 6, 'meta'),
+    ( 7, 'bookmarks'),
+    ( 8, 'prefs'),
+    ( 9, 'tabs'),
+    (10, 'passwords'),
+    (11, 'addons'),
+    (12, 'addresses'),
+    (13, 'creditcards');
+
+
+
 -- batches table
 CREATE TABLE batches (
     user_id BIGINT NOT NULL,
@@ -88,3 +109,4 @@ CREATE TABLE batch_bsos (
         batch_id
     ) ON DELETE CASCADE
 );
+
--- a/syncstorage-postgres/src/db/db_impl.rs
+++ b/syncstorage-postgres/src/db/db_impl.rs
@@ -19,7 +19,7 @@ use syncstorage_settings::Quota;
 use super::{PgDb, TOMBSTONE};
 use crate::{
     bsos_query,
-    db::CollectionLock,
+    db::{CollectionLock, PRETOUCH_DT},
     orm_models::BsoChangeset,
     pool::Conn,
     schema::{bsos, collections, user_collections},
@@ -137,6 +137,7 @@ impl Db for PgDb {
             .select(user_collections::modified)
             .filter(user_collections::user_id.eq(user_id))
             .filter(user_collections::collection_id.eq(collection_id))
+            .filter(user_collections::modified.gt(PRETOUCH_DT))
             .for_update()
             .first(&mut self.conn)
             .await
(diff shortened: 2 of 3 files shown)


syncstorage-postgres/src/db/batch_impl.rs
```diff
@@ -222,31 +221,25 @@ pub async fn do_append(
...
-            ))
-            .do_update()
-            .set((
-                batch_bsos::sortindex.eq(excluded(batch_bsos::sortindex)),
-                batch_bsos::payload.eq(excluded(batch_bsos::payload)),
-                batch_bsos::ttl.eq(excluded(batch_bsos::ttl)),
-            ))
-            .execute(&mut db.conn)
-            .await?;
+        sql_query(
```
taddes, line 258: Good call on the raw query here, makes it simple in this case for sure

syncstorage-postgres/src/db/batch_impl.rs
```diff
@@ -160,7 +159,7 @@ impl BatchDb for PgDb {
              WHERE user_id = $1 AND batch_id = $5
              ON CONFLICT (user_id, collection_id, bso_id) DO UPDATE SET
                  sortindex = COALESCE(EXCLUDED.sortindex, bsos.sortindex),
-                 payload = COALESCE(EXCLUDED.payload, bsos.payload),
+                 payload = COALESCE(NULLIF(EXCLUDED.payload, ''), bsos.payload),
```
pjenvey: Unfortunately this introduces a different, somewhat similar edge case where a request to update an existing bso's payload to an empty string would fail to occur. 

At a glance I'm not sure postgres's upsert can handle this in a simple fashion.. 🤔

chenba: Good catch.  I was relying on the tests 😅 

It's using two statements now instead of an upsert.

syncstorage-postgres/src/db/batch_impl.rs
```diff
@@ -142,7 +141,40 @@ impl BatchDb for PgDb {
...
+                     CASE
+                         WHEN src.ttl IS NOT NULL THEN $3::TIMESTAMP + (src.ttl || ' seconds')::INTERVAL
+                         ELSE NULL
+                     END,
+                     bsos.expiry
+                 )
+             FROM (
+                 SELECT batch_bso_id, sortindex, payload, ttl
+                 FROM batch_bsos
+                 WHERE user_id = $1 AND batch_id = $5
```
pjenvey:
```suggestion
                 WHERE user_id = $1 AND collection_id = $2 AND batch_id = $5
```

syncstorage-postgres/src/db/batch_impl.rs
```diff
@@ -142,7 +141,40 @@ impl BatchDb for PgDb {
...
 
+        sql_query(
+            "UPDATE bsos
+             SET
+                 sortindex = COALESCE(src.sortindex, bsos.sortindex),
+                 payload = COALESCE(src.payload, bsos.payload),
+                 modified = $3::TIMESTAMP,
+                 expiry = COALESCE(
+                     CASE
+                         WHEN src.ttl IS NOT NULL THEN $3::TIMESTAMP + (src.ttl || ' seconds')::INTERVAL
```
pjenvey: btw, do we need any of the casts in either of these queries other than the `INTERVAL`s (even $1::BIGINT and ''::TEXT)? We probably don't want to cast this to `TIMESTAMP` now that it's `TIMESTAMPTZ`, anyway.

```suggestion
                 modified = $3,
                 expiry = COALESCE(
                     CASE
                         WHEN src.ttl IS NOT NULL THEN $3 + (src.ttl || ' seconds')::INTERVAL
```

chenba: 🤦 Good catch on the `TIMESTAMP`.  I'll double check the casts.  Thanks

pjenvey: np, I should have fixed it myself w/ the `TIMESTAMPTZ` change

syncstorage-db/src/tests/batch.rs
```diff
@@ -314,3 +314,53 @@ async fn test_append_async_w_null() -> Result<(), DbError> {
...
+    db.commit_batch(params::CommitBatch {
+        user_id: hid(uid),
+        collection: coll.to_owned(),
+        batch,
+    })
+    .await?;
+
+    let bso_1_updated = db.get_bso(gbso(uid, coll, bid_0)).await?.unwrap();
+    assert!(
+        bso_1_updated.payload.is_empty(),
```
pjenvey: Oops, one more nit

```suggestion
    let bso_0_updated = db.get_bso(gbso(uid, coll, bid_0)).await?.unwrap();
    assert!(
        bso_0_updated.payload.is_empty(),
```

PR #1944, opened 2025-12-17
Use real exit codes for Postgres tests

chenba:
Let failing test fail CI workflow.

PR #1950, opened 2025-12-19
bug: address reserve and custom collections

taddes:
## Description

We have a shared constant `FIRST_CUSTOM_COLLECTION_ID` that is used to define collections >= 101 as being custom. We have defined the standard static collections but need a mechanism to ensure added collections insert past the custom value. Spanner utilized this in this create_collection calls. 

MySQL does not do this and simply used a database migration entry to define the next value. We want to make sure we address this in Postgres.

This involves changing how we manage `create_collection` and `get_or_create_collection_id` , compared to how it’s handled in Spanner and MySQL.

Alternatively, there may be a way to address this directly with the setup of the Postgres db. This PR first tests in the migrations and then determines if the programmatic addition needs to land.

## Testing

CI test results. 

## Issue(s)

Closes [STOR-429](https://mozilla-hub.atlassian.net/browse/STOR-429).

[STOR-429]: https://mozilla-hub.atlassian.net/browse/STOR-429?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/syncstorage-postgres/migrations/2025-10-20-155711_create_schema/up.sql
+++ b/syncstorage-postgres/migrations/2025-10-20-155711_create_schema/up.sql
@@ -62,7 +62,9 @@ INSERT INTO collections (collection_id, name) VALUES
     (12, 'addresses'),
     (13, 'creditcards');
 
-
+-- Set subsequent values to insert after 100.
+ALTER TABLE collections 
+ALTER COLUMN collection_id RESTART WITH 101;
 
 -- batches table
 CREATE TABLE batches (


pjenvey (approved):
with the tests passing you could also replace the postgres Makefile test targets `exit 0` w/ `exit $$exit_code` but we can do that later

taddes:
I see @chenba has https://github.com/mozilla-services/syncstorage-rs/pull/1944/files, so I'll let him rebase and then push that to give us our glorious exit codes of success.

Issue #1964, opened 2025-12-29
Feature request: forward proxy implementation


FabianZihlmann:
Dear all

Thanks a lot for your work. I run my own server and it works like a charme. 
In our production environment we have several security zones. So I'd like to run the server in a higher security zone. So I found out, that the server needs direct access to the internet.

If there is a forward proxy in beetween, I get the following error:
Error requesting remote jwks: error sending request for url (https://oauth.accounts.firefox.com/v1/jwks)

I didn't found any information about "how to set a forward proxy" in the configuration options.
Is there a possibility to implement such a feature?

Thanks a lot in advance!
Best regards
Fabian

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/STOR-442)

FabianZihlmann:
solved with pull request [#1966 ](https://github.com/mozilla-services/syncstorage-rs/pull/1966)

taddes:
Hello @FabianZihlmann ! Thanks for your kind regards and also providing the pull request. Much of us are out due to the holiday break, so we will review and get back to you next week. The change looks simple enough, we just to evaluate the implications and see if there's any required testing. Thank you!

FabianZihlmann:
Hello taddes
Thanks a lot for your comment. No hurry up, everything's fine.
It might be a good idea to improve the patch with some debugging information, in case the proxy is not well configured. I'm not experienced with rust so maybe someone else could care about.
For me it's just perfect, we can operate the syncstorage server in our security environment. Direct access to internet is a no go for any services expect proxy servers ... ;-)
Best regards and happy landing in the new year
Fabian

chenba:
@FabianZihlmann what do you think of setting up the proxy to proxy all the requests to `SYNC_TOKENSERVER__FXA_OAUTH_SERVER_URL` / `tokenserver.fxa_oauth_server_url` through to the actual destination?  e.g. token server -> http://localproxy/v1/jwks -> https://oauth.accounts.firefox.com/v1/jwks.  Then you can set the env var or toml value of `SYNC_TOKENSERVER__FXA_OAUTH_SERVER_URL` / `tokenserver.fxa_oauth_server_url` without having to patch the service.  Is there some advantage to proxying a single url?

FabianZihlmann:
Dear chenba
[...]

PR #1966, opened 2025-12-30
Add forward proxy for verify auth token

FabianZihlmann:
Add forward proxy for checking auth token. Optional proxy could be like http://proxy.domain.com:8080

## Description
The http_client is now using a forward proxy if this is configured in the config.toml. So you're able to operate the syncstorage-rs server in a security environment where a server has no direct access to internet.
In this case the server is not able to verify the auth token from for example https://oauth.accounts.firefox.com and the you'll get the follwing error message in the log:
Error requesting remote jwks: error sending request for url (https://oauth.accounts.firefox.com/v1/jwks)

This resolves the follwing isseu [#1964](https://github.com/mozilla-services/syncstorage-rs/issues/1964) 

## Testing
Add the following line to config.toml
tokenserver.fxa_oauth_request_proxy = "http://proxy.domain.com:8080"


chenba:
@FabianZihlmann I'm sorry about such a delayed response.  The feature and implementation make sense.  Can you add a unit test, and perhaps documentation to https://mozilla-services.github.io/syncstorage-rs/config.html?  Thank you.

FabianZihlmann:
Dear chenba
Thanks for your comment. Unfortunately, I'm not familiar how to make these unit tests. Do you might any other option?
Docs added, hope that is detailled enough? There was some mess up with not actual local code. I hope you can fix that. Sorry, I'm not really familiar with! (shame on me!)
Best regards
Fabian

PR #1969, opened 2025-12-31
bug: downgrade mdbook version for preprocessor

taddes:
## Description
As noted in the mdbook-mermaid repo in [this pr](https://github.com/badboy/mdbook-mermaid/pull/61), we were/are having the same issue reported on Nov 18 with the mermaid preprocessor failing:

```
 INFO Book building has started
Unable to parse the input
 WARN Error writing the RenderContext to the backend, Broken pipe (os error 32)
ERROR The "mermaid" preprocessor exited unsuccessfully with exit status: 1 status
```

Downgrading to supported v.0.5 might hopefully fix the issue.

## Testing

Ensure workflow succeeds.

### DIFF
--- a/.github/workflows/publish-docs.yaml
+++ b/.github/workflows/publish-docs.yaml
@@ -14,7 +14,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     env:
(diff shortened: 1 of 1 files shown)


jrconlin:
May want to note why we're not bumping past `0.5.1`? Helps prevent future rediscovery.

PR #1976, opened 2026-01-06
chore(ci): use gh actions to build and deploy to GAR and ghcr

chenba:
With this patch we move to Github Actions for building and publish container images to GAR.  Additionally we publish those images to ghcr.  (At the moment the ghcr images are private due to some org level settings.)

DockerHub pushes will remain in CircleCI at the moment, since a) https://github.com/mozilla-it/deploy-actions does not support DockerHub, and b) we should consider removing DockerHub support(?).

Closes STOR-412


.github/workflows/mozcloud-publish.yaml
```diff
@@ -19,6 +28,7 @@ jobs:
     permissions:
       contents: read
       id-token: write
+      packages: write
     uses: mozilla-it/deploy-actions/.github/workflows/build-and-push.yml@1b87069d293273436a84dff04954a8950d3ff9ca # v6.1.0
     with:
       image_name: syncstorage-rs
```
chenba, line 34: I didn't add a '-spanner' postfix here, but I'm not against it.  afaict we only need to update a couple of values in webservices-infra if we do make the change.

taddes: Exactly, I think the intention will be to flip that in webservices-infra when we push our next deploy if we've moved to using `-spanner`. Doesn't hurt to leave it as is for now. I intentionally kept the GAR one without the postfix, like you are doing here, specifically so that we'd not have to depend on an infra change were we to need to deploy/sync with argo.

taddes:
I'm also going to note that we get a lot of noise on the PRs here with the action, noting that we have 145 comments on this PR alone. Perhaps there's a way to reduce that noise with the preview build notifications.

Additionally, can we implement the checks being contingent on test results from CI? I know in the future we may move our whole deployment workflow to actions, but if we're passing back and forth between the two, that makes it tricky... I'd have to look more into if we can use signals from CI to affect the action, in the case of waiting for tests to pass.

chenba:
>     * We'd be now building the final docker image twice (until we migrate more of the circleci workflow to actions) -- I believe the github actions run parallel with circleci's run?
> 

Yes.  

>     * Publishing to GAR would always occur regardless of any tests failing in circleci
> 

I did briefly thought about triggering the github action only when the circleci workflow is successful, but then I remembered that our PRs require the branch to be up to date before merging, so I think we can rely on the PR's test results.

pjenvey (approved):

taddes (approved):
👍

PR #1981, opened 2026-01-07
test: fix flake in test_users_with_the_same_batch_id_get_separate_data

pjenvey:
due to _switch_user picking up old test data

## Testing

This might slow down `test_users_with_the_same_batch_id_get_separate_data` (which is already slow, according to circleci Test Insights, taking up to 90 seconds at least on uuid based batch id spanner/postgres backends). 

I might log a separate issue to disable this test entirely on those backends, especially if this change adds a significant slowdown, since they shouldn't be generating same batch ids anyway

## Issue(s)

Closes STOR-415


pjenvey:
Still only took about 90 seconds in this PR's run

PR #2003, opened 2026-01-13
feat: optimize batch_commit w/ MERGE INTO

pjenvey:
## Issue(s)

Closes STOR-457

### DIFF
--- a/syncstorage-mysql/src/db/diesel_ext.rs
+++ b/syncstorage-mysql/src/db/diesel_ext.rs
@@ -1,13 +1,8 @@
-use std::{fmt::Debug, marker::PhantomData};
-
 use diesel::{
-    backend::Backend,
-    insertable::CanInsertInSingleQuery,
     mysql::Mysql,
-    query_builder::{AstPass, InsertStatement, QueryFragment, QueryId},
+    query_builder::{AstPass, QueryFragment, QueryId},
     query_dsl::methods::LockingDsl,
     result::QueryResult,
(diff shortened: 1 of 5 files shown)


syncstorage-postgres/src/db/batch_commit.sql
```diff
@@ -0,0 +1,33 @@
...
+                      payload = COALESCE(batch.payload, bsos.payload),
+                      modified = $4,
+                      expiry = COALESCE(
+                          CASE
+                              WHEN batch.ttl IS NOT NULL THEN $4 + (batch.ttl || ' seconds')::INTERVAL
+                              ELSE NULL
+                          END,
+                          bsos.expiry
+                      )
+WHEN NOT MATCHED THEN
```
chenba, line 24: nit: indentation looks a little odd.

chenba (approved):
👍

syncstorage-postgres/src/db/batch_commit.sql
pjenvey: How so? I attempt to follow https://www.sqlstyle.guide/ (as do our spanner queries) -- you get this sort of "river down the middle" but I'll admit it's not always totally clear for every case

chenba: Just not used to seeing later parts of a statement so far less indented than the beginning.  But "river down the middle" does make sense (and easy to remember).  Thanks.

chenba (approved):

PR #2006, opened 2026-01-15
feat: kill long removed browserid references

pjenvey:
## Issue(s)

Closes STOR-458


tools/tokenserver/loadtests/requirements.txt
```diff
@@ -1,6 +1,5 @@
 authlib
 cryptography
 locust
-pybrowserid
```
pjenvey, line 4: @taddes do you recall if we need any of these `requirements.txt` files checked in? As we switched to having poetry generate them all in the Dockerfile

taddes: I've left them intact as we hadn't run the load tests in a while using Poetry. Also, this is one of the utilities that isn't built and generated by Docker, but run manually. Though the docs are updated with Poetry, so I do think we can kill the `requirements.txt` file. The only change I would then make in the `poetry.toml` file is to update `sqlalchemy = "1.4.46"` to have a carat in front to resolve to most recent version, like so `sqlalchemy = "^1.4.46"`, as I recall that causing an error at some point.

taddes: I also have a PR that's part of the spike related to load tests that will add a sample config.toml file with the env vars. Speaking of, how are you dealing with the env vars when you run the load tests? Where are you setting them when you initiate the run? Just pass them into the command line or are you passing the path to a file when you run them via poetry?

taddes (approved):
👍

PR #2011, opened 2026-01-20
chore: move MySQL build and test to GH actions

chenba:
This patch allows us to build and test the MySQL build with GH actions.  It will also start publishing to ghcr.

One minor downside compared to a single CircleCI workflow is that we are using mozilla-it/deploy-actions's `build-and-push` to publish the image.  That means the image is built twice on push to master or a tag, once for the e2e tests and another by `build-and-push`.  We'll see how well this works in practice and improve if necessary.  (I'm not worried that the `build-and-push` workflow runs simultaneously to the build and tests since we require the PR branch to be up to date with master before merging.)

Closes STOR-454

### DIFF
--- a/.github/workflows/mozcloud-publish.yaml
+++ b/.github/workflows/mozcloud-publish.yaml
@@ -114,3 +114,29 @@ jobs:
       dockerfile_path: tools/postgres/Dockerfile
       image_build_context: tools/postgres
       should_tag_ghcr: true
+
+  build-and-push-syncstorage-rs-mysql:
+    if: >
+      github.event_name == 'workflow_dispatch' ||
+      (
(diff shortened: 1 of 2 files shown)


chenba:
The test result upload sections are temporarily commented out since this repo cannot use `google-github-actions/upload-cloud-storage` until that action is allowed by the admins.

chenba:
> The test result upload sections are temporarily commented out since this repo cannot use `google-github-actions/upload-cloud-storage` until that action is allowed by the admins.

`google-github-actions/upload-cloud-storage` is allowed now.

chenba:
> I did just have some slight thoughts on optimizations

If they are already written down somewhere, please share.  Thanks

PR #2012, opened 2026-01-20
chore: migrate to GH action for Postgres build and tests

chenba:
This patch allows us to build and test the Postgres build with GH actions.

One minor downside compared to a single CircleCI workflow is that we are using mozilla-it/deploy-actions's `build-and-push` to publish the image.  That means the image is built twice on push to master or a tag, once for the e2e tests and another by `build-and-push`.  We'll see how well this works in practice and improve if necessary.  (I'm not worried that the `build-and-push` workflow runs simultaneously to the build and tests since we require the PR branch to be up to date with master before merging.)

Closes STOR-453


chenba:
The `test_authorization.py::TestAuthorization::test_disallow_reusing_old_client_state` e2e test failed twice before passing on the third run.  A little worrying.  The test passed in CircleCI on the first try.

PR #2013, opened 2026-01-20
feat: optimize post_bsos w/ MERGE INTO

pjenvey:
ideally we'd pass data to MERGE INTO's USING w/ a VALUES statement,
but rendering VALUES here w/ diesel's QueryFragment trait is painful

instead it's UNNEST'ed from a single bind parameter of post_bso[] (an
Array of a new, named composite type). this composite type must be
explicitly named/declared as postgres disallows unnamed/anonymous
composite types to be used as bind parameters

Closes [STOR-423](https://mozilla-hub.atlassian.net/browse/STOR-423)

[STOR-423]: https://mozilla-hub.atlassian.net/browse/STOR-423?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/setup-python/action.yml
+++ b/.github/actions/setup-python/action.yml
@@ -1,0 +1,42 @@
+name: Setup Python Environment
+description: Sets up Python with Poetry dependencies
+
+inputs:
+  python-version:
+    description: "Python version to use"
+    required: false
+    default: "3.12" # PYTHON_VER
+  workspace-path:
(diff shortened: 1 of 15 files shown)


chenba (approved):
I like it

taddes (approved):

PR #2015, opened 2026-01-21
feat: migrate to GitHub Actions for Spanner build and tests

taddes:
## Description
Moves to GitHub Actions for Spanner build and test. Evaluating whether a discrete action is better than installing the Spanner emulator from GCP.  At present, that action is not supported in Mozilla enterprise and is not an official Google action. Will also have to address the issue of some environment variables being in place for test upload. Separate SRE ticket filed for that.

## Issue(s)

Closes [STOR-455](https://mozilla-hub.atlassian.net/browse/STOR-455).

[STOR-455]: https://mozilla-hub.atlassian.net/browse/STOR-455?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/spanner.yml
+++ b/.github/workflows/spanner.yml
@@ -5,7 +5,7 @@ on:
     types: [opened, synchronize]
   push:
     branches:
-      - "**"
+      - master
     tags:
       - "**"
   workflow_dispatch: {}


.github/actions/setup-python/action.yml
```diff
@@ -5,7 +5,7 @@ inputs:
   python-version:
     description: "Python version to use"
     required: false
-    default: "3.12" # PYTHON_VER
+    default: "3.12" # PY_VER
```
chenba, line 8: Is there a practical difference?  Or is it just convention?

chenba:
> Evaluating whether a discrete action is better than installing the Spanner emulator from GCP. At present, that action is not supported in Mozilla enterprise

Do we expect that action to be "supported in Mozilla enterprise" at some point?  I guess I don't fully understand the alternative mentioned.

.github/workflows/spanner.yml
```diff
@@ -0,0 +1,226 @@
...
+  push:
+    branches:
+      - master
+    tags:
+      - "**"
+  workflow_dispatch: {}
+
+env:
+  RUST_VERSION: "1.89" # RUST_VER
+  PYTHON_VERSION: "3.12" # PY_VER
```
pjenvey, line 15: Where are these actually utilized? It looks like we call the `setup-{rust,python}` actions without them, so the actions' defaults are used.

If they aren't necessary they would be nice to kill so there's less lines to update when we bump the versions.

pjenvey (commented):
I'm also confused by "Mozilla enterprise" and the action here -- do you mean splitting out the spanner emulator into its own action?

.github/workflows/spanner.yml
taddes: Will do, I can take them out in the subsequent PR that adds the dockerhub stuff.

taddes:
@chenba @pjenvey There is a 3rd party action that somebody made to spin up the Spanner emulator. We'd have to go though an approval process to get that particular external action added, since it's not maintained by GitHub but 3rd party. Since I don't want that to hold us up and it's just as easy to set up the emulator here. Sorry for any confusion, the Mozilla Enterprise message was from when I tried to use it and got a failure message since it's not permitted:

```The action cloudspannerecosystem/setup-spanner-emulator@v0.2.0 is not allowed in mozilla-services/syncstorage-rs because all actions must be from a repository owned by your enterprise, created by GitHub, or match one of the patterns: PyO3/maturin-action@v1*, actions-rs/toolchain@v1, astral-sh/setup-uv@*, codecov/codecov-action@v2, dependabot/fetch-metadata@*, docker/*, dtolnay/rust-toolchain@*, golangci/golangci-lint-action@*, google-github-actions/auth@*, google-github-actions/setup-gcloud@*, google-github-actions/upload-cloud-storage@*, messense/maturin-action@v1, pypa/gh-action-pypi-publish@*, pypa/gh-action-pypi-publish@release/*, slackapi/slack-github-action@*, splunk/appinspect-cli-action@*, terraform-docs/gh-actions@*, uraimo/run-on-arch-action@v2, zizmorcore/zizmor-action@e639db99335bc9038abc0e066dfcd72e23d26fb4.```

Issue #1428, opened 2022-10-20
Installation in docker?


WAdama:
Hi all,

I tried to install syncstorage-rs via docker, but I can't get my head around it with this documentation.

For one it seems the prerequisite mozilla-rust-sdk is now google-cloud-rust, is that correct?

Maybe someone has already got this to run and can provide me with a more step by step documentation...

Thanks
Ingo

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DISCO-2056)

ictabc:
Same here, managed to compile the syncstorage-rs with a bit of a strugle, but after that just a big black hole. No good documentation, missing a lot of parts in the documentation to get things working. Guess this documentation is build by people that forgot the parts that other people need to get things running. Don't want to use the antiquated syncserver code again. And I'm not allowed to sync my data to the cloud, specially if it's in the US.

Found a docker image in the docker hub, but absolutely no data on how to use it. https://hub.docker.com/r/mozilla/syncstorage-rs

As nobody has documented anything yet on how to use it with docker compile or docker cli, totally lost on what to do. I'm not a noob, but but with current documentation it's impossible to do anything.

Maybe someone from the community can write something, how to use the docker container. Or point me to a page where it is described, at least better than the non information that's in the readme posted in this repository.

jakobkukla:
I managed to get it working a few days ago. See my docker-compose.yml file down below. **Don't forget to change the secrets, mysql credentials/urls and paths.**

```
version: "3.8"

services:
  firefox-sync:
    image: mozilla/syncstorage-rs:0.12.4
    container_name: firefox-sync
    depends_on:
      - firefox-sync-syncstorage-db
      - firefox-sync-tokenserver-db
    environment:
      SYNC_HOST: 0.0.0.0
      SYNC_HUMAN_LOGS: 1
      SYNC_MASTER_SECRET: MY_SECRET
      SYNC_DATABASE_URL: mysql://MY_SYNC_MYSQL_USER:MY_SYNC_MYSQL_USER_PASSWORD@firefox-sync-syncstorage-db:3306/syncstorage
      SYNC_TOKENSERVER__ENABLED: "true"
      SYNC_TOKENSERVER__RUN_MIGRATIONS: "true"
      SYNC_TOKENSERVER__NODE_TYPE: mysql
[...]

PR #2019, opened 2026-01-26
docs: add how-to on deploying with docker compose

chenba:
Update docs with a simple how-to on deploying with docker compose.

Closes STOR-118 / #1428

### DIFF
--- a/tools/tokenserver/loadtests/README.md
+++ b/tools/tokenserver/loadtests/README.md
@@ -77,16 +77,16 @@ Ex. `poetry run python locustfile.py`
 
    ```sh
    # Should be set to the "n" component of the JWK
-   SYNC_TOKENSERVER__FXA_OAUTH_PRIMARY_JWK_N
+   SYNC_TOKENSERVER__FXA_OAUTH_PRIMARY_JWK__N
    # Should be set to the "e" component of the JWK (this value should almost always be "AQAB")
-   SYNC_TOKENSERVER__FXA_OAUTH_PRIMARY_JWK_E
-   SYNC_TOKENSERVER__FXA_OAUTH_PRIMARY_JWK_KTY=RSA
-   SYNC_TOKENSERVER__FXA_OAUTH_PRIMARY_JWK_USE=sig
-   SYNC_TOKENSERVER__FXA_OAUTH_PRIMARY_JWK_ALG=RS256
+   SYNC_TOKENSERVER__FXA_OAUTH_PRIMARY_JWK__E
+   SYNC_TOKENSERVER__FXA_OAUTH_PRIMARY_JWK__KTY=RSA
+   SYNC_TOKENSERVER__FXA_OAUTH_PRIMARY_JWK__USE=sig
+   SYNC_TOKENSERVER__FXA_OAUTH_PRIMARY_JWK__ALG=RS256
 
    # These two environment variables don't affect the load tests, but they need to be set:
-   SYNC_TOKENSERVER__FXA_OAUTH_PRIMARY_JWK_KID=""
-   SYNC_TOKENSERVER__FXA_OAUTH_PRIMARY_JWK_FXA_CREATED_AT=0
+   SYNC_TOKENSERVER__FXA_OAUTH_PRIMARY_JWK__KID=""
+   SYNC_TOKENSERVER__FXA_OAUTH_PRIMARY_JWK__FXA_CREATED_AT=0
    ```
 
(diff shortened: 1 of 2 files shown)


docs/src/how-to/how-to-run-fxa.md
```diff
@@ -1,126 +1 @@
-<a id="howto_run_fxa"></a>
-
-# Run Your Own Mozilla Accounts Server (Outdated)
```
chenba, line 3: No reason to this around.

docs/src/how-to/how-to-run-with-docker.md
```diff
@@ -0,0 +1,117 @@
...
+```
+
+## Configuring Firefox
+
+Firefox itself needs to be configured to use the self-hosted Sync server.
+
+1. Go to `about:config` in Firefox.
+1. Find the `identity.sync.tokenserver.uri` configuration.
+1. Change the value to `http://localhost:8000/1.0/sync/1.5`.
+1. Restart Firefox.
```
taddes, line 114:
```suggestion
1. Go to `about:config` in Firefox.
2. Find the `identity.sync.tokenserver.uri` configuration.
3. Change the value to `http://localhost:8000/1.0/sync/1.5`.
4. Restart Firefox.
```

docs/src/how-to/how-to-run-fxa.md
taddes: Maybe we should delete the whole file then?

docs/src/how-to/how-to-run-with-docker.md
```diff
@@ -0,0 +1,117 @@
...
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:8000/__heartbeat__"]
+      interval: 30s
+      timeout: 10s
+      retries: 3
+      start_period: 60s
+```
+
+Note that multiple values will be read from the environment:
+- `SYNC_MASTER_SECRET`: a secret used in cryptographic operations
```
taddes: In the e2e tests, we use `secret0`.  Perhaps we can suggest here that for active use, it should be a proper crypto key, but can be something simpler like a string password. It may possibly confuse someone that this value has to be generated somehow or needs to be something special. Kinda like what you did below with `use_your_own_secret_4d3d3d3d`, just a clarifying sentence.

docs/src/how-to/how-to-run-with-docker.md
```diff
@@ -0,0 +1,117 @@
...
+- `SYNC_SYNCSTORAGE__DATABASE_URL`: database URL for syncstorage, e.g. `mysql://sync:test@example.io/syncstorage` or `postgres://testo:@localhost/syncdb`
+- `SYNC_TOKENSERVER__DATABASE_URL`: database URL for tokenserver, e.g. `mysql://sync:test@example.io/tokenserver` or `postgres://testo:@localhost/syncdb`
+
+The values can be directly written into the yaml as well.
```
taddes, line 62: If someone will use env variables, perhaps there should be a suggestion or recommendation on how best to pass said ENV vars? Putting the settings in the yaml is certainly easiest. We may not need to decide this here, but we should probably come up with a best practice on how to set and pass env vars, whether it be in the `local.toml` or when invoking the service in terminal, or `dotenv`.

docs/src/how-to/how-to-run-fxa.md
chenba: oh oops, I thought I did it.  Not sure how that happened.  Thanks.

docs/src/how-to/how-to-run-with-docker.md
chenba: The markdown renderer can handle it.

docs/src/how-to/how-to-run-with-docker.md
chenba: The reader can see that on https://github.com/mozilla-services/syncstorage-rs/pull/2019/changes#diff-694b9247321f6a4ec48bd348e6f4a5c3ea3da468e3eb8e8ee30c0f1b536a03c5R66-R69 already though?

taddes: Fair enough, I'll leave it up to you. I've been zealously wanting to make stuff as crystal clear as possible after wading though so much unclear documentation heh

chenba: I believe by the time someone wants to self-host Sync they likely have experience with environment/shell variables.  Or if we do not believe that to be true, then we should (resource permitting) provide something they can copy and paste to get the service running.  I don't think teaching about env vars (or reverse proxy, or certs, etc.) should be a goal here.

taddes: Alright 👍

PR #2025, opened 2026-01-27
docs: add one-shot docker compose yaml to how-to

chenba:
Add a one-shot docker file to start the db with `docker compose` also.  Follow-up to STOR-118 / #1428 / PR #2019.

### DIFF
--- a/docs/src/how-to/how-to-run-with-docker.md
+++ b/docs/src/how-to/how-to-run-with-docker.md
@@ -9,28 +9,29 @@ Images are available for both
 [MySQL](https://github.com/mozilla-services/syncstorage-rs/pkgs/container/syncstorage-rs%2Fsyncstorage-rs-mysql)
 and
 [PostgreSQL](https://github.com/mozilla-services/syncstorage-rs/pkgs/container/syncstorage-rs%2Fsyncstorage-rs-postgres)
-as the database.  The sample code will focus on MySQL.  Differences in
-configuration or deployment steps will be noted. 
+as the database.  Differences in configuration or deployment steps will be
+noted. 
 
 > **Note:** At the time of writing, there are no tagged release builds
 > available on ghcr.io.  This guide will use a build from the main development
 > branch.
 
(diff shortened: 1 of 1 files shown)


docs/src/how-to/how-to-run-with-docker.md
```diff
@@ -104,6 +104,99 @@ Restart the service with
...
+## Docker Compose, One-Shot with PostgreSQL
+
+Alternatively, the database can be started through `docker compose` as well.  The real service URL can be set with the `NODE_URL` environment variable.
+
+Save the yaml below into a file, e.g. `docker-compose.one-shot.yaml`.
+
+```yaml
+services:
+  syncserver:
+    image: ghcr.io/mozilla-services/syncstorage-rs/syncstorage-rs-postgres:11659d98f9
```
taddes: I'll leave it up to you, but perhaps just a note to clarify that the hash value after `syncstorage-rs-postgres:` corresponds to a particular build in the aforementioned link (https://github.com/mozilla-services/syncstorage-rs/pkgs/container/syncstorage-rs%2Fsyncstorage-rs-postgres) and for someone to select that accordingly.

chenba: I was relying on the note at https://github.com/mozilla-services/syncstorage-rs/blob/master/docs/src/how-to/how-to-run-with-docker.md?plain=1#L15-L17 but yeah that is not exactly explicit.  I can definitely mention the tag hash but there's no good, succinct advice on how to pick a tag; they are just merges on the main branch after all.

I was planning on updating the tags in the doc once we tagged the next release.

pjenvey: I think we should probably consider a separate issue to add additional docker tags to help with such setups, like a typical "latest", and "\<major\>", "\<major\>.\<minor\>" available on many of the popular dockers.

For now I'm fine with the hash, or how about `syncstorage-rs-postgres:${SYNCSERVER_VERSION:-11659d98f9}`

PR #2029, opened 2026-01-29
chore: Update GAR Image Tagging to MozCloud Spec

taddes:
## Description

Per MozCloud’s spec on Container Images and Continuous Deployment, we want to align our image tagging in GAR and GHCR to match the org-wide standard to include the release date.

Tags must consist of:

-  A short Git commit SHA (10 characters)

-  A UTC timestamp in the format YYYYMMDDTHHMMSS

`<short-sha>-<YYYYMMDDTHHMMSS>`

## Testing

Output from workflow (will verify in branch)

## Issue(s)

Closes [STOR-451](https://mozilla-hub.atlassian.net/browse/STOR-451).

[STOR-451]: https://mozilla-hub.atlassian.net/browse/STOR-451?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/mozcloud-publish.yaml
+++ b/.github/workflows/mozcloud-publish.yaml
@@ -1,3 +1,8 @@
+# Mozilla Deploy Actions url: <https://github.com/mozilla-it/deploy-actions>
+# Note: even though Mozilla maintains the above actions, it is still suggested
+# when upgrading to use the full commit SHA and comment with version.
+# See <https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions>
+# Ex. `mozilla-it/deploy-actions/.github/workflows/build-and-push.yml@4784cb70739a4f32ce010921f60fb1ebbc791a38 # v6.2.2`
 name: Build, Tag and Push Container Images to GAR Repository
 
 on:
@@ -9,7 +14,7 @@ on:
     branches:
       - master
     tags:
-      - "**"
+      - '**'
   workflow_dispatch: {}
 
 jobs:
@@ -39,7 +44,7 @@ jobs:
       contents: read
       id-token: write
       packages: write
-    uses: mozilla-it/deploy-actions/.github/workflows/build-and-push.yml@v6.2.2
+    uses: mozilla-it/deploy-actions/.github/workflows/build-and-push.yml@4784cb70739a4f32ce010921f60fb1ebbc791a38 # v6.2.2
     with:
       image_name: syncstorage-rs
       gar_name: sync-prod
@@ -67,7 +72,7 @@ jobs:
       contents: read
       id-token: write
       packages: write
-    uses: mozilla-it/deploy-actions/.github/workflows/build-and-push.yml@v6.2.2
+    uses: mozilla-it/deploy-actions/.github/workflows/build-and-push.yml@4784cb70739a4f32ce010921f60fb1ebbc791a38 # v6.2.2
     with:
       image_name: syncstorage-rs-postgres
       gar_name: sync-prod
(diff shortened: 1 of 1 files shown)


taddes:
Thought I'd tag you on this one @grahamalama in case I was missing something with the mozilla-it maintained action. I just created a job to gen the timestamp and then pass it in the `image_tag_metadata`, which I gleaned from the deploy actions docs: https://github.com/mozilla-it/deploy-actions/blob/main/.github/workflows/docs/build-and-push.md. If there's another way services has done it, lmk.

It does work successfully, but given the way the workflow goes, it adds two `-`, see [here](https://github.com/mozilla-services/syncstorage-rs/pkgs/container/syncstorage-rs%2Fsyncstorage-rs-spanner-python-utils). Any way to avoid that or is that fine? I'm just aligning with the MozCloud spec as we move away from DockerHub and to the org standard. Thanks!

.github/workflows/mozcloud-publish.yaml
```diff
@@ -9,11 +9,21 @@ on:
     branches:
       - master
     tags:
-      - '**'
+      - "**"
```
chenba: Doesn't this make it less consistent with the rest of file/PR?  I see `'master'` and `'refs/tags/'` for example.

chenba (dismissed):

pjenvey (dismissed):

.github/workflows/mozcloud-publish.yaml
```diff
@@ -29,7 +39,7 @@ jobs:
       contents: read
       id-token: write
       packages: write
-    uses: mozilla-it/deploy-actions/.github/workflows/build-and-push.yml@1b87069d293273436a84dff04954a8950d3ff9ca # v6.1.0
+    uses: mozilla-it/deploy-actions/.github/workflows/build-and-push.yml@v6.2.2
```
grahamalama: I would still [recommend pinning at the full SHA](https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions:~:text=Pin%20actions%20to%20a%20full%2Dlength%20commit%20SHA)and adding the semver tag at the end. Dependabot knows how to update the comments along with the SHA. Though we're a "trusted creator", it's still something [Zizmor](https://docs.zizmor.sh/) will flag (if you're using that).

.github/workflows/mozcloud-publish.yaml
taddes: For some reason my auto-formatter in VSCode keeps demanding that, I'll see if I can disable the prefs for the double quotes.

.github/workflows/mozcloud-publish.yaml
taddes: Thanks so much for chiming in @grahamalama ! I'll go ahead and make those changes too and get a PR for you folks in global-platform-admin for those `image_regex` values. Appreciate you pointing me right to them.

taddes:
Sry @pjenvey @chenba , rebase and changes from @grahamalama made review approval stale. Need approval again.

.github/workflows/mozcloud-publish.yaml
chenba: Ah, no worries.  Just thought the change was a bit odd; makes more sense now knowing it was an auto-formatter.

pjenvey:
> want to log an issue for the image_regex change?

eh it could just be part of this STOR-451 though

taddes:
> > want to log an issue for the image_regex change?
> 
> eh it could just be part of this [STOR-451](https://mozilla-hub.atlassian.net/browse/STOR-451) though

Yes good call. I've got it in [STOR-434](https://mozilla-hub.atlassian.net/browse/STOR-434) already as the general update, but will note @grahamalama 's comment there too 👍 

[STOR-451]: https://mozilla-hub.atlassian.net/browse/STOR-451?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[STOR-434]: https://mozilla-hub.atlassian.net/browse/STOR-434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Issue #1850, opened 2025-10-18
Db unit tests fail against Spanner emulator 


data-sync-user:
We don’t explicitly roll back test transactions against spanner (we rely on diesel `test_transactions` for this, of which there’s no equivalent on our spanner impl). This seems to trigger a “bug” in the spanner emulator where it doesn’t clean up the previous transaction, then later errors out that the emulator only supports handling of one transaction at a time.

We can fix this by manually rolling back the transactions ourselves in the test, we would need something along the lines of the `db_pool.transaction_http/http` methods that wraps around every db unit test, which would begin a transaction and roll it all back at the end.

This would also let us cleanup some of the unfortunate implicit transaction handling in the spanner implementation.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/STOR-388)

data-sync-user:
➤ Barry Chen commented:

{quote}This would also let us cleanup some of the unfortunate implicit transaction handling in the spanner implementation.{quote}

Philip Jenvey can you tell me a bit more about this?

data-sync-user:
➤ Philip Jenvey commented:

Barry Chen We do an implicit begin to start a transaction from sql_request for the tests here ( https://github.com/mozilla-services/syncstorage-rs/blob/9f700d642e5bd21ad40bce552bf4eedd5ea17118/syncstorage-spanner/src/db/mod.rs#L164 ). (TBH I think this code used to be even more of a kludge in the past but was cleaned up a bit). 

Regardless it would still be clearer to explicitly create the transaction and roll it back in the tests, at least on Spanner. Then this implicit begin could be removed entirely: it should only be used by the unit tests.

data-sync-user:
➤ Philip Jenvey commented:

We also noop commit when use_test_transactions is set. We do call commit in a couple of tests, so I’m not sure we can remove that nooping on Spanner..

data-sync-user:
➤ Philip Jenvey commented:

If we’re making this all more explicit, I’d probably consider changing the commits in the test to rollback. Basically you can’t easily have much unit test coverage for commit, so we probably shouldn’t bother calling it all there.

PR #2045, opened 2026-02-06
test: rollback transaction in tests relying on Spanner emulator

chenba:
We need to explicitly rollback transactions in tests that rely on the Spanner emulator.

Closes STOR-388 / https://github.com/mozilla-services/syncstorage-rs/issues/1850

### DIFF
--- a/.github/workflows/mozcloud-publish.yaml
+++ b/.github/workflows/mozcloud-publish.yaml
@@ -18,6 +18,9 @@ on:
   workflow_dispatch: {}
 
 jobs:
+  # Note: we are moving towards renaming all images `syncserver`, the union of sync and tokenserver.
+  # This presently remains for the time being to simplify deploys by maintaining `image_name: syncstorage-rs`.
+  # Historical context: <https://github.com/mozilla-services/syncstorage-rs/pull/1306>
   build-and-push-syncstorage-rs:
     if: >
       github.event_name == 'workflow_dispatch' ||
@@ -44,7 +47,7 @@ jobs:
         MYSQLCLIENT_PKG=libmysqlclient-dev
       should_tag_ghcr: true
 
-  build-and-push-syncstorage-rs-postgres:
+  build-and-push-syncserver-postgres:
     if: >
       github.event_name == 'workflow_dispatch' ||
       (
@@ -62,7 +65,7 @@ jobs:
       packages: write
     uses: mozilla-it/deploy-actions/.github/workflows/build-and-push.yml@4784cb70739a4f32ce010921f60fb1ebbc791a38 # v6.2.2
     with:
-      image_name: syncstorage-rs-postgres
+      image_name: syncserver-postgres
       gar_name: sync-prod
       project_id: moz-fx-sync-prod
       docker_build_args: |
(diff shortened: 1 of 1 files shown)


chenba:
> In the original issue I had an idea for making a wrapper similar to `db_pool.transaction_http/http` -- do you think that's feasible?
> 
I'll take a look.

> It would be more explicit than requiring the `Drop` and requiring the tests to call `rollback` themselves. One issue with the manual `rollback` is if any part of the test fails for whatever reason, that `rollback` won't be called.
> 
I misinterpreted your request for making things explicit.  `Drop` works without the other changes.  As it is in this PR, the Drop is conditional so that if the test failed to rollback then Drop handles it.  But we could just as well remove the "explicit" rollback in the test and let Drop handle it every time.

chenba:
Added `with_transaction`, similar to `db_pool.transaction_http`, that wraps a test in a transaction, and ensures the rollback.  The test updates are mostly whitespace from the additional indentation.  h/t @pjenvey

syncstorage-db/src/tests/batch.rs
```diff
@@ -172,34 +172,37 @@ async fn quota_test_create_batch() -> Result<(), DbError> {
...
-    let result = db.create_batch(cb(uid, coll, bsos2)).await;
-    if settings.enforce_quota {
-        assert!(result.is_err());
-    } else {
-        assert!(result.is_ok());
-    }
-
-    Ok(())
+    with_transaction(
+        Some(settings.clone()),
```
pjenvey: nit/not blocking this PR/more of an FYI: there's a Rust idiom where you can avoid the need for Some() here: [Into<Option<_>>](https://deterministic.space/elegant-apis-in-rust.html#intooption_)

syncstorage-db/src/tests/support.rs
```diff
@@ -36,6 +36,30 @@ pub async fn test_db(pool: DbPoolImpl) -> Result<Box<dyn Db<Error = DbError>>, D
...
+
+    db.begin(true).await?;
+    let result = test(&mut *db).await;
+    let rollback_result = db.rollback().await;
+
+    // the rollback itself might fail
+    match (result, rollback_result) {
+        (Ok(()), rollback) => rollback,
+        (Err(test_err), Ok(())) => Err(test_err),
+        (Err(test_err), Err(_rollback_err)) => Err(test_err),
```
pjenvey: These should be equivalent
```suggestion
        (Err(test_err), _) => Err(test_err),
```

so I believe this simplifies further into

```suggestion
    if result.is_err() {
        result
    } else {
        // the rollback itself might fail
        rollback_result
    }

```

syncstorage-db/src/tests/support.rs
```diff
@@ -36,6 +36,30 @@ pub async fn test_db(pool: DbPoolImpl) -> Result<Box<dyn Db<Error = DbError>>, D
     Ok(db)
 }
 
+/// Wrap a test in a DB transaction that will be rolled back.  Read-only tests
+/// do not need this.
+pub async fn with_transaction<T>(
```
pjenvey: Maybe call this `test_transaction` or `with_test_transaction`? Especially as it's along similar lines of the `use_test_transactions` setting which calls diesel's `begin_test_transaction`.

Which reminds me, diesel even has essentially its own version of this it just calls [test_transaction](https://docs.diesel.rs/2.3.x/diesel/connection/trait.Connection.html#method.test_transaction) FYI

PR #2051, opened 2026-02-10
chore: update image names and jobs to syncserver

taddes:
## Description

Updates the naming convention of images to `syncserver`, our historically appropriate title for sync + tokenserver. Only references to image publishing are updated, other things like test metrics and whatnot are kept as `syncstorage-rs` for clarity.

Keeping the spanner/mysql image the same so as to not have to change it for current deployment and not needing to change the global platform admin name. Added a note for the future when we decide to change it. 

## Issue(s)

Closes [STOR-478](https://mozilla-hub.atlassian.net/browse/STOR-478).

[STOR-478]: https://mozilla-hub.atlassian.net/browse/STOR-478?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/mozcloud-publish.yaml
```diff
@@ -36,15 +36,15 @@ jobs:
       packages: write
     uses: mozilla-it/deploy-actions/.github/workflows/build-and-push.yml@4784cb70739a4f32ce010921f60fb1ebbc791a38 # v6.2.2
     with:
-      image_name: syncstorage-rs
+      image_name: syncserver
```
chenba: Since we are changing name might as well change this to syncserver-spanner?

chenba (approved):

.github/workflows/mozcloud-publish.yaml
taddes: I was torn on that, I had actually changed it, but changed it back before pushing since I thought maybe keeping it consistent with the repo name was better? 😰  I am happy to change it if you think we should do that across the board. Maybe even renaming the repo eventually.

.github/workflows/mozcloud-publish.yaml
```diff
@@ -18,7 +18,7 @@ on:
   workflow_dispatch: {}
 
 jobs:
-  build-and-push-syncstorage-rs:
+  build-and-push-syncserver:
```
pjenvey: Per the slack thread we discussed holding off on renaming the spanner build because it will require the image rename change on the infra side, limiting our stage/prod env to only new images. We can deal with this later -- simplest to just rename the mysql/postgres images for now (which don't have much usage yet)

taddes: Ok, I'll revert to keep things simple and we can reassess.

taddes:
Just need fresh approvals since I reverted the spanner `image_name` @chenba @pjenvey

chenba (dismissed):

taddes:
> Sorry, also: it might be worth adding a comment in here somewhere about the naming inconsistency, just stating that we're moving over time to "syncserver".
> 
> FYI here's Ethan's first change to "syncserver" which explains the thinking a bit [#1306 (comment)](https://github.com/mozilla-services/syncstorage-rs/pull/1306#issue-1229988533)

Ok good call, I will add that.

PR #2061, opened 2026-02-12
chore: `cargo upgrade` deps

chenba:
And updated slog calls to get rid of deprecation warnings...which is the majority of the changes by file count.  I can move that into another commit/PR if that's easier to review.

Closes the rust deps part of STOR-436

### DIFF
--- a/syncstorage-db/src/tests/batch.rs
+++ b/syncstorage-db/src/tests/batch.rs
@@ -374,3 +374,20 @@ async fn test_append_async_w_empty_string() -> Result<(), DbError> {
     })
     .await
 }
+
+#[tokio::test]
+async fn pretouch() -> Result<(), DbError> {
+    let settings = Settings::test_settings().syncstorage;
+    with_test_transaction(settings, async |db: &mut dyn Db<Error = DbError>| {
+        let uid = 1;
+        let coll = "clients";
+        db.delete_storage(hid(uid)).await?;
+        let _batch = db.create_batch(cb(uid, coll, vec![])).await?;
+        // client with no timestamp (or PRETOUCH on some backends) should act as if
+        // no timestamp exists
+        let ts = db.get_storage_timestamp(hid(uid)).await?;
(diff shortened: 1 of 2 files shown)


syncserver-common/src/middleware/sentry.rs
```diff
@@ -94,17 +94,25 @@ where
                             // add an info here, temporarily turn on info level debugging on a given server,
                             // capture it, and then turn it off before we run out of money.
                             maybe_emit_metrics(&metrics, reportable_err);
-                            debug!("Sentry: Not reporting error (service error): {:?}", error);
+                            slog::debug!(
```
pjenvey: Doh, this again! Autopush was hit by [this](https://mozilla-hub.atlassian.net/browse/PUSH-579) here (sorry I totally forgot about this).

Per my comment there I thought their new release was going to solve this so this change wouldn't be necessary, which they finally released last month?

chenba: @pjenvey nice catch, thanks.  I captured the original `cargo upgrade` output in the jira and looks like it didn't upgrade slog-scope.  I just tried it again and it did upgrade slog-scope.  Not sure what the deal is but I think the updated PR is ready now.

chenba:
sorry I screwed up a rebase.  Setting this back to a draft for now.

Cargo.toml
```diff
@@ -78,23 +78,23 @@ serde = "1.0"
...
   "max_level_trace",
   "release_max_level_info",
   "dynamic-keys",
 ] }
-slog-async = "2.5"
+slog-async = "2.8"
 slog-envlogger = "2.2.0"
 slog-mozlog-json = "0.1"
-slog-scope = "4.3"
+slog-scope = "4.4"
```
chenba, line 89: wait WHAT I'm dying here

chenba: oh right never mind, 4.4.1 is installed in `Cargo.lock`.

PR #2073, opened 2026-02-18
feat: log when Sentry is not configured

chenba:
When Sentry is not configured, the sentry middleware writes the event to the error log instead.

Closes STOR-490

### DIFF
--- a/tools/tokenserver/loadtests/poetry.lock
+++ b/tools/tokenserver/loadtests/poetry.lock
@@ -598,14 +598,14 @@ test = ["pytest (>=6)"]
 
 [[package]]
 name = "flask"
-version = "3.1.2"
+version = "3.1.3"
 description = "A simple framework for building complex web applications."
 optional = false
 python-versions = ">=3.9"
 groups = ["main"]
 files = [
-    {file = "flask-3.1.2-py3-none-any.whl", hash = "sha256:ca1d8112ec8a6158cc29ea4858963350011b5c846a414cdb7a954aa9e967d03c"},
-    {file = "flask-3.1.2.tar.gz", hash = "sha256:bf656c15c80190ed628ad08cdfd3aaa35beb087855e2f494910aa3774cc4fd87"},
+    {file = "flask-3.1.3-py3-none-any.whl", hash = "sha256:f4bcbefc124291925f1a26446da31a5178f9483862233b23c0c96a20701f670c"},
+    {file = "flask-3.1.3.tar.gz", hash = "sha256:0ef0e52b8a9cd932855379197dd8f94047b359ca0a78695144304cb45f87c9eb"},
 ]
 
 [package.dependencies]
@@ -2184,14 +2184,14 @@ test = ["pytest", "websockets"]
 
 [[package]]
 name = "werkzeug"
-version = "3.1.5"
+version = "3.1.6"
 description = "The comprehensive WSGI web application library."
 optional = false
 python-versions = ">=3.9"
 groups = ["main"]
 files = [
-    {file = "werkzeug-3.1.5-py3-none-any.whl", hash = "sha256:5111e36e91086ece91f93268bb39b4a35c1e6f1feac762c9c822ded0a4e322dc"},
-    {file = "werkzeug-3.1.5.tar.gz", hash = "sha256:6a548b0e88955dd07ccb25539d7d0cc97417ee9e179677d22c7041c8f078ce67"},
+    {file = "werkzeug-3.1.6-py3-none-any.whl", hash = "sha256:7ddf3357bb9564e407607f988f683d72038551200c704012bb9a4c523d42f131"},
+    {file = "werkzeug-3.1.6.tar.gz", hash = "sha256:210c6bede5a420a913956b4791a7f4d6843a43b6fcee4dfa08a65e93007d0d25"},
 ]
 
 [package.dependencies]


syncserver-common/src/middleware/sentry.rs
```diff
@@ -266,3 +287,25 @@ pub fn exception_from_reportable_error(err: &dyn ReportableError) -> sentry::pro
         ..Default::default()
     }
 }
+
+/// Log the error event instead when Sentry is not configured.
+fn log_event(event: &sentry::protocol::Event<'static>) {
+    let first_exception = event.exception.first();
```
pjenvey: These are reversed so the original error is last
```suggestion
    let first_exception = event.exception.last();
```

syncserver-common/src/middleware/sentry.rs
```diff
@@ -266,3 +287,25 @@ pub fn exception_from_reportable_error(err: &dyn ReportableError) -> sentry::pro
...
+    let error_type = first_exception.map_or("UnknownError", |e| e.ty.as_str());
+    let error_value = first_exception
+        .and_then(|e| e.value.as_deref())
+        .unwrap_or("No error message");
+    error!(
+        "{}", error_value;
```
pjenvey: I think it'd be worth adding the stacktrace here, separated by a newline (and ideally only when `RUST_BACKTRACE` is set to 1/full, which we always have) -- we want as much info as possible for when this comes from a customer's deployment, which we probably won't be able to inspect much ourselves.

It seems like the only nice output we can get from the sentry event's `Stacktrace` is json, so we might want to just pass the original error/or at least its stacktrace in to this fn for this.

Maybe this is a 2 story pointer..

syncserver-common/src/middleware/sentry.rs
chenba: 🤦

chenba:
> I'll also note actix-web claims to logs errors to WARN https://actix.rs/docs/errors/#error-logging but I haven't seen it do so, I wasn't able to see any in my attempt at running locally 😕

I changed it a `warn!`.

pjenvey:
> I changed it a `warn!`.

I think error makes more sense but more importantly, syncstorage's default `RUST_LOG` on prod is set to `error` (whereas tokenserver is apparently `info`), so we do want a more severe than warn.

pjenvey:
@chenba JR reminded me that the sentry client also catches Rust panics. I believe those should just work because they're already printed to stderr for us, but I think it would be a good idea to add a `panic!` to the error endpoint so we have the ability to easily verify that. autoconnect/endpoint already does so, e.g. https://github.com/mozilla-services/autopush-rs/blob/2b1a2991abf38c556ec0a190b37062c15b1a2609/autoendpoint/src/routes/health.rs#L142

chenba:
> @chenba JR reminded me that the sentry client also catches Rust panics. 

I don't think that's true in our case, since `panic` is not in https://github.com/mozilla-services/syncstorage-rs/blob/master/Cargo.toml#L70-L75

> I believe those should just work because they're already printed to stderr for us, but I think it would be a good idea to add a `panic!` to the error endpoint

👍

syncserver-common/src/middleware/sentry.rs
```diff
@@ -118,9 +138,16 @@ where
...
                 let event = event_from_actix_error::<E>(error);
-                let event_id = hub.capture_event(event);
-                trace!("event_id = {}", event_id);
+
+                if sentry_enabled {
+                    let event_id = hub.capture_event(event);
+                    trace!("event_id = {}", event_id);
+                } else {
+                    // log if Sentry is not configured
+                    log_event::<E>(&event, error, backtrace_enabled);
```
taddes, line 149: Hm, I wonder why the compiler didn't catch this, but the error should be a reference, given the signature calls for `error: &actix_web::Error`.

syncserver-common/src/middleware/sentry.rs
```diff
@@ -266,3 +293,42 @@ pub fn exception_from_reportable_error(err: &dyn ReportableError) -> sentry::pro
...
+        error
+            .as_error::<E>()
+            .and_then(|e| e.backtrace())
+            .map(|bt| format!("\n{bt:?}"))
+            .unwrap_or_default()
+    } else {
+        String::new()
+    };
+
+    error!(
```
taddes, line 321: Yeah, my question is should we be calling `error!` for every and all cases, or is `warn!` warranted in some instances.

pjenvey:
> I don't think that's true in our case, since `panic` is not in https://github.com/mozilla-services/syncstorage-rs/blob/master/Cargo.toml#L70-L75

Good catch, we should be. Logged https://mozilla-hub.atlassian.net/jira/software/c/projects/STOR/boards/3074

syncserver-common/src/middleware/sentry.rs
pjenvey: It's already a reference because it's from [response().error()](https://docs.rs/actix-web/latest/actix_web/struct.HttpResponse.html#method.error) here.

syncserver-common/src/middleware/sentry.rs
```diff
@@ -266,3 +293,42 @@ pub fn exception_from_reportable_error(err: &dyn ReportableError) -> sentry::pro
...
+        "{}{}", error_value, backtrace;
```
pjenvey, line 322: nit: maybe a separator?
```suggestion
        "{}{}{}", error_value, if backtrace.is_empty() { "" } else { "\n" }, backtrace;
```

syncserver/src/server/test.rs
```diff
@@ -841,3 +840,56 @@ async fn lbheartbeat_ttl_check() {
...
+            let log_msg = String::from_utf8(buffer.lock().unwrap().clone()).unwrap();
+            println!("{}", log_msg);
+
+            assert!(
+                log_msg.contains("Oh Noes!"),
+                "Unexpected error message: {log_msg}",
+            );
+            assert!(
+                log_msg.contains("test_error"),
+                "Unexpected backtrace in error message: {log_msg}"
```
pjenvey:
```suggestion
                "Unexpected error message: {log_msg}"
```

syncserver/src/server/test.rs
```diff
@@ -841,3 +840,56 @@ async fn lbheartbeat_ttl_check() {
...
+    use std::io::Write;
+    use std::sync::{Arc, Mutex};
+
+    temp_env::async_with_vars(
+        [("SENTRY_DSN", None::<&str>), ("RUST_BACKTRACE", Some("1"))],
+        async {
+            let app = init_app!().await;
+
+            struct CaptureDrain(Arc<Mutex<Vec<u8>>>);
+            impl slog::Drain for CaptureDrain {
```
pjenvey, line 856: Thanks for adding this test, I was going to suggest adding one here but didn't find such a capturing drain  anywhere so I gave up (not realizing it would be this simple to make one)

syncserver-common/src/middleware/sentry.rs
taddes: ah ok, thx

syncserver-common/src/middleware/sentry.rs
chenba: There's a `\n` in the format string on line 315 above.  Unless you want an empty line in between, which I'm also in favor of.

syncserver/src/server/test.rs
```diff
@@ -841,3 +840,56 @@ async fn lbheartbeat_ttl_check() {
...
+            let logger = slog::Logger::root(drain, slog::o!());
+            let _guard = slog_scope::set_global_logger(logger);
+
+            let req = test::TestRequest::get().uri("/__error__").to_request();
+            let resp = test::call_service(&app, req).await;
+
+            assert_eq!(resp.status(), StatusCode::INTERNAL_SERVER_ERROR);
+
+            let log_msg = String::from_utf8(buffer.lock().unwrap().clone()).unwrap();
+            println!("{}", log_msg);
```
chenba: Oops, I'll remove.

syncserver/src/server/test.rs
chenba: I understand the suggestion and your suggested message is correct...but I wonder if without that context, it won't be as clear to the readers that the test is look for the fn name in the backtrace.  I'll add a comment.

syncserver-common/src/middleware/sentry.rs
chenba: How do we determine when to call which macro?

syncserver-common/src/middleware/sentry.rs
pjenvey: my fault, I missed that one

syncserver/src/server/test.rs
pjenvey: I must have missed this was looking for the fn name, thanks, the comment helps. feel free to switch the message back if you want

PR #2076, opened 2026-02-19
feat: postgres node type

taddes:
## Description

Tokenserver includes a `node_type` field in its response payload, originally added for the client to report to telemetry: to track the usage of the previous syncstorage mysql implementation vs spanner (while migrating usage from the former to the latter).

This metric probably isn’t used any longer, but could still potentially be useful. We are reporting the incorrect default of spanner with the postgresql syncstorage backend, which should be adjusted.

This adds a new NodeType variant, simple helper function for Posgres, and set SYNC_TOKENSERVER__NODE_TYPE accordingly in the docker files and actions for proper builds and testing

## Testing

E2E tests should pass with correct node_type

## Issue(s)

Closes [STOR-467](https://mozilla-hub.atlassian.net/browse/STOR-467).

[STOR-467]: https://mozilla-hub.atlassian.net/browse/STOR-467?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tokenserver-common/src/lib.rs
+++ b/tokenserver-common/src/lib.rs
@@ -15,13 +15,3 @@ pub enum NodeType {
     #[serde(rename = "postgres")]
     Postgres,
 }
-
-impl NodeType {
-    pub fn spanner() -> Self {
-        Self::Spanner
-    }
-
-    pub fn postgres() -> Self {
-        Self::Postgres
-    }
-}
--- a/tokenserver-settings/src/lib.rs
+++ b/tokenserver-settings/src/lib.rs
@@ -38,7 +38,7 @@ pub struct Settings {
     /// The rate at which capacity should be released from nodes that are at capacity.
     pub node_capacity_release_rate: Option<f32>,
     /// The type of the storage nodes used by this instance of Tokenserver.
-    #[serde(default = "NodeType::spanner")]
+    #[serde(default)]
     pub node_type: NodeType,
     /// The label to be used when reporting Metrics.
     pub statsd_label: String,


tokenserver-common/src/lib.rs
```diff
@@ -11,12 +11,18 @@ pub enum NodeType {
     MySql,
     #[serde(rename = "spanner")]
     Spanner,
+    #[serde(rename = "postgres")]
+    Postgres,
 }
 
 impl NodeType {
```
chenba: What's the reason we have spanner and postgres but not mysql here?

taddes: Easy enough to add if we want it. Technically, I don't think don't need any of these since impls since it's just a setter function that could achieve the same with a single line. We don't actively build and deploy MySQL, so it didn't seem relevant to add.

chenba: Where do we use the `postgres` fn?

taddes: We don't. The original `spanner`  fn  is used by serde as a default for the serialization logic 😅. Was just matching the trait impl between the two. The `Default` takes care of making spanner the default. We may want it/not want it in the future.

pjenvey: I think we can kill the `spanner` fn and that `#[serde(default = "NodeType::spanner")]` -- I don't see why that Settings field needs it at all when the struct uses `#[serde(default)]` and its `Default` impl uses `NodeType::Spanner`

taddes: Removed, and updated to use the easier default macros.

taddes:
Need to land https://github.com/mozilla-services/syncstorage-rs/pull/2082 so CI will work, plus Phil's patch https://github.com/mozilla-services/syncstorage-rs/pull/2083

PR #2083, opened 2026-02-23
fix: re-add tokenserver/spanner scripts to the final docker

pjenvey:
## Description

poetry export doesn't install them, so these still need to be copied over. deployments expect the tokenserver tools in
/app/tools/tokenserver

the spanner scripts are not in active use yet but adding them back anyway

and install cargo nextest via binaries (as building it requires recent stable rustcs)

## Issue(s)

Closes STOR-489

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -128,8 +128,11 @@ commands:
   install-test-deps:
     steps:
       - run:
-          name: Install test dependencies
-          command: cargo install --locked cargo-nextest cargo-llvm-cov
+          name: Install cargo-nextest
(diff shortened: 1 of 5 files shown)


chenba:
I don't think that's the right issue number in the description

.circleci/config.yml
```diff
@@ -128,8 +128,11 @@ commands:
   install-test-deps:
     steps:
       - run:
-          name: Install test dependencies
-          command: cargo install --locked cargo-nextest cargo-llvm-cov
+          name: Install cargo-nextest
```
taddes, line 131: Good, like you had mentioned that will take care of our nextest issues 👍

PR #2087, opened 2026-02-26
feat: insert the first storage node record with env var

chenba:
Also updated some docs.

Closes STOR-487


tokenserver-postgres/src/pool.rs
```diff
@@ -131,6 +145,16 @@ impl DbPool for TokenserverPgPool {
         // As long as the sync service "sync-1.5" service record is in the database, this query should not fail,
         // unless there is a network failure or unpredictable event.
         let _ = self.init_service_id().await;
+
+        // Init the Sync 1.5 node record if there is an INIT_NODE_URL env var
+        if let Ok(node_url) = env::var("INIT_NODE_URL") {
```
chenba: Maybe we should check `SYNC_TOKENSERVER__RUN_MIGRATIONS` here as well?  Since we rely on that to insert the service record...

pjenvey: That run_migrations config option is more a special case for our existing stage/prod envs that migrated from the old python version, signaling the db is ready to go, so I'd say it isn't necessary.

Ideally that option would default to true, I think it was defaulted to false just to be extra safe. It would be nice to switch it, I think we would probably need to write a 'mocked' migrations table to our other environments before doing so, for again safety reasons.

docs/src/config.md
```diff
@@ -95,6 +95,8 @@ The following configuration options are available.
 
 | Env Var | Default Value | Description |
 | --- | --- | --- |
+| <span id="INIT_NODE_URL"></span>INIT_NODE_URL | None | The storage node URL, protocol + host,  to upsert into the `nodes` table on startup.  This is the origin where the service is hosted, e.g. "http://localhost:8000". |
```
pjenvey: I think these should include the standard tokenserver prefix

tokenserver-mysql/migrations/.diesel_lock
chenba, line 1: errr sure?

tokenserver-db/src/tests.rs
```diff
@@ -1352,28 +1434,5 @@ async fn db_pool() -> DbResult<Box<dyn DbPool>> {
     )?;
     pool.init().await?;
 
-    if settings.tokenserver.database_url.starts_with("mysql://") {
-        // Ensure the "sync-1.5" service
-        // TODO: tokenserver-mysql's migration should add this service
-        // entry for us (if possible)
```
chenba, line 1358: Say no more.

tokenserver-postgres/src/pool.rs
chenba: > Ideally that option would default to true, I think it was defaulted to false just to be extra safe. It would be nice to switch it, I think we would probably need to write a 'mocked' migrations table to our other environments before doing so, for again safety reasons.

Should we file an issue for this?

tokenserver-mysql/migrations/.diesel_lock
pjenvey: Looks like this is from the diesel migrations cli, so we don't need it committed

PR #2094, opened 2026-02-27
chore: stop uploading test results to gcs from circleci

chenba:
Because we've started uploading from gh workflows


chenba:
> 👍 Thanks @chenba . Maybe we'll want to add a little documentation ...

iiuc it was a test engineering thing.  maybe they already have some sort of docs on it?  I haven't checked.

PR #2097, opened 2026-03-02
chore: upgrade jsonwebtoken

chenba:


Cargo.toml
```diff
@@ -55,7 +55,7 @@ hostname = "0.4"
 hkdf = "0.12"
 hmac = "0.12"
 http = "1.4"
-jsonwebtoken = { version = "9.3", default-features = false }
+jsonwebtoken = { version = "10.3", default-features = false, features = ["rust_crypto"] }
```
chenba: 10.x requires the crypto backend to be specified.

Cargo.toml
```diff
@@ -55,7 +55,7 @@ hostname = "0.4"
 hkdf = "0.12"
 hmac = "0.12"
 http = "1.4"
-jsonwebtoken = { version = "9.3", default-features = false }
+jsonwebtoken = { version = "10.3", default-features = false, features = ["aws_lc_rs"] }
```
pjenvey, line 58: Just brain dumping here because I was wondering about this:

It looks like jsonwebtoken only added `aws_lc_rs` in this new 10.x version. It doesn't seem to clearly recommend it vs the alternative `rust_crypto` but it shows `aws_lc_rs` as the primary example in the README.

[rustls](https://github.com/rustls/rustls) does recommend its somewhat newer `aws_lc_rs` backend by default now over `ring`.

- https://www.memorysafety.org/blog/rustls-with-aws-crypto-back-end-and-fips/
- https://users.rust-lang.org/t/why-did-rustls-choose-aws-lc-rs-to-replace-ring-as-its-default-cryptography-library/134559

I wonder which one builds quicker, but there might not be much of a difference with both `aws-lc-rs/ring` relying on a large amount of underlying C code

pjenvey: and I'll note autopush went with `rust_crypto`, not sure why https://github.com/mozilla-services/autopush-rs/blob/101c7fbb1b55c0d815ac51377dde45d879862b78/autoendpoint/Cargo.toml#L24

chenba: @pjenvey how do you feel about pure Rust vs wrapped C code?

pjenvey: Depends on the situation, in some cases pure rust might build quicker than the alternative, which is preferable, but not always.

`aws_lc_rs` here seems fine 👍

PR #2108, opened 2026-03-10
feat: add FxA event webhook endpoint

chenba:
Add webhook endpoint to handle a couple of events from https://github.com/mozilla/fxa/tree/main/packages/fxa-event-broker.  It will succeed https://github.com/mozilla-services/syncstorage-rs/blob/HEAD/tools/tokenserver/process_account_events.py.

Closes STOR-128

### DIFF
--- a/tools/integration_tests/conftest.py
+++ b/tools/integration_tests/conftest.py
@@ -114,22 +114,6 @@ def setup_server_local_testing():
 
 
 @pytest.fixture(scope="session")
-def setup_server_local_testing_with_oauth():
-    """
-    Fixture to set up the server for local testing with OAuth.
-    This fixture sets the necessary environment variables and
-    starts the server.
-    """
-    _set_local_test_env_vars()
-
-    # Set OAuth-specific environment variables
-    os.environ["TOKENSERVER_AUTH_METHOD"] = "oauth"
-
-    # Start the server
-    yield from _server_manager()
-
-
-@pytest.fixture(scope="session")
 def setup_server_end_to_end_testing():
     """
     Fixture to set up the server for end-to-end testing.
--- a/tools/integration_tests/tokenserver/test_authorization.py
+++ b/tools/integration_tests/tokenserver/test_authorization.py
@@ -6,7 +6,7 @@ import unittest
 from integration_tests.tokenserver.test_support import TestCase
 
 
-@pytest.mark.usefixtures("setup_server_local_testing_with_oauth")
+@pytest.mark.usefixtures("setup_server_local_testing")
 class TestAuthorization(TestCase, unittest.TestCase):
     def setUp(self):
         super(TestAuthorization, self).setUp()
--- a/tools/integration_tests/tokenserver/test_misc.py
+++ b/tools/integration_tests/tokenserver/test_misc.py
@@ -9,7 +9,7 @@ from integration_tests.tokenserver.test_support import TestCase
 MAX_GENERATION = 9223372036854775807
 
 
-@pytest.mark.usefixtures("setup_server_local_testing_with_oauth")
+@pytest.mark.usefixtures("setup_server_local_testing")
 class TestMisc(TestCase, unittest.TestCase):
     def setUp(self):
         super(TestMisc, self).setUp()
--- a/tools/integration_tests/tokenserver/test_node_assignment.py
+++ b/tools/integration_tests/tokenserver/test_node_assignment.py
@@ -8,7 +8,7 @@ from integration_tests.tokenserver.test_support import TestCase
 from sqlalchemy.sql import text as sqltext
 
 
-@pytest.mark.usefixtures("setup_server_local_testing_with_oauth")
+@pytest.mark.usefixtures("setup_server_local_testing")
 class TestNodeAssignment(TestCase, unittest.TestCase):
     def setUp(self):
         super(TestNodeAssignment, self).setUp()


syncserver/src/tokenserver/extractors.rs
```diff
@@ -345,9 +344,9 @@ impl FromRequest for DbPoolWrapper {
 }
 
 /// An authentication token as parsed from the `Authorization` header.
-/// OAuth tokens are opaque to Tokenserver and must be verified via FxA.
+/// Signed JWTs can be verified locally or via FxA.
```
chenba, line 347: I did some renaming in numerous spots since JWTs can be used outside of an OAuth context.

tokenserver-settings/src/lib.rs
```diff
@@ -61,6 +63,9 @@ pub struct Settings {
     /// The capacity value for the node record created for `init_node_url`.  Only used if
     /// `init_node_url` is set.
     pub init_node_capacity: i32,
+    /// Whether to enable the FxA webhook endpoint.
+    /// Defaults to false.
+    pub fxa_webhook_enabled: bool,
```
chenba, line 68: I'll update the config.md to add this.

syncserver/src/tokenserver/handlers.rs
```diff
@@ -290,6 +291,54 @@ async fn update_user(
...
+    let service_id = db
+        .get_service_id(tokenserver_db::params::GetServiceId {
+            service: SYNC_SERVICE_NAME.to_owned(),
+        })
+        .await?
+        .id;
+    let events = match claims.events.as_object() {
+        Some(map) => map.clone(),
+        None => return Ok(HttpResponse::Ok().finish()),
+    };
```
pjenvey: looks like the clone isn't needed either
```suggestion
    let Some(events) = claims.events.as_object() else {
        return Ok(HttpResponse::Ok().finish());
    };
```

tokenserver-auth/src/crypto.rs
```diff
@@ -87,14 +88,14 @@ pub enum OAuthVerifyError {
...
-impl OAuthVerifyError {
+impl JWTVerifyError {
     pub fn metric_label(&self) -> &'static str {
         match self {
-            Self::ExpiredSignature => "oauth.error.expired_signature",
-            Self::TrustError => "oauth.error.trust_error",
-            Self::InvalidKey => "oauth.error.invalid_key",
-            Self::InvalidSignature => "oauth.error.invalid_signature",
-            Self::DecodingError => "oauth.error.decoding_error",
+            Self::ExpiredSignature => "jwt.error.expired_signature",
```
pjenvey, line 94: I'll note we don't seem to graph any of these in our dashboards so 👍 to renaming

tokenserver-db/src/mock.rs
```diff
@@ -1,18 +1,26 @@
...
 impl MockDbPool {
     pub fn new() -> Self {
-        MockDbPool
+        MockDbPool::default()
+    }
+
+    pub fn with_capture() -> (Self, Arc<Mutex<Vec<params::PutUser>>>) {
+        let pool = MockDbPool::default();
+        let put_user_calls = Arc::clone(&pool.put_user_calls);
+        (pool, put_user_calls)
```
chenba: Maybe we should consider using `mockall` instead.  But I'll leave that to another PR, if we want to make that change.

tokenserver-auth/src/crypto.rs
```diff
@@ -173,3 +173,75 @@ impl TryFrom<Jwk> for JWTVerifierImpl {
...
+pub struct SETVerifierImpl {
+    key: DecodingKey,
+    validation: Validation,
+}
+
+impl SETVerifierImpl {
+    pub fn new(jwk: &Jwk, client_id: &str) -> Result<Self, JWTVerifyError> {
+        let decoding_key = DecodingKey::from_jwk(jwk).map_err(|_| JWTVerifyError::InvalidKey)?;
+        let mut validation = Validation::new(Algorithm::RS256);
+        validation.set_audience(&[client_id]);
```
pjenvey, line 197: Maybe we should verify `iss` as well?
```suggestion
        validation.set_audience(&[client_id]);
        validation.set_issuer(&[<SYNC_TOKENSERVER__FXA_OAUTH_SERVER_URL?>]);
```

syncserver/src/tokenserver/handlers.rs
```diff
@@ -290,6 +291,54 @@ async fn update_user(
...
+
+    for event_type in events.keys() {
+        match event_type.as_str() {
+            "https://schemas.accounts.firefox.com/event/delete-user" => {
+                db.put_user(tokenserver_db::params::PutUser {
```
pjenvey: We need a new `Db::retire_user` here like the one in `database.py` called by `process_account_events.py`, which sets a `replaced_at` on the record

syncserver/src/tokenserver/handlers.rs
```diff
@@ -290,6 +291,54 @@ async fn update_user(
...
+                    keys_changed_at: None,
+                })
+                .await?;
+            }
+            "https://schemas.accounts.firefox.com/event/password-change" => {
+                if let Some(change_time_ms) = events[event_type]
+                    .get("changeTime")
+                    .and_then(|t| t.as_i64())
+                {
+                    db.put_user(tokenserver_db::params::PutUser {
```
pjenvey: I _think_ we also need a new `Db::update_user_generation` for this (see my comment in https://mozilla-hub.atlassian.net/browse/STOR-128?focusedCommentId=1281438)

chenba:
@pjenvey Thanks for the review!  I think I've addressed all the comments.

syncserver/src/tokenserver/handlers.rs
```diff
@@ -290,6 +291,51 @@ async fn update_user(
...
+            "https://schemas.accounts.firefox.com/event/delete-user" => {
+                db.retire_user(RetireUser {
+                    service_id,
+                    email: format!("{}@{}", claims.sub, state.fxa_email_domain),
+                })
+                .await?;
+            }
+            "https://schemas.accounts.firefox.com/event/password-change" => {
+                if let Some(change_time_ms) = events[event_type]
+                    .get("changeTime")
```
pjenvey, line 306: Just now looking closer at the payload, I fear this value might not be the same `generation` value we're expecting.

I'm assuming the SQS event's generated [here](https://github.com/mozilla/fxa/blob/807c636a4ae60a24f583d883fb61f50567cd145d/packages/fxa-auth-server/lib/routes/password.ts#L450) whereas I'm not sure about the `SET`, probably [here](https://github.com/mozilla/fxa/blob/807c636a4ae60a24f583d883fb61f50567cd145d/packages/fxa-event-broker/src/queueworker/queueworker.service.ts#L196)?

chenba: :'(  I think you're right.  I'll bug FxA about it.  Thank you for being so thorough.

pjenvey: @chenba We can definitely merge as is though because enterprise will currently only send us deletion events

PR #2127, opened 2026-03-13
chore: upgrade Python for all utils and refactor

taddes:
## Description

Testing result of workflow Python update

## Issue(s)

Closes [link](link).

### DIFF
--- a/tools/postgres/test_purge_ttl.py
+++ b/tools/postgres/test_purge_ttl.py
@@ -2,8 +2,6 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-from __future__ import annotations
-
 from argparse import Namespace
 from unittest.mock import MagicMock, Mock, patch
 
--- a/tools/postgres/purge_ttl.py
+++ b/tools/postgres/purge_ttl.py
@@ -2,8 +2,6 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-from __future__ import annotations
-
 import argparse
 import logging
 import os
--- a/tools/spanner/count_expired_rows.py
+++ b/tools/spanner/count_expired_rows.py
@@ -5,8 +5,6 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-from __future__ import annotations
-
 import sys
 import logging
 from typing import Any
--- a/tools/spanner/purge_ttl.py
+++ b/tools/spanner/purge_ttl.py
@@ -4,8 +4,6 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-from __future__ import annotations
-
 import argparse
 import logging
 import os
--- a/tools/spanner/test_count_expired_rows.py
+++ b/tools/spanner/test_count_expired_rows.py
@@ -1,5 +1,3 @@
-from __future__ import annotations
-
 import logging
 from unittest.mock import MagicMock
 
--- a/tools/spanner/test_purge_ttl.py
+++ b/tools/spanner/test_purge_ttl.py
@@ -1,5 +1,3 @@
-from __future__ import annotations
-
 import argparse
 from types import SimpleNamespace
 from unittest import mock
--- a/tools/spanner/test_utils.py
+++ b/tools/spanner/test_utils.py
@@ -1,5 +1,3 @@
-from __future__ import annotations
-
 import pytest
 
 from tools.spanner.utils import ids_from_env
--- a/tools/spanner/utils.py
+++ b/tools/spanner/utils.py
@@ -4,8 +4,6 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-from __future__ import annotations
-
 import os
 from enum import auto, Enum
 from urllib import parse


.github/workflows/checks.yml
```diff
@@ -10,6 +10,10 @@ on:
     branches:
       - "**"
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
```
chenba, line 15: This and other similar edits are unrelated?

docs/src/tools/make_hawk_token.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+See [`tools/examples/put.bash`](https://github.com/mozilla-services/syncstorage-rs/blob/master/tools/examples/put.bash) for a full curl example using this output.
+
+---
+
+## Notes
+
+- The `--secret` value must match the `SYNC_MASTER_SECRET` configured in the running syncstorage-rs instance.
+- The generated token is scoped to the exact URI and method — a token generated for `GET /foo` is not valid for `PUT /foo`.
```
chenba, line 91: This sound right from a quick read but I'm no hawk auth pro.  I also cannot tell if this is related to the Python upgrades.

tools/hawk/make_hawk_token.py
```diff
@@ -7,8 +7,7 @@
 valid_header_with_querystring.
 
 The latter modifies the query string which changes the mac/nonce and
-potentially ts values (in the Hawk header).
-
+potentially its values (in the Hawk header).
```
chenba: The original text was correct.  'ts' as in timestamp.

tools/hawk/pyproject.toml
```diff
@@ -7,7 +7,7 @@ authors = [
 ]
 license = "Mozilla Public License Version 2.0"
 readme = "README.md"
-requires-python = ">=3.9"
+requires-python = ">=3.12"
```
chenba, line 10: I see 3.14 (from 3.12) so far except here.  Why is that?

tools/hawk/test_make_hawk_token.py
```diff
@@ -0,0 +1,208 @@
...
+    result = metrics_hash(make_args(), "somevalue")
+    assert isinstance(result, str)
+    int(result, 16)  # raises ValueError if not valid hex
+
+
+def test_metrics_hash_is_deterministic() -> None:
+    """Same input always produces the same hash."""
+    assert metrics_hash(make_args(), "abc") == metrics_hash(make_args(), "abc")
+
+
```
chenba, line 45: Better than no tests I guess.  But if `metrics_hash` returns the same hex string every time the two tests above would still pass.

chenba: ...and the two tests below as well.

tools/hawk/test_make_hawk_token.py
```diff
@@ -0,0 +1,208 @@
...
+    assert local == email
+
+
+def test_metrics_hash_str_hmac_key_coerced_to_bytes() -> None:
+    """A string hmac_key is coerced to bytes in-place before hashing."""
+    args = make_args(hmac_key="foo")
+    result = metrics_hash(args, "value")
+    assert isinstance(result, str)
+    assert args.hmac_key == b"foo"
+
```
chenba, line 61: If we really want to test `metrics_hash` then we need to compare its output against known values.

.github/workflows/checks.yml
taddes: Yes, this was testing for the other PR. I've enabled it here to make quicker checks and shutting down the runs when making dep checks

tools/hawk/pyproject.toml
taddes: With Poetry, you don't have to explicitly specify the version your using, so long as the constraint, which is greater than 3.12 is sufficient. Sometimes Poetry will scream at you with certain dependencies (like some we use that are outdated) if the version specifier is too high.

docs/src/tools/make_hawk_token.md
taddes: Yeah, I wanted to add the docs in the `tools/` dir to the ones in our main docs repo, and did a little revision across the board.

tools/hawk/test_make_hawk_token.py
```diff
@@ -0,0 +1,208 @@
...
+# --- create_token ---
+
+
+def test_create_token_returns_four_tuple() -> None:
+    """create_token returns (token, key, expires, salt)."""
+    assert len(create_token(make_args())) == 4
+
+
+def test_create_token_expires_in_future() -> None:
+    """The expiry timestamp is strictly after the current time."""
```
chenba: Not so sure about "strictly".  `create_token` doesn't even validate the duration argument.

tools/hawk/test_make_hawk_token.py
```diff
@@ -0,0 +1,208 @@
...
+
+def test_create_token_data_roundtrips() -> None:
+    """Token payload decoded with tokenlib contains the original uid, node, and FxA fields."""
+    args = make_args()
+    token, _, _, _ = create_token(args)
+    decoded = tokenlib.parse_token(token, secret=args.secret)
+    assert decoded["uid"] == args.uid
+    assert decoded["node"] == args.node
+    assert decoded["fxa_uid"] == args.fxa_uid
+    assert decoded["fxa_kid"] == args.fxa_kid
```
chenba, line 101: Why not test the expiration this way too?

tools/postgres/test_purge_ttl.py
```diff
@@ -75,63 +87,67 @@ def test_filters_empty_collection_ids(self):
...
-        actual_condition = get_expiry_condition(args)
-        assert "DATE_TRUNC" in actual_condition
-        assert "UTC" in actual_condition
+        actual = get_expiry_condition(args)
+        assert "DATE_TRUNC" in actual
+        assert "UTC" in actual
 
-    def test_invalid_expiry_mode(self):
+    def test_invalid_expiry_mode(self) -> None:
+        """An unrecognised expiry mode raises an exception."""
```
chenba, line 101: How very British.

taddes: I can't help but _honour_ my Canadian heritage 🇨🇦  @chenba

tools/hawk/test_make_hawk_token.py
taddes: Good point, I then pre-computed a value using `hmac` and `hashlib`, put it in the test and compare against it, so that's better 👍 

basically:
```python
import hmac
import hashlib

res = hmac.new(b'foo', b'value', hashlib.sha256).hexdigest())
```

.github/workflows/mysql.yml
```diff
@@ -10,9 +10,13 @@ on:
...
   workflow_dispatch: {}
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+
 env:
   RUST_VERSION: "1.91" # RUST_VER
-  PYTHON_VERSION: "3.12" # PY_VER
+  PYTHON_VERSION: "3.14" # PY_VER
```
pjenvey: I'll note again -- AFAICT this and `RUST_VERSION` aren't actually used anywhere

taddes: Aha, good point, it's a hangover from the original run of setting up the actions. Only the rust and python setup actions use this. I can flag this on my current workflow refactor and make sure to take it out, since these will move into the root workflow.

taddes: That or I can yank them out here, since they're not doing anything 👍

tools/spanner/pyproject.toml
```diff
@@ -7,26 +7,37 @@ authors = [
...
+mypy = "^1.19.1"
 pydocstyle = "^6.3.0"
-ruff = "^0.12.7"
-black = "^25.1.0"
-bandit = "^1.8.3"
-isort = "^6.0.1"
+ruff = "^0.15.6"
+black = "^26.3.1"
+bandit = "^1.9.4"
+isort = "^8.0.1"
```
chenba, line 27: I'm a Python novice so maybe I simply don't understand how any of this works...but I don't think some of these deps are used?  How and where is `isort` invoked?

taddes: Ah yes, I'm probs the only one using these and running them with poetry... I had meant to add makefile targets to use these utilities.  I will make sure to do that in another PR and include it as a part of our Python checks workflows since yeah, not doing us much good. 😹 

 In short, `ruff` does most of the heavy lifting with format checks and linting, but isort makes the imports PEP8 compliant, bandit checks for vulnerabilities, pydocstyle provides utilities for code documentation, black is a formatter, and mypy gives us type safe-ish Python with proper annotation checks and validation.

tools/postgres/test_purge_ttl.py
```diff
@@ -2,38 +2,47 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-import pytest
-from unittest.mock import Mock, MagicMock, patch
+from __future__ import annotations
+
```
pjenvey: I think these are no longer needed on Python 3.14, which we're moving to?
```suggestion
```

taddes: Yeah good point then, looked it up to be sure and it's the default to support the PEP 649 lazy annotation evaluation. Thx for catching 👍

PR #2139, opened 2026-03-16
chore: rmv oauth tokenserver auth method

taddes:
## Description

Remove the `TOKENSERVER_AUTH_METHOD` env var, since alternating between OAuth and BrowserID is no longer required.

## Issue(s)

Closes [STOR-510](https://mozilla-hub.atlassian.net/browse/STOR-510).

[STOR-510]: https://mozilla-hub.atlassian.net/browse/STOR-510?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tools/integration_tests/conftest.py
+++ b/tools/integration_tests/conftest.py
@@ -114,19 +114,6 @@ def setup_server_local_testing():
 
 
 @pytest.fixture(scope="session")
-def setup_server_local_testing_with_oauth():
-    """
-    Fixture to set up the server for local testing with OAuth.
-    This fixture sets the necessary environment variables and
-    starts the server.
-    """
-    _set_local_test_env_vars()
-
-    # Start the server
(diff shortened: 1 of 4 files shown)


tools/integration_tests/conftest.py
```diff
@@ -122,9 +122,6 @@ def setup_server_local_testing_with_oauth():
     """
     _set_local_test_env_vars()
 
-    # Set OAuth-specific environment variables
-    os.environ["TOKENSERVER_AUTH_METHOD"] = "oauth"
```
pjenvey, line 126: And this should make `setup_server_local_testing_with_oauth` the same as `setup_server_local_testing`, so we can kill the former entirely

Issue #559, opened 2020-04-01
Redo #556 on top of 0.2.8 release


pjenvey:
The intention for #556 was to tag a 0.2.1.1 (0.2.1 + the sole addition of the MAX_TTL fix). However 0.2.1.1 isn't a valid semver so cargo strictly disallows it. We could also use the Rust purge_ttl on prod sooner rather than later.

Rather than doing a weird version number shuffle, instead I'm just going to disable the 0.2.8 bits:

- sort order changes
- encoded offset (which relies on the sort order change): stop rendering and parsing a ':'

And cut a 0.2.9. w/ purge_ttl and the other couple minor features from 0.2.2.

pjenvey:
Closed in #560

PR #2145, opened 2026-03-18
feat: (re-)enable timestamp+offset based pagination optimization

chenba:
https://github.com/mozilla-services/syncstorage-rs/issues/53 enabled 
https://github.com/mozilla-services/syncstorage-rs/issues/559 disabled

And this patch enables it for MySQL and Postgres but not Spanner.

Closes STOR-473

### DIFF
--- a/syncstorage-postgres/src/db/mod.rs
+++ b/syncstorage-postgres/src/db/mod.rs
@@ -326,8 +326,8 @@ macro_rules! bsos_query {
             // returned in those cases
 
             let limit = limit.unwrap_or(-1);
-            let did_overflow = limit  >= 0 && items.len() > limit as usize;
-            if  did_overflow {
+            let did_overflow = limit >= 0 && items.len() > limit as usize;
+            if did_overflow {
                 items.pop();
             }
             (items, did_overflow, limit, numeric_offset)


syncstorage-db-common/src/params.rs
```diff
@@ -68,27 +68,16 @@ pub struct Offset {
 
 impl Display for Offset {
     fn fmt(&self, fmt: &mut Formatter) -> Result<(), fmt::Error> {
```
pjenvey, line 70: I don't recall why we ended up/need a duplicate version of `Offset` here, maybe we can try killing the other one (in a separate issue)

syncstorage-postgres/src/db/mod.rs
```diff
@@ -320,13 +327,11 @@ macro_rules! bsos_query {
             // returned in those cases
 
             let limit = limit.unwrap_or(-1);
-            let next_offset = if limit >= 0 && items.len() > limit as usize {
+            let did_overflow = limit  >= 0 && items.len() > limit as usize;
+            if  did_overflow {
```
pjenvey:
```suggestion
            let did_overflow = limit >= 0 && items.len() > limit as usize;
            if did_overflow {
```

syncstorage-mysql/src/db/db_impl.rs
```diff
@@ -340,14 +349,6 @@ impl Db for MysqlDb {
         // an error. We "fudge" a bit here by taking the id order as a secondary, since
         // that is guaranteed to be unique by the client.
         query = match params.sort {
-            // issue559: Revert to previous sorting
-            /*
-            Sorting::Index => query.order(bso::id.desc()).order(bso::sortindex.desc()),
-            Sorting::Newest | Sorting::None => {
```
pjenvey, line 346: There's a subtle bug, this small change here to make `None` act like `Newest` is I think at least one solution:

`encode_next_offset` returns the encoded offset for every type of sort except `Index`, including `None`. However that special offset isn't going to make much sense if we're not applying the timestamp above (in the `Some(ts)` `_ => {}` arm) -- and we probably also need to enforce some kind of ordering here for the pagination to make sense.

I thought we had a test somewhere for such a case but I guess not?

chenba: Thank you for catching that.  I've made `None` behave like `Newest`.

syncstorage-db-common/src/params.rs
pjenvey: https://mozilla-hub.atlassian.net/browse/STOR-522

PR #2147, opened 2026-03-19
feat: add logging for acct webhook handler

chenba:
Adding the logging for the account event webhook handler.

Closes STOR-515


syncserver/src/tokenserver/handlers.rs
```diff
@@ -293,29 +293,33 @@ pub async fn handle_fxa_events(
...
                     })
                     .await?;
                 }
             }
-            _ => {}
+            _ => {
+                info!(
+                    "Dropping unhandled event type {:?} for {}",
+                    event_type, email
+                );
```
chenba, line 321: In the Python SQS event handler this is a warning.  But since we are knowingly not handling the rest of the event types, I felt that's not quite right.

PR #2149, opened 2026-03-20
chore: build and push "directly" instead of using mozilla-it/deploy-actions

chenba:
We relied on mozilla-it/deploy-actions's build-and-push to push a couple of Docker images to another GAR in GCP.  That duplicate the image builds.  With this patch we'll define the build-and-push actions ourselves.

Additionally, we'll push a 'latest' tag to the non-enterprise image registries on a git tag push.

Closes STOR-499, STOR-509

### DIFF
--- a/.github/workflows/mozcloud-publish.yaml
+++ b/.github/workflows/mozcloud-publish.yaml
@@ -1,15 +1,8 @@
-# Mozilla Deploy Actions url: <https://github.com/mozilla-it/deploy-actions>
-# Note: even though Mozilla maintains the above actions, it is still suggested
-# when upgrading to use the full commit SHA and comment with version.
-# See <https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions>
-# Ex. `mozilla-it/deploy-actions/.github/workflows/build-and-push.yml@4784cb70739a4f32ce010921f60fb1ebbc791a38 # v6.2.2`
-name: Build, Tag and Push Container Images to GAR Repository
+name: Build, Tag and Push Container Images to GAR
 
 on:
   pull_request:
     types: [opened, labeled, unlabeled, synchronize]
-    # paths:
-    #     - '**/sync*/**'
   push:
     branches:
       - master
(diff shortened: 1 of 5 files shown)


.github/workflows/mozcloud-publish.yaml
```diff
@@ -166,14 +197,44 @@ jobs:
...
+      - name: GCP auth
+        uses: google-github-actions/auth@7c6bc770dae815cd3e89ee6cdf493a5fab2cc093 # v3
+        with:
+          workload_identity_provider: "projects/${{ vars.GCPV2_WORKLOAD_IDENTITY_POOL_PROJECT_NUMBER }}/locations/global/workloadIdentityPools/github-actions/providers/github-actions"
+          service_account: "${{ vars.SERVICE_ACCOUNT_NAME || 'artifact-writer' }}@moz-fx-fx-enterprise-prod.iam.gserviceaccount.com"
+          token_format: access_token
+          create_credentials_file: false
+
+      - name: Configure GAR
+        run: gcloud auth configure-docker us-docker.pkg.dev --quiet
```
pjenvey: I notice the mozcloud docker-push action does a `docker/login-action` here instead: https://github.com/mozilla-it/deploy-actions/blob/b9d76236c1d1dd89daab4ea15651cbe76574267a/docker-push/action.yml#L39 I'm not sure it matters?

chenba:
> (I noticed remote-settings basically rolls their own https://github.com/mozilla/remote-settings/blob/main/.github/workflows/publish-ent.yaml) 
> 

Thanks, I'll take a look.

.github/workflows/mozcloud-publish.yaml
taddes: Yeah, this would be good to update since I was planning on breaking those out in the broader workflow updates.

chenba:
The two GAR pushes are on different steps; it appears docker/login-action won't allow two auth sessions to us-docker.pkg.dev simultaneously.  Not a big deal since the later step can rely on cache.

.github/workflows/mozcloud-publish.yaml
```diff
@@ -48,46 +41,116 @@ jobs:
...
-        SYNCSTORAGE_DATABASE_BACKEND=spanner
-        MYSQLCLIENT_PKG=libmysqlclient-dev
-      should_tag_ghcr: true
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          persist-credentials: false
+
+      - uses: ./.github/actions/build-and-push
```
pjenvey: This evolved into more of a `setup-build-and-push` action, I think it could use a renaming to something along those lines?

chenba: ✅

PR #2165, opened 2026-03-26
perf: drop/de-allocate req body str explicitly after parsing

chenba:
In the BsoBodies::from_request extractor, the request body string is dropped at the end of its scope.  But it can be explicitly dropped once it's no longer needed.

Low risk so seems worth trying.

### DIFF
--- a/syncserver/src/web/extractors/bso_bodies.rs
+++ b/syncserver/src/web/extractors/bso_bodies.rs
@@ -134,6 +134,8 @@ impl FromRequest for BsoBodies {
                 return future::err(make_error());
             };
 
+            drop(body);
+
             // Validate all the BSO's, move invalid to our other list. Assume they'll all make
             // it with our pre-allocation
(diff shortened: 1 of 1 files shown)


chenba:
@pjenvey what do you think?  Low risk enough to just try it?

pjenvey:
> @pjenvey what do you think? Low risk enough to just try it?

Yea, definitely. It might be worth refactoring this extractor to use an async block first (or afterwards is ok), then you might be able to easily add a new scope (nested `{ }` block) to do it instead of drop. I thought we had converted all extractors to async/await for https://github.com/mozilla-services/syncstorage-rs/issues/544 but maybe I was wrong..

chenba:
> 
> ... I thought we had converted all extractors to async/await for #544 but maybe I was wrong..

I looked at the few PRs linked from that issue and none of them were extractor related.

Issue #289, opened 2019-10-17
Switch sentry to curl transport


pjenvey:
Sentry's default reqwest transport has issues on some linux environments:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error(NativeTls(Normal(ErrorStack([]))))', src/libcore/result.rs:1084:5
...
  11:     0x556bd2db0e31 - core::result::Result<T,E>::unwrap::h191e585b1bfacfc4
                               at /rustc/625451e376bb2e5283fc4741caa0a3e8a2ca4d54/src/libcore/result.rs:852
  12:     0x556bd2d83b2f - sentry::transport::ReqwestHttpTransport::new_internal::http_client::{{closure}}::h6bd2b0c50fe7e41a
                               at /home/pjenvey/.cargo/registry/src/github.com-1ecc6299db9ec823/sentry-0.17.0/src/transport.rs:315
  13:     0x556bd2d85666 - core::option::Option<T>::unwrap_or_else::hce3c1cddc352ea69
                               at /rustc/625451e376bb2e5283fc4741caa0a3e8a2ca4d54/src/libcore/option.rs:419
  14:     0x556bd2d8376d - sentry::transport::ReqwestHttpTransport::new_internal::http_client::h412c3e582852a265
                               at /home/pjenvey/.cargo/registry/src/github.com-1ecc6299db9ec823/sentry-0.17.0/src/transport.rs:305
  15:     0x556bd2d81c2c - sentry::transport::ReqwestHttpTransport::new_internal::h5b89b375874c3044
                               at /home/pjenvey/.cargo/registry/src/github.com-1ecc6299db9ec823/sentry-0.17.0/src/transport.rs:166
  16:     0x556bd2d81a09 - sentry::transport::ReqwestHttpTransport::new::h90a0d3b4f2fbe5fe
                               at /home/pjenvey/.cargo/registry/src/github.com-1ecc6299db9ec823/sentry-0.17.0/src/transport.rs:147
  17:     0x556bd2d815de - <sentry::transport::DefaultTransportFactory as sentry::transport::TransportFactory>::create_transport::hc2ee5b29a79080bb
                               at /home/pjenvey/.cargo/registry/src/github.com-1ecc6299db9ec823/sentry-0.17.0/src/transport.rs:108
  18:     0x556bd2db9de3 - sentry::client::Client::with_options::h53f6e69f860ead1b
[...]

PR #2166, opened 2026-03-27
chore: capture sentry backtraces

pjenvey:
## Description

(we were missing the "capture" feature)

and switch back to sentry's now longstanding default transport of reqwest+rustls, which avoids issues w/ our grpc client's included boringssl that occurred w/ reqwest+native-tls

## Issue(s)

Closes STOR-493


syncserver/src/main.rs
```diff
@@ -34,16 +34,10 @@ async fn main() -> Result<(), Box<dyn Error>> {
     debug!("Starting up...");
 
     // Set SENTRY_DSN environment variable to enable Sentry.
-    // Avoid its default reqwest transport for now due to issues w/
-    // likely grpcio's boringssl
-    let curl_transport_factory = |options: &sentry::ClientOptions| {
-        Arc::new(sentry::transports::CurlHttpTransport::new(options)) as Arc<dyn sentry::Transport>
-    };
```
chenba, line 41: What changed so we no longer need this?  Sorry I don't have the context.

pjenvey:
> lgtm, though I'll have to get brief context on this from you sometime. grpcio ftw. Is this basically where `reqwest` and `grpcio's` `boringssl` fight over TLS internals, so `libcurl` was used as a side-step since it has its own independent TLS stack that didn't clash?

Yea, exactly. @chenba the original issue is here: https://github.com/mozilla-services/syncstorage-rs/issues/289 as Taddes said, reqwest with its `native-tls` backend (which usually ends up using the openssl installed on your system) somehow seemed to "conflict" w/ the boringssl vendored with our grpc client used for Spanner, so the easy fix was to switch to a different transport.

Apparently sentry changed its default a long time ago, not that long after our original workaround, to reqwest w/ the pure rust `rustls` backend. I probably should have noticed this when autopush started using the same grpc client a couple years ago and never needed the curl transport 😅

PR #2186, opened 2026-04-02
refactor: use configurable running service for Python integration tests

chenba:
Instead of configuring and starting server instances within Python as test fixtures, simply testing against a running server.  A mix of make target and docker-compose changes is used to achieve the same level of test coverage that previously relied on the (re-)configure and (re-)start of the services in conftest.py.

A new `make` target, 'run_local_e2e_tests', can be used to run the integration tests locally.  However, the stage FxA JWT validation tests in test_e2e.py are excluded.  Those tests rely on the JWK configuration of the Token Server and the stage FxA API, making them less "local". Anyone working on that specific integration can certainly invoke those tests themselves.

This patch also:
 - deletes some duplicate docs
 - moves the docker-compose yamls into a dir named 'docker'

Closes STOR-511


Makefile
```diff
@@ -69,61 +69,121 @@ clean:
...
+	env -u SYNC_TOKENSERVER__FXA_OAUTH_PRIMARY_JWK__USE \
+	env -u SYNC_TOKENSERVER__FXA_OAUTH_PRIMARY_JWK__N \
+	env -u SYNC_TOKENSERVER__FXA_OAUTH_PRIMARY_JWK__E \
+	RESULTS_FILENAME=mysql_no_jwk_integration_results.xml \
 	docker compose \
-		-f docker-compose.mysql.yaml \
-		-f docker-compose.e2e.mysql.yaml \
+		-f docker/docker-compose.mysql.yaml \
+		-f docker/docker-compose.e2e.mysql.yaml \
+		-f docker/docker-compose.e2e.mysql.no-jwk-cache.yaml \
```
chenba: Since we  cannot _unset_ env vars with docker-compose, the order here matters, which is why we unset JWK env vars from the invoking shell and run the tests with no local JWKs, and then set the JWK env vars for the tests that follow.

Makefile
```diff
@@ -180,6 +240,14 @@ run_token_server_integration_tests:
 	poetry install --no-root --without dev
 	poetry run pytest tools/tokenserver --junit-xml=${INTEGRATION_JUNIT_XML}
 
+run_local_e2e_tests:
+	PYTHONPATH=$(PWD)/tools \
+	SYNC_MASTER_SECRET=$${SYNC_MASTER_SECRET:-secret0} \
+	SYNC_TOKENSERVER__FXA_OAUTH_SERVER_URL=$${SYNC_TOKENSERVER__FXA_OAUTH_SERVER_URL:-http://localhost:6000} \
+	TOKENSERVER_HOST=$${TOKENSERVER_HOST:-http://localhost:8000} \
+	poetry -C tools/integration_tests \
+		run pytest . --ignore=tokenserver/test_e2e.py
```
chenba, line 249: Naming is hard.  `e2e` for the make target here is consistent with the docker-compose based tests but kinda amusing we are skipping a file named `test_e2e`.

Makefile
```diff
@@ -180,6 +240,14 @@ run_token_server_integration_tests:
 	poetry install --no-root --without dev
 	poetry run pytest tools/tokenserver --junit-xml=${INTEGRATION_JUNIT_XML}
 
+run_local_e2e_tests:
+	PYTHONPATH=$(PWD)/tools \
+	SYNC_MASTER_SECRET=$${SYNC_MASTER_SECRET:-secret0} \
+	SYNC_TOKENSERVER__FXA_OAUTH_SERVER_URL=$${SYNC_TOKENSERVER__FXA_OAUTH_SERVER_URL:-http://localhost:6000} \
+	TOKENSERVER_HOST=$${TOKENSERVER_HOST:-http://localhost:8000} \
```
chenba, line 247: Since we support multiple db backends, I did not include default db urls here.

tools/integration_tests/conftest.py
```diff
@@ -1,136 +1,46 @@
...
-
-
-@pytest.fixture(scope="session")
-def setup_server_end_to_end_testing():
-    """Set up the server for end-to-end testing.
-
-    This fixture sets the necessary environment variables and
-    starts the server.
+@pytest.fixture(scope="session", autouse=True)
+def setup_server():
```
pjenvey: Since the compose does its own healthcheck on `syncserver`, this now only really catches failures outside the compose setup. I'm not sure how important of a guardrail it is -- I'd argue we could kill it?

docker/docker-compose.e2e.mysql.yaml
```diff
@@ -0,0 +1,42 @@
+services:
+  syncserver:
+    environment:
+      SYNC_CORS_MAX_AGE: "555"
+      SYNC_CORS_ALLOWED_ORIGIN: "*"
```
pjenvey: These default to some larger age and origin *, can these be killed?
```suggestion
```

docker/docker-compose.e2e.mysql.yaml
```diff
@@ -0,0 +1,42 @@
...
+    environment:
+      SYNC_SERVER_URL: http://syncserver:8000
+      TOKENSERVER_HOST: http://syncserver:8000
+      SYNC_HOST: 0.0.0.0
+      SYNC_MASTER_SECRET: secret0
+      SYNC_SYNCSTORAGE__DATABASE_URL: mysql://test:test@sync-db:3306/syncstorage
+      SYNC_TOKENSERVER__DATABASE_URL: mysql://test:test@tokenserver-db:3306/tokenserver
+      SYNC_TOKENSERVER__NODE_TYPE: mysql
+      SYNC_TOKENSERVER__FXA_EMAIL_DOMAIN: api-accounts.stage.mozaws.net
+      SYNC_TOKENSERVER__FXA_METRICS_HASH_SECRET: secret0
```
pjenvey: I think some of these can be killed here and elsewhere because they were only used by syncserver:

```suggestion
      SYNC_SERVER_URL: http://syncserver:8000
      TOKENSERVER_HOST: http://syncserver:8000
      SYNC_MASTER_SECRET: secret0
      SYNC_SYNCSTORAGE__DATABASE_URL: mysql://test:test@sync-db:3306/syncstorage
      SYNC_TOKENSERVER__DATABASE_URL: mysql://test:test@tokenserver-db:3306/tokenserver
```

Makefile
pjenvey: I might be missing something here (env vars can be really confusing in docker/compose) but is this intended to guard against any of these variables potentially being set in the invoking shell?

But now that these vars are no longer defined in `e2e.mysql`/`e2e.mysql.no-jwk-cache` (only in `e2e.mysql.jwk-cache.yaml`), shouldn't docker compose *not* inherit any potential values of them set in the invoking shell?

pjenvey (changes requested):

docker/docker-compose.e2e.mysql.yaml
chenba: > These default 

huh?  The reason I have them is that `TestStorage::test_cors_settings_are_set` fails when they are omitted.  I'll check why the defaults aren't working as expected.

Makefile
chenba: Sorry, the implicit thing that I should've made explicit in my comment is that I want devs be able to run the 'docker_run_*_e2e_tests' `make` targets in their env.  That's the reason behind unsetting the JWK env vars.  You're correct that in an CI env that won't be necessary.

docker/docker-compose.e2e.mysql.yaml
chenba: I updated teh assertion so that it isn't checking for `555`, which doesn't match the default value.

pjenvey: Ah, my mistake, I thought I grepped around for all of these env vars but apparently not the cors

pjenvey: But yea, having them assert the default's better anyway so we don't need them here 👍

tools/integration_tests/test_storage.py
```diff
@@ -2311,7 +2310,7 @@ def test_cors_settings_are_set(self):
             },
         )
 
-        self.assertEqual(int(res.headers["access-control-max-age"]), 555)
+        self.assertGreater(int(res.headers["access-control-max-age"]), 0)
```
chenba, line 2313: @pjenvey would you rather this checks for the default value?

pjenvey: Nevermind, that's ok. I guess the original intention of this test was to check if they were correctly set in the config.

tools/integration_tests/conftest.py
```diff
@@ -1,136 +1,8 @@
...
-import pytest
-import requests  # type: ignore[import-untyped]
 import logging
 
-DEBUG_BUILD = "target/debug/syncserver"
-RELEASE_BUILD = "/app/bin/syncserver"
-# max number of attempts to check server heartbeat
-SYNC_SERVER_STARTUP_MAX_ATTEMPTS = 35
-JWK_CACHE_DISABLED = os.environ.get("JWK_CACHE_DISABLED")
+SYNC_SERVER_URL = os.environ.get("SYNC_SERVER_URL", "http://localhost:8000")
```
pjenvey: I don't think this is used any longer, we can probably kill the entire file now (`test_support` does its own line of code for this)

chenba:
> 
> This makes me wonder if we could consolidate the compose files at all ...

I went down that route at some point but I can't remember why I backed out of it.  It was probably the names since they cannot be parameterized.  I'll poke at it; I'll file a follow-up issue if it takes more effort than I imagine.

PR #2192, opened 2026-04-06
test: modernize tests to use pytest

taddes:
## Description
Our integration and e2e tests are a hodgepodge of class-based `unitTest` (circa 2010 Python) and the newer `pytest`. This means we are leaving a lot of features on the table, like parallelized test runs and simpler maintenance with fixtures. This task updates our runs to use `Pytest` fully.

## Testing

Our E2E & integration test runs

## Issue(s)

Closes [STOR-518](https://mozilla-hub.atlassian.net/browse/STOR-518).

[STOR-518]: https://mozilla-hub.atlassian.net/browse/STOR-518?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docker/docker-compose.spanner.yaml
```diff
@@ -38,9 +38,9 @@ services:
       MYSQL_USER: test
       MYSQL_PASSWORD: test
     healthcheck:
-      test: ["CMD-SHELL", "mysqladmin -uroot -p$${MYSQL_ROOT_PASSWORD} version"]
+      test: ["CMD-SHELL", "mysqladmin -h 127.0.0.1 -P 3306 -uroot -p$${MYSQL_ROOT_PASSWORD} ping"]
```
taddes, line 35: This was done due to some recurring timeout issues. Found that forcing the health check to use TCP instead and increasing the max retry fixed it.

mysqladmin version without -h 127.0.0.1 connects via Unix socket, not TCP. So the healthcheck passes (socket is ready) before port 3306 is accepting TCP connections. Seemed we were getting 'Connection refused' on TCP. This race exists in the MySQL e2e too, but that job has a lighter stack without the spanner emulator competing for resources, so MySQL is fully ready before syncserver even starts. With spanner 1.5.52 being heavier, everything starts simultaneously and hits the race, I think.

pjenvey: I see some discussion of this [here on this issue](https://github.com/docker-library/mysql/issues/930#issuecomment-1329741754), so this makes sense. Let's apply it to the other couple mysql health checks in our compose files.

There's another comment there [claiming a 3s timeout as being "crucial"](https://github.com/docker-library/mysql/issues/930#issuecomment-1463541157), who knows if that's true/still the case but we could also bump to that from our 2s as well.

tools/integration_tests/tokenserver/conftest.py
```diff
@@ -0,0 +1,467 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+"""Fixtures and helpers for tokenserver integration tests.
```
pjenvey, line 4: This module replaces most of `test_support.py` I think, should that be killed now (or the non fixture functions could move there, but again I won't block on that)

tools/integration_tests/tokenserver/conftest.py
```diff
@@ -0,0 +1,467 @@
...
+DEFAULT_OAUTH_SCOPE = "https://identity.mozilla.com/apps/oldsync"
+
+NODE_ID = 800
+NODE_URL = "https://example.com"
+FXA_EMAIL_DOMAIN = "api-accounts.stage.mozaws.net"
+TOKEN_SIGNING_SECRET = os.environ.get("SYNC_MASTER_SECRET", "secret0")
+FXA_METRICS_HASH_SECRET = os.environ.get("SYNC_MASTER_SECRET", "secret0")
+
+
+# ── DB-mode helper ───────────────────────────────────────────────────────────
```
pjenvey, line 47: FYI my preference would be to only keep fixtures in this file with the other functions living elsewhere but I won't block on that.

pjenvey (changes requested):

docker/docker-compose.spanner.yaml
taddes: Yeah good call, I even bumped the postgres ones up by a second so it's 3s across the board, unless a reason not to. Wonder if some of the intermittent issues are at all related to this as that link gives pretty compelling reasons.

tools/integration_tests/tokenserver/conftest.py
taddes: Let's land this if it's not a block and when I rebase with the large test_support and test_storage PR, I can remove that and/or move whatever standalone functions out

tools/integration_tests/tokenserver/conftest.py
taddes: Agreed, as above: let's land this if it's not a block and when I rebase with the large test_support and test_storage PR, I can remove that and/or move whatever standalone functions out.

pjenvey (approved):

PR #2194, opened 2026-04-06
Depend on PostgreSQL token server if building PostgreSQL sync server

ntninja:
## Description

Building the PostgreSQL syncserver should also build the PostgreSQL token server, currently one gets a mixed build which is kinda weird and pretty annoying for packaging.

## Testing

How should reviewers test?

```commands
$ cd syncserver
$ cargo build --no-default-features -F pgsql,py_verifier
```

Should build the PostgreSQL Sync- **and** Token-Server.
Not, the PostgreSQL Sync-Server with the MySQL Token-Server.


pjenvey:
This looks reasonable but this breaks some CI assumptions within our Dockerfile, where the default build arg of `TOKENSERVER_DATABASE_BACKEND=mysql` currently doesn't emit any tokenserver backend `--feature`, assuming `tokenserver-db` defaults to mysql.

I think simplifying that to always specifying `--feature=tokenserver-db/$TOKENSERVER_DATABASE_BACKEND` as is done for the `syncstorage-db` backend should fix this

PR #2195, opened 2026-04-06
test: add tokenserver util tests

taddes:
## Description

Addition of Tokenserver and Tokenserver util tests to our e2e Python integration tests.

## Testing

Additional e2e runs for Tokenserver

## Issue(s)

Closes [STOR-526](https://mozilla-hub.atlassian.net/browse/STOR-526).

[STOR-526]: https://mozilla-hub.atlassian.net/browse/STOR-526?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Makefile
```diff
@@ -97,6 +98,7 @@ docker_run_mysql_e2e_tests:
 	exit_code=$$?;
 	docker cp mysql-e2e-tests:/mysql_integration_results.xml ${MYSQL_INT_JUNIT_XML};
 	docker cp mysql-e2e-tests:/mysql_no_jwk_integration_results.xml ${MYSQL_NO_JWK_INT_JUNIT_XML};
+	docker cp mysql-e2e-tests:/mysql_tokenserver_utils_results.xml ${TOKENSERVER_UTILS_JUNIT_XML} || true;
```
chenba: Why do we need the `|| true` fallback here?

docker-compose.e2e.mysql.yaml
```diff
@@ -43,5 +43,6 @@ services:
         PYTHONPATH=/app pytest /app/tools/integration_tests/ --junit-xml=/mysql_integration_results.xml || exit_code=$$?;
         export JWK_CACHE_DISABLED=true;
         PYTHONPATH=/app pytest /app/tools/integration_tests/ --junit-xml=/mysql_no_jwk_integration_results.xml || exit_code=$$?;
+        cd /app/tools/tokenserver && pytest . -v --junit-xml=/mysql_tokenserver_utils_results.xml || exit_code=$$?;
```
chenba: The other pytest calls don't use `-v`.  Do we need it for these tests for some particular reasons?

docker-compose.e2e.spanner.yaml
```diff
@@ -43,5 +43,6 @@ services:
         PYTHONPATH=/app pytest /app/tools/integration_tests/ --junit-xml=/spanner_integration_results.xml || exit_code=$$?;
         export JWK_CACHE_DISABLED=true;
         PYTHONPATH=/app pytest /app/tools/integration_tests/ --junit-xml=/spanner_no_jwk_integration_results.xml || exit_code=$$?;
+        cd /app/tools/tokenserver && pytest . -v --junit-xml=/spanner_tokenserver_utils_results.xml || exit_code=$$?;
```
chenba: I get why the file is named like that but the spanner has nothing to do with the token server 🙃

tools/tokenserver/database.py
```diff
@@ -196,6 +198,25 @@
...
+from
+    nodes
+where
+    service = :service
+    and available > 0
+    and capacity > current_load
+    and downed = 0
+    and backoff = 0
+order by
+    log(NULLIF(current_load, 0)) / log(capacity) NULLS FIRST
```
pjenvey: This made me notice we're using `ln` instead of `log` in the rust code's query, offhand I'm not sure what difference that results in -- is it the same, or should it switch over to log or this query switch to `ln` as well?

chenba: Sorry--I didn't look at the queries that closely since I assumed that they are mostly copy and pasted.

Postgres's `ln` is the equivalent of MySQL's `log`.  The two impls match, and for consistency this query here should be updated to match the Rust code's.

docker-compose.e2e.mysql.yaml
taddes: Makes the output more verbose which I tend to like, perhaps we add it to the others since for output it gives more nuance?

docker-compose.e2e.spanner.yaml
taddes: Yah, totally, I know it's a clunky distinction because yes, it doesn't have anything to do with spanner 🙃 however it's part of the spanner docker build making maybe a little less unclear.

Makefile
taddes: Not necessarily, but I was using it as a fallback, since the primary container output are the actual integration test results and not so much these, that way it doesn't crash should a write err occur, but it's working so I can take it out.

docker/docker-compose.e2e.mysql.yaml
```diff
@@ -33,4 +33,7 @@ services:
       - >-
         PYTHONPATH=/app
         pytest /app/tools/integration_tests/
-        --junit-xml=/${RESULTS_FILENAME}
+        --junit-xml=/${RESULTS_FILENAME};
+        cd /app/tools/tokenserver &&
+        pytest . -v
+        --junit-xml=/mysql_tokenserver_utils_results.xml
```
chenba: Sorry you need to override the entrypoint in `docker-compose.e2e.{no-,}jwk-cache.yaml`.  Also, is it possible to run a single `pytest` command?

taddes: Ah missed those new files, thanks 👍 I see `no-jwk` only runs: `pytest /app/tools/integration_tests/tokenserver/test_e2e.py` , so maybe that doesn't need an override?

chenba: oh true, no override needed there.

clouserw (dismissed):

clouserw (approved):

PR #2207, opened 2026-04-10
wip: trying a TestDb trait

pjenvey:
## Description

Draft a `TestDb` trait which seemed like the best alternative in the STOR-380 options for these test methods, but I'm not in love with the bit of the complication it adds for what `impl<T: stuff> Db for T` is -- I'm leaning towards cancelling this issue entirely (or probably doing the adding some comments option)

## Issue(s)

Issue STOR-380

### DIFF
--- a/tokenserver-db-common/src/lib.rs
+++ b/tokenserver-db-common/src/lib.rs
@@ -43,8 +43,18 @@ impl Clone for Box<dyn DbPool> {
     }
 }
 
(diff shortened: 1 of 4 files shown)


chenba:
I'm in favor of cancelling the issue.  If the more experienced Rust dev doesn't like the complexity introduced in the changes, then, as the less experienced Rust dev, I prefer not having it in the code base.

PR #2239, opened 2026-04-23
test: test_storage test_support pytest refactor

taddes:
## Description
Our integration and e2e tests are a hodgepodge of class-based `unitTest` (circa 2010 Python) and the newer `pytest`. This means we are leaving a lot of features on the table, like parallelized test runs and simpler maintenance with fixtures. This task updates our runs to use `Pytest` fully.

This is related to #2192 , however addresses the some 5000 lines of core integration test code for `test_storage` and `test_support`, not the ancillary tests around tokenserver and other things.

## Testing

Our E2E & integration test runs

## Issue(s)

Closes [STOR-537](https://mozilla-hub.atlassian.net/browse/STOR-537).

[STOR-537]: https://mozilla-hub.atlassian.net/browse/STOR-537?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tools/integration_tests/conftest.py
```diff
@@ -25,3 +36,187 @@ def _logged_do_request(self, req, *args, **kwargs):
...
+            req, auth_state["auth_token"], auth_state["auth_secret"]
+        )
+        return orig_do_request(req, *args, **kwds)
+
+    app.do_request = new_do_request
+
+    root = "/1.5/%d" % auth["user_id"]
+    retry_delete(app, root)
+
+    ctx = {
```
pjenvey, line 158: Does this need to be a dict (a class would be a little nicer)?

tools/integration_tests/conftest.py
```diff
@@ -25,3 +36,187 @@ def _logged_do_request(self, req, *args, **kwargs):
...
+
+@pytest.fixture(scope="function")
+def st_ctx():
+    """Functional test context for storage API tests.
+
+    Sets up a Pyramid configurator, creates a TestApp with hawk signing,
+    authenticates a random user, clears that user's data, and yields a
+    context dict. Tears down configurator on exit.
+    """
+    ini_file = os.environ.get("MOZSVC_TEST_INI_FILE", "tests.ini")
```
pjenvey, line 106: FYI I suspect we don't need most of these `MOZSVC_` env vars any longer -- I suspect the ini file might still be in use with the pyramid based auth policy types, but I think that can probably go away as well once we strip out pyramid.

taddes: Ok good call, we should then file a ticket to remove Pyramid. I'll make a note to capture this in the scope

tools/integration_tests/conftest.py
taddes: I had thought of that, but since I was taking a pretty strict approach to returning simple types with the fixtures, being a bit more idiomatic for Pytest. I could alternatively use `SimpleNamespace` ? That gives attribute access without the same overhead of defining a class. Happy to change to that, but I think it might be ok as-is. Lmk

pjenvey: Why the strict limit for fixture return types?

taddes: Mostly personal preference and experience. Since we've moved away from class based tests entirely, I felt keeping the interface pretty simple. Some of the main setup fixtures are pretty big and have a lot of context (like `st_ctx` here). I'd read unless you're returning a class representation as an object used a test, for normal fixtures that deal with setup and teardown, it's a bit of an anti-pattern. Just personally found it over-engineered or hard to debug (kind of how it was before).

pjenvey (approved):

PR #2252, opened 2026-04-28
fix: apply zizmor autofixes to github actions

fkiriakos07:
## Description

Address misconfiguration in github actions which were reported by zizmor

## Testing

Ensure that github actions are running correctly especially the ones which use caching since zizmor recommended using the parameter `lookup-only: true` when using the cache action

## Issue(s)

Closes [STOR-540](https://mozilla-hub.atlassian.net/browse/STOR-540).

[STOR-540]: https://mozilla-hub.atlassian.net/browse/STOR-540?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2253, opened 2026-04-28
fix: apply zizmor autofixes

taddes:
## Description

Referencing auto fixes applied from Zizmor run on Sync and testing some other possible issues (some actions appear to be having cache restore issues and certain utilities not working). Original PR https://github.com/mozilla-services/syncstorage-rs/pull/2252 

## Issue(s)

Closes [STOR-540](https://mozilla-hub.atlassian.net/browse/STOR-540).

[STOR-540]: https://mozilla-hub.atlassian.net/browse/STOR-540?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/main-workflow.yml
+++ b/.github/workflows/main-workflow.yml
@@ -114,7 +114,6 @@ jobs:
             ~/.cache/pip
             ~/.cache/pypoetry/virtualenvs
           key: ${{ runner.os }}-python-${{ hashFiles('pyproject.toml', 'poetry.lock') }}
-          lookup-only: true
 
       - name: Install Poetry
         run: pip3 install poetry
@@ -151,7 +150,6 @@ jobs:
             ~/.rustup/toolchains
             ~/.rustup/update-hashes
           key: ${{ runner.os }}-rust-toolchain-${{ env.RUST_VERSION }}
-          lookup-only: true
 
       - name: Set Rust toolchain
         run: rustup default ${RUST_VERSION}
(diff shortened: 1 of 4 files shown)


.github/workflows/main-workflow.yml
```diff
@@ -162,7 +160,6 @@ jobs:
         with:
           path: ~/.cargo/bin/cargo-audit
           key: ${{ runner.os }}-cargo-audit-${{ hashFiles('.github/workflows/main-workflow.yml') }}
-          lookup-only: true
```
pjenvey, line 165: I'll note if we want to be pedantic we can probably keep a few of these (all the actual "Caching.." steps vs restore ones) but this doesn't seem terribly important when this workflow only publishes test result files AFAICT

(zizmor seems [only concerned about runtime artifacts here](https://docs.zizmor.sh/audits/#remediation_4)):
```
error[cache-poisoning]: runtime artifacts potentially vulnerable to a cache poisoning attack
   --> ./.github/workflows/main-workflow.yml:238:9
    |
  3 | / on:
  4 | |   push:
  5 | |     branches:
  6 | |       - master
...   |
 19 | |         required: false
 20 | |         default: "1.91" # RUST_VER
    | |__________________________________- generally used when publishing artifacts generated at runtime
...
238 |           uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5
    |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ enables caching by default
    |
    = note: audit confidence → Low
    = note: this finding has an auto-fix

```

pjenvey (approved):

.github/workflows/main-workflow.yml
taddes: Yeah some were causing problems with the caches not being restored, but the removal of some made sense after trying since yeah, they're not doing something or handling something super important.

PR #2266, opened 2026-05-01
docs: Docker HEALTHCHECK honors `SYNC_PORT` from environment

HorlogeSkynet:
## Description

When documentation explicitly sets `SYNC_PORT` environment variable in Compose stack, makes Docker/Podman `HEALTHCHECK` actually honor its value to prevent a wrong "unhealthy" state.

---

Thanks, bye :wave:


pjenvey (approved):

pjenvey:
Thanks @HorlogeSkynet!

PR #2289, opened 2026-05-07
refactor: remove pyramid

taddes:
## Description

Pyramid is a legacy artifact from when syncstorage-rs was a Python web app.  Pyramid survives only in the test infrastructure where it's being used as an overpowered settings container and fake request factory. The actual crypto work (tokenlib, hawkauthlib) has zero framework dependency and is simpler to deal with. Removing it simplifies the test stack and eliminates unnecessary dependencies.

This PR removes Pyramid and depends on simpler implementaiton details while making sure the tests still work as expected.

## Issue(s)

Closes [STOR-561](https://mozilla-hub.atlassian.net/browse/STOR-561).

[STOR-561]: https://mozilla-hub.atlassian.net/browse/STOR-561?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tools/integration_tests/test_support.py
```diff
@@ -28,12 +23,45 @@
 import unittest
 import uuid
 from webtest import TestApp
-from zope.interface import implementer
 
 
 VALID_FXA_ID_REGEX = re.compile("^[A-Za-z0-9=\\-_]{1,64}$")
 
 
+class TestConfig:
```
pjenvey: I *think* we could go even further and kill the settings bits like this class, `get_configurator`, etc. If we were even using anything at all in the `tests.ini` it was probably just the secret. https://github.com/mozilla-services/syncstorage-rs/blob/6a6db57a686fbf14315a1a7e2733bfb7b267a443/tools/integration_tests/tests.ini#L23

And we've long read the secret in from an env var anyways, not the ini file. In fact, looking further, we still have the `TestCase` classes in this file that reference configurator. Can't those be killed as well since now that test_storage was refactored out of the class based structure?

taddes: Absolutely, one step ahead of ya! That's the next commit 👍 dead `TestCase`, `StorageTestCase`, `FunctionalTestCase` etc.

pjenvey (approved):
remove pyramid+, hell yes :fire:

PR #2295, opened 2026-05-11
chore(build): use 8.4-lts's libmysqlclient-dev

chenba:
This commit removes libmysqlclient-dev and defaults to the metapackage of default-libmysqlclient-dev because libmysqlclient-dev isn't available for Trixie.  It effectively uses libmariadb-dev-compat, which was the default in Dockerfile.

The original reason of switching to the MySQL lib was that the MariaDB lib did not support TLS.  That doesn't appear to be the case anymore.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,9 +24,9 @@ RUN apt-get -q update && \
         MYSQL_PKG="$MYSQLCLIENT_PKG"; \
         if [ "$MYSQLCLIENT_PKG" = libmysqlclient-dev ] ; then \
             # First install gnupg and setup MySQL repo
-            # Key ID A8D3785C from https://dev.mysql.com/doc/refman/8.0/en/checking-gpg-signature.html
+            # Key ID A8D3785C from https://dev.mysql.com/doc/refman/8.4/en/checking-gpg-signature.html
             apt-get -q install -y --no-install-recommends gnupg ca-certificates && \
-            echo "deb https://repo.mysql.com/apt/debian/ trixie mysql-8.0" >> /etc/apt/sources.list && \
+            echo "deb https://repo.mysql.com/apt/debian/ trixie mysql-8.4-lts" >> /etc/apt/sources.list && \
             # Fetch and install the MySQL public key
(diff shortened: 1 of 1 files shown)


chenba:
This fixes the currently borked Spanner/MySQL build in main.

Dockerfile
```diff
@@ -22,16 +22,6 @@ RUN apt-get -q update && \
     POSTGRES_DEV_PKG="" && \
     if [ "$SYNCSTORAGE_DATABASE_BACKEND" = "mysql" ] || [ "$TOKENSERVER_DATABASE_BACKEND" = "mysql" ]; then \
         MYSQL_PKG="$MYSQLCLIENT_PKG"; \
-        if [ "$MYSQLCLIENT_PKG" = libmysqlclient-dev ] ; then \
```
taddes, line 25: I exhale with relief seeing these lines disappear.

taddes (dismissed):

pjenvey (changes requested):
> The original reason of switching to the MySQL lib was that the MariaDB lib did not support TLS. That doesn't appear to be the case anymore.

libmariadb itself supported TLS but it required a nonstandard way to configure it (and it had to be explicitly configured to be used last time I checked) --- which diesel didn't support doing.

I found that mysql offers a `mysql-8.4-lts` client, just not `mysql-8.0` https://github.com/mozilla-services/syncstorage-rs/compare/chore/libmysqlclient-8.4 (building succeeded on this branch).

However, now that I think about it -- do we even need the C based mysql client lib any longer? Now that we're fully based on `diesel-async`, the underlying `mysql_async` crate is I believe fully written in Rust -- so we  could try killing it entirely?

pjenvey:
Oops, I forgot about the Python scripts. The purge script, tests etc probably still need it.

chenba:
Changed it to use 8.4-lts.

pjenvey (approved):
Thanks, I guess we could also consider switching to that pure python mysql driver to avoid all this stuff (mysql+pymysql)

PR #2303, opened 2026-05-12
perf: stop count(*) when count is not used

chenba:
For the Spanner backend we count the number of bsos for a given uid+kid+collection_id, but then we ingore the actual value.  This patch updates the query to see if there are any rows.

## Issue(s)

Closes STOR-215

### DIFF
--- a/syncstorage-spanner/src/db/mod.rs
+++ b/syncstorage-spanner/src/db/mod.rs
@@ -354,12 +354,12 @@ impl SpannerDb {
              AND collection_id = @collection_id
            GROUP BY fxa_uid"
         } else {
-            "SELECT COUNT(*)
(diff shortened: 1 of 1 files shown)


pjenvey (approved):
This looks good but going back to the old version of this code before the quota support was added, I notice it wasn't even doing any select on bsos. As when quota's disabled, it's really just doing an upsert to update the `modified` time on `user_collections`.

I logged https://mozilla-hub.atlassian.net/browse/STOR-564 for changing at least the end of this method to an upsert, I think we could improve further here on both by separating out the quota support from the non quota support version of this method (the other impls somewhat do this).

(We explicitly avoid touching the quota columns in this impl though as it adds to the mutation count).

PR #2304, opened 2026-05-13
feat: implements `allow_new_users` setting

### LINKED ISSUE #1429: allow_new_users setting
With the python2 `syncserver` implementation it was possible to disable new user requests as seen below.

```
[syncserver]
...other settings...
allow_new_users = false
```

Is there an equivalent setting in `syncstorage-rs`? I couldn't find anything fitting while looking through the code but it is possible I missed it.

Thanks!

HorlogeSkynet:
## Description

New `allow_new_users` setting, useful for self-hosted nodes.

## Testing

:warning: Not functionally tested at the moment.

Set `SYNC_TOKENSERVER__ALLOW_NEW_USERS` to `false` and no new users allocation should occur.

## Issue(s)

Closes #1429.


tokenserver-db/src/tests.rs
```diff
@@ -1270,6 +1296,47 @@ async fn test_correct_created_at_used_during_user_retrieval() -> DbResult<()> {
...
+        capacity: 100,
+        available: 100,
+        downed: 1,
+        ..Default::default()
+    })
+    .await?;
+
+    // User allocation fails because users allocation is disabled
+    let result = db
+        .allocate_user(params::AllocateUser {
```
chenba: This is not testing the correct function.  `get_or_create_user` is the updated function.

tokenserver-db-common/src/error.rs
```diff
@@ -26,6 +30,10 @@ impl DbError {
         DbErrorKind::PoolTimeout(timeout_type).into()
     }
 
+    pub fn is_user_not_found(&self) -> bool {
+        matches!(&self.kind, DbErrorKind::UserNotFound(_))
+    }
```
chenba, line 35: Since this is used only in tests, it should use the `#[cfg(debug_assertions)]` flag.

tokenserver-db-common/src/lib.rs
```diff
@@ -127,7 +127,12 @@ pub trait Db {
 
         if raw_users.is_empty() {
             // There are no users in the database with the given email and service ID, so
-            // allocate a new one.
+            // allocate a new one if allowed by configuration.
+            if !params.allow_new_users {
+                warn!("New users registration is disabled by configuration"; "email" => &params.email);
+                return Err(DbError::user_not_found(params.email.clone()));
+            }
```
chenba, line 134: Preferably we don't want to log out someone's email address.  Do you need it for debugging, blocking, something else?

chenba (changes requested):

chenba:
@HorlogeSkynet Thank you for the patch.  I just tried running the tests and there is a failure.  See comments above.

tokenserver-db/src/tests.rs
HorlogeSkynet: Nice catch ! Sorry for that, I had copied the wrong test.

tokenserver-db-common/src/lib.rs
HorlogeSkynet: No, no need I guess. I just wanted to display a soft-error message with details before the full traceback. I've opted for rewording of the new `DbErrorKind::UserNotFound`.

tokenserver-db-common/src/error.rs
HorlogeSkynet: Added, thanks.

HorlogeSkynet:
Any news @chenba ? :slightly_smiling_face:

chenba:
> Any news @chenba ? 🙂

I'm sorry.  I got distracted/busy.  I should have time to take a look today.  Thanks for the reminder.

tokenserver-db-common/src/error.rs
```diff
@@ -18,6 +18,10 @@ pub struct DbError {
 }
 
 impl DbError {
+    pub fn user_not_found(user: String) -> Self {
+        DbErrorKind::UserNotFound(user).into()
+    }
```
chenba: What's the reason for a Not Found?  I believe an HTTP 403 is more appropriate.  We are denying an action.

tokenserver-db-common/src/error.rs
```diff
@@ -72,6 +84,11 @@ enum DbErrorKind {
 impl From<DbErrorKind> for DbError {
     fn from(kind: DbErrorKind) -> Self {
         match kind {
+            DbErrorKind::UserNotFound(_) => Self {
+                kind,
+                status: StatusCode::NOT_FOUND,
+                backtrace: Box::new(Backtrace::new_unresolved()),
+            },
```
chenba, line 90: When `impl From<DbError> for TokenserverError` converts an error of this type we'd end up with a HTTP 503.  That's misleading for the clients.  (We don't want them to retry at the very least.)  Whatever error we add here for the feature should become a HTTP 403.

chenba (commented):
Got a couple comments wrt error handling.

tokenserver-db-common/src/error.rs
HorlogeSkynet: Renamed error kind and associated resources to match "User not created".

tokenserver-db-common/src/error.rs
HorlogeSkynet: By explicitly matching `DbErrorKind::UserNotCreated` kind in this context, I guess we prevent `_` from converting it to a HTTP 503 by default ?
I've updated the pattern matching to throw a 403 FORBIDDEN as you proposed. Tell me whether it suits what you had in mind.

HorlogeSkynet: Hey @chenba, this is a little _up_ :innocent:

chenba:
@HorlogeSkynet Sorry about the delay with looking at this patch.  Tomorrow is a holiday here so I'll take a look on Monday.

chenba:
@HorlogeSkynet  `cargo check` is throwing errors (see below).  Do you see that locally?

```
    Checking syncstorage-db-common v0.23.3 (/Users/barrychen/code/syncstorage-rs/syncstorage-db-common)
error[E0592]: duplicate definitions with name `is_user_not_created`
  --> tokenserver-db-common/src/error.rs:34:5
   |
21 |     pub fn is_user_not_created(user: String) -> Self {
   |     ------------------------------------------------ other definition for `is_user_not_created`
...
34 |     pub fn is_user_not_created(&self) -> bool {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ duplicate definitions for `is_user_not_created`

error[E0599]: no function or associated item named `user_not_created` found for struct `DbError` in the current scope
   --> tokenserver-db-common/src/lib.rs:132:37
    |
132 |                 return Err(DbError::user_not_created(params.email.clone()));
    |                                     ^^^^^^^^^^^^^^^^ function or associated item not found in `DbError`
    |
   ::: tokenserver-db-common/src/error.rs:14:1
    |
 14 | pub struct DbError {
    | ------------------ function or associated item `user_not_created` not found for this struct
    |
note: if you're trying to build a new `DbError` consider using one of the following associated functions:
      DbError::is_user_not_created
      DbError::internal
      DbError::pool_timeout
   --> tokenserver-db-common/src/error.rs:21:5
    |
 21 |     pub fn is_user_not_created(user: String) -> Self {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
 25 |     pub fn internal(msg: String) -> Self {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
 29 |     pub fn pool_timeout(timeout_type: deadpool::managed::TimeoutType) -> Self {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
help: there is an associated function `is_user_not_created` with a similar name
    |
132 |                 return Err(DbError::is_user_not_created(params.email.clone()));
    |                                     +++

    Checking syncstorage-mysql v0.23.3 (/Users/barrychen/code/syncstorage-rs/syncstorage-mysql)
Some errors have detailed explanations: E0592, E0599.
For more information about an error, try `rustc --explain E0592`.
error: could not compile `tokenserver-db-common` (lib) due to 2 previous errors

```

tokenserver-db-common/src/error.rs
chenba: Sorry I think my earlier comment was not clear.  When the `DbError` is converted into a `TokenserverError`, the status is modified here: https://github.com/mozilla-services/syncstorage-rs/blob/40cab9df/tokenserver-db-common/src/error.rs#L117.  So you'd need to add an `if db_error.is_user_not_created()` check there to return the 403 response.

tokenserver-db-common/src/lib.rs
chenba: How about directly log the value out instead of having it in the error?  Having that in the error itself could expose it in Sentry, the http response, or whatever error handling/capture get introduced.

tokenserver-db-common/src/error.rs
HorlogeSkynet: Thanks ! It's clear. Added.

HorlogeSkynet:
> @HorlogeSkynet `cargo check` is throwing errors (see below). Do you see that locally?

(apologies for the delay)

Yes sorry it slipped during function renaming. Fixed.

tokenserver-db-common/src/error.rs
HorlogeSkynet: As it's now **not only** used in tests, should the flag be removed ?

**EDIT** : it must.

chenba:
@HorlogeSkynet Looks like there are some build errors.

HorlogeSkynet:
> @HorlogeSkynet Looks like there are some build errors.

Hey @chenba ! Fixed (at least I hope so).

HorlogeSkynet:
Hey @chenba, I can't manage to access error-ed checks logs. Could you tell us what remains to be fixed ?
Thanks, bye :wave:

chenba:
> Hey @chenba, I can't manage to access error-ed checks logs. Could you tell us what remains to be fixed ? Thanks, bye 👋

The most recent failures are not related to your changes afaict. 👍

tokenserver-db-common/src/lib.rs
chenba: I think this got lost in the back and forth.  Let's log the email with `warn!()` but keep it out of the error message.

HorlogeSkynet: You're right ! Warning has been restored and `UserNotCreated` dropped its parameter. I hope this is what you meant.
Thanks, bye :wave:

tokenserver-db-common/src/error.rs
```diff
@@ -59,6 +67,9 @@ impl ReportableError for DbError {
 
 #[derive(Debug, Error)]
 enum DbErrorKind {
+    #[error("Specified user does not exist (new users are disallowed)")]
+    UserNotCreated(),
```
chenba: This can be `UserNotCreated` now

tokenserver-db-common/src/error.rs
chenba, line 106:
```suggestion
    fn from(db_error: DbError) -> Self {
        if db_error.is_user_not_created() {
            return TokenserverError {
                context: db_error.to_string(),
                backtrace: db_error.backtrace.clone(),
                source: Some(Box::new(db_error)),
                ..TokenserverError::new_users_disabled()
            };
        }

```

then in the TokenserverError impl you can add something like 

```
    pub fn new_users_disabled() -> Self {
        Self {
            status: "new-users-disabled",
            location: ErrorLocation::Body,
            description: "New users are not allowed".to_owned(),
            http_status: StatusCode::FORBIDDEN,
            ..Self::default()
        }
    }
```

tokenserver-db-common/src/error.rs
```diff
@@ -98,6 +114,9 @@ impl From<DbError> for TokenserverError {
                 // Use the status code from the DbError if it already suggests an internal error;
                 // it might be more specific than `StatusCode::SERVICE_UNAVAILABLE`
                 db_error.status
+            } else if db_error.is_user_not_created() {
+                // Throw a proper 403 Forbidden HTTP error if user creation is disabled
+                StatusCode::FORBIDDEN
```
chenba: Ok this works in that it changes the http status.  But I think the message in the response body will remain "Server error", which can be slightly confusing.  I've made a suggestion to return a different TokenserverError above.

tokenserver-db-common/src/error.rs
HorlogeSkynet: Fixed.

tokenserver-db-common/src/error.rs
HorlogeSkynet: Your proposal has been applied !

chenba:
@HorlogeSkynet I'm sorry that I didn't notice before that your commits were not signed, but we require [verified commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification).  Can you signed the commit?  Thank you.

HorlogeSkynet:
> @HorlogeSkynet I'm sorry that I didn't notice before that your commits were not signed, but we require [verified commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification). Can you signed the commit? Thank you.

Commit has been signed off @chenba :pray:

chenba (approved):
pjenvey already approved: https://github.com/mozilla-services/syncstorage-rs/pull/2478#pullrequestreview-4867983549

Thank you for patch, @HorlogeSkynet

PR #2313, opened 2026-05-14
fix: purge old record storage node err handling

taddes:
## Description

Revise the `purge_old_records` to better handle failure cases.

## Issue(s)

Closes [STOR-70](https://mozilla-hub.atlassian.net/browse/STOR-70).

[STOR-70]: https://mozilla-hub.atlassian.net/browse/STOR-70?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tools/tokenserver/purge_old_records.py
+++ b/tools/tokenserver/purge_old_records.py
@@ -62,9 +62,16 @@ def purge_old_records(
     """
     logger = logging.getLogger(LOGGER)
     logger.info("Purging old user records")
+    start = time.monotonic()
+    success = False
     try:
         database = Database()
         previous_uids = set()
+        if metrics:
+            metrics.gauge(
+                "purge.backlog.size", database.count_old_user_records(grace_period)
+            )
+        previous_list = []
         # Process batches of <max_per_loop> items, until we run out.
         while True:
             offset = random.randint(0, max_offset)
@@ -89,6 +96,8 @@ def purge_old_records(
                     f" to {uid_range[1] or 'End'}"
                 )
             logger.info(f"Fetched {len(rows)} rows at offset {offset}{range_msg}")
+            if metrics:
+                metrics.incr("purge.batch.fetched", value=len(rows))
             counter = 0
             for row in rows:
                 try:
@@ -185,15 +194,22 @@ def purge_old_records(
                     logger.info("Reached max_records, exiting")
                     if metrics:
                         metrics.incr("max_records")
+                    success = True
                     return True
             if len(rows) < max_per_loop:
                 break
+        logger.info("Finished purging old user records")
+        success = True
+        return True
     except Exception:
         logger.exception("Error while purging old user records")
         return False
-    else:
-        logger.info("Finished purging old user records")
-        return True
+    finally:
+        if metrics:
+            metrics.timing(
+                "purge.run.duration_ms", int((time.monotonic() - start) * 1000)
+            )
+            metrics.incr("purge.run.success" if success else "purge.run.failure")
 
 
 def delete_service_data(
--- a/tools/tokenserver/purge_old_records.py
+++ b/tools/tokenserver/purge_old_records.py
@@ -71,7 +71,6 @@ def purge_old_records(
             metrics.gauge(
                 "purge.backlog.size", database.count_old_user_records(grace_period)
             )
-        previous_list = []
         # Process batches of <max_per_loop> items, until we run out.
         while True:
             offset = random.randint(0, max_offset)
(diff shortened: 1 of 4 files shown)


tools/tokenserver/purge_old_records.py
```diff
@@ -89,85 +95,120 @@ def purge_old_records(
...
-                                # if an override was specifed,
-                                # use that node ID
-                                override_node=override_node,
                                 metrics=metrics,
                             )
+                            if metrics:
+                                metrics.incr("delete_data")
+                            retryable(database.delete_user_record, row.uid)
+                            if metrics:
+                                metrics.incr("delete_user", tags={"type": "not_down"})
```
chenba, line 131: nit: "nodeless" above is self-explanatory but "not_down" is ambiguous.  Maybe either change the name to mention "node" or update the value so that it's obviously about the node.

chenba: Oh I see that it's old code so I guess we'll keep it :D

tools/tokenserver/purge_old_records.py
```diff
@@ -89,85 +95,120 @@ def purge_old_records(
...
+                    # the row to the back of the queue: it
+                    # leaves the eligible set for `grace_period` seconds,
+                    # giving healthy rows a chance to clear.
+                    reason = _classify_failure(exc)
+                    logger.exception(
+                        f"Failed to purge uid {row.uid}; "
+                        f"deferring retry (reason={reason})"
+                    )
+                    if metrics:
+                        metrics.incr("row_failure", tags={"reason": reason})
```
chenba, line 181: I don't think this is possible when `dryrun` is true

tools/tokenserver/purge_old_records.py
```diff
@@ -218,6 +259,26 @@ def retryable(fn, *args, **kwargs):
...
 
+def _classify_failure(exc):
+    """Map a per-row failure to a stable metric tag value.
+
+    Returns one of a small fixed set of strings to keep the metric's tag
+    cardinality bounded.
+    """
+    if isinstance(exc, requests.HTTPError) and exc.response is not None:
+        code = exc.response.status_code
+        if 500 <= code < 600:
```
chenba: nit: `code >= 500`

tools/tokenserver/purge_old_records.py
```diff
@@ -218,6 +259,26 @@ def retryable(fn, *args, **kwargs):
...
+            return "http_5xx"
+        if 400 <= code < 500:
```
chenba: nit: can be `code >= 400` because early return above

tools/tokenserver/purge_old_records.py
taddes: Yeah I originally had the same thought, since yeah, is it referring to the node, record or request, but taken with the other metrics `nodeless`, `node_up`, and `force`, it makese sense, given it's existing instrumentation.

tools/tokenserver/purge_old_records.py
taddes: In dryrun (which doesn't hit any network or db), from my reading, the except block is unreachable except a single case:
```python
                    elif force:
                        delete_sd = not points_to_active(
                            database, row, override_node, metrics=metrics
                        )
                        logger.info(
                            "Forcing tokenserver record delete: "
                            f"{row.uid} on {row.node} "
                            f"(deleting service data: {delete_sd})"
                        )
```
                        
If that DB call fails (connection error, timeout, etc.) with `--dryrun --force,` you land in the except block, and the `if not dryrun` check there correctly prevents `database.replace_user_record` from being called. So basically it's a correctness guard for the `dryrun` + `force` combo.

taddes: So the DB read only fires when both `--force` & `--override_node` are passed and the row's node differs from the override. I am not the authority on this script, but I think the `override_node` flag exists specifically for node migration, where you want to force-delete data on the old node and redirect the delete request to the new one, where you'd set `--dryrun --force --override_node=<new_node`> to prev what would get cleaned up before pulling the trigger.

tools/tokenserver/purge_old_records.py
taddes: Yah, good call

pjenvey (approved):
Sorry for the delay, this looks good. The main purge loop could probably use some cleanup at this point with all the nested scopes, but that can be cleaned up later

PR #2314, opened 2026-05-19
refactor(db): use Spanner INSERT OR UPDATE DML

chenba:
This commit:
 - replaces all inserts-and/or-updates with Spanner's upsert DML where appropriate
 - drops the database_spanner_use_mutations settings as mutations are no longer used

## Issue(s)

Closes STOR-218


syncstorage-spanner/src/db/BATCH_COMMIT.txt
```diff
@@ -9,7 +9,7 @@ those with NULL values won't be touched. When INSERTing bsos, columns with NULL
 values recieve defaults.
 
 The entire move (batch commit) happens in one single transaction. Spanner
-limits the number of writes (or "mutations") within a transaction to 20000
+limits the number of writes (or "mutations") within a transaction to 80000
```
chenba: [Latest mutations per commit limit](https://docs.cloud.google.com/spanner/quotas#limits-for)

syncstorage-spanner/src/db/BATCH_COMMIT.txt
```diff
@@ -30,7 +30,7 @@ The batch commit mutations are as follows:
 - Possibly direct inserts via post_bsos (but these only reduce total mutations
   by avoiding writing to the batch table, so not included here)
 
-- Write 1664 (MAX_TOTAL_RECORDS) to `bsos`: max 19968 (1664 * 12) mutations
+- Write 6665 (MAX_TOTAL_RECORDS) to `bsos`: max 79980 (6665 * 12) mutations
```
taddes: It'll actually be 6,656 since we're preserving the same legacy pattern, basically 1664 * 4, giving us a tiny bit more headroom, as in the webservices-infra PR. No biggie tho

```suggestion
- Write 6656 (MAX_TOTAL_RECORDS) to `bsos`: max 79872 (6656 * 12) mutations
```

syncstorage-spanner/src/db/BATCH_COMMIT.txt
```diff
@@ -52,7 +52,7 @@ The batch commit mutations are as follows:
 
 Totals:
 - quota: False
-  4 + 19968 + 1 = 19973
+  4 + 79980 + 1 = 79985
```
taddes:
```suggestion
  4 + 79872 + 1 = 79877
```

syncstorage-spanner/src/db/BATCH_COMMIT.txt
```diff
@@ -52,7 +52,7 @@ The batch commit mutations are as follows:
...
 
 - quota: True
-  6 + 19968 + 1 + 6 = 19981
+  6 + 79980 + 1 + 6 = 79993
```
taddes:
```suggestion
  6 + 79872 + 1 + 6 = 79885
```

syncstorage-spanner/src/db/mod.rs
```diff
@@ -808,206 +613,196 @@ impl SpannerDb {
...
+
+        let sql = format!(
+            "INSERT OR UPDATE INTO bsos
+                 (fxa_uid, fxa_kid, collection_id, bso_id, modified, payload, expiry{sortindex_col})
+             SELECT
+                 @fxa_uid, @fxa_kid, @collection_id, @bso_id,
+                 {modified_expr},
+                 {payload_expr},
+                 {expiry_expr}{sortindex_expr}
+               FROM UNNEST([1])
```
taddes: Given this is a little hard to understand, as you can't have a LEFT JOIN without a FROM clause to give it a left-hand side, maybe explaining that the UNNEST is the way to do this.

  side.

```suggestion
               FROM UNNEST([1]) --  provides a row source for the LEFT JOIN
```

taddes (commented):
Thanks for doing this! Just a few minor nits.

syncstorage-spanner/src/db/BATCH_COMMIT.txt
chenba: I'm thinking maybe we should delete this file and rely on the doc you are adding in https://github.com/mozilla-services/syncstorage-rs/pull/2327.  No need to duplicate.

taddes: Yeah that works fine I think

syncstorage-spanner/src/db/mod.rs
```diff
@@ -808,206 +613,196 @@ impl SpannerDb {
...
+    async fn put_bso_dml(
         &mut self,
-        bso: params::PutBso,
-    ) -> DbResult<results::PutBso> {
+        user_id: &UserIdentifier,
+        collection_id: i32,
+        bso: params::PostCollectionBso,
+        timestamp: SyncTimestamp,
+    ) -> DbResult<()> {
         use syncstorage_db_common::util::to_rfc3339;
```
pjenvey: nit: can hoist this import out of here to the module level while you're here -- it was probably more convenient here in the past when this was a test only method

syncstorage-spanner/src/db/mod.rs
```diff
@@ -808,206 +613,196 @@ impl SpannerDb {
...
+                .payload
+                .map(IntoSpannerValue::into_spanner_value)
+                .unwrap_or_else(null_value);
+            let ttl = bso
+                .ttl
+                .map(IntoSpannerValue::into_spanner_value)
+                .unwrap_or_else(null_value);
+
+            let mut row = ListValue::new();
+            row.set_values(RepeatedField::from_vec(vec![
```
pjenvey: nit: `RepeatedField` has a `From<Vec<T>>` so you don't even need to use it here, you can just do `into()` on the vecs (it doesn't save much but protobuf is so verbose..)

pjenvey (dismissed):
I'm glad we revisited this, this is great --

It's obvious the benefit of having `INSERT OR UPDATE` available as DML vs solely in mutations -- the latter being so much more limiting not being intertwined with sql.

The only downside I can see to this change is we're now *always* reading/writing in the existing data vs not on selective updates (e.g. `payload` of `None`/`NULL` passed in). Is the Spanner query optimizer smart enough to e.g. not penalize an update that's directed to write the existing payload back into the row? (It'd be interesting to see the query execution plan here but I doubt it would answer that).

If not how much are we paying for that? At the very least it's costing us extra mutations, but that's totally worth the cost (we've been limiting ourselves to 20k and that's getting bumped soon anyway).

Approved but I added a couple nits

pjenvey:
FYI this also closes STOR-565

chenba:
> The only downside I can see to this change is we're now _always_ reading/writing in the existing data vs not on selective updates (e.g. `payload` of `None`/`NULL` passed in). Is the Spanner query optimizer smart enough to e.g. not penalize an update that's directed to write the existing payload back into the row?

Good point/question.  I'm skeptical that'd be the case.

Let's keep an eye on the cost once this is deployed.  🤞

taddes (approved):

PR #2328, opened 2026-05-22
fix: update user_collection_quota

taddes:
## Description

Updates to simplified upsert in SpannerDb::update_user_collection_quotas

Metric note:
Moved the `storage.quota.update_existing_totals` into the `if` expr. I looked into `Metrics::start_timer` and found that because the .clone() produces an unbound temporary, the drop op runs at the end of the statement, so the "timer" records ~0 ms and is effectively a counter, not a duration. The change is therefore a reduction in count volume equal to the rate of quota disabled calls. The label is storage.quota.* and firing it where no quota work happens seemed misleading.

## Issue(s)

Closes [STOR-564](https://mozilla-hub.atlassian.net/browse/STOR-564).

[STOR-564]: https://mozilla-hub.atlassian.net/browse/STOR-564?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/syncstorage-spanner/src/db/mod.rs
+++ b/syncstorage-spanner/src/db/mod.rs
@@ -297,8 +297,6 @@ impl SpannerDb {
                 "fxa_kid" => user.fxa_kid.clone(),
                 "collection_id" => collection_id,
             };
-            // SUM/COUNT without GROUP BY always returns one row; COALESCE
-            // covers the empty-bsos case (collection just emptied 0, 0).
             let mut result = self
                 .sql(
                     "SELECT COALESCE(SUM(BYTE_LENGTH(payload)), 0), COUNT(*)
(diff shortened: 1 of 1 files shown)


syncstorage-spanner/src/db/mod.rs
```diff
@@ -285,105 +287,53 @@ impl SpannerDb {
...
-                SET modified = @modified
-                WHERE fxa_uid = @fxa_uid
-                AND fxa_kid = @fxa_kid
-                AND collection_id = @collection_id"
-            }
-        } else {
-            // Otherwise, there are no BSOs that match, check to see if there are
-            // any collections.
-            let result = self
+            // SUM/COUNT without GROUP BY always returns one row; COALESCE
```
pjenvey: nit: the group by comment isn't really necessary since we're removing it

pjenvey (dismissed):

syncstorage-spanner/src/db/mod.rs
taddes: Cool just took it out

pjenvey (approved):

PR #2336, opened 2026-05-26
chore: auto-rebase open Dependabot PRs on master merge

taddes:
## Summary
- Adds `.github/workflows/dependabot-auto-rebase.yml`.
- On every PR merge into master, the workflow lists open Dependabot PRs targeting master and posts `@dependabot rebase` on each.
- Uses the default `GITHUB_TOKEN` with `pull-requests: write` only — no code checkout, no third-party actions, no PAT.
- **Switches all three dependabot ecosystems (cargo, pip, github-actions) from `weekly` to `monthly` scheduling.**

## Why
Master moves leave open Dependabot PRs un-rebased. Manually commenting `@dependabot rebase` on each PR after every merge is repetitive toil and lets the dependency-update queue drift from mergeable state.

Per [review feedback from @pjenvey](https://github.com/mozilla-services/syncstorage-rs/pull/2336#issuecomment-4558828480), bundling a `weekly` to `monthly` schedule change with the auto-rebase workflow addresses the original concern raised by @chenba that auto-rebasing a large standing queue of Dependabot PRs would burn substantial CI. Going monthly should keep the typical standing queue much smaller (especially since `open-pull-requests-limit: 1` per ecosystem is already in place), while security advisories continue to land on their own cadence outside the schedule.

Plan: close stale pending updates before this merges (most notably #2286, superseded by #2340), then let the new schedule + auto-rebase settle in.

## Security shape
- `permissions: {}` at the workflow level — deny-all default.
- Job-level `pull-requests: write` only.
- Comment body is a fixed string (`@dependabot rebase`), not user-controlled — no comment injection surface.
- Trigger is `pull_request.closed` filtered to merges into master — only fires after a maintainer-approved merge. The PR HEAD is never checked out.
- Concurrency group coalesces overlapping merges; one in-flight run is enough since the survivor re-lists open PRs each time.

## Test plan
- [ ] Land PR.
- [ ] On the next master merge, watch the **Actions** tab for an "Auto-rebase Dependabot PRs" run.
- [ ] Verify it lists any open Dependabot PRs and posts one `@dependabot rebase` comment per PR.
- [ ] Confirm Dependabot reacts (force-pushes the branch / posts its own rebase confirmation) within ~1 minute.
- [ ] If no open Dependabot PRs exist at merge time, the run should log "No open Dependabot PRs targeting master" and exit cleanly.
- [ ] Observe over the next ~30 days that the monthly schedule produces a noticeably smaller standing queue.

Closes [STOR-573](https://mozilla-hub.atlassian.net/browse/STOR-573).

[STOR-573]: https://mozilla-hub.atlassian.net/browse/STOR-573?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


chenba:
I like the idea in general.  But given the number of dependabot PRs we often leave open, this causes a lot of unnecessary CI.

taddes:
> I like the idea in general. But given the number of dependabot PRs we often leave open, this causes a lot of unnecessary CI.

Indeed, that is a consideration for sure. I had thought of that, but was unsure if we're in danger of overusing our Actions Env here by running extra CI, since we also clear out the cache regularly. I mean... in the end, the CI runs anyway since we have to end up rebasing manually, right?

chenba:
> I mean... in the end, the CI runs anyway since we have to end up rebasing manually, right?

The difference is in quantity.  (Unless of course you manually do what the proposed workflow does.)

At the time of writing there are eight open dependbot PRs so let's use that as the example.  Assume we do nothing but focus on closing them.  After merging the first one, 7 rebases with CI.  Then 6, 5, and so on to a sum of 28.  That is the best case scenario.  If we focus on other work first, then it is 8 multiply by whatever the number of merges into main before we get started.

taddes:
> > I mean... in the end, the CI runs anyway since we have to end up rebasing manually, right?
> 
> The difference is in quantity. (Unless of course you manually do what the proposed workflow does.)
> 
> At the time of writing there are eight open dependbot PRs so let's use that as the example. Assume we do nothing but focus on closing them. After merging the first one, 7 rebases with CI. Then 6, 5, and so on to a sum of 28. That is the best case scenario. If we focus on other work first, then it is 8 multiply by whatever the number of merges into main before we get started.

Yeah, fair point. If you prefer we can cancel this then np.

pjenvey:
How about we switch the interval to monthly along with this PR, then try it (and maybe close the pending updates first)? That should in theory reduce the number of typically open dependabot PRs (though we'll still get security updates as soon as they land)

taddes:
> How about we switch the interval to monthly along with this PR, then try it (and maybe close the pending updates first)? That should in theory reduce the number of typically open dependabot PRs (though we'll still get security updates as soon as they land)

Great idea, I'll do that here and we'll get a good idea of perf there. I can close the pending updates too before we merge this.

chenba:
Considering we already have `open-pull-requests-limit: 1` set for the package ecosystems, majority of the open PRs are security updates.

pjenvey (approved):
I also logged https://mozilla-hub.atlassian.net/browse/STOR-577 which should be able to help reduce the number of dependabot PRs as well

pjenvey:
Also dependabot is supposed to merge changes in from the main/master branch when it detects them, though in practice I'm not sure it ends up doing this 100% of the time? If it does then the rebase forcing here just makes this occur quicker?

taddes:
> Also dependabot is supposed to merge changes in from the main/master branch when it detects them, though in practice I'm not sure it ends up doing this 100% of the time? If it does then the rebase forcing here just makes this occur quicker?

Yeah exactly, though I have found at times it doesn't do a good job and still requires some manual input, so hopefully this address that. We'll see if it all helps in the end.

taddes:
> I also logged https://mozilla-hub.atlassian.net/browse/STOR-577 which should be able to help reduce the number of dependabot PRs as well

Cool, I just filed a BugZilla for us to get approval to use this action

PR #2339, opened 2026-05-26
chore: add ssdlc sbom

clouserw:


.github/workflows/sbom.yml
```diff
@@ -0,0 +1,19 @@
+name: ssdlc-sbom
+
+on:
+  release:
+    types: [published]
+  push:
+    branches:
+      - 'release/**' #if triggering on branches, update to your specific branch pattern
```
chenba, line 8: Does this mean we need to push tags to a 'release/' branch from now on?

.github/workflows/sbom.yml
```diff
@@ -0,0 +1,19 @@
...
+  workflow_dispatch:
+
+permissions: {}
+
+jobs:
+  sbom:
+    uses: mozilla/actions/.github/workflows/sbom.yml@5cc2b20e42d53d1643853f12244fbec9ed2f2e0c
```
taddes, line 15:
```suggestion
    uses: mozilla/ssdlc-actions/.github/workflows/ssdlc-sbom.yml@5cc2b20e42d53d1643853f12244fbec9ed2f2e0c
```

Just missing the `ssdlc` part for the correct repo and workflow file: https://github.com/mozilla/ssdlc-actions/blob/main/.github/workflows/ssdlc-sbom.yml

.github/workflows/sbom.yml
```diff
@@ -0,0 +1,19 @@
...
+    permissions:
```
taddes, line 16:
```suggestion
    with:
      sbom_name: ${{ github.event.release.tag_name || github.sha }}
    permissions:
```
Looking at the doc, I think we need the `with block` since the Workflow downstream [needs the `sbom_name`](https://github.com/mozilla/ssdlc-actions/blob/9e64d8c1dc043cee460346789d7beb687fbf3f22/.github/workflows/ssdlc-sbom.yml#L5) as input.

.github/workflows/sbom.yml
```diff
@@ -0,0 +1,19 @@
...
+      actions: read # Required by anchore/sbom-action to read workflow run context.
+      contents: write # Required to upload the SBOM as a release asset.
```
taddes, line 18:
```suggestion
    permissions:
      actions: read       # required by anchore/sbom-action to read workflow run context
      contents: write     # required to upload the SBOM as a release asset
```

This scopes the permissions to the job-level which I think is the intent in the doc and leaving the higher level workflow permission as `{}` is currect.

.github/workflows/sbom.yml
taddes, line 8:
```suggestion
  push:
    branches:
      - 'release/**'
    tags:
      - '[0-9]+.[0-9]+.[0-9]+'
```
Makes sure tag based releases are covered, not only branch pushes. The tag part matches our server convention for reeleases.

taddes: Don't think so. AFAIK tags and branches are independent in GHA trigger filters.  branches: `release/** `match branch names like `release/0.22` (which we have 21 of; `git branch -r | grep release`.  `tags: release/**` would match tag names prefixed with `release/`, but we don't use those. Since our tags are bare semver , we just need to change `tags: release/**` to tags: `'[0-9]+.[0-9]+.[0-9]+'` so the trigger matches our actual tag convention.

taddes (changes requested):
Just some updates required.

taddes:
Closing as we're landing https://github.com/mozilla-services/syncstorage-rs/pull/2368 to add SBOM.

PR #2340, opened 2026-05-26
chore: bump diesel async deadpool

taddes:
## Description

Bumps `diesel-async` 0.7 to 0.9 and `deadpool` 0.12 → 0.13. These were the coordinated-major piece of dependabot PR #2326 . Cargo.toml's existing comments already noted these two must move together. Group A (the safe patch/minor bumps) landed separately under #2334.

No source changes were required:
- `.transaction()` closure signature changed in diesel-async 0.9 (scoped_boxed to real async closures), but this repo uses `TransactionManager` methods directly (`syncstorage-postgres/src/db/db_impl.rs:30-52`), so no call sites are affected.
- `deadpool` `Hook::async_fn` signatures at `syncstorage-{mysql,postgres}/src/pool.rs:80` and both tokenserver pools compile cleanly under the new version.

`cargo audit`'s allowlisted-warning count also dropped from 5 to 4 — the diesel-async bump pulled a vulnerable transitive forward.

### Held back

- `diesel_migrations 2.3.1 to 2.3.2` was originally included in this group but causes an `IntoUpdateTarget` trait-selection failure in `syncstorage-postgres` even with diesel-async 0.9. It's a downstream regression on its own, not a coupling issue. Suggest adding a `dependabot.yml` ignore entry separately so it stops re-appearing in grouped PRs.
- `protobuf 2.28 to 3.7.2` (from #2326) remains held unless  we were to update`google-cloud-rust-raw` to a protobuf-3.x release (unlikely).

Stacked on top of #2334 — rebase after that merges.

## Issue(s)

Closes [STOR-572](https://mozilla-hub.atlassian.net/browse/STOR-572).

[STOR-572]: https://mozilla-hub.atlassian.net/browse/STOR-572?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,4 +1,7 @@
 [workspace]
+# Virtual workspaces don't inherit the resolver from member editions, so v3
+# (the edition 2024 default) must be set explicitly here. See:
+# https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions
 resolver = "3"
 members = [
   "syncserver-common",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace]
-resolver = "2"
+resolver = "3"
 members = [
   "syncserver-common",
   "syncserver-db-common",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,13 +35,13 @@ docopt = "1.1"
 base64 = "0.22"
 
 diesel = "2.3"
-diesel-async = { version = "0.7", features = ["mysql", "deadpool", "migrations", "async-connection-wrapper"] } # worth checking when updating deadpool
+diesel-async = { version = "0.9", features = ["mysql", "deadpool", "migrations", "async-connection-wrapper"] }
 diesel_migrations = "2.3"
 
 cadence = "1.8"
 backtrace = "0.3"
 chrono = "0.4"
-deadpool = { version = "0.12", features = ["rt_tokio_1"] } # deps related to diesel-async, have to wait for them to update version to upgrade
+deadpool = { version = "0.13", features = ["rt_tokio_1"] }
 env_logger = "0.11"
 futures = { version = "0.3", features = ["compat"] }
 futures-util = { version = "0.3", features = [
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1050,7 +1050,7 @@ dependencies = [
  "downcast-rs",
  "itoa",
  "pq-sys",
- "uuid 1.23.1",
+ "uuid 0.8.2",
 ]
 
 [[package]]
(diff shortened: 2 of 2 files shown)


pjenvey (commented):
> diesel_migrations 2.3.1 to 2.3.2 was originally included in this group but causes an IntoUpdateTarget trait-selection failure in syncstorage-postgres even with diesel-async 0.9. It's a downstream regression on its own, not a coupling issue. Suggest adding a dependabot.yml ignore entry separately so it stops re-appearing in grouped PRs.

Are you able to reproduce this? I'm not seeing any build issues with `diesel_migrations` 2.3.2 against this branch

Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace]
-resolver = "2"
+resolver = "3"
```
pjenvey, line 5: I don't think this should be necessary as we target edition 2024 where it should be the default: https://doc.rust-lang.org/edition-guide/rust-2024/cargo-resolver.html

taddes: In my reading of this, our root Cargo.toml is a virtual manifest and there's no `[package]` section, only `[workspace].` The edition = "2024" is under [workspace.package] (line 22-27), which only sets the inherited default for member crates via edition.workspace = true. It is not the root package's own edition, because there is no root package. [](https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions )

taddes: and isn't it that because `default-members = ["syncserver"]` (line 20), the workspace is a pure container; the binary lives in the `syncserver/` member crate, therefore this resolver def becomes an issue? I'll again try and revert and see if I get an err again.

pjenvey: Ah, if the root package isn't opting in then let's fix that. Also now that you mention it, I'm wondering if our current explicit `resolver = 2` setting might be overwriting a new value of 3

taddes: I think it's ok the way it is now, with resolver = "3" commented out, cargo metadata emitted:

```
warning: virtual workspace defaulting to `resolver = "1"` despite one or more workspace members being on edition 2024 which implies `resolver = "3"`
  = note: to use the edition 2024 resolver, specify `workspace.resolver = "3"` in the workspace root's manifest
```

I think it's ok the way it is then, adding  a root [package]  would make the resolver "implied" via edition but doesn't  [package] need a target (ex. src/lib.rs)? Prob not worth it

pjenvey: Got it, makes sense. Thanks, TIL. I wouldn't mind a comment about that above the workspace resolver line

pjenvey (dismissed):

Cargo.toml
taddes: Good call, added one:

```
Virtual workspaces don't inherit the resolver from member editions, so v3
# (the edition 2024 default) must be set explicitly here. See:
# https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions```

pjenvey (approved):

PR #2360, opened 2026-06-03
chore: pre commit hooks 

taddes:
## Description

Adds pre-commit so we catch the dumb stuff (formatting, lint, clippy) locally before it hits a PR and breaks CI. Same checks CI already runs, just earlier. pre-commit is a poetry dev dep and the hooks call the existing Makefile targets, so there's one source of truth with CI.

Two stages:
* commit (fast): whitespace/EOF/yaml/toml hygiene, `cargo fmt`, ruff lint + format on `tools/`
* push (heavier): clippy (mysql always, postgres/spanner only when those crates change), cargo audit (only when deps change), mypy, bandit, pydocstyle

Setup is: `make install && make install-hooks`. Or you can run `poetry run pre-commit`

More detailed instructions are in `Contributing.md`

Added a handy `make clippy-all` command to do a single check.

## Testing

Run `poetry run pre-commit run --all-files` for the commit stage and `poetry run pre-commit run --all-files --hook-stage pre-push` for the heavy stage, or just commit and push something and watch it fire. Verified every hook runs, the backend clippy scoping triggers on the right crates, and cargo audit comes back clean (exits 0).

## Issue(s)

Closes [STOR-588](https://mozilla-hub.atlassian.net/browse/STOR-588)

[STOR-588]: https://mozilla-hub.atlassian.net/browse/STOR-588?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 2.4.1 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.3.2 and should not be changed by hand.
```
chenba: Not sure how much this downgrade matters but ?

taddes: AH good point, that's since my local Poetry is at that version, I'll correct.  ran poetry lock locally on Poetry 2.3.2, while the previous lockfile was generated by a newer Poetry enabled in the virtual env.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,139 @@
...
+#
+# Setup (pre-commit is a poetry dev dependency):
+#   make install                  # installs dev deps, incl. pre-commit, via poetry
+#   make install-hooks            # or: poetry run pre-commit install --install-hooks
+#
+# Run manually against everything:
+#   poetry run pre-commit run --all-files
+#
+# Two tiers, by stage:
+#   pre-commit  -> fast checks, run on every commit (hygiene, fmt, ruff)
```
chenba, line 15: I think these are also running pre-push?  The configs lack the `stages` that exist on the pre-push checks.

taddes: Aha good call,I added a default staging key that will run the lighter checks at pre-commit 👍

chenba (approved):

pjenvey (approved):

PR #2362, opened 2026-06-03
chore(settings): fail fast when DATABASE_URL is unset

taddes:
Previously the syncstorage and tokenserver settings defaulted database_url to "mysql://root@127.0.0.1/syncstorage" and "mysql://root@127.0.0.1/tokenserver" respectively. A missing SYNC_SYNCSTORAGE__DATABASE_URL or SYNC_TOKENSERVER__DATABASE_URL would therefore silently fall back to a local default rather than surfacing the misconfiguration.

The defaults are now empty, and Settings::validate() returns an error at startup when an enabled service has no database_url configured, naming the exact environment variable that must be set. A disabled service does not require its database_url.

Closes [STOR-545](hhttps://mozilla-hub.atlassian.net/browse/STOR-545)

[STOR-545]: https://mozilla-hub.atlassian.net/browse/STOR-545?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/syncserver-settings/src/lib.rs
+++ b/syncserver-settings/src/lib.rs
@@ -148,6 +148,19 @@ impl Settings {
     }
 
     pub fn validate(&self) -> Result<(), ConfigError> {
+        // Fail fast if an enabled service has no DATABASE_URL configured,
+        // rather than silently falling back to a default connection string.
+        if self.syncstorage.enabled && self.syncstorage.database_url.is_empty() {
+            return Err(ConfigError::Message(
+                "SYNC_SYNCSTORAGE__DATABASE_URL must be set".to_owned(),
+            ));
+        }
+        if self.tokenserver.enabled && self.tokenserver.database_url.is_empty() {
+            return Err(ConfigError::Message(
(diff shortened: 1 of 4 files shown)


syncserver-settings/src/lib.rs
```diff
@@ -148,6 +148,19 @@ impl Settings {
...
+        if self.syncstorage.enabled && self.syncstorage.database_url.is_empty() {
+            return Err(ConfigError::Message(
+                "SYNC_SYNCSTORAGE__DATABASE_URL must be set".to_owned(),
+            ));
+        }
+        if self.tokenserver.enabled && self.tokenserver.database_url.is_empty() {
+            return Err(ConfigError::Message(
+                "SYNC_TOKENSERVER__DATABASE_URL must be set".to_owned(),
+            ));
+        }
```
chenba, line 162: nit: would be nice to know about both missing values in one error.

chenba (approved):

PR #2381, opened 2026-06-10
feat: allow quota enforcement for all db backends

chenba:
This commit:
 - implements storage quotas for MySQL and Postgres
 - fixes a bug in Spanner's batch quota calculation

Closes STOR-598


pjenvey (approved):
Thanks Barry, lgtm. 

I remembered the history behind spanner's quota being per collection (and double checked w/ JR). We were more concerned about quota at the collection level due to the original issues encountered with spanner at the time in 2020 -- a small number of users with very large collections weren't able to cleanly separate across a Spanner split

PR #2393, opened 2026-06-15
fix: add settings for FxA acct event SET `aud` and `iss`

chenba:
## Description

Adding two settings for account event SET validation

## Issue(s)

Fixes STOR-606


tokenserver-settings/src/lib.rs
taddes, line 73: Possibly forgive my ignorance,  but if `fxa_webhook_enabled = true` but the new set vars are unset, is it that the webhook accepts requests but has zero verifiers? Would it still work? I wonder if for a follow up it might make sense to have some error handling around that

chenba: An http 401 is returned from the extractor if there are no verifiers.

taddes (approved):
👍

PR #2402, opened 2026-06-18
fix: do not validate `exp` on FxA event tokens

chenba:
SETs from FxA Event Broker do not include a `exp`.  Meanwhile the `jsonwebtoken` crate validates `exp` by default.  This patch disables the `exp` validation.


taddes (approved):
👍

PR #2407, opened 2026-06-22
feat: add payload offloading to GCS read/write paths

pjenvey:
## Description

via a new payload_link column (currently Spanner only)

## Issue(s)

Closes STOR-578
Closes STOR-579


syncserver/src/web/handlers.rs
```diff
@@ -389,10 +407,37 @@ where
...
+                    &coll.user_id,
+                    &coll.collection,
+                    &bso.id,
+                    payload,
+                )
+                .await?;
+                bso.payload = Some(String::new());
+                bso.payload_link = Some(url);
+            }
+        }
```
chenba, line 438: (minor, wishlist) might be nice to upload concurrently and failing fast here?

syncserver/src/web/payload_offload.rs
```diff
@@ -0,0 +1,121 @@
...
+        .build()
+        .await
+        .map_err(|e| ApiErrorKind::Internal(format!("GCS builder error: {e}")).into())
+}
+
+/// Upload `payload` to `bucket` under the key `{fxa_uid}/{collection}/{bso_id}`
+/// and return the resulting `gs://` URL.
+///
+/// The object is written with custom metadata `committed=false` and a
+/// `customTime` of the upload moment.
```
chenba, line 68: Given the fixed key, would that mean on an BSO update where the upload succeeded but DB transaction failed, the reaper process deletes the payload of an existing BSO?  I might be missing how that's addressed.

chenba (dismissed):
r+wc

👍 Patch looks good.  I have a question about data loss on BSO.  Maybe I missed something or it can be addressed separately.

syncserver/src/web/payload_offload.rs
pjenvey: Thanks for pointing this out. The GCS lifecycle policy will catch this (GC on custom-time > 1 day). Downstream of the Spanner change stream process will trigger committed payloads to switch their custom-time to some max sentinel value, avoiding being GC'd.

However I also intended to catch failed transaction commits in the handler, having it trigger immediate GCS delete attempts (since we know off hand they're not used at that point). I forgot to add this here, I'll add this in a separate issue.

syncserver/src/web/transaction.rs
```diff
@@ -164,6 +176,80 @@ impl DbTransactionPool {
...
+            let resource_ts = db
+                .extract_resource(
+                    self.user_id.clone(),
+                    self.collection.clone(),
+                    self.bso_opt.clone(),
+                )
+                .await
+                .map_err(ApiError::from)?;
+
+            if let Some(precondition) = &self.precondition.opt {
```
taddes, line 210: a nit, but the precondition match and X-Last-Modified injection are identical as in the from [transaction_http](https://github.com/mozilla-services/syncstorage-rs/pull/2407/changes#diff-b9c5941fe4cca7a8b52d4285a7f06f0b86cd7ea29dc0e534b68167ec077c798dL121) above, rather than shared, so the two could drift over time. Could the precondition check be factored into a small helper both calls?

syncserver/src/web/transaction.rs
taddes, line 133: Here's where I mean the same check could be de-duplicated. Agian, not a big deal

syncserver/src/web/handlers.rs
```diff
@@ -300,52 +301,70 @@ pub async fn delete_collection(
...
+    // against GCS once the DB transaction has closed
+    db_pool
+        .transaction_http_then(
+            &request,
+            async |db| {
+                coll.emit_api_metric("request.get_collection");
+                handle_not_found(db.get_bsos(params).await).map_err(Into::into)
+            },
+            async |mut bsos: Paginated<results::GetBso>| {
+                for bso in bsos.items.iter_mut() {
```
taddes, line 340: Maybe not an issue, but  If any single download_payload error hits, wouldn't the whole get_collection?full=1 response return 500, even though the DB read succeeded? Acceptable for v1, where one bad/missing object takes down the entire collection fetch.

syncserver/src/web/payload_offload.rs
```diff
@@ -0,0 +1,128 @@
...
+    collection: &str,
+    bso_id: &str,
+    payload: String,
+) -> Result<String, ApiError> {
+    let object_name = format!(
+        "{}/{}/{}/{}",
+        user_id.fxa_uid,
+        collection,
+        bso_id,
+        Uuid::new_v4().hyphenated()
```
taddes, line 82: Just gonna take note that we're planning on applying a UUID suffix at the end here? So for every new upload/PUT, there's a unique UUID? Reasonable for it making things unique, but I know we were thinking about how exactly to manage the lifecycle rules and this might factor into that decision. Necessary?

syncserver/src/web/payload_offload.rs
```diff
@@ -0,0 +1,128 @@
...
+use crate::{
+    error::{ApiError, ApiErrorKind},
+    server::ServerState,
+};
+
+const COMMITTED_METADATA_KEY: &str = "committed";
+
+/// Return the GCS bucket name if `collection` is opted into payload off-load
+/// and a bucket is configured. `None` disables off-load for this request.
+pub fn offload_bucket<'a>(state: &'a ServerState, collection: &str) -> Option<&'a str> {
```
taddes, line 34: Maybe some sort of backend check or startup config validation (via a protocol state) that confirms we're dealing with a Spanner backend here. I feel some misconfig or setup where `gcs_payload_bucket` is set on a Postgres/SQL backend, things get messy. I see in teh tests there's the `uses_spanner()` check, so something along those lines here?

pjenvey:
> I have a question about data loss on BSO. Maybe I missed something or it can be addressed separately.

I'll note that Barry's question pointed out a bug on updates which https://github.com/mozilla-services/syncstorage-rs/pull/2407/commits/920588e44f84805a71deb93bfd4810f82f9c29f6 fixed

syncserver/src/web/handlers.rs
pjenvey: That's right -- I don't think there's any alternative? If we can't fulfill the request it should fail (we can't partially give back results e.g.).

syncserver/src/web/payload_offload.rs
pjenvey: It needs to be unique as the `UPDATE` path adds a new GCS object -- there could be multiple GCS objects for a BSO at any time (but of course only one canonical one Spanner's `payload_link`). I'm totally not attached to this specific filename format otherwise.

syncserver/src/web/payload_offload.rs
pjenvey: Yup, this is sort of half baked, I logged https://mozilla-hub.atlassian.net/browse/STOR-627 for this. I originally prefixed all the config knobs with "spanner_" but arguably this feature could apply to other backends, so I renamed them to be less specific. But agreed we should disallow their usage on other backends for now.

syncserver/src/web/transaction.rs
pjenvey: Logged https://mozilla-hub.atlassian.net/browse/STOR-634 for this -- I made a very quick attempt to look at consolidating them last week but failed, there's probably something we can do here though.

syncserver/src/web/handlers.rs
pjenvey: https://mozilla-hub.atlassian.net/browse/STOR-633

syncserver/src/web/payload_offload.rs
pjenvey: https://mozilla-hub.atlassian.net/browse/STOR-623

taddes (approved):
🚀

PR #2413, opened 2026-06-26
chore: pilot parallel install steps in postgres CI job

taddes:
## What

Pilots the new GitHub Actions [parallel-steps feature](https://github.blog/changelog/2026-06-25-actions-steps-can-now-be-run-in-parallel/) (`background` / `wait-all`) on a **single job** `build-and-unit-test-postgres` to measure time saving and confirm the syntax behaves as documented.

## Change

- Marks the `apt-get` PostgreSQL client install as `background: true` so it overlaps the `cargo-nextest` / `cargo-llvm-cov` installs that follow.
- Adds a `wait-all` barrier before `Create Tokenserver database` (which needs the client).

`dpkg`, `cargo`, and `curl` use independent locks, so there is no contention.

## Scope / safety

- Scoped to one job; **no `permissions:`, secrets, or action pins changed**.
- Purely step reordering + two new keywords.

## What we're watching

- Whether the `background` / `wait-all` keywords parse and run as expected (feature is brand new; syntax not yet in the official workflow-syntax reference).
- Wall-clock delta on the Postgres unit-test job vs master, especially on cargo-tool cache-miss runs.

If the result looks good we'll file a STOR ticket and template the pattern across the other backend jobs.


pjenvey (approved):
lgtm, I'd lean towards adding a JIRA issue for tracking this even though it's a minor change

PR #2432, opened 2026-06-30
fix(sentry): redact sensitive request header values in captured events

taddes:
The Sentry middleware copied every inbound request header value verbatim onto captured error events. This redacts the values of headers that may carry credentials or user-identifying data (`Authorization`, `Proxy-Authorization`, `Cookie`, `X-Client-State`, `X-KeyID`, `X-Forwarded-For`) with a `[redacted]` placeholder before the request is built, leaving ordinary headers intact.

Adds unit tests covering both redaction (case-insensitive) and pass-through of ordinary headers.

## Testing
```
cargo test -p syncserver-common
```

Closes STOR-617


pjenvey (dismissed):
This works but going by the [sentry docs](https://develop.sentry.dev/sdk/foundations/data-scrubbing/#sensitive-data) they note the [before_send](https://docs.rs/sentry/latest/sentry/struct.ClientOptions.html#structfield.before_send) hook is a good place for such filtering (and more resilient if we were to start emitting sentry events elsewhere)

taddes:
> This works but going by the [sentry docs](https://develop.sentry.dev/sdk/foundations/data-scrubbing/#sensitive-data) they note the [before_send](https://docs.rs/sentry/latest/sentry/struct.ClientOptions.html#structfield.before_send) hook is a good place for such filtering (and more resilient if we were to start emitting sentry events elsewhere)

Really good call. Had a look and added coverage for `before_send`. Leaving the other in place so it's some defense in depth, though again... we're defending against ourselves a bit here lol.

taddes:
> > This works but going by the [sentry docs](https://develop.sentry.dev/sdk/foundations/data-scrubbing/#sensitive-data) they note the [before_send](https://docs.rs/sentry/latest/sentry/struct.ClientOptions.html#structfield.before_send) hook is a good place for such filtering (and more resilient if we were to start emitting sentry events elsewhere)
> 
> Really good call. Had a look and added coverage for `before_send`. Leaving the other in place so it's some defense in depth, though again... we're defending against ourselves a bit here lol.

Lmk if this looks like an ok solution to you @pjenvey

pjenvey (approved):
> we're defending against ourselves a bit here lol.

yea, I'd argue you don't need both but I won't block on that. Thanks

PR #2441, opened 2026-07-03
feat: add the change stream pipeline

pjenvey:
## Description

- payload_link_changes: New payload_link Spanner change stream
- payload-link-dataflow: Customized dataflow flex template in Java that consumes the change stream, producing pub/sub events. Derived from GCP's Cloud_Spanner_Change_Streams_to_PubSub template w/ a small change to ignore non payload_link changes (see the upstream-customization.patch diff)
- payload-reconciler: Consumes pub/sub events, producing reconciliation operations on GCS: either marking new payload objects as committed or deleting payload objects deleted in Spanner

## Testing

https://mozilla-hub.atlassian.net/browse/STOR-632

## Issue(s)

Closes STOR-583
Closes STOR-624
Closes STOR-625


tools/payload-link-dataflow/upstream.txt
```diff
@@ -0,0 +1 @@
+TBD_REPLACE_WITH_UPSTREAM_SHA
```
chenba: Since we have a patch above we should have a real value here?

tools/payload-reconciler/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Local install / tests
+
+```bash
+cd tools/payload-reconciler
+poetry install
+poetry run pytest
+```
+
+`poetry.lock` is generated on first install (not yet checked in).
```
chenba: poetry.lock is checked in; I see it above.

tools/payload-reconciler/reconcile_payload_links.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+def delete_object(gcs_client: storage.Client, bucket: str, name: str) -> None:
+    """Delete a GCS object. 404 is treated as success."""
+    blob = gcs_client.bucket(bucket).blob(name)
+    try:
+        blob.delete()
+        statsd.incr("payload_reconciler.orphan_deletes")
+    except gax_exceptions.NotFound:
+        log.warning("delete 404: gs://%s/%s", bucket, name)
```
chenba: Maybe I'm confused what this 404 means but does this warrant a warning when the comment says 404 is treated as success?

chenba (dismissed):
Left a few nitpicky comments but looks good overall

taddes (dismissed):
Good work. Most effort was reading and getting up to speed with implementation, but looks solid.

tools/payload-link-dataflow/upstream.txt
pjenvey: thanks, forgot about this. I've added a recent value, cleaned up the diff and its documentation (to clarify it's not the full diff) and added a script that generates the full diff, for if it's ever wanted

.gitignore
```diff
@@ -10,7 +10,6 @@ build
 ddb
 dist
 docs/_build
-*.xml
```
chenba, line 13: I think this means the test result xmls will stopped being ignored locally.  Maybe we can specifically ignore those *.xml?

chenba (dismissed):
👍

pjenvey:
Oops, totally forgot to push https://github.com/mozilla-services/syncstorage-rs/pull/2441/commits/344c43b172c0924c4e242f4ec0b28a340f399887 a couple days ago

chenba (approved):

PR #2450, opened 2026-07-08
feat: add optional per-collection storage limit overrides

chenba:
This commit:
 - adds an optional `limits.collections` map in settings that overrides
   the default global limits
 - enforces the per-collection max_record_payload_bytes limit

Closes STOR-587

### DIFF
--- a/syncserver-settings/src/lib.rs
+++ b/syncserver-settings/src/lib.rs
@@ -168,6 +168,16 @@ impl Settings {
             ));
         }
 
+        // overriding limits must be > 0.
+        for (name, overrides) in &self.syncstorage.limits.collections {
+            if overrides.max_record_payload_bytes == Some(0) {
+                return Err(ConfigError::Message(format!(
+                    "Invalid SYNC_SYNCSTORAGE__LIMITS__COLLECTIONS: \
+                     \"{name}\" max_record_payload_bytes must be greater than 0"
+                )));
+            }
+        }
+
         if let Some(init_node_url) = &self.tokenserver.init_node_url {
             let url = Url::parse(init_node_url).map_err(|e| {
                 ConfigError::Message(format!("Invalid SYNC_TOKENSERVER__INIT_NODE_URL: {e}"))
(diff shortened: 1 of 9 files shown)


syncstorage-db-common/src/lib.rs
```diff
@@ -233,6 +235,9 @@ pub trait Db: BatchDb {
 
     #[cfg(debug_assertions)]
     fn set_quota(&mut self, enabled: bool, limit: usize, enforce: bool);
+
+    #[cfg(debug_assertions)]
+    fn set_collection_limits(&mut self, _limits: std::collections::HashMap<i32, usize>) {}
```
taddes: Had a thought that since we're destructuing all over in the format (for example):

`if !STD_COLLS.iter().any(|(_, std_name, _)| std_name == name) {`

might be nice to add tidy named struct (Ex. s`truct StdColl { id, name, default_limit })` that would read better and survive the next field addition without touching every call site. Can be left as-is, but occurred to me as a possible improvement.

syncserver-settings/src/lib.rs
```diff
@@ -168,6 +169,16 @@ impl Settings {
             ));
         }
 
+        // Per-collection limits may only target standard collections in STD_COLLS.
+        for name in self.syncstorage.limits.collections.keys() {
```
taddes: we check the name here but never the value. a limit of 0 gets accepted and then every write to that collection trips the ceiling, since the check is >=. Maybe can we reject 0, and warn when a limit is larger than `max_request_bytes` so it can never actually be hit in one request?

syncserver/src/server/mod.rs
```diff
@@ -529,6 +528,29 @@ fn calculate_worker_max_blocking_threads(count: usize) -> usize {
...
+    let collection_limits = if quota_enabled {
+        limits.resolved_collection_limits_by_name(&STD_COLLS)
+    } else {
+        Default::default()
+    };
+    let mut limits_value = serde_json::to_value(limits).expect("ServerLimits failed to serialize");
+    if !collection_limits.is_empty()
+        && let Some(obj) = limits_value.as_object_mut()
+    {
+        obj.insert(
```
taddes, line 537: this advertises the collection limits whenever `enable_quota` is on, but they only get enforced with `enforce_quota`. so during the observe phase a client sees a limit we won't actually reject on. Maybe a sentence in config.md so nobody reads the advertised number as a hard guarantee or that it is confusing?

syncstorage-spanner/src/db/mod.rs
```diff
@@ -530,12 +534,23 @@ impl SpannerDb {
...
-        // return the over quota error.
+    pub(super) fn get_quota_limited_collection_ids(&self) -> Vec<i32> {
+        self.coll_limits.keys().copied().collect()
+    }
+
+    pub(super) fn get_collection_limit(&self, collection_id: i32) -> Option<usize> {
+        self.coll_limits.get(&collection_id).copied()
+    }
+
+    pub(super) fn emit_at_limit(&self, collection: &str, total_bytes: usize) -> DbResult<()> {
```
taddes: I believe this shifts spanner behavior on the batch path. before we only emitted storage.quota.at_limit when enforcing. now we always emit, then warn.  lines spanner up with mysql and pg which is good, but a warn only deployment will start emitting this metric where it didn't before. something to keep in mind for our logging.

taddes (dismissed):
Nice work @chenba , approving. left a few small notes (reject a 0 limit, and a heads up that the spanner batch path now emits storage.quota.at_limit in warn only mode where it didn't before) but none of them block.

pjenvey (changes requested):
Sorry I probably wasn't clear enough in the original https://mozilla-hub.atlassian.net/browse/STOR-587 -- my thought here about "limits" was in general (as we refer to everything published in `/info/collections` as [ServerLimits](https://github.com/mozilla-services/syncstorage-rs/blob/dc6907145e28a7e4968c8124718c2e9b4e234f30/syncstorage-settings/src/lib.rs#L198)). So in theory we could overwrite *any* of these limits on a per collection basis, but what's more important for expanded payloads is specifically `max_record_payload_bytes` (I only mentioned it here for the loadtester work https://mozilla-hub.atlassian.net/browse/STOR-585?focusedCommentId=1492389).

The idea would be say a specific collection such as "newtab-images" would set its `max_payload_record_bytes` to something larger than the default. This tells the client it can exceed the spanner max payload size and we would in turn offload those payloads. And the load tester would also utilize that information from the published limits.

chenba:
@pjenvey Thanks.  `max_record_payload_bytes_` makes more sense.  I should've put more thoughts into it or at least checked with you.

syncserver-settings/src/lib.rs
```diff
@@ -168,6 +169,22 @@ impl Settings {
             ));
         }
 
+        // Only standard collections in STD_COLLS and limits > 0 allowed
+        for (name, overrides) in &self.syncstorage.limits.collections {
+            if !STD_COLLS.iter().any(|(_, coll_name)| coll_name == name) {
```
pjenvey: Thanks, this looks good, though we might want to relax this restriction that they must be in `STD_COLLS` -- for now anyway (I think it will make testing easier).

pjenvey (dismissed):

syncserver-settings/src/lib.rs
chenba: > relax this restriction that they must be in `STD_COLLS` -- for now anyway (I think it will make testing easier).

Sounds good.  I'll update.

pjenvey (approved):

PR #2460, opened 2026-07-20
feat: attempt deleting gcs objects if db transaction failed

chenba:
This commit:
 - attempts to delete the offloaded gcs objects if the db transaction failed in put_bso or post_collection (only Spanner is supported at the moment)
 - uses the StorageControl struct since delete is an admin action
   - note that StorageControl uses gRPC instead of REST

Closes STOR-623


syncserver/src/web/handlers.rs
```diff
@@ -480,7 +482,15 @@ pub async fn post_collection(
                     "failed": coll.bsos.invalid,
                 })))
         })
-        .await
+        .await;
+
+    if resp.is_err() {
+        for url in offload_urls {
+            let _ = delete_payload(&url).await;
```
pjenvey, line 489: I wonder if it's worth emitting a metric upon such deletions? It could even tag whether or not it was successful. Not gonna block on that though (it could be a separate issue)

syncserver/src/web/payload_offload.rs
```diff
@@ -117,6 +117,31 @@ pub async fn download_payload(state: &ServerState, gs_url: &str) -> Result<Strin
         .map_err(|e| ApiErrorKind::Internal(format!("invalid utf-8 in GCS payload: {e}")).into())
 }
 
+async fn gcs_control_client() -> Result<StorageControl, ApiError> {
+    StorageControl::builder()
```
pjenvey, line 119: In retrospect I should have put a singleton `Storage` client on `ServerState` 🤦‍♂️ ([docs](https://docs.rs/google-cloud-storage/latest/google_cloud_storage/client/struct.Storage.html#pooling-and-cloning)). I'll log an issue for that to change it on both clients.

pjenvey (approved):

syncserver/src/web/handlers.rs
chenba: I'll file an issue.

PR #2469, opened 2026-07-23
refactor: convert BsoBodies extractor to async await

pjenvey:
## Issue(s)

Closes STOR-649

### DIFF
--- a/syncserver/src/error.rs
+++ b/syncserver/src/error.rs
@@ -101,6 +101,10 @@ impl ApiError {
         ApiErrorKind::Internal(msg.into()).into()
     }
 
+    pub fn no_server_state() -> Self {
+        ApiErrorKind::NoServerState.into()
+    }
+
     pub fn is_sentry_event(&self) -> bool {
         // Should we report this error to sentry?
         self.status.is_server_error()
--- a/syncserver/src/error.rs
+++ b/syncserver/src/error.rs
@@ -97,6 +97,10 @@ impl ApiErrorKind {
 }
 
 impl ApiError {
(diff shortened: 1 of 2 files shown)


syncserver/src/web/extractors/bso_bodies.rs
```diff
@@ -36,112 +33,80 @@ impl FromRequest for BsoBodies {
...
-            Some(s) => s,
-            None => {
-                error!("⚠️ Could not load the app state");
-                return Box::pin(future::err(
+                .into());
+            }
+
+            // Grab the max sizes
+            let Some(state) = req.app_data::<Data<ServerState>>() else {
+                return Err(ApiError::internal("No app_data").into());
```
chenba: There's an `ApiErrorKind::NoServerState`.  Can that be used here?

chenba (approved):

syncserver/src/error.rs
```diff
@@ -97,6 +97,14 @@ impl ApiErrorKind {
 }
 
 impl ApiError {
+    pub fn internal(msg: impl Into<String>) -> Self {
```
taddes, line 100: I see there are no callers now for this particular error. Is the intention it's one used down the road?  it's a pub fn on a pub struct, so the compiler won't emit dead_code, but was just curious. Makes sense as a companion to no_server_state()

taddes (approved):
lgtm, just had one question

syncserver/src/error.rs
pjenvey: Yea, that is the intention

PR #2472, opened 2026-07-23
docs: run clippy and tests after rust changes

taddes:
## Description

Adds one line to CLAUDE.md so the agent runs clippy and unit tests for the affected backend before finishing rust work, rather than leaning on CI to catch it. Points at the existing rust-code-smell skill for multi backend coverage.

No code or CI changes.


chenba (approved):
🤖

PR #2476, opened 2026-07-24
feat: upload/download to/from gcs concurrently

chenba:
This commit
 - performs uploads and downloads against gcs concurrently, up to some configured number of concurrent requests, when offloading is enabled
 - builds shared singletons of the Storage and StorageControl clients

Closes STOR-633, STOR-651

### DIFF
--- a/syncserver/src/server/mod.rs
+++ b/syncserver/src/server/mod.rs
@@ -29,7 +29,7 @@ use syncstorage_db::{DbError, DbPool, DbPoolImpl};
 use syncstorage_settings::{Deadman, ServerLimits};
 
 use crate::{
-    error::{ApiError, ApiErrorKind},
+    error::ApiError,
     tokenserver,
     web::{
         handlers, middleware,
--- a/syncserver/src/server/mod.rs
+++ b/syncserver/src/server/mod.rs
@@ -2,7 +2,6 @@
 
 use std::{convert::Infallible, num::NonZeroUsize, sync::Arc, time::Duration};
 
-use crate::error::ApiError;
 use actix_cors::Cors;
 use actix_web::{
     App, FromRequest, HttpRequest, HttpResponse, HttpServer,
@@ -15,6 +14,11 @@ use actix_web::{
 use cadence::{Gauged, StatsdClient};
 use futures::future::{self, Ready};
 use glean::server_events::GleanEventsLogger;
+use google_cloud_storage::client::{Storage, StorageControl};
+use tokio::{sync::RwLock, time};
+use utoipa::OpenApi;
+use utoipa_swagger_ui::SwaggerUi;
+
 use syncserver_common::{
     BlockingThreadpool, BlockingThreadpoolMetrics, Metrics, Taggable,
     middleware::sentry::SentryWrapper,
@@ -23,12 +27,15 @@ use syncserver_db_common::GetPoolStatus;
 use syncserver_settings::Settings;
 use syncstorage_db::{DbError, DbPool, DbPoolImpl};
 use syncstorage_settings::{Deadman, ServerLimits};
-use tokio::{sync::RwLock, time};
-use utoipa::OpenApi;
-use utoipa_swagger_ui::SwaggerUi;
 
-use crate::tokenserver;
-use crate::web::{handlers, middleware};
+use crate::{
+    error::{ApiError, ApiErrorKind},
+    tokenserver,
+    web::{
+        handlers, middleware,
+        payload_offload::{build_client, build_control_client},
+    },
+};
 
 pub const BSO_ID_REGEX: &str = r"[ -~]{1,64}";
 pub const COLLECTION_ID_REGEX: &str = r"[a-zA-Z0-9._-]{1,32}";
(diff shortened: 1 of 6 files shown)


syncserver/src/server/mod.rs
```diff
@@ -74,11 +81,35 @@ pub struct ServerState {
...
+    }
+
+    /// Shared GCS control-plane client.  Available when GCS off-load is enabled.
+    pub fn gcs_control_client(&self) -> Result<&StorageControl, ApiError> {
+        self.gcs_control_client.as_ref().ok_or_else(|| {
+            ApiErrorKind::Internal(
+                "GCS off-load enabled but control client not initialized".to_owned(),
+            )
+            .into()
+        })
```
pjenvey: nit (and I just added this method the other day):
```suggestion
        self.gcs_client.as_ref().ok_or_else(|| {
            ApiError::internal("GCS off-load enabled but client not initialized")
        })
    }

    /// Shared GCS control-plane client.  Available when GCS off-load is enabled.
    pub fn gcs_control_client(&self) -> Result<&StorageControl, ApiError> {
        self.gcs_control_client.as_ref().ok_or_else(|| {
            ApiError::internal("GCS off-load enabled but control client not initialized")
        })
```

syncserver/src/server/mod.rs
```diff
@@ -397,7 +428,17 @@ impl Server {
...
+        let (gcs_client, gcs_control_client) = match gcs_payload_bucket {
+            Some(_) => {
+                let endpoint = settings.syncstorage.gcs_endpoint.as_deref();
+                (
+                    Some(build_client(endpoint).await?),
+                    Some(build_control_client(endpoint).await?),
+                )
+            }
+            None => (None, None),
+        };
```
pjenvey: Not important/can leave as is: this block's only about the existence of `gcs_payload_bucket` so an if/else's maybe a little more to the point:

```suggestion
        let (gcs_client, gcs_control_client) = if gcs_payload_bucket.is_some() {
            let endpoint = settings.syncstorage.gcs_endpoint.as_deref();
            (
                Some(build_client(endpoint).await?),
                Some(build_control_client(endpoint).await?),
            )
        } else {
            (None, None)
        };
```

syncserver/src/web/handlers.rs
```diff
@@ -422,23 +443,39 @@ pub async fn post_collection(
...
+                let client = client.clone();
+                async move {
+                    upload_payload(&client, bucket, user_id, collection, &bso_id, payload)
+                        .await
+                        .map(|url| (i, url))
+                }
+            })
+            .buffer_unordered(state.gcs_payload_max_concurrency)
+            .try_collect()
+            .await?;
```
pjenvey, line 471: Might be worth noting in a comment here that any individual upload failure could leave other payloads orphaned, but the bucket will eventually delete them

pjenvey (dismissed):

syncserver/src/server/mod.rs
chenba: This is why `if` should've left out of the lang altogether ^_-

syncserver/src/server/mod.rs
```diff
@@ -74,11 +81,31 @@ pub struct ServerState {
...
-    /// Unset in prod; opt-in via `SYNC_SYNCSTORAGE__GCS_ENDPOINT` for
-    /// local/e2e testing against a fake-gcs-server or httptest mock.
-    pub gcs_endpoint: Option<String>,
+    /// Shared GCS client built at startup. `None` when GCS off-load is disabled.
+    pub gcs_client: Option<Storage>,
+
+    /// Shared GCS control-plane client built at startup.  `None` when GCS off-load is disabled.
+    pub gcs_control_client: Option<StorageControl>,
+
+    /// Maximum GCS uploads/downloads to run concurrently.
```
taddes, line 90: minor nit, but perhaps this should also be an Option?

```suggestion
    /// Maximum GCS uploads/downloads to run concurrently.  `None` when GCS off-load is disabled.
     pub gcs_payload_max_concurrency: Option<usize> ,
```

pjenvey (dismissed):

pjenvey:
just needs another clippy fix

syncserver/src/web/handlers.rs
```diff
@@ -337,10 +340,28 @@ pub async fn get_collection(
                 handle_not_found(db.get_bsos(params).await).map_err(Into::into)
             },
             async |mut bsos: Paginated<results::GetBso>| {
-                for bso in bsos.items.iter_mut() {
-                    if let Some(link) = bso.payload_link.take() {
-                        bso.payload = download_payload(&state, &link).await?;
-                    }
+                if let Some(client) = state.gcs_client.as_ref() {
```
taddes: If gcs_client is None we skip the whole block, so payload_link never gets taken and the response goes out with an empty payload plus the gs:// url sitting in the json. 

get_bso later in the file uses `state.gcs_client()? `and 500s instead. Can we do the same thing here? Otherwise if the bucket ever gets unset, I think clients quietly get blank records back and treat that as the real data.

syncserver/src/web/handlers.rs
```diff
@@ -422,23 +443,41 @@ pub async fn post_collection(
...
+            .map(|(i, bso_id, payload)| {
+                let client = client.clone();
+                async move {
+                    upload_payload(&client, bucket, user_id, collection, &bso_id, payload)
+                        .await
+                        .map(|url| (i, url))
+                }
+            })
+            .buffer_unordered(state.gcs_payload_max_concurrency)
+            .try_collect()
```
taddes, line 470: I think try_collect gives up on the first error, so any uploads that already finished get dropped on the floor and offload_urls never gets filled in. Then the cleanup later on has nothing to delete.

I wonder If GCS gets flaky partway through a big batch we end up leaving a pile of objects in the bucket that nothing in the db points at, and the client retry uploads a fresh copy of all of them. Could we collect everything, keep the urls that succeeded, and delete those before returning the error?

Maybe overkill for this feat or I'm overthinking it... can file a follow up

taddes (dismissed):
lgtm; have some nits and thoughts, but shouldn't be blocking.

syncserver/src/server/mod.rs
chenba: What does that get us?

chenba: hmmm actually I think I'll change it to a NonZeroUsize.

syncserver/src/web/handlers.rs
chenba: > Then the cleanup later on has nothing to delete.

Which cleanup are you referring to here?  The GCS lifecycle will clean up the uncommitted objects I do believe.

syncserver/src/server/mod.rs
taddes: Ok that's good! It was just consistency, since all GCS upload related values were Options as well

taddes (approved):

PR #2481, opened 2026-07-30
chore: quiet reconciliation e2e emulator logging

taddes:
The reconciliation e2e stack dumps a ton of emulator noise into the CI log, mostly per-request lines from fake-gcs and the pub/sub emulator, which buries the logs you actually want to read.

This quiets both emulators without touching what the tests do:

- **fake-gcs**: drop to `-log-level=warn`. Override with `FAKE_GCS_LOG_LEVEL=info` when you need the per-request output back.
- **pub/sub emulator**: it's a JVM process, so point it at a `java.util.logging` config at `WARNING` via `JAVA_TOOL_OPTIONS`, and disable netty's leak detector (it spits SEVERE ByteBuf-leak stack traces that are just an emulator quirk).

Pretty sure this is the cleanest way to do this from what I looked at 🤞 

Checked locally against the two emulators with the same traffic:
- fake-gcs: ~one line per request-ish, 0 lines at warn
- pub/sub: holds at its 7 startup lines even under load (was flooding a ton of INFO + leak traces before)

Closes [STOR-654]()https://mozilla-hub.atlassian.net/browse/STOR-654

[STOR-654]: https://mozilla-hub.atlassian.net/browse/STOR-654?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docker/docker-compose.e2e.reconciliation.yaml
+++ b/docker/docker-compose.e2e.reconciliation.yaml
@@ -20,6 +20,17 @@ services:
         gcloud beta emulators pubsub start
         --host-port=0.0.0.0:8085
         --project=test-project
+    # The emulator is a JVM process; JAVA_TOOL_OPTIONS points it at a
+    # java.util.logging config that raises its level to WARNING so it
+    # stops logging every RPC at INFO. leakDetection=disabled silences
+    # netty's SEVERE ByteBuf-leak stack traces, which are an emulator
+    # internal quirk we don't care about in a throwaway test. STOR-654.
+    environment:
+      JAVA_TOOL_OPTIONS: >-
+        -Djava.util.logging.config.file=/pubsub-logging.properties
+        -Dio.netty.leakDetection.level=disabled
+    volumes:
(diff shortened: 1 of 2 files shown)


taddes:
The whole job log went from 7,274 lines down to 2,082, so about 71% smaller.

Almost all of that is the two emulators this ticket was about. fake-gcs dropped from 3,962 lines to 11, and the pubsub emulator dropped from 1,038 to 28. They used to be roughly 69% of the log and now they're around 2%.

Nothing useful got lost. The reconciler, publisher, syncserver, spanner and the pytest output all have basically the same line counts as before, and  the emulator noise is gone. 🥳

chenba (approved):

pjenvey (approved):
thanks, and may I apologize again for Java

taddes:
> thanks, and may I apologize again for Java

We will one day be punished for using Java, I know it.

PR #2484, opened 2026-07-30
feat: allow per collection overrides for max_request/post_bytes

pjenvey:
## Description

I'll note this only checks `max_request_bytes` against the `Content-Length` header whereas `PayloadConfig`/`JsonConfig` additionally check against the actual payload reads as well. I'm not sure how critical that is but I'll likely log another issue for it -- it's a little bit involved (basically reimplementing actix body parsers), especially the Json version.

## Issue(s)

Closes STOR-642


syncstorage-settings/src/lib.rs
```diff
@@ -235,15 +239,58 @@ pub struct ServerLimits {
     pub collections: HashMap<String, CollectionLimitOverride>,
 }
 
+impl ServerLimits {
+    /// The effective
+    /// `max_record_payload_bytes`/`max_post_bytes``max_request_bytes` for
```
chenba:
```suggestion
    /// `max_record_payload_bytes`/`max_post_bytes`/`max_request_bytes` for
```

syncserver/src/web/extractors/utils.rs
```diff
@@ -69,6 +72,32 @@ pub fn get_accepted(req: &HttpRequest, accepted: &[&str], default: &'static str)
...
+        .and_then(|v| v.to_str().ok())
+        .and_then(|v| v.parse::<usize>().ok())
+        && len > max_request_bytes
+    {
+        return Err(ValidationErrorKind::FromDetails(
+            "size-limit-exceeded".to_owned(),
+            RequestErrorLocation::Header,
+            Some("Content-Length".to_owned()),
+            Some("request.validate.request_bytes_exceeded"),
+        )
```
chenba, line 95: This isn't converted into content too large http 413 correct?  We probably should return that to the client when the per collection limit's exceeded.

chenba (approved):

taddes (approved):

PR #2486, opened 2026-07-31
fix: stop the reconciler deleting live payload files on batch commit

taddes:
When a batch upload commits, the reconciler was deleting the GCS payload file that the committed record still points at, so later reads of that record failed with a 500. In CI this shows up as the flaky reconciliation e2e failures on the batch tests.

Why it happens: a batch write stores the payload link on a temporary batch_bsos row. On commit that same link is copied into the permanent bsos row and the batch_bsos rows are deleted. The change stream reports both tables, so the reconciler saw the batch_bsos deletion, treated it as an orphaned file, and deleted an object the bsos row still needs.

This change makes the reconciler ignore batch_bsos records and act only on bsos records. Committed batch files still get finalized through the bsos write. Files from batches that are abandoned or overwritten stay uncommitted and get swept by the GCS lifecycle policy, so nothing leaks. The table name is already in the message the reconciler reads, so no publisher or sync changes are needed.

Closes STOR-657

### DIFF
--- a/tools/payload-reconciler/reconcile_payload_links.py
+++ b/tools/payload-reconciler/reconcile_payload_links.py
@@ -137,6 +137,23 @@ def handle_message_body(
     success), so re-delivery is safe.
     """
     record: dict[str, Any] = json.loads(body)
+    table_name = record.get("tableName")
+
+    # A batch_bsos payload_link is a staging pointer. syncstorage only removes a
+    # batch_bsos row inside the commit transaction, and that same commit copies
+    # the link into the permanent bsos row, so deleting the object on that
+    # removal would drop one the committed bsos row still points at. We skip the
+    # delete for a batch_bsos row removed by a client transaction (the commit
+    # handoff). A batch_bsos row removed by a TTL system transaction is instead a
+    # genuinely abandoned batch, so its object should go. Spanner tags TTL
+    # row deletion policy deletes with transaction_tag "RowDeletionPolicy" and
+    # is_system_transaction true; a client delete has neither. bsos changes and
+    # batch_bsos overwrites are unaffected. See STOR-657.
+    is_ttl_delete = (
+        bool(record.get("isSystemTransaction"))
+        and record.get("transactionTag") == "RowDeletionPolicy"
+    )
+
     ops_performed = 0
     for mod in record.get("mods", []):
         old_values_str = mod.get("oldValues") or "{}"
@@ -151,6 +168,13 @@ def handle_message_body(
             ops_performed += 1
 
         if old_link and old_link != new_link:
+            # A batch_bsos row removed (new_link None) by a non-TTL transaction
+            # is the commit handoff: the object now belongs to bsos, so skip.
+            if table_name == "batch_bsos" and new_link is None and not is_ttl_delete:
+                metrics.incr("batch_commit_skips")
+                # Recognized and intentionally skipped, not filter noise.
+                ops_performed += 1
+                continue
             bucket, name = parse_gs_url(old_link)
             _require_bucket(bucket, expected_bucket)
             delete_object(gcs_client, bucket, name)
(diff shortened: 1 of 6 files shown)


tools/payload-reconciler/reconcile_payload_links.py
```diff
@@ -126,6 +126,19 @@ def handle_message_body(
...
+
+    # Ignore batch_bsos change records. A batch_bsos payload_link is a staging
+    # pointer: on batch commit it is copied into the permanent bsos row (which
+    # the change stream also reports, so the object is finalized via that bsos
+    # record) and the batch_bsos row is then deleted. Acting on that delete
+    # would remove a GCS object the committed bsos row still points at. Objects
+    # from batches that are abandoned or overwritten never reach bsos, stay
+    # committed=false, and are reclaimed by the GCS lifecycle policy. See
+    # STOR-657.
+    if record.get("tableName") == "batch_bsos":
```
chenba: If we always ignore then maybe should consider removing the change stream altogether.

pjenvey: I think we still want to treat deletes on `batch_bsos` via TTLs as actual deletes. syncstorage otherwise only triggers deletes of batches upon committing them (in the same transaction) so any TTL delete would have occurred outside that transaction, rendering their `payload_link`s as abandoned.

So the logic should be: the case of deletion of `batch_bsos` `payload_links` should only occur when the record has a `transaction_tag` of "RowDeletionPolicy" and `is_system_transaction` of `true`

pjenvey (commented):
sorry I forgot to submit this review the other day

tools/payload-reconciler/reconcile_payload_links.py
```diff
@@ -151,6 +168,13 @@ def handle_message_body(
             ops_performed += 1
 
         if old_link and old_link != new_link:
+            # A batch_bsos row removed (new_link None) by a non-TTL transaction
+            # is the commit handoff: the object now belongs to bsos, so skip.
+            if table_name == "batch_bsos" and new_link is None and not is_ttl_delete:
```
chenba: This doesn't cover cascading deletes resulting from `DELETE FROM user_collections`.  Those 'batch_bso' deletes should delete the gcs object referenced by the old link.

chenba: hmm the deletion policy is on 'batches.expiry' so `is_ttl_delete` might _always_ be false here.

pjenvey: > This doesn't cover cascading deletes resulting from `DELETE FROM user_collections`. Those 'batch_bso' deletes should delete the gcs object referenced by the old link.

Good catch. So we still need to special case ignore batch commit deletes, but treat TTL and the cascading deletes as deletes to GCS.

We can tag our transactions to distinguish between these cases, just like the TTL deletes tag their own ops, so it probably makes the most sense to tag the batch commit itself, treating anything without that specific tag as the latter case (delete to GCS).

We marked this issue closed a while ago so we should merge this as is and make this a separate issue.

> hmm the deletion policy is on 'batches.expiry' so `is_ttl_delete` might _always_ be false here.

I don't follow, can you elaborate?

chenba: > I don't follow, can you elaborate?

The first condition is `if table_name == "batch_bsos"` but the deletes on 'batch_bsos' is never (directly) from a deletion policy; they are cascading deletes resulting from a deletion policy.  So `not is_ttl_delete` is always true for deletes on "batch_bsos".

pjenvey: Got it. I'm guessing because they're part of the same transaction as the cascading delete on the parent that it would result in them being correctly marked (`is_ttl_delete` being True in the case we expect) -- we could veirfy this but I'm proposing we change our logic to fix the other issue anyway

taddes: K I filed https://mozilla-hub.atlassian.net/browse/STOR-668 to address as follow up and will get on it now. Meanwhile, made the interim fix skipping the batch_bsos removal.

pjenvey (approved):

PR #2504, opened 2026-08-05
fix: prevent new bso with empty payload and payload_link

chenba:
Closes STOR-660

### DIFF
--- a/tools/integration_tests/test_storage.py
+++ b/tools/integration_tests/test_storage.py
@@ -1376,27 +1376,20 @@ def test_update_of_ttl_without_sending_data(st_ctx):
     bso = {"payload": "x", "ttl": 1}
     retry_put_json(app, root + "/storage/xxx_col2/TEST1", bso)
     retry_put_json(app, root + "/storage/xxx_col2/TEST2", bso)
-    # Before those expire, update ttl on one that exists
-    # and on one that does not.
+    # Before those expire, update ttl.
     time.sleep(0.2)
     bso = {"ttl": 10}
     retry_put_json(app, root + "/storage/xxx_col2/TEST2", bso)
-    retry_put_json(app, root + "/storage/xxx_col2/TEST3", bso)
     # Update some other field on TEST1, which should leave ttl untouched.
     bso = {"sortindex": 3}
     retry_put_json(app, root + "/storage/xxx_col2/TEST1", bso)
-    # If we wait, TEST1 should expire but the others should not.
+    # If we wait, TEST1 should expire but TEST2 should not.
     time.sleep(0.8)
     items = app.get(root + "/storage/xxx_col2?full=1").json
     items = dict((item["id"], item) for item in items)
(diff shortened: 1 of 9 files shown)


taddes (dismissed):
👍

pjenvey (dismissed):
Can this be applied to the other write methods as well (post_bsos/batch)?

chenba:
> Can this be applied to the other write methods as well (post_bsos/batch)?

should be possible since THEN RETURN WITH ACTION yields one action per row.

chenba:
> > Can this be applied to the other write methods as well (post_bsos/batch)?
> 
> should be possible since THEN RETURN WITH ACTION yields one action per row.

It does add a DB read since we won't know the state of the bsos until the client tries to commit the batch.

tools/integration_tests/test_storage.py
```diff
@@ -1954,7 +1942,8 @@ def test_batch_ttl_is_based_on_commit_timestamp(st_ctx):
     resp = retry_post_json(app, collection + "?batch=true", [], status=202)
     batch = resp.json["batch"]
     endpoint = collection + "?batch={0}".format(batch)
-    retry_post_json(app, endpoint, [{"id": "a", "ttl": 3}], status=202)
+    # A payload is required
+    retry_post_json(app, endpoint, [{"id": "a", "payload": "x", "ttl": 3}], status=202)
```
pjenvey, line 1946: Shoot -- only seeing the need for this test change did I realize the ramifications of adding this validation and I think I've lead us down the wrong path here, as this test should continue to work per the API contract.

How it's treated currently (and we continue doing this with this change, we just trigger an error afterwards): we treat new bsos' missing fields as defaults, treating this as `"payload": ""`.

So this brings up a good edge case I didn't previously realize: for an offloaded collection we're continuing to use the `payload` column for this empty string default case, whereas the intention was for offloaded collections to never use `payload` at all (letting us even drop this column one day when all its values become NULL). Having an empty file in GCS instead is maybe a bit odd, but I think for simplicity sake that'd be preferable to confusing ourselves using `payload` for only this case.

..But then in this case we can't upload an empty file to GCS before hand, in the handler, because we don't know if we're inserting a new bso or just updating an existing one while ignoring its existing payload. 

Doh. Does this make sense? I guess I'm proposing we close this as done without merging (or maybe we can utilize some of it, albeit a bit differently?), but I'm not sure what to do with this edge case. Continue using `payload` here? Special case an empty string in `payload_link` for this case instead?

chenba: Yeah, it make sense.

I'm in favor of closing without merging.  As for the default payload url, maybe we could use a const url.  It can and should be a url to an actual empty, and not-writable, file.  But the url itself can be treated like a sentinel in the code.

pjenvey: @chenba Yea, I was thinking about exactly such a sentinel as I wrote the above comment -- it sounded better when you said it 😆 that's probably the way to go

PR #2507, opened 2026-08-06
feat: add a fine grained access role for reading the change stream

pjenvey:
## Description

This is for use by the access defined in https://github.com/mozilla-services/cloudops-infra/pull/6952

## Issue(s)

Issue STOR-655


taddes (approved):
🚀

PR #2509, opened 2026-08-07
feat: use statsd with tags in the payload link reconciler

chenba:
This commit
 - moves Metrics from tokenserser/util.py to a shared module
 - updates the statsd calls in the reconciler to use tags
 - updates the purge script to use the shared module

Closes STOR-645


taddes (approved):
Good call on shifting into shared util mod. Lgtm

pjenvey (approved):

PR #2520, opened 2026-08-13
refactor: extract and share common code between transaction_http{,_then}

chenba:
Closes STOR-634

### DIFF
--- a/syncserver/src/web/transaction.rs
+++ b/syncserver/src/web/transaction.rs
@@ -1,6 +1,6 @@
 use actix_http::{BoxedPayloadStream, Error, HttpMessage, Method, StatusCode, header::HeaderValue};
 use actix_web::dev::Payload;
-use actix_web::http::header;
+use actix_web::http::header::HeaderName;
 use actix_web::web::Data;
 use actix_web::{FromRequest, HttpRequest, HttpResponse};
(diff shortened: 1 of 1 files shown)


syncserver/src/web/transaction.rs
```diff
@@ -130,42 +158,36 @@ impl DbTransactionPool {
                 .await
                 .map_err(ApiError::from)?;
```
pjenvey:
```suggestion
.await?;
```

syncserver/src/web/transaction.rs
```diff
@@ -130,42 +158,36 @@ impl DbTransactionPool {
...
+                // See if we already extracted one and use that if possible
+                ensure_last_modified(&mut resp, resource_ts);
+                resp
+            }
         };
 
-        let (resp, mut db) = self
-            .transaction_internal(request, check_precondition)
-            .await?;
         // match on error and return a composed HttpResponse (so we can use the tags?)
```
pjenvey: this comment is kind of odd, probably should be deleted
```suggestion
```

pjenvey (dismissed):
Great, thanks. Noted a couple nits with the original code if you want to do any of these while you're in there

pjenvey (approved):

PR #2522, opened 2026-08-14
fix: SpannerIO requires the v2 runner

pjenvey:
## Description

and the v2 runner requires streaming engine

also pin the base Docker for reproducable builds

## Issue(s)

Closes STOR-669


chenba (approved):

taddes (approved):
👍

PR #2528, opened 2026-08-17
feat: add metrics for gcs payload cleanup on failed db transactions

taddes:
## Description

STOR-623 added a best-effort GCS delete for payloads that were offloaded before a
write's database transaction failed, but it was fire and forget, so there's no way
to tell from prod whether those deletes actually land. This adds a counter around
them.

The metric is `storage.gcs.payload.cleanup`, tagged with:

- `handler`: `put_bso` or `post_collection`
- `result`: `success` or `failure`
- `reason` on failures: `gcs_error` if GCS rejected the delete, `invalid_url` if the
  `gs://` URL didn't parse (a code bug rather than an infra problem)

Both are emitted inside `delete_payload`, so any future call site gets them for free.

Tests use cadence's `SpyMetricSink` over the existing `StorageControl` stubs, one per
branch. Commits are split so each addition is reviewable on its own.

Note on local verification: `cargo fmt --check` is clean, clippy is clean for
`syncserver`, and the `web::payload_offload` unit tests pass. The full
`make clippy_mysql` can't run on arm64 here because `grpcio-sys` fails to build, and
it also trips two pre-existing `needless_as_bytes` errors in `tokenserver-auth` that
are already on master, so CI is the real check for those.

## Review round

Rebased onto master, so the merge commit is gone and the history is linear.

- Dropped the `skipped` result entirely. Both GCS clients are built together at
  startup when a payload bucket is configured, so there is no state where the control
  client is missing but the upload succeeded. The counter could only ever read zero.
  Back to `if let Ok(client)` with a comment recording why it holds.
- Swapped the hand-rolled recording sink for cadence's `SpyMetricSink`, which needed
  `crossbeam-channel` as a dev-dependency to name the receiver. It was already in the
  lock via cadence.
- Dropped the `.config/nextest.toml` retry commit as out of scope. Worth noting for
  whenever it gets picked up separately: nextest profiles inherit from
  `[profile.default]`, so `test-threads = 1` does apply to `profile.ci`, which means
  parallel tests contending on shared rows can't be the cause of that abort.

I did not take the `gcs_control_client()?` variant. At the cleanup site the `?` would
return the GCS client error in place of the database error that actually caused the
rollback, so the real cause would be lost from both the response and Sentry.

## Testing

No behaviour change beyond the metric. `cargo test -p syncserver --lib web::payload_offload`
covers success, GCS failure, and unparseable URL.

## Issue(s)

Closes [STOR-653](https://mozilla-hub.atlassian.net/browse/STOR-653).

[STOR-653]: https://mozilla-hub.atlassian.net/browse/STOR-653?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/syncserver/src/web/handlers.rs
+++ b/syncserver/src/web/handlers.rs
@@ -29,7 +29,8 @@ use crate::{
             HeartbeatRequest, MetaRequest, ReplyFormat, TestErrorRequest,
         },
         payload_offload::{
-            delete_payload, download_payload, offload_bucket, reattach_by_index, upload_payload,
+            CLEANUP_METRIC, cleanup_tags, delete_payload, download_payload, offload_bucket,
+            reattach_by_index, upload_payload,
         },
         transaction::DbTransactionPool,
     },
@@ -526,12 +527,20 @@ pub async fn post_collection(
         })
         .await;
 
-    if resp.is_err()
-        && !offload_urls.is_empty()
-        && let Ok(client) = state.gcs_control_client()
-    {
-        for url in offload_urls {
-            let _ = delete_payload(client, &url, &metrics, "post_collection").await;
+    if resp.is_err() && !offload_urls.is_empty() {
+        match state.gcs_control_client() {
+            Ok(client) => {
+                for url in offload_urls {
+                    let _ = delete_payload(client, &url, &metrics, "post_collection").await;
+                }
+            }
+            // No client to clean up with: the objects are left to the
+            // reconciler, so count them rather than dropping the signal.
+            Err(_) => metrics.count_with_tags(
+                CLEANUP_METRIC,
+                offload_urls.len() as i64,
+                cleanup_tags("post_collection", "skipped"),
+            ),
         }
     }
 
@@ -864,9 +873,17 @@ pub async fn put_bso(
 
     if resp.is_err()
         && let Some(gcs_url) = &payload_link
-        && let Ok(client) = state.gcs_control_client()
     {
-        let _ = delete_payload(client, gcs_url, &metrics, "put_bso").await;
+        match state.gcs_control_client() {
+            Ok(client) => {
+                let _ = delete_payload(client, gcs_url, &metrics, "put_bso").await;
+            }
+            // No client to clean up with: the object is left to the
+            // reconciler, so count it rather than dropping the signal.
+            Err(_) => {
+                metrics.incr_with_tags(CLEANUP_METRIC, cleanup_tags("put_bso", "skipped"));
+            }
+        }
     }
     resp
 }
--- a/syncserver/src/web/handlers.rs
+++ b/syncserver/src/web/handlers.rs
@@ -29,8 +29,8 @@ use crate::{
             HeartbeatRequest, MetaRequest, ReplyFormat, TestErrorRequest,
         },
         payload_offload::{
-            CLEANUP_METRIC, CleanupHandler, CleanupResult, cleanup_tags, delete_payload,
-            download_payload, offload_bucket, reattach_by_index, upload_payload,
+            CleanupHandler, delete_payload, download_payload, offload_bucket, reattach_by_index,
+            upload_payload,
         },
         transaction::DbTransactionPool,
     },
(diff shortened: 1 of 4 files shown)


syncserver/src/web/payload_offload.rs
```diff
@@ -126,16 +133,50 @@ pub async fn build_control_client(endpoint: Option<&str>) -> Result<StorageContr
...
-        .map_err(|e| ApiErrorKind::Internal(format!("cannot delete GCS object: {e}")).into())
+        .await;
+
+    match &result {
+        Ok(_) => metrics.incr_with_tags(CLEANUP_METRIC, cleanup_tags(handler, "success")),
+        Err(e) => {
+            warn!("gcs payload cleanup failed for {gs_url}: {e}");
+            metrics.incr_with_tags(CLEANUP_METRIC, cleanup_failure_tags(handler, "gcs_error"));
+        }
+    }
```
chenba: You don't need to break up the chain for the side effect.  You can do

```rust      
      .inspect(|_| metrics.incr_with_tags(CLEANUP_METRIC, cleanup_tags(handler, "success")))
      .inspect_err(|e| {
          warn!("gcs payload cleanup failed for {gs_url}: {e}");
          metrics.incr_with_tags(CLEANUP_METRIC, cleanup_failure_tags(handler, "gcs_error"));
      })
```

syncserver/src/web/payload_offload.rs
```diff
@@ -126,16 +133,50 @@ pub async fn build_control_client(endpoint: Option<&str>) -> Result<StorageContr
...
+    metrics: &Metrics,
+    handler: &str,
+) -> Result<(), ApiError> {
+    let (bucket, object) = match parse_gs_url(gs_url) {
+        Ok(parsed) => parsed,
+        Err(e) => {
+            metrics.incr_with_tags(CLEANUP_METRIC, cleanup_failure_tags(handler, "invalid_url"));
+            return Err(e);
+        }
+    };
```
chenba: Since you are returning the Result as is (`parsed` -> `parsed`, `Err(e)` -> `Err(e)`) you can use `inspect_err` for the side effect.

syncserver/src/web/payload_offload.rs
```diff
@@ -126,16 +133,50 @@ pub async fn build_control_client(endpoint: Option<&str>) -> Result<StorageContr
...
+    let mut tags = cleanup_tags(handler, "failure");
+    tags.insert("reason".to_owned(), reason.to_owned());
+    tags
+}
+
+pub async fn delete_payload(
+    client: &StorageControl,
+    gs_url: &str,
+    metrics: &Metrics,
+    handler: &str,
```
chenba: Maybe make an enum instead of allowing any string.

syncserver/src/web/payload_offload.rs
taddes: ok cool, good suggestion

.config/nextest.toml
```diff
@@ -14,6 +14,10 @@ slow-timeout = { period = "300s" }
 
 [profile.ci]
 fail-fast = false
+# The Spanner emulator aborts a test transaction with a Conflict (503) when it
+# hits lock contention on the shared test rows. That is transient, so retry in
+# CI rather than failing the whole job; local runs keep the strict retries = 0.
+retries = 2
```
chenba: This is a bit out of scope.  But also, line 6 above is `test-threads = 1` so I'm not sure the lock contention is even possible?

taddes: It was a bug I was running into locally and that was the suggested fix 🤷 I can look a bit more closely at that part

syncserver/src/web/handlers.rs
```diff
@@ -857,9 +874,20 @@ pub async fn put_bso(
...
+                let _ = delete_payload(client, gcs_url, &metrics, CleanupHandler::PutBso).await;
+            }
+            // No client to clean up with: the object is left to the
+            // reconciler, so count it rather than dropping the signal.
+            Err(_) => {
+                metrics.incr_with_tags(
+                    CLEANUP_METRIC,
+                    cleanup_tags(CleanupHandler::PutBso, CleanupResult::Skipped),
+                );
+            }
```
pjenvey: I'd argue we don't need these `Err` blocks and their metrics at all because we should never have anything offloaded in the first place if there's no gcs_control_client (there should never be a lack of gcs_control_client if there is a gcs_client).

Or at least we could do this more simpler by just propagating the error (`gcs_control_client()?`) -- similarly to the above `state.gcs_client()?` -- if we're set to offload and there's no gcs clients, just fail with an error.

syncserver/src/web/payload_offload.rs
```diff
@@ -178,6 +268,34 @@ mod tests {
 
     use super::*;
 
+    /// Sink that keeps every statsd line so tests can assert on emissions.
+    #[derive(Debug)]
+    struct RecordingSink(Arc<Mutex<Vec<String>>>);
```
pjenvey: FYI cadence ships [a similar sink for testing](https://docs.rs/cadence/latest/cadence/struct.SpyMetricSink.html) -- pretty similar but it gives you a https://docs.rs/crossbeam-channel/0.5.15/crossbeam_channel/struct.Receiver.html you can iterate over (here's [an e.g. of it used in contile](https://github.com/mozilla-services/contile/blob/6035d1fb4cb6f0a0242585da2ac0c4774a9881f2/src/web/test.rs#L841))

pjenvey (approved):

PR #2525, opened 2026-08-17
docs: add the df job's metadata db to the arch diagram

pjenvey:
## Issue(s)

Issue STOR-636

PR #2538, opened 2026-08-20
docs: diagram and document the change stream system

taddes:
## Description

We had two solid docs on payload offload but nothing that says what the system
is at a high level, and no flow diagrams. This adds an overview page with
diagrams and cleans up drift in the existing docs.

New stuff:

- A Payload Offload overview page, promoted to its own section in the book
  instead of hanging off the Spanner DB page. Four diagrams on it: context,
  write path, read path, and the object lifecycle state machine.
- Redraws the reconciler's ASCII sketch as mermaid so it matches the rest of
  the book. Six mermaid diagrams total.
- Documents the `payload_link_reader` Spanner role. It was in neither doc
  despite being a live Dataflow launch param, so anyone standing up stage
  would have hit it cold.
- Documents the `tmpl-pub` CI identity, the cronjob invariants (the four
  settings you cannot change without breaking an assumption), and the
  reconciler metrics with what to alert on.
- Adds a stage and prod readiness checklist.

Fixes:

- The infra diagram declared `subgraph v1` twice, so the metadata database box
  never rendered separately. Came in with #2525.
- The reconciler doc said syncserver does an inline finalize when a write
  fails. It does a delete, and on an upload failure it does nothing inline at
  all.
- The Dataflow service account roles did not match the Terraform.
- Pins the lifecycle window at 30 days instead of "N days".
- The Spanner table reference was missing `payload_link` and still said
  `payload` is `NOT NULL`.
- Glossary entries and a pointer from architecture.md, since offload was not
  mentioned anywhere above these two pages.

Also turns on `--markdown-linebreak-ext` for the trailing-whitespace hook.
Without it, touching glossary.md strips the two-space line breaks and collapses
every term into its own definition.

Two things I found but left alone: SUMMARY.md points "Hawk Token Generator" at
`tools/hawk_make_token.md` but the file is `tools/make_hawk_token.md`, so
mdbook silently generates an empty page. And syncserver still emits no metrics
on the offload path, which is now written down in the docs but not ticketed.

## Testing

Docs only, no code changes.

```
make doc-prev
```

Click through Payload Offload, Reconciliation Pipeline, and GCP Infrastructure
and check the six diagrams render. `mdbook test docs/` passes.

## Issue(s)

Closes [STOR-631](https://mozilla-hub.atlassian.net/browse/STOR-631).

[STOR-631]: https://mozilla-hub.atlassian.net/browse/STOR-631?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/src/payload-offload/overview.md
```diff
@@ -0,0 +1,261 @@
...
+    stream --> dataflow
+    dataflow --> topic
+    topic --> reconciler
+    reconciler -->|"finalize or delete"| gcs
+```
+
+**Synchronous, in the request path.** On a write, syncserver uploads the
+payload to GCS *before* it opens the database transaction, then commits a row
+carrying the resulting URL. On a read, it fetches the row first, then
+downloads the payload from GCS *after* the transaction commits, and swaps it
```
pjenvey:
```suggestion
downloads the payload from GCS *after* the database read transaction is completed, and swaps it
```

docs/src/payload-offload/overview.md
```diff
@@ -0,0 +1,261 @@
...
+
+This is a work in progress. As of this writing:
+
+- Only the dev environment is built out. Stage and prod need the same set of
+  resources plus a dedicated Spanner metadata database.
+- A `batch_bsos` row removal is currently skipped wholesale rather than
+  distinguishing a batch commit handoff from a genuine delete, which leaks
+  the object in the genuine case. See STOR-668.
+- Syncserver emits no metrics on the offload path itself. Upload and download
+  latency, and the rollback cleanup, are invisible.
```
pjenvey: Let's kill these points as they'll be resolved soon

docs/src/payload-offload/overview.md
```diff
@@ -0,0 +1,261 @@
...
+- Change stream storage cost in prod has not been measured. See STOR-639.
+- The Dataflow job is launched with `on_delete = "cancel"` for testing.
+  Production wants `"drain"`.
```
pjenvey: I learned yesterday that the SpannerIO sink doesn't support draining at all, so this will be kept as `cancel` going forward (I've noted it in https://github.com/mozilla/webservices-infra/pull/12218)

https://docs.cloud.google.com/spanner/docs/change-streams/use-dataflow#draining

docs/src/tools/payload-offload-infrastructure.md
```diff
@@ -139,19 +145,76 @@ Names are derived from `${application}-${realm}-${environment}` in
...
+1. **Terraform.** Copy `bucket.tf`, `pubsub.tf`, `dataflow.tf` and
+   `dataflow_iam.tf` from `sync/tf/dev/` into the target environment
+   directory in webservices-infra.
+
+2. **A dedicated Spanner metadata database.** The change stream
+   connector keeps its partition state in a Spanner database of its own.
+   Dev shares `syncdb-dev` for convenience; stage and prod should each
+   get a separate database, for example `syncdb-pldf-meta-stage`, so the
+   connector holds no write access to the syncstorage database. This is
+   a cloudops-infra change.
```
pjenvey:
```suggestion
2. **A dedicated Spanner metadata database.** The change stream
   connector keeps its partition state in a Spanner database of its own.
   Each environment should get a separate database, for example 
   `syncdb-pldf-meta-stage`, so the connector holds no write access
   to the syncstorage database. This is a cloudops-infra change.
```

docs/src/tools/payload-offload-infrastructure.md
```diff
@@ -139,19 +145,76 @@ Names are derived from `${application}-${realm}-${environment}` in
...
+5. **Template spec.** Add a `write-payload-link-dataflow-spec-<env>` job
+   to `.github/workflows/mozcloud-publish.yaml`, and make sure the
+   environment's `-tmpl-pub` service account exists first. The spec has
+   to be in the bucket before the Terraform job resource is applied.
+
+6. **Job deletion behaviour.** Dev runs with `on_delete = "cancel"` and
+   `skip_wait_on_job_termination = true` so test iterations are fast.
+   Stage and prod want `on_delete = "drain"` and the provider default for
+   the wait, so a replaced job finishes publishing what it has already
+   read instead of dropping it.
```
pjenvey: Have to note we need on_delete = "cancel" going forward.

pjenvey (changes requested):
awesome, thanks for this

chenba:
I don't think we need to include all the jira issue numbers.  The jira number for enabling the feature in dev is neither here nor there, e.g..  Or if you think they are important, then maybe convert them to the mirrored gh issue numbers.

taddes:
> I don't think we need to include all the jira issue numbers. The jira number for enabling the feature in dev is neither here nor there, e.g.. Or if you think they are important, then maybe convert them to the mirrored gh issue numbers.

Sure happy to take them out, though I left them there as a pretty blatant reminder to update the docs since we've run into stuff getting forgotten after changes. It's a good idea to roll them into follow up issues like you suggest.

docs/src/tools/payload-offload-infrastructure.md
taddes: Ah right, good catch

docs/src/payload-offload/overview.md
taddes: Ah ok that's good to know, I'll make that update. I was basing it off the setting we had originally set when we were trying to run the atlantis apply

docs/src/payload-offload/overview.md
taddes: Ok wfm

pjenvey (approved):

PR #2552, opened 2026-08-26
feat: hard limit payload reads in extractors

taddes:
## Description

STOR-642 added per collection `max_request_bytes`, but we only ever compared it against the `Content-Length` header. Actix's `PayloadConfig`/`JsonConfig` limits are set at startup, before we know the collection, so they have to be sized to the largest limit across every collection. 

Both BSO body extractors now read the body with `to_bytes_limited` at the collection's own limit, so an oversized body is cut off as it arrives whilst streaming rather than after it has all been buffered. Made sure it follows same 413 and weave error code the header check already returned.

The PUT path used to use actix's `Json` extractor and now parses the bytes itself. I think this is ok. Status, response body and metric labels are unchanged, the only difference is the text that lands in logs. `Json` was also turning away `application/newlines` for us, so there's a check for that to keep the same response.

One caveat worth raising (mentioned in the ticket...): this may be overkill and ended up being a bit more complex than I would have liked. The ticket calls it minor, the load balancer caps request size ahead of us, and the server still has a hard overall limit, so the exposure was somewhat handled already. Fine by me if we'd rather not carry the extra code.

Docs for the actix pieces:

- `Payload::to_bytes_limited`: https://docs.rs/actix-web/4.13.0/actix_web/web/struct.Payload.html#method.to_bytes_limited
- `body::to_bytes_limited`, the one used here: https://docs.rs/actix-web/4.13.0/actix_web/body/fn.to_bytes_limited.html
- `BodyLimitExceeded`, the error meaning the cap was hit: https://docs.rs/actix-web/4.13.0/actix_web/body/struct.BodyLimitExceeded.html

Closes [STOR-672](https://mozilla-hub.atlassian.net/browse/STOR-672)

[STOR-672]: https://mozilla-hub.atlassian.net/browse/STOR-672?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/syncserver/src/web/extractors/bso_bodies.rs
+++ b/syncserver/src/web/extractors/bso_bodies.rs
@@ -90,8 +90,15 @@ impl FromRequest for BsoBodies {
                 }
                 Err(_) => return Err(size_limit_exceeded().into()),
             };
-            let Ok(body) = std::str::from_utf8(&body) else {
-                warn!("⚠️ Payload read error: body is not utf-8");
+            // Decode with the request's charset instead of assuming utf-8,
+            // as `String`'s extractor did for us before
+            let encoding = req.encoding().map_err(|e| {
+                warn!("⚠️ Payload read error: {:?}", e);
+                unreadable_body()
+            })?;
+            let Some(body) = encoding.decode_without_bom_handling_and_without_replacement(&body)
+            else {
+                warn!("⚠️ Payload read error: could not decode the body");
                 return Err(unreadable_body());
             };
 
@@ -108,7 +115,7 @@ impl FromRequest for BsoBodies {
                     }
                 }
                 bsos
-            } else if let Ok(json_vals) = serde_json::from_str::<Vec<Value>>(body) {
+            } else if let Ok(json_vals) = serde_json::from_str::<Vec<Value>>(&body) {
                 json_vals
             } else {
                 // Per Python version, BSO's must json deserialize
--- a/syncserver/src/web/extractors/bso_bodies.rs
+++ b/syncserver/src/web/extractors/bso_bodies.rs
@@ -2,6 +2,7 @@ use std::collections::{HashMap, HashSet};
 
 use actix_web::{
     Error, FromRequest, HttpMessage, HttpRequest,
+    body::{BodyStream, to_bytes_limited},
     dev::Payload,
     http::header::{ContentType, Header},
     web::Data,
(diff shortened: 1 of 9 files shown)


syncserver/src/web/extractors/bso_bodies.rs
```diff
@@ -74,18 +75,25 @@ impl FromRequest for BsoBodies {
...
-                        RequestErrorLocation::Header,
-                        None,
-                        None,
-                    )
-                })?;
+                    return Err(unreadable_body());
+                }
+                Err(_) => return Err(size_limit_exceeded().into()),
+            };
+            let Ok(body) = std::str::from_utf8(&body) else {
```
pjenvey: One missing bit I see vs actix-web's [FromRequest for String](https://docs.rs/actix-web/latest/src/actix_web/types/payload.rs.html#205) that I think we want compatibility with: it doesn't assume utf-8, [checks the charset from the request](https://docs.rs/actix-web/latest/src/actix_web/types/payload.rs.html#219) and if something other than utf-8, calls [encoding.decode_without_bom_handling_and_without_replacement(&body) instead](https://docs.rs/actix-web/latest/src/actix_web/types/payload.rs.html#255)

pjenvey (changes requested):
Cool, I thought this was going to be more involved -- I didn't notice actix-web exposed such reading bits publicly (I just looked at the FromRequest impls, there's probably some more code reuse they could improve on by using these calls)

syncserver/src/web/extractors/bso_body.rs
```diff
@@ -91,18 +99,25 @@ impl FromRequest for BsoBody {
             let coll_limits = state.limits.limits_for(collection.as_deref());
 
             check_content_length(&req, coll_limits.max_request_bytes as usize)?;
-            let bso = <actix_web::web::Json<BsoBody>>::from_request(&req, &mut payload)
```
chenba, line 94: I _think_ this means we can remove those `web::JsonConfig::default()`s from `syncserver/src/server/mod.rs`

syncserver/src/web/extractors/bso_bodies.rs
```diff
@@ -74,18 +75,32 @@ impl FromRequest for BsoBodies {
...
+                }
+                Err(_) => return Err(size_limit_exceeded().into()),
+            };
+            // Decode with the request's charset instead of assuming utf-8,
+            // as `String`'s extractor did for us before
+            let encoding = req.encoding().map_err(|e| {
+                warn!("⚠️ Payload read error: {:?}", e);
+                unreadable_body()
+            })?;
+            let Some(body) = encoding.decode_without_bom_handling_and_without_replacement(&body)
```
pjenvey, line 99: I'll note actix-web prefers `std::str::from_utf8` for the probably more common case here but I'm not sure it matters, this method returns a `Cow` for utf-8 (and so shouldn't allocate)

pjenvey (approved):

PR #2553, opened 2026-08-27
test: docker rig for offloaded payload load tests

taddes:
Docker rig for running the molotov load tests against the whole offload pipeline on emulators, plus a runbook to replace the Confluence page (its molotov and locust steps had gotten interleaved and unreadable).

```
make loadtest-offload-image
make loadtest-offload-handler       # STOR-629, 100% offloaded
make loadtest-offload-report
make loadtest-offload-changestream  # STOR-628, mixed
make loadtest-offload-report
```

No GCP project or credentials needed. Also fixes the loadtest Dockerfile, which hasn't built since poetry 2 dropped `--no-dev`, and drops the `loadtest-docker` target that pointed at a repo folded into this one years ago.

Three things turned up from actually running it:

- `max_record_payload_bytes` can only be raised per collection, for collections  in the offload list. Spanner caps a `STRING(MAX)` value at 2.5 MiB, so raising  it globally fails in the database on every non-offloaded collection.
- The Spanner emulator emits `transaction_tag` on change records but never fills it in, so the reconciler can't tell a batch-commit handoff from a real removal  and deletes the object the committed `bsos` row points at. Repro in `docker/batch-commit-probe.py`. Reads like an emulator gap rather than a code bug, but it means the STOR-668 skip can't be verified locally.
- compose never rebuilds `payload-link-publisher` when its source changes, so a stale image quietly runs against new server code. These targets pass
  `--build`; `docker_run_reconciliation_e2e_tests` doesn't.

STOR-629 is covered. STOR-628 is half covered: the NULL-filtering half checks
out (1340 published vs 1311 filtered on a mixed run), the batch-commit skip
needs a dev run with `STRICT_BATCH_COMMIT=1`.

Closes STOR-629
Refs STOR-628

### DIFF
--- a/docs/src/tools/load-testing.md
+++ b/docs/src/tools/load-testing.md
@@ -1,0 +1,738 @@
+# Load Testing
+
+## Quickstart: load test the change stream system (STOR-628)
+
+Four commands. Needs Docker, plus `curl` and `jq`. No GCP account, no
+credentials, no VM.
+
+```console
+make loadtest-offload-image          # once: builds the server image
+make loadtest-offload-changestream   # run the load test
+make loadtest-offload-report         # did the pipeline keep up?
+make loadtest-offload-down           # tear it all down
+```
+
+The report is the answer. Three numbers matter:
+
+| Look for | Meaning |
+| --- | --- |
+| `filtered` non-zero | The pipeline correctly dropped change records that have no payload file attached. This is the thing STOR-628 is asking about. |
+| `handler errors` zero | Nothing failed while finalizing or deleting objects. |
+| `noop_skips` zero | Nothing inert slipped past the filter. |
+
+It prints `OK: pipeline drained clean` at the end, or a list of `FAIL` lines
+explaining what went wrong, and exits non-zero on a real problem so it can gate
+a script.
+
+One thing to expect: a `NOTE` saying `batch_commit_skips` is 0. That is a known
+emulator limitation rather than your run going wrong. See
+[The emulator drops transaction tags](#the-emulator-drops-transaction-tags).
+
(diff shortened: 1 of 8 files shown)


pjenvey:
> * The Spanner emulator emits `transaction_tag` on change records but never fills it in, so the reconciler can't tell a batch-commit handoff from a real removal  and deletes the object the committed `bsos` row points at. Repro in `docker/batch-commit-probe.py`. Reads like an emulator gap rather than a code bug, but it means the [STOR-668](https://mozilla-hub.atlassian.net/browse/STOR-668) skip can't be verified locally.

Sigh: https://github.com/GoogleCloudPlatform/cloud-spanner-emulator/issues/280

[STOR-668]: https://mozilla-hub.atlassian.net/browse/STOR-668?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

docs/src/tools/load-testing.md
```diff
@@ -0,0 +1,738 @@
...
+```
+
+Then open `localhost:8090`. See `tools/tokenserver/loadtests/README.md` for
+the rest.
+
+---
+
+## Gaps worth knowing about
+
+- **The offload read/write path is not timed.** `payload_offload.rs` emits
```
pjenvey, line 726: Good point, I logged https://mozilla-hub.atlassian.net/browse/STOR-708

pjenvey (approved):

docs/src/tools/load-testing.md
taddes: Cool, I'll take care of it

PR #2560, opened 2026-09-01
docs: refresh the payload offload docs

taddes:
## Description

Refreshes the three payload offload docs.

Two are done, but one will need a touch up later:

- STOR-668 (tag-based batch commit detection) landed in #2518 , the metrics table listing `batch_bsos_skips` was missed. The code emits `batch_commit_skips`, and the "while the blanket skip is in place" note was stale.
- Syncserver offload metrics had no ticket when STOR-681 was filed. They became STOR-653 and landed in #2528  That added one counter, we needed to consider`storage.gcs.payload.cleanup`, on the rollback cleanup. Upload and download are still uninstrumented (tbd), so the docs now say that plainly rather than implying (and so we don't forget).
- STOR-639 (change stream storage sizing) is still in progress, so I've made note on that ticket to update docs when we make the final choice.

Two things added (since more stuff has been merged since tkt was filed_:

- `spannerMetadataTableName` (STOR-683, #2543) was missing from the launch-parameter list. Worth reading if nothing else: unpinned, a `--update` or cancel-then-relaunch resumes from `Timestamp.now()` against a fresh metadata table and silently drops every record committed in between. 
- The doc claimed the dev metadata table shares `syncdb-dev`, but not so... It uses the dedicated `syncdb-pldf-meta-dev`, which also makes the service account grant wrong.

Added  "batch commit is a handoff, not a delete" invariant to the overview. 

## Testing

`mdbook test docs/` passes clean.

Values document were checked against the live dev resources rather than inferred (or left up to my memory or glancing around). The running Dataflow job pins `spannerMetadataTableName: Metadata_payload_link`, and the rest of the infra doc's dev-resources table matches the job's pipeline options. Both Pub/Sub subscriptions match the documented ack deadline, retention, and delivery attempts:

To check:
```
gcloud dataflow jobs describe <JOB_ID> --project=moz-fx-sync-nonprod --region=us-west1 --full --format=json
```

I had an initial hiccup... the `--full` flag matters here, since the default view omits pipeline options. 😕 

## Issue(s)

Closes [STOR-681](https://mozilla-hub.atlassian.net/browse/STOR-681).

[STOR-681]: https://mozilla-hub.atlassian.net/browse/STOR-681?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/src/tools/payload_link_reconciler.md
+++ b/docs/src/tools/payload_link_reconciler.md
@@ -374,10 +374,16 @@ real dev Spanner instance).
 
 ## Metrics
 
-The reconciler emits statsd counters under the `payload_reconciler`
-namespace. Everything the pipeline can tell you about itself is in this
-list, so it is worth knowing what each one looks like when things are
-working.
+Two processes report on this system, and they are instrumented to
+different depths. The reconciler covers the asynchronous arm end to end.
+The syncserver side has one counter, on the rollback cleanup, and
+nothing on upload or download.
+
+### Reconciler
+
+Statsd counters under the `payload_reconciler` namespace. Everything the
+asynchronous arm can tell you about itself is in this list, so it is
+worth knowing what each one looks like when things are working.
 
 | Metric | Meaning | Healthy shape |
 |---|---|---|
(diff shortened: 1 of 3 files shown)


docs/src/tools/payload_link_reconciler.md
```diff
@@ -404,6 +454,8 @@ Worth alerting on:
 | Sustained `payload_reconciler.gcs_404` with `op:delete` | Object was already deleted (redelivery or concurrent cleanup) | Acceptable; idempotent by design. |
 | Messages in `payload-link-changes-dlq` | Repeated handler exceptions on the same message after 5 retries (malformed JSON, cross-bucket link, GCS auth failure) | Inspect the DLQ payload; fix and re-publish or discard. The main subscription continues to drain. |
 | `payload_reconciler.errors` with `kind:handler` non-zero | Same as above before reaching DLQ. | Same. |
+| `storage.gcs.payload.cleanup` with `result:failure` | The compensating delete after a failed write transaction did not land | The object is stranded at `committed=false` and the 30 day lifecycle policy is the only thing left to reap it. Tolerable in ones; sustained means the write path is failing and cleanup is failing with it. |
```
chenba: > Tolerable in ones;

I know what you mean but that sounds so odd.  "Occasional failures are tolerable" or something like that?

taddes: Yeah that does read better

taddes: Swapped it for "Occasional failures are tolerable" in 642cdc17.

pjenvey (approved):

PR #2564, opened 2026-09-01
feat: update off-loaded gcs object name format

chenba:
Use the gcs object name as recommended by the doc in stor-581.

Closes STOR-692

### DIFF
--- a/syncserver/src/web/extractors/test_utils.rs
+++ b/syncserver/src/web/extractors/test_utils.rs
@@ -80,6 +80,7 @@ pub fn make_state_with_limits(limits: Arc<ServerLimits>) -> ServerState {
         gcs_payload_bucket: None,
         gcs_payload_max_concurrency: syncstorage_settings.gcs_payload_max_concurrency,
         gcs_payload_offload_collections: Arc::new(Vec::new()),
+        gcs_payload_prefix: Arc::new(syncstorage_settings.gcs_payload_prefix.clone()),
     }
 }
 
--- a/syncserver/src/web/payload_offload.rs
+++ b/syncserver/src/web/payload_offload.rs
@@ -10,6 +10,8 @@
 //! resolved by downloading from GCS after the database transaction commits,
 //! and `payload_link` is cleared before the response is rendered.
 //!
+//! Object names are formatted as `{gcs_payload_prefix}/{fxa_uid}/{uuid}`.
+//!
 //! Objects are written with the custom metadata `committed=false` and a
 //! `customTime` set to upload time; a later step flips `committed` to `true`
 //! once the database row is durably visible.
@@ -65,24 +67,22 @@ pub async fn build_client(endpoint: Option<&str>) -> Result<Storage, ApiError> {
         .map_err(|e| ApiErrorKind::Internal(format!("GCS builder error: {e}")).into())
 }
 
-/// Upload `payload` to `bucket` under the key `{fxa_uid}/{collection}/{bso_id}`
-/// and return the resulting `gs://` URL.
+/// Upload `payload` to `bucket` under the key `{prefix}/{fxa_uid}/{uuid}` and
+/// return the resulting `gs://` URL.
 ///
 /// The object is written with custom metadata `committed=false` and a
 /// `customTime` of the upload moment.
 pub async fn upload_payload(
     client: &Storage,
     bucket: &str,
+    prefix: &str,
     user_id: &UserIdentifier,
-    collection: &str,
-    bso_id: &str,
     payload: String,
 ) -> Result<String, ApiError> {
     let object_name = format!(
-        "{}/{}/{}/{}",
+        "{}/{}/{}",
+        prefix,
         user_id.fxa_uid,
-        collection,
-        bso_id,
         Uuid::new_v4().hyphenated()
     );
 
--- a/syncserver/src/web/payload_offload.rs
+++ b/syncserver/src/web/payload_offload.rs
@@ -14,7 +14,8 @@
 //!
 //! Objects are written with the custom metadata `committed=false` and a
 //! `customTime` set to upload time; a later step flips `committed` to `true`
-//! once the database row is durably visible.
+//! once the database row is durably visible. The objects also carry
+//! `bso_id`, `fxa_kid`, `original_size`, and `format_version`.
 
 use std::{collections::HashMap, time::SystemTime};
 
@@ -30,6 +31,14 @@ use crate::{
 };
 
 const COMMITTED_METADATA_KEY: &str = "committed";
+
+/// Custom metadata keys
+const BSO_ID_METADATA_KEY: &str = "bso_id";
+const FXA_KID_METADATA_KEY: &str = "fxa_kid";
+/// Byte length of the payload as received
+const ORIGINAL_SIZE_METADATA_KEY: &str = "original_size";
+const FORMAT_VERSION_METADATA_KEY: &str = "format_version";
+const FORMAT_VERSION: &str = "1";
 
 /// Counter for the best-effort GCS cleanup that runs when the database
 /// transaction of an offloading write fails. Tagged with the `handler` that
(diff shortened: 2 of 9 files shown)


syncserver/src/web/extractors/test_utils.rs
```diff
@@ -73,6 +73,7 @@ pub fn make_state() -> ServerState {
         gcs_payload_bucket: None,
         gcs_payload_max_concurrency: syncstorage_settings.gcs_payload_max_concurrency,
         gcs_payload_offload_collections: Arc::new(Vec::new()),
+        gcs_payload_prefix: Arc::new("v1".to_owned()),
```
taddes: Nit: the fields around this pull from syncserver_settings, this one hardcodes "v1"... not a dealbreaker, but something worth considering.

syncstorage-settings/src/lib.rs
```diff
@@ -137,6 +137,10 @@ pub struct Settings {
     /// off-load path for all collections.
     pub gcs_payload_offload_collections: Vec<String>,
 
+    /// Prefix off-loaded GCS object names, which are formatted as
+    /// `{gcs_payload_prefix}/{fxa_uid}/{uuid}`.
+    pub gcs_payload_prefix: String,
```
taddes, line 142: nit: making the version string a setting instead of a hardcoded v1/ ? .

syncserver/src/web/payload_offload.rs
taddes, line 95: This is the spot for the rest of the metadata listed in #9 of our format doc (format-version, bso_id, fxa_kid, collection_id, original-size).

syncserver/src/web/payload_offload.rs
```diff
@@ -65,24 +67,22 @@ pub async fn build_client(endpoint: Option<&str>) -> Result<Storage, ApiError> {
...
     bucket: &str,
+    prefix: &str,
     user_id: &UserIdentifier,
-    collection: &str,
-    bso_id: &str,
     payload: String,
 ) -> Result<String, ApiError> {
     let object_name = format!(
-        "{}/{}/{}/{}",
+        "{}/{}/{}",
```
taddes, line 83: This matches our doc spec's Option C / #9 , though under metadata it also lists bso_id, fxa_kid, collection_id, format-version and original-size as object metadata, and we only write "committed" string in `set_metadata` atm. I think for us to have better traceability, that's where we'd want it.

Follow-up ticket is probs best, I think just want it to exist. However, you'd have to put the  bso_req.bso in put_bso; or bso.id in post_collection that was removed here to get it, but it's free. As for fxa_kid that's there in `UserIdentifier`. Format version is just "1" and `original-size` can just be calling  `payload.len()`before the String moves into `upload_payload`. Looking at it though, `collection_id` may be tricky. Either way, this might be the wrong place for this...

taddes (commented):
thanks for doing this, just some nits:

Main thing in looking back at our  docs (# 9 for our selected choice) is the metadata section lists bso_id, fxa_kid, collection_id, format-version and original-size as object metadata... atm we only write "committed" for COMMITTED_METADATA_KEY.  bso_id, fxa_kid, format-version, and original-size are all sitting right there at both call sites; collection_id is the only awkward one since it doesn't exist above the DB layer and the uploads run pre-transaction, so the collection name might be the better call.  Happy for it to be a follow-up, just want the ticket to exist (if it matters to us). I wrote more in the inline section.

Looking back, # 10 in our doc has "fxa_uid in names: plaintext vs HMAC(uid)" as an open question and this impl uses plaintext. Did we settle that fer sher?

final nit about docs: the Syncstorage Payload offload table needs a `SYNC_SYNCSTORAGE__GCS_PAYLOAD_PREFIX` row added.
(when I looked, it seems we're alsomissing SYNC_SYNCSTORAGE__GCS_PAYLOAD_MAX_CONCURRENCY and SYNC_SYNCSTORAGE__GCS_ENDPOINT ☹️ )  Might as well catch all three up while you're in there?

syncstorage-settings/src/lib.rs
chenba: This is in `Settings`.  I'm afraid I don't understand your comment. :-/

syncserver/src/web/payload_offload.rs
chenba: > under metadata it also lists ...

Where in that doc does it explain how those metadata fields are going to be used?  I'm having a hard time trying to understand where they came from.

taddes: How it's exactly used I don't 100% recall when we made the doc, but I think we thought it would be a nice to have

syncstorage-settings/src/lib.rs
taddes: Gah ignore me, meant by comment above. but it did make me realize (and this really might not matter) but  it's an unvalidated String, so "v1/" or "" from config gives you `v1//{uid}/{uuid} `or `/{uid}/{uuid},`. Would require quite a config blunder to do so... A trim of '/' on load would cover it. Low stakes, feel free to skip. I'll drop a suggestion.

chenba:
> Looking back, # 10 in our doc has "fxa_uid in names: plaintext vs HMAC(uid)" as an open question and this impl uses plaintext. Did we settle that fer sher?

#10 doesn't explain why we'd want to hmac or what we hmac it with.  Where can I find that?

syncstorage-settings/src/lib.rs
taddes: A trim in `Settings::normalize` handles the slash typo, (will leave suggestion on ln 194). The empty case has to go in `syncserver-settings/src/lib.rs` `validate()`, near the existing offload check around line 212, since "" trims to "" and still gives a leading slash:

```rust
let prefix = &self.syncstorage.gcs_payload_prefix;
if prefix.is_empty() || prefix.trim_matches('/') != prefix {
    return Err(ConfigError::Message(
        "SYNC_SYNCSTORAGE__GCS_PAYLOAD_PREFIX must be non-empty and must not \
         have leading or trailing slashes"
            .to_owned(),
    ));
}
```

syncstorage-settings/src/lib.rs
taddes, line 194:
```suggestion
        }

        // Object names are joined as `{prefix}/{fxa_uid}/{uuid}`, so a leading
        // or trailing slash on the configured prefix yields an empty segment.
        self.gcs_payload_prefix = self.gcs_payload_prefix.trim_matches('/').to_owned();
```

taddes: (totally optional)

taddes (dismissed):
🚀

syncserver/src/web/payload_offload.rs
pjenvey: What's potentially useful about those fields is the ability to "reconstruct" the primary key from the object in GCS (the doc mentions "reconstructing" but only from the perspective of constructing the filename from the PK) -- I think that's potential useful when we consider alternative means of reconciliation (not using the data pipeline).

Or it could be useful for reporting purposes, imagine a job against GCS that gives us a report about usage. Maybe we wouldn't need bso_ids for such a report, but collection_id might be nice. 

I've been looking more closely at the GCS inventory reports, and noticed that they can't read custom metadata, making me wonder more about the recommended #9 proposal..

Not that I'm sold on using an alternative to the dataflow job for reconciliation, but I could totally see it being useful for possibly having something that can audit the correctness of our current pipeline. Or say as a fallback if there was a bug found in the pipeline, we could use the report as a basis for a more manual way of checking/fix up the data between GCS + spanner

chenba: > I've been looking more closely at the GCS inventory reports, and noticed that they can't read custom metadata

right, so _if_ we want those ids, we might've as well jam them into the object name?

chenba:
Chatted with Phil on this.  We are going to add the various ids into the object metadata and add the collection id into the object name.

chenba:
hmmm at the time of upload we don't have all the ids we want in the metadata.  it's not as simple as I imagined.

chenba:
I need a db connection to get the collection id, so either a) I get a connection for the metadata, release it, and then the transaction gets another connection after the upload, or b) get a connection and hold it for the upload and the transaction.  b sounds terrible.  a feels ok; getting a connection from the pool is relatively low cost.  I guess there is the third option of exposing a collection id cache directly.

chenba:
I think what I'll do is incorporate all the changes except those require the collection id here in this PR.  Then if we don't like the PR for the collection id changes, then it'd be easier to reject/revert.

pjenvey:
> I think what I'll do is incorporate all the changes except those require the collection id here in this PR. Then if we don't like the PR for the collection id changes, then it'd be easier to reject/revert.

I'm fine holding off on including a collection id in the name, I'm not totally sure how useful it could potentially be. It might also be worth switching to the `simple` uuid format to save a few characters

taddes (dismissed):

taddes (approved):

PR #2584, opened 2026-09-14
feat: add reconciler finalize lag and run telemetry

taddes:
## Description

Two metrics for the reconciler:

`finalize_age` is the Spanner-commit-to-finalize lag, from the change record's `commitTimestamp`. That is the window an object sits at `committed=false` where the lifecycle policy can still reach it. 

`runs`, `drain_duration`, `messages_processed` and `budget_exhausted` to cover the drain. Every existing counter only fires on the event it names/is related to.

Pulling the loop out into `_drain_loop` is what lets `drain` report duration and volume in a `finally` branch. It also made the loop more testable.

## Issue(s)

Closes [STOR-697](https://mozilla-hub.atlassian.net/browse/STOR-697).

[STOR-697]: https://mozilla-hub.atlassian.net/browse/STOR-697?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


chenba (changes requested):
Doesn't seem to implement everything in STOR-697, e.g. "Upload and download counters plus timers in syncserver."

taddes:
> Doesn't seem to implement everything in [STOR-697](https://mozilla-hub.atlassian.net/browse/STOR-697), e.g. "Upload and download counters plus timers in syncserver."

@chenba I should have just closed Phil's ticket and kept it all to one.  That one's in #2585... this PR is the reconciler side, #2585 is the syncserver side, and between the two all five items from STOR-697 are covered.

My fault for the confusion, 697 is a broad list I wrote spanning both halves, and then 708 got filed for just the timers, so neither ticket maps cleanly onto one PR... 

[STOR-697]: https://mozilla-hub.atlassian.net/browse/STOR-697?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

tools/payload-reconciler/reconcile_payload_links.py
```diff
@@ -221,11 +307,42 @@ def drain(
...
         )
 
+    # Every other counter here is conditional on the event it names, so a run
+    # that finalizes nothing emits nothing at all. `runs` is the unconditional
+    # series that makes "quiet" distinguishable from "not running" on a
+    # dashboard.
+    metrics.incr("runs")
+    started = time.monotonic()
+
+    processed = 0
```
chenba: This stays 0 if `_drain_loop` throws an error, then that causes `messages_processed` to miscount.

taddes: Ah, yeah, thanks for pointing that out.

taddes:
Is it ok to get final approval on this one @chenba ?

tools/payload-reconciler/reconcile_payload_links.py
```diff
@@ -262,6 +377,7 @@ def drain(
             sub_client.acknowledge(
                 request={"subscription": sub_path, "ack_ids": ack_ids}
             )
+        metrics.incr("messages_processed", value=len(response.received_messages))
```
chenba: If the ack above failed then this metric is skip even though the messages have been processed.

chenba (approved):

PR #2585, opened 2026-09-14
feat: emit timing metrics for offloading

taddes:
## Description

Timers on the offload paths, which had none.

- `storage.gcs.payload.upload` / `.download` - per object, tagged `result`
- `storage.gcs.payload.batch` - per request, tagged `op` and `handler`
- `storage.gcs.payload.upload.bytes` / `.download.bytes` - counters

The batch timer is not derivable from the per-object ones. Uploads and downloads run concurrently up to `gcs_payload_max_concurrency`, so the latency a request actually pays is neither the sum nor the max of its parts.

No operation counters, per the ticket. Statsd timers carry their own count. Bytes are counted because latency is not interpretable without size on a project about raising a size ceiling.

`Metrics::start_timer` holds one timer that fires on `Drop`, which cannot express concurrent operations, so this adds `timing_with_tags` to emit directly.

## Issue(s)

Closes [STOR-708](https://mozilla-hub.atlassian.net/browse/STOR-708).

[STOR-708]: https://mozilla-hub.atlassian.net/browse/STOR-708?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


chenba (changes requested):
I don't see items 4 and 5 from STOR-697 in this patch either.

taddes:
> I don't see items 4 and 5 from [STOR-697](https://mozilla-hub.atlassian.net/browse/STOR-697) in this patch either.

K, adding in 4 now. FYI item 5 is on reconciler side #2584 Between the two PRs all five items from 697 are done (like I said over there). Should have closed one of the two tickets instead of splitting them out.

Rather than re-scoping anything, it's all there between the two. Thx for the review. 

[STOR-697]: https://mozilla-hub.atlassian.net/browse/STOR-697?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

syncserver-common/src/metrics.rs
```diff
@@ -109,6 +110,60 @@ impl Metrics {
...
+            }
+            match tagged.try_send() {
+                Err(e) => {
+                    // eat the metric, but log the error
+                    warn!("⚠️ Metric {} error: {:?} ", label, e; MetricTags(mtags));
+                }
+                Ok(v) => trace!("☑️ {:?}", v.as_metric_str()),
+            }
+        }
+    }
```
chenba, line 169: There are three of these `*_with_tags` functions now.  Can you consolidate the code?

syncserver/src/web/handlers.rs
```diff
@@ -350,14 +350,22 @@ pub async fn get_collection(
...
+                            async move {
+                                download_payload(&client, &link, &metrics)
+                                    .await
+                                    .map(|p| (i, p))
+                            }
                         })
                         .buffer_unordered(state.gcs_payload_max_concurrency.get())
                         .try_collect()
                         .await?;
+                    record_batch(&metrics, OP_DOWNLOAD, "get_collection", started.elapsed());
```
chenba: The `?` returns when there's an error so I think we should record the metric before propagating the error.  Same with the upload payload metric.

chenba:
oh the other PR reminded me, we might need to update docs to include the new metrics?

taddes:
> oh the other PR reminded me, we might need to update docs to include the new metrics?

Aha, yes! Always easy to forget. Good call. Want me to roll them in or file new quick tkt?

chenba:
> > oh the other PR reminded me, we might need to update docs to include the new metrics?
> 
> Aha, yes! Always easy to forget. Good call. Want me to roll them in or file new quick tkt?

let's just include them with the code changes

taddes:
> > > oh the other PR reminded me, we might need to update docs to include the new metrics?
> > 
> > 
> > Aha, yes! Always easy to forget. Good call. Want me to roll them in or file new quick tkt?
> 
> let's just include them with the code changes

Done, they're all in one MR to make rebasing easier

syncserver/src/web/payload_offload.rs
```diff
@@ -505,4 +634,211 @@ mod tests {
...
+            "expected a 7ms download timing, got: {emitted}"
+        );
+        assert!(
+            emitted.contains("result:error"),
+            "expected result:error, got: {emitted}"
+        );
+    }
+
+    #[test]
+    fn record_batch_carries_op_and_handler() {
```
chenba, line 673: nonblocking, but it would nice to have handler level test.  we moved the metric recording call to run before error propagation but a test that directly calls `record_batch` doesn't really test that.

chenba (approved):

